PR #206, opened 2020-04-22
Schema check

milen-sage:
Reintroduced code that returns validation error when a csv with the wrong schema is uploaded; the returned error array in this case has format

[["", "Manifest with wrong schema provided", "", errorMessage]]


milen-sage:
Closing this PR since #211 also addresses the schema check.

Note that if the schema is wrong, the returned column in the error list is "Wrong schema".

PR #219, opened 2020-05-23
Schema update to v9

milen-sage:
- includes all the fixes indicated in https://github.com/sage-bionetworks/htan-coordination/issues/128


config.py
```diff
@@ -22,15 +22,15 @@
         "googleManifest":{
             #required columns/cells background color
             "reqBgColor":{
-                        'red': 231.0/255,
+                        'red': 235.0/255,
                         'green': 248.0/255,
                         'blue': 250.0/255
             },
             #optional columns/cells background color
             "optBgColor":{
```
xdoan, line 30: the optional color scheme still looks weird to me (since there are two colors and white it isn't clear what's required vs optional vs nothing at all now) and interested to see how this tests.

milen-sage: oh I totally agree, but people requested and I put it in; we can always just switch back to white :)

PR #226, opened 2020-06-04
Translated "schema generator" to OOP

sujaypatil96:
This PR addresses the following issues:

1. Laid out the skeleton/barebones of the application/directory structure.
2. Translated/converted code within existing "schema generator" module from functional paradigm to object-oriented paradigm.

Request for feedback and review: @milen-sage.


README.md
```diff
@@ -1,4 +1,52 @@
...
+# HTAN Data Ingress Pipeline
+
+## Directory Structure
+
+A bird's eye view of structure and classification of various modules / packages within this repository:
+    
+    HTAN-data-pipeline/
+                       data/
+                       docs/
+                       data-ingress-pipeline/
```
jaeddy, line 10: @sujaypatil96, @milen-sage I was thinking about this organization, and I'm wondering if we could go with a more logical / semantically intuitive package structure:

```
.
├── manifest
│   # package & modules for handling manifests
│   ├── __init__.py
│   └── generator.py
├── models
|    # package & modules for wrapping data models
│   └── __init__.py
├── schemas
|    # packages & modules for handling schemas
│   ├── __init__.py
│   ├── explorer.py
│   └── generator.py
├── synapse
|    # package & modules for interacting with Synapse
│   ├── __init__.py
│   └── store.py
└── utils
```

For example, you might have an import statement in the `manifest.generator` module like this:

```python
from ..schemas import explorer, generator
```

(or something similar)

I'm sure there's room for improvement in my naming scheme, but the overall structure feels a bit cleaner and friendlier to work with (to me). Let me know what you think.
```

Comment (sujaypatil96) [reply README.md:10]: Hi @jaeddy! This looks like a good organization of the directories which (further) groups the packages we had defined earlier.

One quick question about the naming though --

Say, we are importing modules from both `manifest` and `schemas` inside a module in `model` like:

    from ..schemas import explorer, generator
    from ..manifest import explorer

This will cause a conflict isn't it?

Comment (jaeddy) [reply README.md:10]: Thanks, @sujaypatil96 - you're completely right. You probably wouldn't want to import the entire modules either, but specific objects. So, instead, it would look more like this:

```
from ..schemas.explorer import SchemaExplorer
from ..schemas.generator import SchemaGenerator
from ..manifest.generator import ManifestGenerator
```

I'm not sure what we want to name the top level package, but something like this might be even better:
```
from ingresspipe.schemas.explorer import SchemaExplorer
from ingresspipe.schemas.generator import SchemaGenerator
from ingresspipe.manifest.generator import ManifestGenerator
```

sujaypatil96: Ah, yes, you're right, we're importing specific classes rather than whole modules.👍

Yes, we can also consider renaming the top-level package to `ingresspipe` (`ingresspipeline` might be a bit long).

So, the directory structure would now be:

.
|── data
|── docs
|── ingresspipe
|         |── manifest
|         |         # package & modules for handling manifests
|         |         |── __init__.py
|         |        └── generator.py
|         |── models
|         |         # package & modules for wrapping data models
|         |        └── __init__.py
|         |── schemas
|         |          # packages & modules for handling schemas
|         |          |── __init__.py
|         |          |── explorer.py
|         |         └── generator.py
|         |── synapse
|         |         # package & modules for interacting with Synapse
|         |         |── __init__.py
|         |        └── store.py
|         |── utils    
|── tests
└── setup

milen-sage: Starting at the bottom first - we can name the top level package provisionally (e.g. ingresspipe) for now and decide on actual name later? (I think VS code has a feature that helps with automatic refactor due to name changes?). I'd opt for a proper noun for a name instead of functionality/features-driven one, since features often change past tool/package names :) 

I like @jaeddy  idea of shortening the package names (for better style and ease of use) while changing the grouping of classes within a package by topic, so that each package contains all features related to a topic (e.g. manifest or schemas); it would be more user friendly. 

From my perspective, a good consideration when structuring the package hierarchy would be a design enabling "convenient" API calls/routes: ensuring the respective code modules are abstracted sufficiently to provide the "right" granularity for calls that serve the *use cases for data coordination/curation*. That would tend to ensure that, even before we have an API, the methods & classes within each package are packaged at the right abstraction level and are easy to use.

@sujaypatil96, @xdoan, and @jaeddy - we can setup a meeting to discuss the most salient use cases we want to support? Most of these are reflected (although admittedly somewhat ad hoc) in the usage example/test scripts (e.g.  https://github.com/Sage-Bionetworks/HTAN-data-pipeline/blob/dev/metadata_model_test_driver.py and https://github.com/Sage-Bionetworks/HTAN-data-pipeline/blob/dev/storage_test_driver.py) that @sujaypatil96 has  already gone through; and definitely also in the data curator features  - e.g. can take a look at https://github.com/Sage-Bionetworks/HTAN_data_curator/blob/shiny-server/synStoreFuns.py and https://github.com/Sage-Bionetworks/HTAN_data_curator/blob/shiny-server/metadataModelFuns.py. Those curator function provide some intuition for the required level of method/API calls granularity.

PR #230, opened 2020-06-08
New package/directory structure and complete schema generator

sujaypatil96:
This PR addresses the following issues:

1. New package/module structure proposed per PR (closed) #226 
2. Translated/converted code within existing "schema generator" module from functional paradigm to object-oriented paradigm.

Per Milen's comments in #226 , we can can consider renaming `ingresspipe` to `src` in the future, if you don't want to use functionality-specific directory names (but `ingresspipe` should be okay for now).

Request for feedback and review: @milen-sage.


ingresspipe/models/metadata.py
```diff
@@ -0,0 +1,262 @@
...
+from jsonschema import Draft7Validator, exceptions, validate, ValidationError
+
+# allows specifying explicit variable types
+from typing import Any, Dict, Optional, Text, List
+
+# handle schema logic; to be refactored as SchemaExplorer matures into a package
+# as collaboration with Biothings progresses
+
+import sys
+sys.path.append("..") # Adds higher directory to python modules path.
```
jaeddy, line 15: This is a pretty common pattern for running and testing a Python package _before_ you've created your `setup.py` to make it installable. With that working, you can use `pip[3] install -e .` to install the package in 'developer' mode, which allows you to keep making changes to the code while being able to run like any other installed Python library.

Unless you're accessing files outside the main package folder (e.g., configs, data, etc.), you shouldn't need this line.

ingresspipe/models/metadata.py
```diff
@@ -0,0 +1,262 @@
...
+            inputMModelLocationType: specifier to indicate where the metadata model resource can be found; (e.g. 'local' if file/JSON-LD is on local machine)
+        """
+        # extract extension of 'inputMModelLocation'
+        # ensure that it is necessarily pointing to a '.jsonld' file
+        if inputMModelLocation.rpartition('.')[-1] == "jsonld":
+            self.inputMModelLocation = inputMModelLocation
+
+            self.sg = SchemaGenerator(inputMModelLocation)
+        else:
+            print("Can't create object of SchemaGenerator class. Please make sure the 'inputMModelLocation' argument is pointing to a JSON-LD file.")
```
jaeddy, line 50: Just a note, not a blocker: we should use the Python `logging` library for console messages instead of `print`.

ingresspipe/schemas/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .explorer import *
```
jaeddy, line 1: `import *` is generally advised against in Python best practices. Even though its more verbose, it's good to explicitly list out which objects are being imported — e.g.:

```python
from .explorer import SchemaExplorer, SchemaValidator
from some_other_module import SomeClass, some_function
```

sujaypatil96: Thanks @jaeddy! I've made changes accordingly in [this commit](https://github.com/Sage-Bionetworks/HTAN-data-pipeline/pull/230/commits/8d58b75344aefc991c4c142667bdaac05139c5c8).

ingresspipe/models/metadata.py
sujaypatil96: Thanks! Will make sure to change it everywhere accordingly.

ingresspipe/models/metadata.py
sujaypatil96: Yup, you're right. I added the setup.py file (i.e., made the app installable) and tested it without the above lines, and it works without the need to add `..` into the modules PATH.

Thank you!

ingresspipe/models/metadata.py
jaeddy: Thanks, Sujay! OK to save for a separate PR, if you want to get this one wrapped up more quickly.

sujaypatil96:
Hi @jaeddy! Re-requesting a review so we can merge the changes on this PR.😀

Issue #141, opened 2019-11-04
Programmatic end-to-end example of data upload and curation


milen-sage:
We have documented the Data Curator App usage and provided data upload examples for Synapse (both via WUI and client): https://github.com/Sage-Bionetworks/HTAN-data-pipeline/blob/dev/DOCUMENTATION.md

Once we package the backend (#82) we should provide fully-programmatic data upload & data curation example as well, and link it in the documentation above (at least one site has requested support for this use case: OHSU (#91).

milen-sage:
Marking as feature since this is both documentation and a user interface.

ychae:
@milen-sage this is covered in the current HTAN docs. The schematic backend documentation task is being tracked in #362 -- would it be okay to close this?

milen-sage:
yes, let's close - though I do want eventually to see a notebook containing a working end-to-end example... wish list.

ychae:
Notebook ticket created :) #427

PR #242, opened 2020-06-19
Schema Explorer module and Synapse Storage module example(s)

sujaypatil96:
Added example(s) to illustrate the usage of the various methods within `ingresspipe.schemas.explorer` in this PR.

See the examples here: https://github.com/sujaypatil96/HTAN-data-pipeline/tree/organized-into-packages/ingresspipe/schemas

Also added examples to illustrate the usage of the methods in `ingresspipe.synapse.store` module.

See here: https://github.com/sujaypatil96/HTAN-data-pipeline/tree/organized-into-packages/ingresspipe/synapse

Request for review/feedback: @milen-sage / @jaeddy / @xdoan.


jaeddy:
@sujaypatil96 I like the examples in general, but I'm curious: What's the motivation of including examples as Python modules? Do we expect that users might run (or modify and run) the code? If so, something like an IPython/Jupyter notebook might be a better format. In terms of illustrating usage, I think the content you have in the READMEs addresses that goal just fine on its own. @milen-sage, any thoughts?

milen-sage:
Having the examples in a IPython/Jupyter notebook is a good idea -  in addition to the readme info (perhaps in a separate PR).

@sujaypatil96 could we also cover #141 in the examples (separate PR)? We can discuss what to include in it with more detail.

PR #244, opened 2020-06-20
Use paginated REST call 

milen-sage:
Do not use the default results of a non-paginated REST call to retrieve user's project; the default returns only 50 projects.


SynapseStorage.py
```diff
@@ -127,15 +127,19 @@ def getStorageProjects(self) -> List[str]:
...
-        currentUserProjects = self.syn.restGET('/projects/user/{principalId}'.format(principalId=currentUserId))
-        
+        # get a list of projects from Synapse 
+        currentUserProjects = self.getPaginatedRestResults(currentUserId)
+
         # prune results json filtering project id
         currentUserProjects = [currentUserProject.get('id') for currentUserProject in currentUserProjects["results"]]
 
+        print(currentUserProjects)
+        print(storageProjects)
```
sujaypatil96: Hi @milen-sage! I think these debugging/logging statements need to be removed?

PR #245, opened 2020-06-23
cleanup within 'schemas' package

sujaypatil96:
Cleanup of files/re-organization of code within `schemas` package.

Changes made:
1. Moved all required files to main `data` folder (in `ingresspipe` package)
2. Tests to be moved to main `tests` folder
3. Moved `SchemaValidator` to separate `validator.py` module (in `schemas` package)

Question here: at the moment we are using validation methods within `base.py`, but there are similar methods (now) in `validator.py` as well - do we intend to use those methods instead of the ones from `base.py` in the future?

4. Moved `utils` to main `utils` folder (in `ingresspipe` package)

Request for review: @milen-sage.


milen-sage:
> Cleanup of files/re-organization of code within `schemas` package.
> 
> Changes made:
> 
> 1. Moved all required files to main `data` folder (in `ingresspipe` package)
> 2. Tests to be moved to main `tests` folder
> 3. Moved `SchemaValidator` to separate `validator.py` module (in `schemas` package)
> 
> Question here: at the moment we are using validation methods within `base.py`, but there are similar methods (now) in `validator.py` as well - do we intend to use those methods instead of the ones from `base.py` in the future?
> 
> 1. Moved `utils` to main `utils` folder (outside `ingresspipe` package)
> 
> Request for review: @milen-sage.

That's a good question: let's review the meta-schema validation done in base.py and the metadata validation done in validation.py; if we can apply the methods in validation.py on the schemas used in base.py and get the desired validation output I see no reason to not do that.

sujaypatil96:
No, I haven't refactored the code in `validator.py` yet, just moved the code for `SchemaValidator` class which was previously in the same file as `SchemaExplorer` into the `validator.py` file.

The methods in `validator.py` seem to be more than the ones in `base.py`, but as far the ones that are being used in the app, only the ones in `base.py` seems to suffice, which is why I had that question.

Yes, I think it would be a good idea to add the manifest validation logic to `validator.py`. It will also help reduce the size of the method in `manifest.generator`, making it more readable.

milen-sage:
Ok - yes I agree, we should use the code in validator.py instead of the one in base.py. Just double check the validation logic and output ends up working the same (if not, we can add whatever logic is needed to validator.py). 

Wrt manifest validation, I meant this:
https://github.com/Sage-Bionetworks/HTAN-data-pipeline/blob/a0eb46d1108dcfecf09bed50a22a2e4fb9dd46b9/ingresspipe/models/metadata.py#L193
Do you think there is logic in the manifest generator that belongs to the validator as well?

sujaypatil96:
Oh no, my apologies I got confused with the schema validation logic and google api request logic in manifest generator.

Yes definitely, I will verify the outputs of `base` and `validator`.

sujaypatil96:
Cleaned up code in `ingresspipe.schemas`. The `base` and `utils` file (former) was moved to files in `ingresspipe.utils` package.

ingresspipe/utils/general.py
```diff
@@ -1,117 +1,41 @@
...
-    """
-    return [record["rdfs:label"] for record in schema["@graph"] if record['@id'] == uri][0]
-
-
-def find_duplicates(_list):
-    """Find duplicate items in a list
-    """
-    return set([x for x in _list if _list.count(x) > 1])
\ No newline at end of file
+def visualize(edges, size=None):
```
milen-sage, line 33: I see we have df_utils module; do you think it's justified to also have viz_utils that contain functions related to visualization - we might add more options in the future there?

ingresspipe/utils/load_utils.py
```diff
@@ -0,0 +1,42 @@
+import os
```
milen-sage, line 1: not sure what the best name is for this utils file - it does have some export functionality too... do you think something like io_utils might be ok?

ingresspipe/utils/general.py
sujaypatil96: Yes, we could add a `viz_utils` module, right now, it's just sitting in `general`, and it's probably a good idea to move it out.

ingresspipe/utils/load_utils.py
sujaypatil96: Hm..yes, you're right, `io_utils` sounds better. I thought about it for a while, and couldn't come up with a name better than `load_utils` which is why I went with that. But yes, I will change it to `io_utils`.

PR #246, opened 2020-06-29
included ManifestGenerator example

sujaypatil96:
Example (module and README) added for executing utilities/methods within the manifest generator module.


sujaypatil96:
Thanks @milen-sage! I plan to add more content to it based on what gets covered in the upcoming `metadata_model` example.

PR #250, opened 2020-07-08
Support for validation rules + minor bug fixes

milen-sage:
- added support for validation rules in schema definitions
- added multi-value validation rule (e.g. validation rule 'list') as an example of validation rules
--- added HIV schema as an example of how to use the 'list' validation rule (schema is in data/schema_org)

- fixed a bug in refactored synapse/store where the REST call to retrieve projects was using obsolete code
- fixed a bug in refactored manifest/schema generator where attributes were reordered alphabetically even though schema-based ordering was requested
- noted a bug in refactored schema utils: export_schema function is not working (e.g. references self object where there isn't any in the refactored code); so I moved the export_schema function where it was earlier: in schema/explorer.py This is suboptimal and we need to fix the refactored code


milen-sage:
@sujaypatil96 please hold on on merging this - I need to clean a bunch of files I accidentally added.

sujaypatil96:
Hi @milen-sage! It seems that you've pushed code from both the `dev` branch as well as the `organized-into-packages` branch. Not sure if you want to remove the files associated with `dev` branch and issue another PR?

milen-sage:
ok @sujaypatil96 looks better now :) sorry about the messy PR; I accidentally switched my keyboard layout and messed up my staging... I think I removed everything unneeded.

sujaypatil96:
Thanks @milen-sage! I'll test the newly added code with different use-cases (to see of it breaks), and add some comments/feedback to this PR by tomorrow/day after.

HTAN_schema.py
```diff
@@ -0,0 +1,34 @@
...
+
+import pandas as pd
+
+from schema_explorer import SchemaExplorer
+
+from csv_2_schemaorg import create_schema_classes 
+
+
+# path to base schema
+base_schema_path = "./data/schema_org_schemas/biothings_schema.jsonld"
```
sujaypatil96, line 12: Change path to `./data/schema_org_schemas/biothings.jsonld` or any other json-ld file we decide rename the file.

HTAN_schema.py
```diff
@@ -0,0 +1,34 @@
+import os
+import json
+
+import pandas as pd
+
+from schema_explorer import SchemaExplorer
```
sujaypatil96, line 6: Update this with new path to `SchemaExplorer`.

HTAN_schema.py
milen-sage: > Not sure if we want to put `csv_2_schemaorg` module into a package within the `utils` directory/sub-package like `ingresspipe.utils.csv_utils`?

That's a good idea - I'll add it there.

> Also need to update `./data/schema_org_schemas/HTAN.csv` to include a blank `Validation Rules` column.

Will do. 

> Suggestion: let's create a separate directory within `./data/` that contains the csv files to use (test/custom) say, `./data/csv` and also create a config parameter/object to allow configuration of the path to this. The objective of this would be to try and avoid as much hard-coding within the modules as possible.

Will handle that too.

HTAN_schema.py
milen-sage: Yes - that one slipped under the radar... will update.

milen-sage: Where do you think files like HTAN_schema.py should live?
I don't want them to hang around in the main directory. 

These are in a way examples of how to use the csv_2_schema script to create schemas, so they are sort of on the border between schema utils and schema creation examples?

sujaypatil96: Since `HTAN_schema.py` is essentially an example to test the `csv_2_schemaorg.py` module perhaps we can follow a similar structure as we did with other sub-packages and rename `HTAN_schema.py` to `csv_to_schemaorg_usage` and put it into an `examples` sub-folder?

Like this:
ingresspipe
- config
- manifest
- models
- schemas
- utils
    * other utils *
    - csv_utils
    - examples
       - csv_to_schemaorg_usage

milen-sage: I like this idea - could you handle that one?

sujaypatil96: Sure! Will do.

PR #251, opened 2020-07-10
synapse client package upgrade

sujaypatil96:
Upgrade `synapseclient==2.1.0` rather than using _version_ `1.9.2`.

Change made: potentially breaking/deprecating change in `getAnnotations()` and `setAnnotations()` methods. Latest version of the client uses `get_annotations()` and `set_annotations()` instead.

### DIFF
--- a/ingresspipe/synapse/store.py
+++ b/ingresspipe/synapse/store.py
@@ -413,7 +413,17 @@ class SynapseStorage(object):
 
                 metadataSyn[keySyn] = v
 
-            self.syn.setAnnotations(entityId, metadataSyn)
+            # set annotation(s) for the various objects/items in a dataset on Synapse
+            annos = self.syn.get_annotations(entityId)
+
+            for anno_k, anno_v in annos.items():
+                if anno_k in metadataSyn:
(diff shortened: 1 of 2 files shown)


sujaypatil96:
Yes, I've tested the above changes/code by using a dummy manifest file and associating it with a synapse folder/dataset.

Perhaps you could test it once on your end as well - just the `associatedMetadataWithFiles()` method?

I'm yet to test it with the Shiny app, will let you know once I do.

milen-sage:
Sounds good! Let's not merge until we verify the Shiny app works with the token auth mechanism and the latest synapseclient. I think that might be the main breaking change. Switching to oauth won't be hard and is actually preferable long term anyway, so if the token auth doesn't work with the latest synapseclient, I'd vote for prioritizing switching to oauth. However, if the token auth still works we don't need to switch to oauth until later.

sujaypatil96:
Just verified that the Shiny app works with this updated piece of code as well. We are good to go with the merge.

You can test the curator app here: https://shinypro.synapse.org/users/spatil/HTAN_data_curator/

PR #253, opened 2020-07-17
support for csv to schemaorg operation

sujaypatil96:
Based on our discussion in PR #250 (closed), I've organized the module(s) responsible for csv to schema.org conversion. It also includes an example for the same.

Note: I have also thought about a refactor strategy to improve the performance of conversion (changing implementation of `edit_class() / update_class()` which we can discuss over call.

I've tested the modules here, but feel free to test once on your end before merging if you'd like.


sujaypatil96:
I think we should be okay to merge this @milen-sage! I've tested manifest generation (with multi-value input) based on the HIV data model/schema.

milen-sage:
@sujaypatil96 feel free to merge!

PR #256, opened 2020-07-28
external config integration

sujaypatil96:
This PR addresses the following issue: Allow configuration of core module parameters through external config file (`config.yml`) file.


sujaypatil96:
- Thanks for noting the point about things like `class_to_node`, I had created that to test the performance of `update_class()` / `edit_class()` methods. I will issue another commit that removes those changes.

- Yup, I think it's fine to do this until we have a first release of the package. Once that's there, it's easy to use that as a reference to create new/different branches for new features after that.

milen-sage:
Oh no worries, you can keep them in for now. Just in general it's easier to review if all the code for a feature is in the same PR.

milen-sage:
Looks good!

PR #259, opened 2020-08-10
Method that edits class on nx graph

sujaypatil96:
A new method `edit_class_nx()` has been added the `explorer.py` module. This method does the same work as the `edit_class()` method (which can also be found in `explorer.py`, but except it operates directly on the _networkx_ graph rather than parsing and reconstructing the entire graph from the JSON-LD again.

In order to test this method, do the following (example used: `HIV.jsonld`)

1. Identify a class from the _HIV.jsonld_ schema and make a modification to some of the attributes of that class.

Note: Making changes to attributes like "comment", "displayName" will not affect the edges. However, changes made to "subClassOf", "requiresDependency" and other attributes will affect the in-edges and out-edges accordingly.

2. Iterate over the graph, isolate the node and print the contents of the node from the graph before and after modification. You will see the contents of the node has changed. Also verify by printing in-edges and out-edges.

3. You can also use `graphviz` to visualize the graph.


sujaypatil96:
To be added: method to sync the changes to the graph, to the JSON-LD file as well (that's very simple, and is independent of the above method).

milen-sage:
@sujaypatil96 I started reviewing the code, meanwhile could you try running what you showed us today on this schema:
https://github.com/Sage-Bionetworks/HTAN-data-pipeline/blob/organized-into-packages/data/schema_org_schemas/HTAN.jsonld

Given the HTAN schema is large, I am curious how long does it take to update it (I wonder how long it'd take to visualize it as well... :)?

Also, when you get the code at a stable point it would be great if you can try making some edits in the portions of the schema^ that are used in a manifest and check if these edits are reflected in the manifest properly (you can work on a copy of the schema to avoid accidental update of the HTAN schema :). E.g. you can try something along the lines of adding a new node "DoubleStepBrother" (a fake family relationship) and indicating this is a new valid value for relationshipType (see bts:RelationshipType).  It'd be great testing the update code using the csv2schemaorg script as well.

ingresspipe/schemas/explorer.py
```diff
@@ -439,4 +438,129 @@ def get_digraph_by_edge_type(self, edge_type):
...
+                            if target_node != replace_node:
+
+                                # make note of the fact that we are changing in-edges here
+                                schema_graph_nx.add_edge(target_node, replace_node, relationship = "parentOf")
+
+                        # once the edges have been added, change the contents of the node
+                        schema_graph_nx.nodes[node]["subClassOf"] = node_to_replace.nodes[replace_node]["subClassOf"]
+
+                    # "requiresDependency" key related edge manipulation
+                    if "requiresDependency" in data and "requiresDependency" in replace_data:
```
milen-sage, line 507: Just a quick question - if the new/update node has "requiresDependency", but the node_to_replace doesn't have "requriesDependency" why wouldn't we proceed and add "requiresDependency" types of edges? (Similarly to other kinds of edges?

ingresspipe/schemas/explorer.py
```diff
@@ -439,4 +438,129 @@ def get_digraph_by_edge_type(self, edge_type):
...
+                            target_node = extract_name_from_uri_or_curie(_parent["@id"])
+
+                            # label to be associated with "subClassOf" keys is "parentOf"
+                            if target_node != replace_node:
+
+                                # make note of the fact that we are changing in-edges here
+                                schema_graph_nx.add_edge(target_node, replace_node, relationship = "parentOf")
+
+                        # once the edges have been added, change the contents of the node
+                        schema_graph_nx.nodes[node]["subClassOf"] = node_to_replace.nodes[replace_node]["subClassOf"]
```
milen-sage, line 504: Looks like this code (and variations of it) is repeated where the main difference is the type of edge (e.g. subclassOf, requiresDependency, etc.). @sujaypatil96 do you think it would make sense to abstract as a function with arguments that determine the particular types of edges to be added?

ingresspipe/utils/schema_utils.py
```diff
@@ -86,7 +89,7 @@ def load_schema_into_networkx(schema):
                 else:
                     node["required"] = False
 
-
+            # not sure if this is required?
```
milen-sage, line 92: Let's keep it in for now; in the future we might want to handle validation rules in a different way. Or @sujaypatil96 do you think that codepath is not used?

sujaypatil96:
Great! Thank you for the comments @milen-sage!

Yes, absolutely. I am currently testing the code/method(s) on the `HTAN.jsonld` schema itself. I am able to generate visualizations for the same, and you'll be glad to hear it doesn't take too much time, let's say 2-3 mins.

milen-sage:
Do you think you'd be able to try the csv_2_schemaorg script using the updated methods and the HTAN csv schema to check how much time the overall schema generation takes?

ingresspipe/schemas/explorer.py
sujaypatil96: Oh right, that's a great point. I was only thinking of replacing the attributes if they're there in both the existing graph as well as the new replacement node. But this makes sense, I will change the code to accommodate adding these attributes irrespective of whether or not the attribute is present in the existing graph.

sujaypatil96:
Sure! There are still a few changes that I need to make to this method (and their analogous methods) as well. I'll try to complete the implementations for these and then test it with `csv_to_schemaorg`. Does that sound okay?

ingresspipe/schemas/explorer.py
sujaypatil96: Yes, that's exactly right. There's a tonne of redundancy in the code, which I will definitely be removing/abstracting out with a user-defined method. I'm just writing it this way so we can check the behaviour is as expected, and if it works, there are a few more changes I will be making to improve readability, etc.

ingresspipe/utils/schema_utils.py
sujaypatil96: Yup, I feel the `"validationRules"` handler is not required. I can illustrate/demo why during our code review call.

ingresspipe/schemas/explorer.py
milen-sage: Yes - that works great!

milen-sage:
Yes - that works great!

ingresspipe/schemas/explorer.py
milen-sage: cool - feel free to merge after that!

PR #269, opened 2020-09-10
Added handler for quoted valid values

sujaypatil96:
This PR addresses the following issue: add handler for handling "valid values" based on two case scenarios.

Case 1: If it is a comma-separated list, then split by comma (" , ")
Case 2: If it is a comma-separated list, with each value enclosed within double-quotes, then extract all the contents from within the double-quotes

### DIFF
--- a/schematic/utils/csv_utils.py
+++ b/schematic/utils/csv_utils.py
@@ -7,6 +7,8 @@ import string
 import pandas as pd
 import numpy as np
 
+import re
+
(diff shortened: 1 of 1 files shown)


milen-sage (approved):

milen-sage:
great - I was mildly concerned that we are storing the quotes along with the content, but happy that the content is actually extracted from within the quotes; feel free to merge!

PR #271, opened 2020-09-15
Added support for "schema:domainIncludes"

sujaypatil96:
This PR addresses the following issue: There are different edge types being created in the networkx graph - `subClassOf`, `requiresDepedency`, `requiresComponent`, `rangeIncludes`. All of these were being handled except for `domainIncludes`, so this modification adds support for that. An additional edge with edge label "domainValue" will be created between nodes on which the "domainIncludes" attribute exists.

Design:
The edges will run from the Class to the Property (ex. `MolecularEntity -> affectsSynthesisOf`), and this is to indicate that there are bunch/list/set of Properties that are "mapped to" (in other words, "in the domain of") a particular Class.

### DIFF
--- a/schematic/utils/schema_utils.py
+++ b/schematic/utils/schema_utils.py
@@ -85,14 +85,14 @@ def load_schema_into_networkx(schema):
             domain_nodes = record["schema:domainIncludes"]
             if type(domain_nodes) == list:
                 for _domain_node in domain_nodes:
-                    n1 = record["rdfs:label"]  
-                    n2 = extract_name_from_uri_or_curie(_domain_node["@id"]) 
+                    n1 = extract_name_from_uri_or_curie(_domain_node["@id"])
+                    n2 = record["rdfs:label"]
                     # do not allow self-loops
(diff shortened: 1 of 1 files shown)


milen-sage (approved):
@sujaypatil96 sorry this review took longer - I initially saw the amount of change and postponed so that I can go in more detail, but it looks like it's just a bunch of spacing differences and the main changes are just adding the domainIncludes block...

milen-sage:
@sujaypatil96  feel free to merge when ready - just one request in line with our discussion on gradually adopting gitflow practices: could you in the future create changes on the develop-feature-name branch and then PR them into the develop branch in the upstream repo? That way we could have multiple PRs opened (one for each feature branch).

sujaypatil96:
Right! Will do @milen-sage , thanks! I'll create new branches as develop-feature-name before issuing a PR.

Yup, exactly, it was just some "problem" with github's diff tool I guess. But basically, it's just a handler for "domainIncludes" like you said.

PR #273, opened 2020-09-17
docs updates

allaway:
I got a little confused with these docs and eventually realized they are older docs as compared to the HTAN_data_curator repo. I made some changes to make it easier (i think :) ) for someone new to the project to understand what to do. 

I think the main repo has some more instructions not covered here, but I thought it would be helpful to make sure these ones were consistent (e.g. no `conda` vs `virtualenv`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,81 +1,54 @@
-# HTAN Data Ingress Pipeline
+# `schematic` Data Ingress Pipeline
 
 ## Usage
 
-### Virtual Environment Setup
+### Data Curator App Setup (backend)
 
-Python 3 has built-in support for virtual environments (using `venv` module). Perform the following steps:
+_Note: It is assumed that you are running all the below commands from the main/root (HTAN-data-pipeline) directory._
 
-_Note: It is assumed that you are running all the below commands from the main/root (`HTAN-data-pipeline`) directory._
+Clone this repository. 
 
-After cloning the git repository, navigate into the `HTAN-data-pipeline` directory and run the command as below:
+      git clone https://github.com/nf-osi/schematic
 
-```bash
-python[3] -m venv .venv
-```
+If you haven't already, create a conda environment in the cloned directory from the `environment.yml` file which has all the required package dependencies. The conda environment name `data_curator_env` is already defined in the `environment.yml` file.:
 
-This creates a Python3 virtual environment (within the `root` folder/package), with its own site directories (isolated from the system site directories).
+    conda env create -f environment.yml
 
-To activate the virtual environment, run:
+Activate the `data_curator_env` environment:
 
-```bash
-source .venv/bin/activate
-```
+    conda activate data_curator_env
+ 
+Install the package/bundle/application:
 
-_Note: You should now see the name of the virtual environment to the left of the prompt._
-
-### Install App/Package
-
-To install the package/bundle/application:
-
-```bash
-pip[3] install -e .
-```
+      pip install -e .
 
 To verify that the package has been installed (as a `pip` package), check here:
 
-```bash
-pip[3] list
-```
+      pip list
 
 Now, your environment is ready to test the modules within the application.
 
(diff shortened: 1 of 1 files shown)


milen-sage:
@sujaypatil96  we can incorporate this PR, but looking at @allaway's documentation it sounds like we mainly want to ensure 
- documentation for schematic can both be used to install it as a backend for the data curator (assuming shiny server environment for now) *and* as a standalone codebase; we can document the two scenarios explicitly if one set of instructions is not sufficient to work for both: installation via venv and installation via conda in linux shiny server environments. We'd probably note that the latter is what should be used for frontend apps like the Data Curator App (there could be other shiny frontends)
- the Data Curator docs then can just point to schematic's installation docs (instead of maintaining separate instructions)

What do you think?

In the very near future we should just register schematic with pypi...

sujaypatil96:
1. Ideally, it would be better to separate the Data Curator's docs from `schematic`'s docs. It would be nice if the README for `schematic` only had instructions to install the backend. We could point to the Data Curator's README with a link (as resource) if need be.

2. It's okay for the Data Curator to point to `schematic`'s installation instructions, since it is one of the (many) application(s) which uses/will use the schematic backend.

@allaway I think the source of your confusion is the fact that we're using `venv` on the backend, and `conda` for on the frontend. I feel like there's a design decision to be made here @milen-sage to ensure consistency (use of either `venv` or `conda`).

README.md
```diff
@@ -1,81 +1,54 @@
...
-```
-
-To run any of the example file(s), go to your root directory and execute/run python script in the following way:
-
-Let's say you want to run the `metadata_usage` example - then do this:
-
-```bash
-python[3] schematic/models/examples/metadata_usage.py
-```
+      conda deactivate
```
sujaypatil96, line 33: Ideally, we want to avoid associating schematic's README with information about the Data Curator app, since we're developing `schematic` to be a standalone package to be used not only by the data curator app, but other projects that want to leverage this infrastructure. I think it would be a good idea to simply add a link to the Data Curator's README.

milen-sage: @sujaypatil96 I agree with both 1 and 2 above. 

Do we know why conda is required for the shiny data curator app and wouldn't venv be sufficient? If venv is sufficient then, it would be much better to just be consistent and use venv both for the backend and the frontend. @xdoan perhaps you are aware of venv limitations on the shiny server?

xdoan: you can use venv I just don't personally since I use conda. Reticulate can use python, venv, conda env, and there are some cursory instructions for how to change that.

milen-sage: We could have "official" instructions and note that advanced users can use conda? The official instructions could be based on venv since it's more standard with python, @sujaypatil96 @xdoan ?

xdoan: I currently use conda to manange both python and other dependencies, e.g. added linux libraries for the shiny server and was planning to add the R libraries needed as well. Does venv allow for listing of non python packages?

milen-sage: I see. A lot of this will be resolved by 
- register schematic with pypi (which works fine with pip install w/in conda environments): 1-2 months
- build conda package for the pypi package (optional): 2-3 months
- stand schematic API: ~6 months

Short-term, @sujaypatil96 do you think it is feasible to test the conda based instructions for schematic (e.g. considering @allaway's PR, still using 'pip install', etc.)? I still like the idea @sujaypatil96 suggested of having the data curator app usage instructions
1. point to schematic's install instructions; the latter updated to a conda-based environment
and next 
2. guide through data-curator-specific dependencies installation in conda, assuming schematic has been installed in the conda-based environment (as per 1); that way as @xdoan mentioned, she can manage other kinds of dependencies.

The result is that users can install stand-alone schematic or data curator + schematic, using the same conda environment; the instructions are compatible but maintained separate for clarity.

Thoughts?

PR #277, opened 2020-09-18
added create_nx_schema_objects() method in csv_utils

sujaypatil96:
This PR adds the feature that improves the speed of the "CSV to schema.org" parser by operating on the networkx graph directly rather than on the JSON-LD schema.

How to test it?
Simply replace the call to `create_schema_classess()` with `create_nx_schema_objects()` and run `python[3] examples/csv_to_schemaorg.py` on the terminal.

Right now, you can test it with `example.csv` and no other csv file because it is failing due to some inconsistencies in label name provided in csv file ("0"/"1"/"2"/etc. specifically).

### DIFF
--- a/schematic/schemas/explorer.py
+++ b/schematic/schemas/explorer.py
@@ -98,7 +98,6 @@ class SchemaExplorer():
     def find_class_specific_properties(self, schema_class):
         """Find properties specifically associated with a given class
         """
-        #print(schema_class)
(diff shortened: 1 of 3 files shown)


milen-sage:
awesome! I'll give it a shot. Let's poke around the "\0"... issue; I'll take a look on my end as well but let me know if you figure it before me

PR #280, opened 2020-09-22
fixing a bug where manifest is not found 

### LINKED ISSUE #279: develop vers of schematic does not return previously submitted manifests
Testing the develop version of the schematic with the data curator app on the dev server here: https://www.synapse.org/#!Wiki:syn20930961/ENTITY and when getting the template previously updated manifests are not returned, an empty new manifest is instead returned.

milen-sage:
The manifest filename in the config is now a path; store.py should extract the manifest filename to successfully locate existing manifest files in dataset folders on synapse.

@sujaypatil96 and @xdoan could you verify this fix resolves #279 on your ends?

### DIFF
--- a/schematic/synapse/store.py
+++ b/schematic/synapse/store.py
@@ -240,7 +240,7 @@ class SynapseStorage(object):
         """
 
         # get a list of files containing the manifest for this dataset (if any)
-        manifest = self.getFilesInStorageDataset(datasetId, fileNames = [self.manifest])
+        manifest = self.getFilesInStorageDataset(datasetId, fileNames = [os.path.basename(self.manifest)])
         
         if not manifest:
             return []


sujaypatil96 (approved):
Tested the code on the backend with a dataset from HTAN CenterA ("syn21973647"), it works.

`manifest_id = syn_store.getDatasetManifest("syn21973647")` correctly returns `('syn21989523', 'synapse_storage_manifest.csv')` which is the expected tuple to be returned.

The PR is good to be merged @milen-sage !

milen-sage:
@xdoan looks like you can update the data curator app to resolve the manifest issue.

xdoan:
yup tested on dev app and it works @milen-sage. will this get merged into the main soon or should the prod version also use develop?

milen-sage:
In general, let's merge develop to main only when we have stable code in develop (which we can aim to be most of the time, given we PR in develop from different branches). 

In this case, I think develop is stable enough to merge to main.

milen-sage:
@xdoan to resolve communication, let's have all backend updates merged to main undergo review, so that, at least, the three of us are on board when new code is pushed.

PR #282, opened 2020-09-23
Add command-line interface to csv_to_schemaorg.py

BrunoGrandePhD:
- Remove hard-coded values (like "HTAN").
- Look for BioThings JSON-LD file relative to script. If absent, require the user to provide a base schema via the CLI.
- Provide input and output files via argparse CLI.
- Enable execute permission on the script and add shebang line.
- Default to output JSON-LD alongside CSV.


examples/csv_to_schemaorg.py
```diff
@@ -1,37 +1,55 @@
...
-# schema extension definition csv files
-schema_extensions_csv = [
-                        os.path.join(DATA_PATH, '', 'csv/HTAN.csv')
-                        ]
+# Constants (to avoid magic numbers)
+FIRST = 0
+
+# Try to find BioThings JSON-LD file relative to script
+script_dir = os.path.dirname(os.path.realpath(__file__))
+biothings  = os.path.join(script_dir, "..", "data", "schema_org_schemas",
```
sujaypatil96, line 19: Great PR @BrunoGrandePhD , thank you!

Quick question here, any particular reason why you've chosen to go with relative paths? Our philosophy behind creating and using the `config_data` object was that it would load all the configurations from the `config.yml` file and would save us the trouble of modifying any hardcoded paths within any of the scripts.

My thoughts on the above lines are that if ever we decide to change the names of the folders within the `data/` then we would have to change it in this `csv_to_schemaorg.py` file as well, in addition to changing it in the `config.yml` file.

sujaypatil96 (changes requested):

examples/csv_to_schemaorg.py
BrunoGrandePhD: Moving away from using `config_data` was the change I was most hesitant about committing. The reason why I opted to go ahead with the change was the fact that the script imports from the `definitions.py` module, which assumes that the script is being run in the root of the `schematic` repo. Prompted by your question here, I just tried running the script in a different directory and surprisingly, it worked. 

Having looked into this further, the only reason why this has worked so far is because using `pip install -e` adds the repo directory to the Python module path (_i.e._ `import sys; print(sys.path)`). As a result, the `definitions` module can be imported. If you install `schematic` without the `-e` option, running the script results in a `ModuleNotFoundError: No module named 'definitions'` error. 

Maybe we can still rely on the config file but use a path provided at the command line instead of relying on the `definitions` module. What do you think, @sujaypatil96?

sujaypatil96: Right, so `definitions.py` was created with the intent that it would be the file that included all the "constants". All constants that are currently hardcoded (in the `examples/` folder) should be ideally live in that file. `definitions.py` was inspired by Django because that's how Django maintains constants.

> Having looked into this further, the only reason why this has worked so far is because using `pip install -e .` adds the repo directory to the Python module path (i.e. `import sys; print(sys.path)`). As a result, the `definitions` module can be imported. If you install `schematic` without the -e option, running the script results in a `ModuleNotFoundError: No module named 'definitions' error`.

Yup, you're exactly right, and makes sense. But this `definitions.py` file contains constants that are used by most of the example files in the project, do you think it would be a good idea to simply move the `definitions.py` file into the `schematic` package as well and maintain it there? What do you think about that approach @BrunoGrandePhD?

BrunoGrandePhD: I've just got off a call with @sujaypatil96. Here's a quick summary of our conclusions:

- Given that the goal is to submit the schematic package to PyPI, the current implementation of the `definitions.py` script will be problematic because it will refer to file paths within the Python installation (where packages are installed). This assumes that `definitions.py` gets moved within the schematic package. This is because of the use of the `__file__ special attribute in `definitions.py`. 
- It's conceivable that a user might have one installation of schematic (installed from PyPI) but multiple projects to manage. To handle this case, all file paths that are project specific (such as the various account credentials) need to be user-configurable. Currently, they are relative to the package, and thus only only project could be managed per schematic installation. 
- To address these issues, the default values in `definitions.py` will be moved to a section in the `config.yml` file. This section will be annotated with a comment along the lines of "Don't edit this section unless you know what you're doing". Additionally, every user-facing script in `examples/` (which will eventually be exposed as command-line tools when installed via pip) will have a `--config` argument to specify the project configuration YAML file. 

I'm going to take a first stab at this and submit a PR to solicit additional comments on the design.

config.yml
```diff
@@ -1,30 +1,38 @@
----
```
BrunoGrandePhD, line 1: Most of the file looks altered because I removed the `---` at the top (it didn't break anything as far as I can tell) and I de-indented everything so that the top-level keys are "flush" with the left margin.

schematic/configuration.py
```diff
@@ -0,0 +1,98 @@
+import os
+import yaml
+
+
+class Configuration(object):
```
BrunoGrandePhD, line 5: I decided to implement the configuration as a class for a few reasons:
- Apparently, once you import a constants in Python, the value can't be changed later by the module. For instance, if `schematic.TOKEN_PICKLE` was set to `None` when `schematic` is imported but is later updated to the actual value when the user runs `load_config()`, `schematic.TOKEN_PICKLE` will still be set to `None`. There might be a workaround that would involve putting some import statements later in the Python file, but that's unconventional. 
- You can use the special object methods (such as `__getattribute__()` and `__getitem__()`) to do some cool tricks. See comments below for details. 
- When imported in another module or script, everything is encapsulated in one object. You don't have to import each config value separately or load the YAML file over and over again.

schematic/configuration.py
```diff
@@ -0,0 +1,98 @@
...
+        self.DATA_PATH = None
+        # path to service account credentials file
+        self.SERVICE_ACCT_CREDS = None
+        # path to synapse config file
+        self.SYNAPSE_CONFIG_PATH = None
+        # entire configuration data
+        self.DATA = None
+
+
+    def __getattribute__(self, name):
```
BrunoGrandePhD, line 26: This method essentially implements "lazy loading" of the configuration. Whenever the user or the module tries to access an attribute using the dot syntax (_e.g._ `schematic.TOKEN_PICKLE`), this method will check if the value is `None`, which means that nothing has been loaded so far. In most cases, @sujaypatil96 confirmed that in most cases, `schematic` will be used via one of the provided scripts, which should load a config file as demonstrated in `csv_to_schemaorg.py`. 

As these interfaces are implemented, I added a quick way of passing that argument without a CLI using the environment variable `SCHEMATIC_CONFIG`. The two commands below achieve the same thing, but you can see in the output of the second command, the config was pulled from the environment variable. 

```
$ python ~/repos/schematic/examples/csv_to_schemaorg.py --config ~/repos/schematic/config.yml ~/repos/schematic/data/csv/example.csv && echo "Exit code: $status"

[...]
Exit code: 0


$ SCHEMATIC_CONFIG=~/repos/schematic/config.yml python ~/repos/schematic/examples/csv_to_schemaorg.py ~/repos/schematic/data/csv/example.csv && echo "Exit code: $status"

Loading config YAML file specified in 'SCHEMATIC_CONFIG' environment variable: /Users/bgrande/repos/schematic/config.yml
[...]
Exit code: 0
```

schematic/configuration.py
```diff
@@ -0,0 +1,98 @@
...
+            )
+        return value
+
+
+    def __getitem__(self, key):
+        return self.DATA[key]
+
+
+    @staticmethod
+    def load_yaml(file_path: str) -> dict:
```
BrunoGrandePhD, line 53: I moved the `load_yaml()` function to here because I would run into circular import errors otherwise, and it seemed like an inconsequential change since there was only that one function in the module where it came from.

schematic/manifest/examples/generator_usage.py
```diff
@@ -4,13 +4,10 @@
...
 from schematic.utils.google_api_utils import download_creds_file
-from schematic.utils.config_utils import load_yaml
 
-from definitions import CONFIG_PATH, DATA_PATH
+from schematic import CONFIG
 
-config_data = load_yaml(CONFIG_PATH)
-
-PATH_TO_JSONLD = os.path.join(DATA_PATH, config_data["model"]["input"]["location"])
+PATH_TO_JSONLD = os.path.join(CONFIG.DATA_PATH, CONFIG["model"]["input"]["location"])
```
BrunoGrandePhD, line 10: Most of the edits across the module files look like this. 

1. The config file is no longer loading in modules individually. Now, it will get loaded once as soon as one of the attributes is accessed. 
2. Any of the raw variables are prefixed with `CONFIG.`. 
3. Anytime the config dictionary is used, the `CONFIG` object is indexed directly. This is just a shortcut for `CONFIG.DATA[key]`.

schematic/utils/google_api_utils.py
```diff
@@ -8,13 +8,8 @@
...
 from google.oauth2 import service_account
 
-from schematic.utils.config_utils import load_yaml
-from definitions import ROOT_DIR, CONFIG_PATH, CREDS_PATH, TOKEN_PICKLE, CREDS_PATH, SERVICE_ACCT_CREDS
+from schematic import CONFIG
 
-config_data = load_yaml(CONFIG_PATH)
-
-# synapse ID of the 'credentials.json' file, which we need in order to establish communication with gAPIs/services
-SYN_CREDS = config_data["synapse"]["api_creds"]
```
BrunoGrandePhD, line 17: Because the config might not be loaded as soon as the module is imported, I moved any attempts to use the config within the relevant functions. This way, an attempt to use the config can come after the config is loaded.

schematic/configuration.py
```diff
@@ -0,0 +1,112 @@
...
+        with open(file_path, 'r') as stream:
+            try:
+                config_data = yaml.safe_load(stream)
+            except yaml.YAMLError as exc:
+                print(exc)
+                return None
+        return config_data
+
+
+    def normalize_path(self, path):
```
BrunoGrandePhD, line 58: I thought it might be intuitive to have all non-absolute file paths in the config file to be relative to the config file itself. In most cases, I imagine a folder with the config file and all other associated files (_e.g._ credentials).

schematic/configuration.py
```diff
@@ -0,0 +1,112 @@
...
+                "set yet, presumably because the schematic.CONFIG.load_config() "
+                "method hasn't been run yet. Alternatively, you can re-run this "
+                "code with the 'SCHEMATIC_CONFIG' environment variable set to "
+                "the config.yml file, which will be automatically loaded."
+                % name
+            )
+        return value
+
+
+    def __getitem__(self, key):
```
BrunoGrandePhD, line 43: Just a small shortcut so users can use `CONFIG[key]` instead of having to know that the config dictionary is in `CONFIG.DATA`.

examples/csv_to_schemaorg.py
```diff
@@ -1,37 +1,58 @@
...
+parser.add_argument("schema_csv_list", nargs="+", metavar="schema_csv",
+                    help="Input CSV schema files.")
+parser.add_argument("--output_jsonld", "-o", help="Output JSON-LD schema file.")
+parser.add_argument("--config", "-c", help="Configuration YAML file.")
+args = parser.parse_args()
+
+# Load configuration
+config_data = CONFIG.load_config(args.config)
+biothings_jsonld = CONFIG["model"]["biothings"]["location"]
+base_schema_path = os.path.join(CONFIG.DATA_PATH, biothings_jsonld)
```
BrunoGrandePhD, line 34: This is what I imagine would be in every user-facing `schematic` script. Alternatively, the user can specify the config using the `SCHEMATIC_CONFIG` environment variable (see comments in `schematic/configuration.py` for more details).

BrunoGrandePhD:
Alright, this is ready for review, @sujaypatil96! The changes are substantial, so please feel free to critique my approach. I'm open to trying a different method.

config.yml
sujaypatil96: Yup, great! Since we don't have any YAML directives in our config file, the presence (or absence)  of the dashes ("---") doesn't make a difference here.

config.yml
```diff
@@ -1,30 +1,38 @@
...
+    master_template_id: '1LYS5qE4nV9jzcYw5sXwCza25slDfRA1CIg3cs-hCdpU'
+
+# Do not change the 'definitions' section unless you know what you're doing
+definitions:
+  root_dir: "./"
+  synapse_config: ".synapseConfig"
+  creds_path: "credentials.json"
+  token_pickle: "token.pickle"
+  data_path: "data/"
+  service_acct_creds: "quickstart-1560359685924-198a7114b6b5.json"
```
sujaypatil96, line 38: Trivial note:

Do you think it would make more sense to move this "definitions" section to the top of the YAML file? I'm suggesting this because all paths below it (for ex. `[model][input][location]`) is/will be relative to the `root_dir` / `data_path`.

examples/csv_to_schemaorg.py
```diff
@@ -1,37 +1,58 @@
...
+from schematic import CONFIG
+
+# Constants (to avoid magic numbers)
+FIRST = 0
+
+# Try to find BioThings JSON-LD file relative to script
+script_dir = os.path.dirname(os.path.realpath(__file__))
+biothings  = os.path.join(script_dir, "..", "data", "schema_org_schemas",
+                          "biothings.jsonld")
+is_biothings_absent = not os.path.exists(biothings)
```
sujaypatil96, line 21: Is this piece of code still necessary? I believe L#32 and L#33 take care of the same.

schematic/__init__.py
```diff
@@ -0,0 +1 @@
+from .configuration import CONFIG
```
sujaypatil96, line 1: Do we want to replace this with `from schematic.configuration import CONFIG`? The way it is right now is good enough, just checking what would be best.

sujaypatil96 (approved):
All the modifications look great @BrunoGrandePhD! Thank you for thinking through all it and significantly improving the credentials management process.👌

examples/csv_to_schemaorg.py
BrunoGrandePhD: You would be correct! I forgot to remove this bit of code.

schematic/__init__.py
BrunoGrandePhD: Agreed. It's better to be consistent with the rest of the package.

config.yml
BrunoGrandePhD: Because we've moved away from everything being in the same folder relative to the repo, maybe we can drop the concept of a root folder. All paths in the config should be absolute or relative to the config file. What do you think?

BrunoGrandePhD: Here's a quick search for `ROOT_DIR` in the `schematic/` subfolder. If we exclude the matches in `configuration.py` or Markdown files, there aren't many other references to `ROOT_DIR`. It would be easy to remove these. 

```
9 results - 5 files

schematic/configuration.py:
    8          # this is the project's root directory
    9:         self.ROOT_DIR = None
   10          # path to config.yml file

   92          # Update module-level configuration constants
   93:         self.ROOT_DIR = self.DATA["definitions"]["root_dir"]
   94          self.CONFIG_PATH = config_path

  100          # Normalize all file paths as absolute file paths
  101:         self.ROOT_DIR = self.normalize_path(self.ROOT_DIR)
  102          self.CONFIG_PATH = self.normalize_path(self.CONFIG_PATH)

schematic/models/README.md:
  12  
  13: _Note: `CONFIG.CONFIG_PATH`, `CONFIG.ROOT_DIR`, `CONFIG.DATA_PATH` (constants) are loaded from the configuration YAML file._
  14  

  32      syn.login()
  33:     syn.get(SYN_CREDS, downloadLocation = CONFIG.ROOT_DIR)
  34      print("Stored Google API credentials.")

schematic/models/examples/metadata_usage.py:
  37      syn.login()
  38:     syn.get(API_CREDS, downloadLocation = CONFIG.ROOT_DIR)
  39      print("Stored Google API credentials.")

schematic/synapse/examples/store_usage.py:
  8  
  9: PATH_TO_SYN_CONF = os.path.join(CONFIG.ROOT_DIR, CONFIG.SYNAPSE_CONFIG_PATH)
  10  

schematic/utils/google_api_utils.py:
  70          syn.login()
  71:         syn.get(SYN_CREDS, downloadLocation = CONFIG.ROOT_DIR)
  72          print("Downloaded Google API credentials file.")
```

sujaypatil96: That makes sense @BrunoGrandePhD!

So we'll get rid of the `root_dir`attribute from `definitions`? But we still want to allow users to configure other parameters in definitions, like the `data_path` and stuff right?

BrunoGrandePhD: > So we'll get rid of the `root_dir` attribute from `definitions`? 

Exactly, that's what I had in mind. Otherwise, it'll be confusing for users to know if non-absolute paths are relative to the config file or the specified root directory. In the code, I convert all file paths under `definitions` to absolute ones relative to the config file (see [code](https://github.com/BrunoGrandePhD/schematic/blob/add_csv_to_schemaorg_cli/schematic/configuration.py#L58-L65
)). I should consider extending this to all file paths that appear in the config file, not just the ones under `definitions`. Otherwise, we still face inconsistent behavior. 

On this note, I should look into all instances of `os.path.join()` in `schematic` to make sure that the paths are all being updated consistently. 

> But we still want to allow users to configure other parameters in definitions, like the `data_path` and stuff right?

👍

sujaypatil96: That sounds great. Yes, it's important to ensure consistency (with respect to file paths) across all the modules in the app. Feel free to open another PR for that if you'd like. Else, new commits on this should work just as well. Thank you!

BrunoGrandePhD: Actually, I want to go back on what I said about `data_path`. I noticed that most of the `os.path.join()` calls are concatenating `data_path` with another attribute from the config (_e.g._ `CONFIG["model"]["input"]["location"]`). In fact, that's all `data_path` is ever used for: for being concatenated with another attribute in the config (see search results below), except for `csv_to_schemaorg_usage.py`, which needs to be updated to use a CLI. 

I propose that we remove `data_path` from the config, and replace all instances of `os.path.join(CONFIG.DATA_PATH, CONFIG["X"]["Y"]["Z"])` with just `CONFIG["X"]["Y"]["Z"]`.

```
schematic/manifest/examples/generator_usage.py:
  10: PATH_TO_JSONLD = os.path.join(CONFIG.DATA_PATH, CONFIG["model"]["input"]["location"])

schematic/models/examples/metadata_usage.py:
  13: MM_LOC = os.path.join(CONFIG.DATA_PATH, CONFIG["model"]["input"]["location"])
  54: HTAPP_VALIDATION_SCHEMA = os.path.join(CONFIG.DATA_PATH, CONFIG["model"]["demo"]
  59: HTAPP_SCHEMA = os.path.join(CONFIG.DATA_PATH, CONFIG["model"]["demo"]["location"])
  72: MANIFEST_PATH = os.path.join(CONFIG.DATA_PATH, CONFIG["model"]["demo"]["valid_manifest"])

schematic/schemas/generator.py:
  652:         json_schema_log_file = os.path.join(CONFIG.DATA_PATH, '', CONFIG["model"]["input"]["log_location"])

schematic/schemas/examples/explorer_usage.py:
  10: PATH_TO_JSONLD = os.path.join(CONFIG.DATA_PATH, CONFIG["model"]["input"]["location"])
  47: gv_digraph.render(os.path.join(CONFIG.DATA_PATH, '', 'viz/example-GV'), view=True)

schematic/schemas/examples/generator_usage.py:
  7: PATH_TO_JSONLD = os.path.join(CONFIG.DATA_PATH, CONFIG["model"]["input"]["location"])

schematic/synapse/store.py:
  80:             self.manifest = os.path.join(CONFIG.DATA_PATH, CONFIG["synapse"]["manifest_filename"])

schematic/synapse/examples/store_usage.py:
  46: MANIFEST_LOC = os.path.join(CONFIG.DATA_PATH, '', CONFIG["synapse"]["manifest_filename"])

schematic/utils/examples/csv_to_schemaorg_usage.py:
  13: base_schema_path = os.path.join(CONFIG.DATA_PATH, '', CONFIG["model"]["biothings"]["location"])
  19  schema_extensions_csv = [
  20:                         os.path.join(CONFIG.DATA_PATH, '', 'csv/example.csv')
  21                          ]
```

sujaypatil96: Right, that makes sense. Initially, we were allowing the "folder" to be configured in `definitions.py` and the "relative file path" (relative to the directory specified in `definitions`) to be configured in the `config.yml`.

But you're right, what you're suggesting makes complete sense. We can remove the `data_path` parameter, and start specifying full paths (ex. `data/schema_org_schemas/example.jsonld`) and such) for other parameters in the `config.yml` file.

milen-sage: I agree integrating paths in config.yml would make it more streamlined .

Looks like this PR is approved but there is still more work on it, so I assume it's not ready to be merged yet?

BrunoGrandePhD: Yeah, it's not ready yet. I need to wrap up the following:

- [ ] Replace remaining references to the `definitions.py` variables
- [ ] Remove `root_dir` and `data_path` references
- [ ] Track down all non-python files that need to be distributed with the package
  - [ ] Refer to these files using `pkg_resources` for portability
- [ ] Add CLI to remaining command-line tools (at least the most commonly used scripts)

BrunoGrandePhD:
In `examples/schema_explorer.py` and in `schematic/schemas/README.md`, there's a reference to `'viz/HTAN-GV'`. How should I handle this file path, which doesn't exist in the repo? I came across it while looking for instances of `DATA_PATH` being joined with other paths.

BrunoGrandePhD:
What I thought was going to be a small PR has grown into many more changes than I'm comfortable merging without testing more thoroughly (or a very thorough review by someone, but there are so many files that have changed that I wouldn't want to subject anyone to that). Ultimately, I defer to @sujaypatil96 and @milen-sage on deciding if they feel comfortable merging with `develop`. IIRC, Sujay is planning on developing a test suite over the coming weeks, so he could either merge this PR with `develop` and identify any bugs I may have introduced while working on the tests. Alternatively, if you prefer keeping `develop` a bit more stable, @sujaypatil96 could branch off of this branch for working on his test suite. 

Right now, I still have a few things left to do:

- [ ] Add CLI to remaining command-line tools (at least the most commonly used scripts)
- [ ] Track down all non-python files that need to be distributed with the package
  - [ ] Refer to these files using `pkg_resources` for portability
  - [ ] Deal with references to `schema_org_schemas` and `validation_schemas`

First, I would like to know which scripts should be given a command-line interface. Which ones are expected to be used by the user? 

Second, I'm waiting to hear more from this [thread](https://sagebionetworks.slack.com/archives/C01ANC02U59/p1601076059077900) on Slack in order to figure out which files should be bundled in the package on PyPI and which files should be in the repo as examples without being bundled with the package.

milen-sage:
@BrunoGrandePhD the main operations, currently supported, that a user would likely run are
- generate a json-ld from given an input csv 
- validate that the metadata in a given csv manifest match the requirements defined for the manifest's respective data-type in the json-ld schema (example for that currently is in model/examples/metadata_usage.py). We can wait and see if people would want to validate using a custom provided jsonschema (which may or may not be generated from the json-ld); for now that's something that we probably don't need to expose via CLI, but it would be easy to add later (example for it currently also in model/examples/metadata_usage.py)
- generate an empty google sheet manifest template, given a data-type present in the json-ld schema (example in manifest/examples)
- prepopulate a google sheet manifest template with data in a csv manifest (I think there is an example in manifest/examples)
- store a manifest on Synapse (example in synapse/examples/)

milen-sage:
@BrunoGrandePhD regarding references to schema_org_schemas and validation_schemas, I left some comments on the [slack thread](https://sagebionetworks.slack.com/archives/C01ANC02U59/p1601076059077900) you mentioned. Let me know if I missed something there.

sujaypatil96:
Thanks @BrunoGrandePhD! I'm going to go ahead and merge this PR into `develop`. Looks like all the configuration module(s) related changes are working well.

> First, I would like to know which scripts should be given a command-line interface. Which ones are expected to be used by the user?

All the scripts within the `examples/` directory will be the user-facing scripts, so it'd be nice to convert all of them to CLI-based scripts (exposing the configurable parameters within them as CLI args), and ultimately string all of them together into a `bash` script.

PR #286, opened 2020-09-29
HTAN schema update

milen-sage:
Updated HTAN schema with biospecimen and minimal metadata for an assay not covered in RFCs.


milen-sage:
Process: merge PR in develop when approved; test with data curator app; merge develop in main after that; deploy in production. Please treat as *highest priority* based on requests from DCC and external collaborators.

milen-sage:
technically, I should have pushed this PR to ncihtan since its HTAN specific; but I'm waiting out until we finalize our project-specific repos setup

milen-sage:
@xdoan note the schema labels we can generate manifests for are:
- Biospecimen
- OtherAssay

OtherAssay generates a minimal metadata manifest for an assay not covered in the existing RFCs.

xdoan:
currently testing, @milen-sage to clarify do you mean the schema_name is Biospecimen and OtherAssay or is that the display name you wanted?

milen-sage:
yep, schema_name is Biospecimen and OtherAssay; display name can be more descriptive (e.g. 'Biospecimen - All Tiers', and 'Other Assay Type').

xdoan:
So I'm running into issues bc you are merging into develop and @sujaypatil96 already merged the command line utils stuff that is not compatible with the frontend, so I get errors such as "ModuleNotFoundError: No module named 'schematic.utils.config_utils'"

milen-sage:
yep

milen-sage:
Aside from the above two, here are the additional schema names:

BulkRNA-seqLevel3

BulkWESLevel1
BulkWESLevel2
BulkWESLevel3

ScRNA-seqLevel2
ScRNA-seqLevel3
ScRNA-seqLevel4

ScATAC-seqLevel1

@xdoan you can use the same display name pattern that you already use for assays.

xdoan (approved):
it technically works, but there needs to be QC of the actual content per schema

PR #287, opened 2020-09-30
Fix to ensure new Configuration compatibility

sujaypatil96:
We need to update the paths in the `config.yml` file such that they are paths relative to the `config.yml` file. So now, they will be full paths, like `data/schema_org_schemas/HTAN.jsonld` instead of relative to the `data` folder (as it was before).

There was a small update in `io_utils.py` which was necessary for the working of schema explorer module.


milen-sage:
@sujaypatil96 just a quick question is this for compatibility with the data curator app or generally to make the config work with schematic?

PR #288, opened 2020-09-30
Revert "Add command-line interface to csv_to_schemaorg.py"

milen-sage:
Reverts Sage-Bionetworks/schematic#282

Let's revisit these changes on a separate branch (e.g. develop-cli); let's try avoiding setting up env paths. We might want to take a slightly different approach for either the config portion or how we have structured our code (e.g. examples). 

@sujaypatil96 if the other open PR #287 is related to this code please point to develop-cli branch, provided we decide to work on this reverted PR further on a develop-cli branch (please feel free to create the branch as well).


BrunoGrandePhD:
@milen-sage: Should I submit a new PR from my fork and enable “Allow edits by maintainers”? Alternatively, you can grant me write-access to this repo and I can push my changes to the `develop-cli` branch.

sujaypatil96:
Thanks @milen-sage! Will do. 

I'll create a new branch `develop-config` for the Config related changes, and another branch called `develop-cli` for the CLI related additions.

sujaypatil96 (approved):
Looks good! This is good to revert. All of the 30 files that were modified in the original PR seem to be reverted.

BrunoGrandePhD:
@sujaypatil96: I suggest keeping everything in one branch because the CLIs somewhat depend on the new config setup.

sujaypatil96:
@BrunoGrandePhD: Right, that's why I created only a `develop-config` branch on `schematic` repo for now. Once we've finalized everything that has to do with `develop-config`, maybe I/you/Milen can either merge that to `develop`, and then we can create a new branch from `develop` for `develop-cli`, or we can just create a new branch `develop-config`. Does that make sense?

milen-sage:
Hmm, let's track all changes related to enabling CLI (including the CLI logic itself + the config related changes + any other necessary setup or code changes) on a single branch. 

@sujaypatil96 if you think that the the work you propose to track on develop-config is independent of the CLI and so can be merged into develop separately, let's do that. However, if the CLI code depends on that work, we should do that merge into the develop branch soon, so that we can start the develop-cli branch tracking the CLI work consequently. On the other hand, if the config work depends on work in the CLI (reverted PR) let's keep everything on the same branch straight away (e.g. develop-cli).

Either way, at the end of the day we want to have a single develop-cli branch that we can review and merge into develop when ready.

sujaypatil96:
@milen-sage I feel like the changes that have to do with `config` can be tracked independently of the changes that have to do with `cli`, but @BrunoGrandePhD will be a better judge of this than I. I thought it would be better to make all the changes that have to with modifying `config.yml`, adding the new Config class, etc. first, and test it with the rest of the core modules. Once that's ready, we can use it in each of the `examples`. That's what I was thinking, but working on a single `develop-cli` branch works just as well too, so anything is fine. Do let me know if we want to go with the single `develop-cli` branch.

milen-sage:
Ok, so this sounds more like scenario 1 (updated comment above); let's do that.

BrunoGrandePhD:
> I'll create a new branch `develop-config` for the Config related changes, and another branch called `develop-cli` for the CLI related additions.

@sujaypatil96: This quote gave the impression that the two branches would be developed independently. In theory, it should be possible to wrap up the changes to the config without breaking the CLI scripts, and then update the scripts afterwards.

PR #290, opened 2020-09-30
Corrections to HTAN biospecimen schema

milen-sage:
@xdoan the PR contains schema updates that we discussed with Nithya, so that she can review changes in the generated metadata template.


xdoan (approved):
technically works, waiting on nithya to review contents

PR #293, opened 2020-10-01
Add a function to get a subgraph by edge type

milen-sage:
Just realized that I haven't pushed that function yet.

### DIFF
--- a/schematic/schemas/generator.py
+++ b/schematic/schemas/generator.py
@@ -39,6 +39,7 @@ class SchemaGenerator(object):
         Returns:
             None
         """
+
         # create an instance of SchemaExplorer
         self.se = SchemaExplorer()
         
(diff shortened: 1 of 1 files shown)


BrunoGrandePhD:
I'll wait until after we meet on Tuesday to discuss `schematic` before reviewing this PR, if that's okay.

milen-sage:
sounds good

schematic/schemas/generator.py
```diff
@@ -105,6 +106,31 @@ def get_adjacent_nodes_by_relationship(self,
         return sorted(list(nodes))
 
 
+    def get_subgraph_by_edge_type(self,
+                                  graph: nx.MultiDiGraph,
+                                  relationship:str) -> nx.DiGraph:
```
BrunoGrandePhD, line 111: I noticed that `self` isn't used in the function body. Based on future use cases, does it make more sense to use `self.se.schema_nx` (or some other `self` attribute) within the function instead of the `graph` argument or to make this function a standalone utility function? Alternatively, we could drop the `self` argument by making this a [static method](https://docs.python.org/3/library/functions.html#staticmethod) rather than an instance method. I defer to @milen-sage to determine which approach is ideal.

milen-sage: good point @BrunoGrandePhD 

I'm leaning towards dropping the self attribute and going static short-term (we shouldn't assume we always work with the full schema graph; e.g. we could pass a subgraph of the schema). Longer-term the "generator" class will likely need to be refactored; at that point, there will likely be some methods that go into utils.

Issue #292, opened 2020-10-01
scATAC-seq level 1 data curator testing


milen-sage:
- [ ] Check if the generated template and validation in the data curator matches the scATAC-seq RFC requirements.

milen-sage:
@xdoan the scATAC-seq template is ready to go in prod. 

The only people that can further verify the requested metadata works well for their actual data submission are data contributors (and some of them already requested trying that).

@xdoan, could you add scATAC-seq to the prod drop down? (I don't think we need to wait for all recently added data types to be verified, to start adding the templates one by one as we confirm they are ready).

xdoan:
testing generation/validation/errors on dev and will add to prod once it passes

xdoan:
I am seeing columns on the manifest that are not in the the RFC, Attribute Table, or in the Changelog. If they were added after the RFC was closed can you add them to the Changelog? In lieu of having the official data standard the RFC is the place to look up attribute descriptions, and some like, "Nuclei Barcode Length | Nuclei Barcode Read" don't have any descriptions when you hover over them in the spreadsheet and are not in the RFC with a description either. 

https://docs.google.com/spreadsheets/d/13wmIocQtg8bwWXUB_GygUJwQfHBfAliKi7Ln5c964WE/edit#gid=0

https://docs.google.com/spreadsheets/d/1soV03G4-rXiG-XAzbs6CrIvz-n2RJc6jRajyNsGY_Mo/edit#gid=0

milen-sage:
These are added to the manifest because they are required columns specified for the attribute Nucleus Identifier in the RFC (also here: https://docs.google.com/spreadsheets/d/1soV03G4-rXiG-XAzbs6CrIvz-n2RJc6jRajyNsGY_Mo/edit#gid=0&range=12:12)

True, in follow up revisions the RFC editors need to add descriptions for these columns and define them as separate attribtues; for now I assumed most people working with epigenomics know what these columns mean (so the description is the most generic possible...). If we start getting questions I will be reaching out to the co-editors on this RFC.

xdoan:
[...]

PR #294, opened 2020-10-01
Changes to bulk WES to account for ctDNA and cfDNA; corrections to scATA-seq level 1

milen-sage:
and added changes for file format to capture various data formats.


milen-sage:
@xdoan - updated scATAC-seq level 1 - please review/test

xdoan:
commented in #292

xdoan (approved):

PR #299, opened 2020-10-05
Table update fix

milen-sage:
Ensure not adding duplicate rows for existing synapse entities. Addresses https://github.com/sage-bionetworks/htan-coordination/issues/188


xdoan:
works, note that if you update and do not supply an entity ID for an existing object (e.g. clinical folder) then it will create a new one even though the HTAN ID is fixed. Assuming this will be fixed with checks to the HTAN ID? 

Will have to add note to docs than when updating to always get the previously uploaded manifest through the app and edit that since it has entityIds.

xdoan (approved):

Issue #214, opened 2020-05-22
In JIRA: Reflect attribute dependencies logic in manifest column ordering


milen-sage:
Currently columns are ordered based on the schema definition, placing all unconditionally required columns first. However, the order of the conditionally required columns does not reflect their relationship with unconditionally required columns. We can try to improve the column ordering logic to address that. Some thought's needed wrt what would be most user friendly.

BrunoGrandePhD:
Just to confirm we're on the same page, imagine you had the following CSV data model:

```
Attribute            ,Valid Values         ,DependsOn                  ,Required
ClinicalData         ,                     ,"PatientId,TODO"           ,FALSE
PatientId            ,                     ,                           ,TRUE
Diagnosis            ,"Cancer,Healthy"     ,                           ,TRUE
Cancer               ,                     ,"CancerType,DiagnosisDate" ,FALSE
CancerType           ,"Skin,Lung,Colon"    ,                           ,TRUE
Skin                 ,CancerClassification ,                           ,FALSE
CancerClassification ,"Benign,Malignant"   ,                           ,TRUE
DiagnosisDate        ,                     ,                           ,TRUE
Sex                  ,"Female,Male,Other"  ,                           ,TRUE

milen-sage:
> Just to confirm we're on the same page, imagine you had the following CSV model for the `ClinicalData` data type (spaces added for readability). Currently, `schematic` would order the columns in the template manifest as follows: `Diagnosis`, `PatientId`, `Sex`, and `FavoriteColor` (_i.e._ required columns first and then alphabetical order). For simplicity, I'm ignoring the `Filename` and `entityId` special columns in this comment.
> 

The default expected order for this example should be slightly different:
PatientId,Diagnosis,Sex,FavoriteColor
i.e. required attributes for ClinicalData should follow the user-entered order in DependsOn 
@BrunoGrandePhD do you see a different order? If so which branch are you on and could you pull the latest code on for manifest/generator either from develop or main?

> ```
> Attribute            ,Valid Values         ,DependsOn                               ,Required
> ---------------------|---------------------|----------------------------------------|--------
> ClinicalData         ,                     ,"PatientId,Diagnosis,Sex,FavoriteColor" ,FALSE
> Diagnosis            ,"Cancer,Healthy"     ,                                        ,TRUE
> PatientId            ,                     ,                                        ,TRUE
> Sex                  ,"Female,Male,Other"  ,                                        ,TRUE
[...]

PR #310, opened 2020-10-15
Improve get_manifest() method

sujaypatil96:
This PR adds the feature/enhancement as described in [issue #214](https://github.com/Sage-Bionetworks/HTAN-coordination/issues/214).

Test feature by running:
`python3 examples/manifest_generator.py` and supplying different parameters based on the different use-cases.

_Note: Remember to make sure Synapse credentials have been set in `.synapseConfig` file._

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -2,20 +2,17 @@ from __future__ import print_function
 import pickle
 import os.path
 import collections
-
 import pandas as pd
-
-from typing import Any, Dict, Optional, Text, List
-
+from typing import Any, Dict, Optional, Text
 import pygsheets as ps
+import synapseclient
 
 from schematic.schemas.generator import SchemaGenerator
-
-from schematic.utils.google_api_utils import build_credentials, execute_google_api_requests
-
+from schematic.utils.google_api_utils import build_credentials, build_service_account_creds, execute_google_api_requests
 from schematic.utils.config_utils import load_yaml
+from schematic.synapse.store import SynapseStorage
 
-from definitions import CONFIG_PATH, DATA_PATH, CREDS_PATH, TOKEN_PICKLE
+from definitions import CONFIG_PATH, DATA_PATH, CREDS_PATH, TOKEN_PICKLE, SERVICE_ACCT_CREDS, SYNAPSE_CONFIG
 
 config_data = load_yaml(CONFIG_PATH)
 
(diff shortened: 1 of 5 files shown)


schematic/manifest/generator.py
```diff
@@ -589,6 +573,72 @@ def get_manifest(self, json_schema = None):
 
         return manifest_url
 
+    def get_manifest(self, dataset_id: str, sheet_url: bool):
```
milen-sage, line 576: @sujaypatil96 I see you updated functions that previous called get_manifest() as needed to get_empty_manifest or get_manifest(<dataset_id>, <sheet_url>).However, I was wondering, do you think it would make sense to have default argument values and behavior for the new get_manifest(<dataset_id>, <sheet_url>) function: dataset_id = None, sheet_url = True. In this default case, we can just replicate the behavior of the old get_manifest: call get_empty_manifest(). That way we preserve backward compatibility and don't need to change the signature of existing function calls.

sujaypatil96: Yes, we could do that as well. In that case, the `get_manifest()` signature would change to `get_manifest(<dataset_id> = None, <sheet_url> = None, <json_schema> = None)`.

In addition to all the handlers I've written now, we can have a "default" else/elif block which returns `get_empty_manifest()` when either only `<json_schema>` argument is provided or no arguments are provided.

This sounds reasonable, and I think we can implement the behaviour this way. Should I go ahead and issue a new commit that does this @milen-sage?

milen-sage: Yes, all that sounds good. Just a question - why not have sheet_url = True by default?

sujaypatil96: Hm..I'm not sure, I think it's okay to leave `sheet_url = None`. The reason is because if it's defaulting to `True`, then we probably should have a handler for the case when the user sets it to `False`. In our default case, we are just making a call to `get_empty_manifest()`, which in any case returns a URL only.

milen-sage: Ok - sounds good; I will test on my end, but please test on your end as well (it will likely take you less time!) :) 
Reminder: work hours boundaries :) apologies for posting the comment late - I saw yours and decided to reply. Thanks!

sujaypatil96: No worries at all. Yes, I've tested it with different datasets on Synapse (RNAseqAlignment, FamilyHistory, etc. some of which have manifests and some of which don't), and also used different values for the `sheet_url` argument.

I've mentioned how you can test the method in the first PR message, you can just feed different values to the `examples/manifest_generator.py` script, so it should be fairly quick and easy. And when you're ready after the test, feel free to approve the PR and I can merge.

milen-sage:
@sujaypatil96 sorry - could you reopen #310?

PR #313, opened 2020-10-16
Schema updates - imaging level 1 and 2; biospecimen final updates

milen-sage:
This PR adds support for imaging level 1 and level 2 data.

It also addresses https://github.com/sage-bionetworks/htan-coordination/issues/193


milen-sage:
@xdoan, I ended up adding imaging metadata to this PR too. So both final changes to biospecimen and imaging levels 1 and 2 are in this PR schema updates.

Imaging level 1 and 2 schema name is ImagingLevel1and2

milen-sage:
@xdoan latest commit addresses comments here https://github.com/sage-bionetworks/htan-coordination/issues/233

Note: imaging schema name is now ImagingLevel2

xdoan:
is imaging channels in the schema now too?

xdoan (approved):

milen-sage:
Yes, we want to include it in the schema, but (for now at least) not show it in the dropdown of the data curator app. A couple of reasons:
- Dave may validate channels metadata on his end, but we still want to store the information on Synapse as well, which requires it to be part of the data model
- we may want to validate channel metadata on our end as well
- eventually we might be able to support channel metadata submission through the data curator app or backend, in which case we definitely want the channel metadata in our schema

xdoan:
sounds good

PR #323, opened 2020-10-23
Adjusted dropdown generation 

milen-sage:
Accounting for even stricter limits on dropdowns in excel (num characters of items in a dropdown); this resolves 'file format' dropdowns export.


milen-sage:
@xdoan tested a bit on my end and it looks like we are good to go - but might want to take a look on your end too.

xdoan (approved):

PR #326, opened 2020-10-27
Improve get_manifest() method -- `develop`

sujaypatil96:
This PR adds the feature/enhancement as described in [issue #214](https://github.com/Sage-Bionetworks/HTAN-coordination/issues/214).

Test feature by running:
`python3 examples/manifest_generator.py` and supplying different parameters based on the different use-cases.

_Note: Remember to make sure Synapse credentials have been set in `.synapseConfig` file._

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1,8 +1,13 @@
 ---
   synapse:
     master_fileview: 'syn20446927'
+    manifest_folder: 'manifests'
     manifest_filename: 'manifests/synapse_storage_manifest.csv'
     api_creds: 'syn21088684'
+
+  schema:
+    csv: 'csv/HTAN.csv'
+    name: 'HTAN'
 
   model:
     input:
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -2,20 +2,17 @@ from __future__ import print_function
 import pickle
 import os.path
 import collections
-
 import pandas as pd
-
 from typing import Any, Dict, Optional, Text, List
-
 import pygsheets as ps
+import synapseclient
 
 from schematic.schemas.generator import SchemaGenerator
-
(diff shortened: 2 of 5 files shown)


schematic/manifest/generator.py
```diff
@@ -178,27 +175,8 @@ def callback(request_id, response, exception):
         batch.execute()
 
 
-    def _get_valid_values_from_jsonschema_property(self, prop:dict) -> List[str]:
```
milen-sage, line 181: Note: this function needs to stay in develop; similarly there is logic below that needs to be preserved. 

@sujaypatil96 any chance you can apply your changes over the latest code in develop?

sujaypatil96:
@milen-sage I've updated this PR with the latest code from `develop`. I think it's good for another test/review (and merge).

config.yml
```diff
@@ -1,9 +1,14 @@
 ---
   synapse:
     master_fileview: 'syn20446927'
+    manifest_folder: 'manifests'
     manifest_filename: 'manifests/synapse_storage_manifest.csv'
```
milen-sage, line 5: I notice manifest folder got added as a config param; and we have manifest filename which still contains the 'manifests' folder in the path. Is this intentional?

sujaypatil96: `manifest_filename` is being used in `schematic.synapse.store` on [L#83](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/synapse/store.py#L83), to explicitly point to the file in the `manifests` folder. The `manifests` folder (new parameter added to the config) is to point to the folder where the generated manifest (after running `$ python examples/manifest_generator.py`) should sit.

So in summary, the two config parameters are unrelated and we want to keep both of them.

PR #342, opened 2020-11-06
Fixing extra apostrophe added by google sheet API on string values

milen-sage:
String values in data validation dropdowns when the latter are derived from workbook range were prefixed by ' in google API. That is now resolved. 

Addresses point 1 in https://github.com/Sage-Bionetworks/HTAN-coordination/issues/240#issuecomment-722706631

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -240,6 +240,7 @@ class ManifestGenerator(object):
                         "majorDimension":"COLUMNS",
                         "values":[values]
             }
+            
             target_range = 'Sheet2!' + target_col_letter + '2:' + target_col_letter + str(len(values) + 1)
             valid_values = [
                             { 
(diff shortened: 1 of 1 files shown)


allaway:
Re: this issue, I noticed that for some values, like the genotype annotations `+/+` and `+/-`, googles throw errors in the sheet because it interprets them as a formula and does not escape them to `'+/+` and `'+/-`. Do you think this is related to the change you just made? (I've been meaning to file an issue, but keep forgetting).

milen-sage:
Yes - that may be very related; feel free to try out after this PR is merged.
If the '+/-' issue is still unresolved after that, file a bug :)

xdoan:
@milen-sage Do you have suggestions for which template dropdowns to test this with?

milen-sage:
Yes - please use the template that Nithya pointed to in the issue above (I think it was https://github.com/Sage-Bionetworks/HTAN-coordination/issues/240#issuecomment-722706631 and the template is Breast Cancer - but you can confirm.)

xdoan:
tested and the allred col seems to just show integers! https://docs.google.com/spreadsheets/d/1j7f5UWAnM_uq_ZTh4B_QxbGSD-uY1UpNOkWZPypZAcQ/edit#gid=0

xdoan (approved):

Issue #283, opened 2020-09-24
Rename the `Requires` (with an 's') CSV schema column


BrunoGrandePhD:
The motivation for renaming the `Requires` (with an `s`) column is to avoid confusion with the `Required` (with a `d`) column. Quick glances at the documentation can miss that one-letter difference. Alternative names include:

- `DependsOn`
- `Implies`
- `ImpliedAttributes`
- `RelevantAttributes`
- `AssociatedWith`

I think one important factor to consider when picking a new name is that the listed attributes **can** be set if the current entity come up, not that they **must** be set. This is another reason why `Requires` isn't great at conveying this message, and why I suggest changing it rather than `Required`.

milen-sage:
Need to decide whether to 
- drop support for Requires once this is implemented and tested
- leave Requires in the code base, remove from examples, and mark it as obsolete in docs (so that people don't create new schemas using it)
- support Requires providing backward compatibility for existing schemas (so that people don't need to regenerate new schemas)

I lean towards either of the first two options since it's easy enough to regenerate a schema with the renamed field, so no backward compatibility is needed. Option 1 is easier wrt refactor and code maintenance; option 2 seems more user friendly :)

BrunoGrandePhD:
Since it's a single-field fix that users only need to do once per schema floating around (which are limited in number), I vote for option 1 with a meaningful error message if `Requires` is ever detected in the CSV header. The error message can be something along the lines of:

> The input CSV schema file contains the `Requires` column header. To minimize confusion, this column was renamed to `DependsOn`. Please re-run with the new column name.

BrunoGrandePhD:
This will be done in the `bgrande/rename-requires-column` branch.

BrunoGrandePhD:
Solved by #345.

PR #345, opened 2020-11-10
Renaming the 'Requires' column in the CSV schema definition to 'DependsOn'

BrunoGrandePhD:
This will close issue #283. 

Here's an example error when you run the `csv_to_schemaorg.py` script using a CSV file that still has `Requires`:

```
$ python examples/csv_to_schemaorg.py --config ./config.yml data/csv/example.csv

The input CSV schema file contains the 'Requires' column header. To minimize confusion with 'Required', this column was renamed to 'DependsOn'. Please re-run with the new column name.
Schema extension headers: 
{'Validation Rules', 'Description', 'Requires', 'Valid Values', 'Parent', 'Source', 'Required', 'Requires Component', 'Attribute', 'Properties'}
do not match required schema headers: 
{'Validation Rules', 'Description', 'Valid Values', 'Parent', 'DependsOn', 'Source', 'Required', 'Requires Component', 'Attribute', 'Properties'}
ERROR: could not add extension to schema!
```

### DIFF
--- a/data/csv/HIV.csv
+++ b/data/csv/HIV.csv
@@ -1,4 +1,4 @@
-Attribute,Description,Valid Values,Requires,Properties,Required,Parent,Requires Component,Source,Validation Rules
+Attribute,Description,Valid Values,DependsOn,Properties,Required,Parent,DependsOn Component,Source,Validation Rules
 Filename,REQUIRED: Name of file ,,,,TRUE,,,,
 Experiment,,,,,FALSE,,,,
 Experiment Date,REQUIRED: Experiment date. Format: YYYYMMDD.,,,,TRUE,,,,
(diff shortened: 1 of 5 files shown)


BrunoGrandePhD:
@milen-sage: Should I also rename the `Requires Component` column name to `DependsOn Component` for consistency? Or perhaps you have a better name?

milen-sage (approved):

milen-sage:
> @milen-sage: Should I also rename the `Requires Component` column name to `DependsOn Component` for consistency? Or perhaps you have a better name?
> 

Yes, I think that sounds good.

> Also, note to self: I need to also update the other CSV files in the `data/csv/` folder.

Yes, good point I need to adjust my schema inputs as well...

milen-sage (approved):
Looks good!

Issue #344, opened 2020-11-09
Bundle accessory JSON(-LD) files (and others) as package resources


BrunoGrandePhD:
**Is your feature request related to a problem? Please describe.**
The current setup requires that the user downloads/clones the repo in order to use accessible JSON(-LD) files included therein. However, this approach won't work once the package is published on PyPI, unless you force the user to also still download the repo, which would be a terrible UX. Another reason is that addressing this issue would permit the developers to remove a bunch of hard-coded (or needlessly configurable) file paths from the package. 

**Describe the solution you'd like**
In preparation for an eventual publication on PyPI, I propose that we use package resources as a means to bundle or "ship" accessory JSON(-LD) files with the module. Files that never change should be bundled with the package, _e.g._ using [`pkg_resources`](https://kiwidamien.github.io/making-a-python-package-vi-including-data-files.html). Files that should or might be changed should remains user-configurable in the configuration file (as achieved in #295). The lists below is compiled based on discussions [on Slack](https://sagebionetworks.slack.com/archives/C01ANC02U59/p1601076059077900). 

  - **Bundled files**
    - [ ] class_json_schema.json
    - [ ] property_json_schema.json
    - [ ] schema.json
    - [ ] biothings.jsonld
    - [ ] all_layer.jsonld
  - **User-configurable files**
    - [ ] biothings.jsonld
    - [ ] all_layer.jsonld

**How important is this feature?** Select from the options below:
• Low - it's an enhancement but not crucial for work (PyPI deployment is still some time away, so this isn't urgent.)

**When will use cases depending on this become relevant?** Select from the options below:
• Mid-term - 2-4 months (I like to think that deployment to PyPI will happen in the next few months.)

BrunoGrandePhD:
I suggest creating a `etc/` subdirectory within the `schematic/` subdirectory.

BrunoGrandePhD:
Consider renaming the bundled files to use more intuitive names. For example, `schema.json` is fairly generic.

milen-sage:
I agree with the proposed solution above. Let's add example.jsonld schema to the bundled jsonld files - that would enable users to play around with the package immediately after installation. We can add more examples of schemas for different use cases later as well.

sujaypatil96:
Addressed in PR #347.

Issue #328, opened 2020-10-27
CLI scripts architecture discussion


sujaypatil96:
Issue to discuss where the CLI scripts will fit in, given our current directory/folder structure of our package.

Overall objective: Identify atomic functionalities, i.e., the functionalities should make sense as a "unit" by itself, and add as a script to the CLI module.

Example of "atomic functionality": `$ python cli/generate_manifest.py pos_args --opt_args`

The above command generates a manifest based on supplied arguments.

_Structure of the package_:
```
data/
docs/
cli/ -- new cli folder to be added here with atomic functionalities
examples/ (consider renaming to scripts) -- also CLI based, but still needs improvement
schematic/
tests/
.
.
.
```
List of "atomic functionalities" to be added to cli (list to be expanded):

| Core Module      | Module Functionality | CLI Utility |
| ----------- | ----------- | ----------- |
| Metadata Model      | getModelManifest       | get_manifest       |
| Metadata Model   | validateModelManifest        | validate_manifest       |
| Metadata Model   | populateModelManifest        | populate_manifest       |
|   |         |        |
| Manifest Generator   | get_manifest        | get_manifest       |
|   |         |        |
| Schema Explorer   | explore_class        | explore_class       |
| Schema Explorer   | explore_property        | explore_property       |
| Schema Explorer   | edit_class        | edit_class       |
| Schema Explorer   | update_class        | update_class       |
| Schema Explorer   | edit_property        | edit_property       |
| Schema Explorer   | update_property        | update_property       |
|   |         |        |
| Schema Generator   | get_json_schema_requirements        | get_validation_schema       |

Note:
- `get_manifest` (is the same in both the _Manifest Generator_ as well as _Metadata Model_) will make a call to the generic `get_manifest` method (found in _Manifest Generator_).

One of the projects I've been referring to is Sage's project `Genie`. Looking at the [cli implementation](https://github.com/Sage-Bionetworks/Genie/blob/master/genie/__main__.py) there, we could also think of implementing it within `__main__.py` in `schematic/` and indicating that as an "entry point" to the application.

What are your thoughts?

sujaypatil96:
Results of discussion:

1. Some of the scripts will be moved from the `examples/` folder (in their current state) to the `cli/` folder. Some others will need modification(s) before moving.

[...]

PR #347, opened 2020-11-10
package resources (pkg_resources) integration

sujaypatil96:
Based on @BrunoGrandePhD's recommendation(s) in issue #344, I've adopted the use of `pkg_resources` to bundle JSON(-LD) files that are specific to the package.

Do let me know if you have better names in mind for any of the files in the `etc/` folder @milen-sage, and I can push another commit if need be.

Else, I've tested it on my end and all looks well. I would recommend running the `manifest_generator.py` example to test it out. Once this is done, issue #328 is the next logical issue to tackle.


schematic/loader.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __str__(self):
+        return '{}({}), "{}" of {}'.format(self.message, self.errno, self.requested_uri, self.namespace)
+
+
+    def __repr__(self):
+        return self.__str__()
+
+        
+class Loader(object):
```
BrunoGrandePhD, line 32: Where was this `Loader` pattern taken from? This seems like a lot of code for just loading files from within the package.

setup.py
```diff
@@ -25,12 +25,16 @@
     author_email='milen.nikolov@sagebase.org',
     url='https://github.com/Sage-Bionetworks/HTAN-data-pipeline/tree/develop',
     packages=find_packages(include=['schematic', 'schematic.*']),
-    python_requires='>=3.0',
+    python_requires='>=3.6',
```
BrunoGrandePhD, line 28: YAY! We can use f-strings!

schematic/utils/validate_utils.py
```diff
@@ -3,26 +3,33 @@
 from jsonschema import validate
 
 from schematic import CONFIG
+from schematic.loader import Loader, InvalidResourceError
+
+# call Loader() and pass `schematic`, which is the global package namespace
+loader = Loader('schematic', prefix='etc')
```
BrunoGrandePhD, line 9: It seems like you are creating a loader for every submodule. I'm still unsure why we need a loader (as opposed to a single function, like the one below), but if this pattern is useful, then you might want to consider doing like what I did for `CONFIG` and instantiate one loader as a module-level constant. 

```
def get_bundled_file(filename):
    return pkg_resources.resource_stream(__name__, f'etc/{filename}')
```

BrunoGrandePhD (commented):
I left a few general questions as comments on the code. 

Otherwise, I think we can still improve the naming and organization of these accessory files. For example, could the `validation` folder be renamed to `validation_schemas`? I like how the files within the `validation` folder explicitly mention that they're JSON Schemas. I also still don't understand what `all_layer.jsonld` is supposed to convey. Instead of using `schema_org` as the folder name, what would better describe the kind of schemas in this folder? Something like `validation_schemas`.

schematic/loader.py
sujaypatil96: It's not a standard pattern, it was inspired by an example I found which showed the use of `setuptools` / `pkg_resources`. So `Loader()` is a library, there are many ways to load package resources -- `resource_string`, `resource_stream`, etc. so there are wrappers for each of those methods. If someone requires say, loading all the contents from a package resource in the future then they can just use `resource_string` wrapper (as `Loader.read()`) rather than relying on `Loader.filename()` (which uses `resource_stream()`).

schematic/utils/validate_utils.py
sujaypatil96: Awesome, addressed this in a new commit.

Reasoning for adopting `Loader()` pattern detailed in above comment.

setup.py
```diff
@@ -18,7 +18,7 @@
 setup(
     name='schematic',
     version=version['__version__'],
-    description='Packages responsible for data ingress from HTAN collaborators',
+    description='Package for Data Ingress from HTAN Collaborators',
```
BrunoGrandePhD, line 21: Should we consider removing the reference to HTAN here? Maybe something like:

> Tools for Managing (Meta)Data Models and Validation Schemas for Data Ingress

Although, I don't know if the "for Data Ingress" qualifier is even required here.

BrunoGrandePhD: @milen-sage: You should probably have a say in describing your own package. 

Speaking of authorship, @sujaypatil96, there's probably a way of listing contributors alongside authors in the `setup.py` file.

milen-sage: I agree if possible, we should add contributors. A note to self, we should add folks from Scripps as contributors.

I can add a description, but can worry about that later :)

milen-sage: Could we change the description, even if preliminary, to:
"Package for data model and data ingress management"

Mainly not constrained to HTAN.

BrunoGrandePhD (changes requested):
I don't know if it's GitHub's PR interface, but I'm still seeing files ending with `_json_schema.json`, which should end with `.schema.json`. 

```
schematic/etc/validation_schemas/class_json_schema.json
schematic/etc/validation_schemas/property_json_schema.json
```

Also, do you mind adding a quick README for the `etc/` folder in this PR since it consolidates the files there? Thanks!

sujaypatil96:
Ah, my bad. I'm not sure why the class and property validation filenames hadn't changed. I've resolved it now @BrunoGrandePhD. 

I've also added a preliminary README.

config.yml
```diff
@@ -17,8 +17,6 @@ model:
     file_type: 'local'
     validation_schema: 'data/validation_schemas/HTAN_validation_schema_FamilyHistory.json'
     log_location: 'data/json_schema_logs/json_schema_log.json'
-  biothings:
```
BrunoGrandePhD, line 20: Good catch!

schematic/utils/io_utils.py
```diff
@@ -1,8 +1,11 @@
 import os
 import json
 import urllib.request
+import pkg_resources
+from os import path, pathsep
+from errno import ENOENT
```
BrunoGrandePhD, line 6: Do we need to load these packages here? Don't they just need do be loaded in the loader.py submodule?

sujaypatil96: Yup, you're exactly right. I left them there when I had the Loader() being called in every module. I've removed the imports now.

BrunoGrandePhD (approved):

Issue #317, opened 2020-10-20
Logging to stderr


milen-sage:
Currently schematic outputs all errors, warning and informational messages to stdout.

This is not suitable for debugging w/in the context of shiny apps (there, only logs of stderr are stored) as we found out over the last couple of days with @xdoan; but is relevant to other contexts as well (server environments in general).

How about replacing all print statements with calls to a logging library handlers so that we have flexibility to print both to stdout, stderr and log files, at potentially different levels of verbosity (the latter is lower priority).

@sujaypatil96 could you take a look at how the synapse client currently handles that and see if we can have similar functionality in schematic?  https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/core/logging_setup.py

sujaypatil96:
PR #406 resolves this issue.

PR #353, opened 2020-11-18
added submit_metadata_manifest() to metadata.py module

sujaypatil96:
Wrap methods that are responsible for _validation_ of manifests for a given component, and _association_ of the same manifest file with a specified dataset, into a single method called `submit_metadata_manifest()`. We will need this method to expose through the CLI (and eventually the REST API).

### DIFF
--- a/schematic/models/metadata.py
+++ b/schematic/models/metadata.py
@@ -15,7 +15,6 @@ from schematic.schemas.explorer import SchemaExplorer
 from schematic.manifest.generator import ManifestGenerator
 from schematic.schemas.generator import SchemaGenerator
 from schematic.synapse.store import SynapseStorage
-from schematic import CONFIG
 
 class MetadataModel(object):
     """Metadata model wrapper around schema.org specification graph.
@@ -237,28 +236,20 @@ class MetadataModel(object):
         return mg.populate_manifest_spreadsheet(manifestPath, emptyManifestURL)
 
 
-    def submit_metadata_manifest(self, manifest_path: str, dataset_id: str, validate_component: str = None, json_schema: str = None) -> bool:
+    def submit_metadata_manifest(self, manifest_path: str, dataset_id: str, validate_component: str = None) -> bool:
         """Wrap methods that are responsible for validation of manifests for a given component, and association of the 
         same manifest file with a specified dataset.
-
         Args:
             manifest_path: Path to the manifest file, which contains the metadata.
             dataset_id: Synapse ID of the dataset on Synapse containing the metadata manifest file.
             validate_component: Component from the schema.org schema based on which the manifest template has been generated.
-
         Returns:
             True: If both validation and association were successful.
-            False: If the function could not successfully validate (and hence associate) the manifest file with the dataset.
-
         Exceptions:
             ValueError: When validate_component is provided, but it cannot be found in the schema.
-            NameError: When validate_component is not provided, but json_schema is.
+            ValidationError: If validation against data model was not successful.
         """
-        # create object of SynapseStorage() class to associate manifest file with dataset
-        syn = synapseclient.Synapse(configPath=CONFIG.SYNAPSE_CONFIG_PATH)
-        syn.login()
-
-        syn_store = SynapseStorage(syn=syn)
+        syn_store = SynapseStorage()
 
         # check if user wants to perform validation or not
         if validate_component is not None:
@@ -273,16 +264,8 @@ class MetadataModel(object):
                 raise ValueError("The component {} could not be found "
                                  "in the schema.".format(validate_component))
 
-            # list of all the validation errors present in the manifest file
-            val_errors = []
-
-            # JSON schema need not be provided always, it is an optional argument
-            # if JSON schema is provided, then use it
-            if json_schema:
-                val_errors = self.validateModelManifest(manifestPath=manifest_path, rootNode=validate_component, jsonSchema=json_schema)
-            else:
-                # automatic JSON schema generation and validation with that JSON schema
-                val_errors = self.validateModelManifest(manifestPath=manifest_path, rootNode=validate_component)
+            # automatic JSON schema generation and validation with that JSON schema
+            val_errors = self.validateModelManifest(manifestPath=manifest_path, rootNode=validate_component)
 
             # if there are no errors in validation process
             if not val_errors:
@@ -294,12 +277,7 @@ class MetadataModel(object):
                 return True
             else:
                 print(val_errors)
-                return False
-
-        # handle the case-scenario when only `json_schema` argument is provided but not `validate_component`
-        if validate_component is None and json_schema is not None:
-            raise NameError("For `json_schema argument to be accepted, the "
-                            "`validate_component` argument must be specified too.")
+                raise ValidationError("Manifest could not be validated under provided data model.")
 
         # no need to perform validation, just submit/associate the metadata manifest file
         syn_store.associateMetadataWithFiles(metadataManifestPath=manifest_path, datasetId=dataset_id)
--- a/schematic/models/metadata.py
+++ b/schematic/models/metadata.py
@@ -201,10 +201,8 @@ class MetadataModel(object):
             # annotations manifest to a list and strip each value from leading/trailing spaces
             if "list" in self.sg.get_node_validation_rules(col): 
                 manifest[col] = manifest[col].apply(lambda x: [s.strip() for s in str(x).split(",")])
-                print(manifest[col])
 
         annotations = json.loads(manifest.to_json(orient='records'))
-        print(annotations) 
         for i, annotation in enumerate(annotations):
             v = Draft7Validator(jsonSchema)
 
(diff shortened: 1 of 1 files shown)


schematic/models/metadata.py
```diff
@@ -235,3 +237,45 @@ def populateModelManifest(self, title, manifestPath: str, rootNode: str) -> str:
...
+        # list of all the validation errors present in the manifest file
+        val_errors = []
+
+        # JSON schema need not be provided always, it is an optional argument
+        # if JSON schema is provided, then use it
+        if json_schema:
+            val_errors = self.validateModelManifest(manifestPath=manifest_path, rootNode=validate_component, jsonSchema=json_schema)
+
+        # automatic JSON schema generation and validation with that JSON schema
+        val_errors = self.validateModelManifest(manifestPath=manifest_path, rootNode=validate_component)
```
milen-sage, line 264: Perhaps put in an else branch

schematic/models/metadata.py
```diff
@@ -235,3 +237,45 @@ def populateModelManifest(self, title, manifestPath: str, rootNode: str) -> str:
...
+
+        # create object of SynapseStorage() class to associate manifest file with dataset
+        syn = synapseclient.Synapse(configPath=CONFIG.SYNAPSE_CONFIG_PATH)
+        syn.login()
+
+        syn_store = SynapseStorage(syn=syn)
+
```
milen-sage, line 271: We should provide an option for the user to indicate whether they require validation of metadata or just want to submit metadata w/o validation (in the latter case validate_component can be set to None).

schematic/models/metadata.py
```diff
@@ -235,3 +237,45 @@ def populateModelManifest(self, title, manifestPath: str, rootNode: str) -> str:
         emptyManifestURL = mg.get_manifest()
 
         return mg.populate_manifest_spreadsheet(manifestPath, emptyManifestURL)
+
+
+    def submit_metadata_manifest(self, manifest_path: str, dataset_id: str, validate_component: str, json_schema: str = None) -> bool:
```
milen-sage, line 242: if validate_component is None and json_schema is None, then just store metadata manifest to the dataset folder.

schematic/models/metadata.py
```diff
@@ -235,3 +237,45 @@ def populateModelManifest(self, title, manifestPath: str, rootNode: str) -> str:
...
+        """Wrap methods that are responsible for validation of manifests for a given component, and association of the
```
milen-sage, line 243: If validate_component is specified and the component exists in the schema, then do automatic JSON schema generation and validation. If the component doesn't exist then raise an exception.

schematic/models/metadata.py
```diff
@@ -235,3 +237,45 @@ def populateModelManifest(self, title, manifestPath: str, rootNode: str) -> str:
         emptyManifestURL = mg.get_manifest()
 
         return mg.populate_manifest_spreadsheet(manifestPath, emptyManifestURL)
+
+
+    def submit_metadata_manifest(self, manifest_path: str, dataset_id: str, validate_component: str, json_schema: str = None) -> bool:
+        """Wrap methods that are responsible for validation of manifests for a given component, and association of the 
+        same manifest file with a specified dataset.
```
milen-sage, line 244: If if validate_component is None and the jsonschema is provided, then validate using the provided jsonschema.

milen-sage (changes requested):

schematic/models/metadata.py
```diff
@@ -235,3 +237,45 @@ def populateModelManifest(self, title, manifestPath: str, rootNode: str) -> str:
...
+        # JSON schema need not be provided always, it is an optional argument
+        # if JSON schema is provided, then use it
+        if json_schema:
+            val_errors = self.validateModelManifest(manifestPath=manifest_path, rootNode=validate_component, jsonSchema=json_schema)
+
+        # automatic JSON schema generation and validation with that JSON schema
+        val_errors = self.validateModelManifest(manifestPath=manifest_path, rootNode=validate_component)
+
+        # create object of SynapseStorage() class to associate manifest file with dataset
+        syn = synapseclient.Synapse(configPath=CONFIG.SYNAPSE_CONFIG_PATH)
```
BrunoGrandePhD, line 267: Should we allow Synapse-specific code to (further?) creep outside of `synapse/store.py`? It might be worth considering how to tweak this function's  signature (and/or internals) to better accommodate different asset stores.

BrunoGrandePhD (approved):
This PR is harder for me to review because I'm still familiarizing with the mechanics of `schematic`. I defer to @milen-sage, who has already given this a thorough review. 

🚀

schematic/models/metadata.py
sujaypatil96: See [L#285](https://github.com/Sage-Bionetworks/schematic/blob/develop-submit_metadata_manifest/schematic/models/metadata.py#L285).

schematic/models/metadata.py
sujaypatil96: See method prototype on [L#240](https://github.com/Sage-Bionetworks/schematic/blob/develop-submit_metadata_manifest/schematic/models/metadata.py#L240) now.

schematic/models/metadata.py
sujaypatil96: In the case-scenario where `validate_component` and `json_schema` both are `None`, the manifest file is just pushed to the dataset folder. Handler on [L#305](https://github.com/Sage-Bionetworks/schematic/blob/develop-submit_metadata_manifest/schematic/models/metadata.py#L305).

schematic/models/metadata.py
sujaypatil96: [L#268](https://github.com/Sage-Bionetworks/schematic/blob/develop-submit_metadata_manifest/schematic/models/metadata.py#L268) checks that `validate_component` exists in the schema. If not, it raises a `ValueError` exception.

schematic/models/metadata.py
sujaypatil96: My thoughts on this are that if `validate_component` is not specified, then the user doesn't want us to validate the manifest at all, which would mean no JSON Schema generation (and validation). So I thought it would be a good idea to handle it like on [L#300](https://github.com/Sage-Bionetworks/schematic/blob/develop-submit_metadata_manifest/schematic/models/metadata.py#L300). If we still want to do JSON Schema validation I can change the code.

milen-sage: I was thinking that we can provide the option for the user to give us their own jsonschema (e.g. could be outside of the data model) and running the validator on the related manifest. But I agree this is a use case that can be deprioritized, especially given development in Synapse supporting jsconschema validation. So, I'd agree with your approach; in this case do we still even want to expose the jsonschema as an argument option (i.e. only the validate_component is relevant)?

schematic/models/metadata.py
milen-sage: I agree with @BrunoGrandePhD here - can we implement everything related to synapse in a synapse/store method and just call it here, using the store module as an interface?

schematic/models/metadata.py
sujaypatil96: Right, so the `validateModelManifest()` method expects both `rootNode` (`validate_component`) and (`jsonSchema) `json_schema`. So if we want to provide the option of validating without `rootNode`, I'll need to change the signature of `validateModelManifest()` and `get_json_schema_requirements()` (which is not too bad). But you mentioned that use-case can be de-prioritized, so we'll ignore that for now.

So the one case-scenario in which exposing the `json_schema` argument becomes useful is when the user provides a `rootNode` / `validate_component`, but wants custom JSON Schema validation rules. Do we want that flexibility? If not, I can deprecate the use of `json_schema`.

schematic/models/metadata.py
sujaypatil96: Right, so we're creating an instance of a `synapseclient.Synapse()` object, using it to login and passing it to an instance of `SynapseStorage`. Since we're not "tweaking"/modifying any `SynapseStorage` class functionalities, I'm not sure it fits in `synapse/store.py`? Perhaps a `utils/synapse_utils.py`?

milen-sage: Yes, let's move these two lines of code in the else-part of the SynapseStorage class constructor (we actually had that pathway/option in synapse/store module when working on schematic initially).

schematic/models/metadata.py
milen-sage: Yes, we don't want to support scenarios (at least at present) where the user provides a rootNode but provides custom json_schema; so let's go ahead and deprecate the use of json_schema. If needed we can introduce the option in validate_manifest later.

sujaypatil96: The `json_schema` argument has been deprecated in this [commit](https://github.com/Sage-Bionetworks/schematic/pull/353/commits/7aa1fd021a194a767f77271baa68d61cceaf4524).

schematic/models/metadata.py
```diff
@@ -2,7 +2,7 @@
 import networkx as nx
 import json
 import re
-
+import synapseclient
```
milen-sage, line 5: looks like we no longer need synapseclient import?

sujaypatil96: Yup, good catch, you're right. Removed `synapseclient` and `re` imports, both of which weren't being used.

milen-sage (approved):
Looks good! 

Need to prioritize moving to logger instead of using print statements... #317
(but for now let's proceed with the first version of CLI).

The more we accumulate print statements when catching exceptions, the more risk we are running with downstream systems not recording stdout to logs (e.g. shiny server's logs store stderr) and us missing spotting bugs.

sujaypatil96:
Yup, I'll take up issue #317 and move print statements used in every module to `logging` (logger) once the CLI implementation for this is complete.

Issue #355, opened 2020-11-30
Bug in metadata manifest association because of out-of-sync Synapse table


sujaypatil96:
**Describe the bug**
When there are updates to the JSON-LD schema (due to changes made in the CSV file that generates it), for example, a class that had some name previously now has a different name. One example, is the latest schema update changed the name of a column, `HTAN Patient ID` to `HTAN Participant ID`, and that has affected many methods downstream such as the `associateMetadataWithFiles` and `update_df` functions/methods. The problem in the code, when `associateMetadataWithFiles` tries to associate a metadata manifest file with a Synapse folder is that it checks to see if the Synapse table is in sync with the manifest file.

**To Reproduce**
1. Run `export SCHEMATIC_CONFIG='path_to_config_yml'`
2. Use this piece of driver code just for testing purposes:

```python
syn_store.associateMetadataWithFiles(metadataManifestPath=manifest_path, datasetId=dataset_id)
```

This error was realized when testing the `submit_metadata_manifest` method which is still in review in a PR. We can't move forward with that PR until this bug has been resolved.

3. However, if you don't want to test it with `submit_metadata_manifest` simply run `associateMetadataWithFiles` with appropriate arguments.

**Expected behavior**
Based on our discussion, it was decided that since we don't see any immediate use-case for the Synapse table, we can retire/deprecate code that is dealing with/handling Synapse table updates in the `associateMetadataWithFiles` function/method.

sujaypatil96:
Closing this issue since it was handled in PR #358.

PR #356, opened 2020-12-01
Bugfix in associateMetadataWithFiles()

sujaypatil96:
Per issue #355, there was some discrepancy in the updation of Synapse table from the newly updated JSON-LD schema. So I've resolved the issue by adding a method called `update_table_schema()`. This method takes an argument called `col_mapper`. The way `col_mapper` is represented generally is `{OLD_COLUMN_NAME: {NEW_COLUMN_NAME, COLUMN_TYPE}}`.

For example, there's a change in the newer versions of the HTAN schema where `HTAN Patient ID` has been changed to `HTAN Participant ID`. The way you would pass that to `associateMetadataWithFiles()` is this: `associateMetadataWithFiles("./data/manifests/synapse_storage_manifest.csv", "syn21973647", col_mapper={'HTAN Patient ID': ('HTAN Participant ID', 'INTEGER')})`

Note: Pass the `col_mapper` argument only when there is a clash in column names with the Synapse table. If the table columns match the manifest file you have locally, then there's no need to pass a mapper.

_Please ignore the changes that you see in `metadata.py`._

### DIFF
--- a/schematic/models/metadata.py
+++ b/schematic/models/metadata.py
@@ -201,10 +201,8 @@ class MetadataModel(object):
             # annotations manifest to a list and strip each value from leading/trailing spaces
             if "list" in self.sg.get_node_validation_rules(col): 
                 manifest[col] = manifest[col].apply(lambda x: [s.strip() for s in str(x).split(",")])
-                print(manifest[col])
 
         annotations = json.loads(manifest.to_json(orient='records'))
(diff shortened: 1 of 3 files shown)


sujaypatil96:
Deprecating Syanpse table functionalities to replace it with Synapse fileviews.

PR #358, opened 2020-12-03
Remove code handling synapse table logic

sujaypatil96:
Based on our discussion to resolve issue/bug #355, it was decided that we would deprecate the code that dealt with updating Synapse table. More details in the issue.

### DIFF
--- a/schematic/synapse/store.py
+++ b/schematic/synapse/store.py
@@ -392,8 +392,6 @@ class SynapseStorage(object):
         # check if there is an existing manifest
         existingManifest = self.getDatasetManifest(datasetId)
 
-        existingTableId = None
-
         if existingManifest:
 
             # update the existing manifest, so that existing entities get updated metadata and new entities are preserved;
@@ -402,9 +400,6 @@ class SynapseStorage(object):
             # (it is ok if the entities ID in the new manifest are blank)
             manifest['entityId'].fillna('', inplace = True)
             manifest = update_df(manifest, existingManifest, "entityId")
-
-            # retrieve Synapse table associated with this manifest, so that it can be updated below
-            existingTableId = self.syn.findEntityId(datasetName + "_table", datasetParentProject)
     
         # if this is a new manifest there could be no Synapse entities associated with the rows of this manifest
         # this may be due to data type (e.g. clinical data) being tabular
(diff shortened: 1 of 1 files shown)


BrunoGrandePhD (commented):
@milen-sage: Do we want to wait before implementing the auto-creation of file views? I'm assuming that there aren't many users who are clamoring for query-able tables right this instant, but I could be wrong. We would also need to figure out the auto-creation of pseudo-files/folders for clinical data too. 

IHMO, I'm happy to merge this before tackling the file views, which in the interim can always be created manually if need be.

sujaypatil96:
@BrunoGrandePhD: The reason why this needs to be merged into `develop` (with or without the fileview creation logic) is because it affects calls to `associateMetadataWithFiles()` (and other methods, like `submit_metadata_manifest()` which call `associateMetadataWithFiles()`). We wanted to get the `submit_metadata_manifest()` into `develop` so we can start writing the CLI (second method to be exposed via CLI).

BrunoGrandePhD (approved):

BrunoGrandePhD:
Sounds good! 🚀

milen-sage:
I don't think we need to wait for automatic fileview creation before merging this and hopefully we'll get to generate fileviews soon :) + the app in use right now pulls from main which still has the table logic. 

@sujaypatil96 I see your point about having the associateMetadataWithFiles stable before exposing it through CLI, but just a note that having the table logic in the method or not having it there shouldn't affect the CLI call - tables were intended to be used downstream.

I see @BrunoGrandePhD's point about taking care of patient/biosample data - we currently create folders to account for these entities, and apply annotations to these folders, so a fileview accounts for these entities and their annotations, if they are in the fileview scope. In the very near future we'd have to add a normalized view of patients/biosamples types of entities (e.g. so that we don't create multiple folders describing the same patient; right now we create a separate entity for each patient or biosample record). That though has been the case with or without tables.

Another minor issue to consider is that Synapse has limitations on the strings used as keys for annotations (that column headers in tables don't); e.g. no spaces in names - so annotations currently use schema objects' labels and not their display names (whereas the table headers accepted spaces)...  but that's a discrepancy between annotations and tables we always had to deal with.

Issue #312, opened 2020-10-14
Pre-populate manifest template with existing Synapse annotations (only applies if manifest is missing)


allaway:
I think that populating the manifest with annotations that exist on the files already (for keys that are in the manifest, and for valid values) could be useful in a few use cases: 

1) user is annotating old files that may have partial/noncompliant annotations (this is our biggest use case at present)
2) user inadvertently annotates files via a different mechanism with some annotations such as syncToSynapse
3) user programmatically annotates data and wants to do final touch up/validation/manual annotation in app
4) schema changes in the future, and user or data curator wants to update annotations to make compliant with schema changes

BrunoGrandePhD:
> user is annotating old files that may have partial/noncompliant annotations (this is our biggest use case at present)

Given that this is the biggest use case at the moment, the first version of this feature could exist as a toggle—as @allaway suggested—that would cause `schematic` to build a "draft manifest" from all annotations of the corresponding files rather than considering any existing CSV manifest in the folder in question. 

At the implementation level, this "draft manifest" would be exported to a CSV file and re-imported using `manifest.generator.populate_manifest_spreadsheet()` as if it was the CSV manifest from the root of the folder. The function that would pull the annotations from Synapse could be stored in a file alongside the `synapse/store.py` submodule (e.g._ `synapse/get.py` or `synapse/retrieve.py`). The JSON-LD schema would be used to map the Synapse annotation key names (`@id`) to the display name that usually appears in the CSV manifest (`sms:displayName`). 

Outstanding questions:

- If there a value exists for a given file in both the existing CSV manifest and Synapse annotation, which one should take precedence? While this conflict resolution might not be included in the first version of this feature, I'm curious to know to what others would expect. 
- Should we subset the columns in the "draft manifest" to those appearing in the schema? Based on a quick test, it appears that additional columns aren't flagged as invalid, but they're silently dropped when uploaded to Synapse. 

@milen-sage: Thoughts?

milen-sage:
> > user is annotating old files that may have partial/noncompliant annotations (this is our biggest use case at present)
> 
> Given that this is the biggest use case at the moment, the first version of this feature could exist as a toggle—as @allaway suggested—that would cause `schematic` to build a "draft manifest" from all annotations of the corresponding files rather than considering any existing CSV manifest in the folder in question.

Yes - that's a neat idea! Assuming the toggle would be reflected in the the corresponding CLI/API calls (e.g. the modules you suggest below) as well

[...]

PR #361, opened 2020-12-07
Synchronize Synapse annotations (PR 1)

BrunoGrandePhD:
----

### **Attention!** Review and merge #396 first. 

----

This PR aims to address #312. In general, this would improve `schematic`'s handling of Synapse annotations in a few ways. Because of the amount of work involved, I think it's best to incorporate these features in separate PRs to ease with code review. Here's how I propose to break down the PRs: 

## Pull Request 1

Leverage existing Synapse annotations when generating manifest templates.

- Generate a temporary "annotations CSV file" from existing dataset annotations upon template creation. 
- If a CSV manifest doesn't exist, pre-populate the manifest template using this "annotations CSV file", ensuring that any missing attributes (_i.e._ model attributes not among annotations) and extra annotations (_i.e._ annotations not among model attributes) are included in the template.
- If a CSV manifest already exists, compare it with the "annotations CSV file" using `csv-diff` and warn user about any discrepancies affecting model attributes. If this occurs, include additional columns in manifest template containing the annotation values for attributes affected by discrepancies. The annotation column names would have to be suffixed with something like `FromAnnotation` to avoid clashes in the CSV header. Include additional annotations (_i.e._ whose names don't appear in the data model) in the template. 

## Pull Request 2

Prevent potential concurrency issues when updating Synapse annotations. 

- Include etags in the original manifest template. 
- Compare etags between the moment the template was generated and the moment the manifest is submitted. If any are different, warn the user with the list of Synapse IDs of the affected entities so they can explore why. That said, they can decide to proceed with the update anyways, overwriting any changes to the annotations. 
- Before updating annotations, download an "annotations CSV file" in case an error arises and the annotations need to be rolled back. This "annotations CSV file" could be temporarily stored on Synapse in case the server crashes (perhaps under a user-specific filename to prevent clashes).
- Update annotations using `syncToSynapse()` instead of updating Synapse entities individually. 

## Pull Request 3

Handle deleted files.

- If a CSV manifest exists, it can be subset to existing files before anything else (instead of assuming that all files still exist).


sujaypatil96:
@BrunoGrandePhD: This draft PR (soon to be ready) seeks to merge code into `main`. I think you might want to close this and create another draft PR that merges code into `develop`.

BrunoGrandePhD:
@sujaypatil96: I appreciate the heads-up! I noticed the error after opening the PR (since `main` is the default branch on the repo). However, I can't change the base/target branch of the PR until I make a commit since this branch is in sync with `develop`. I'll fix it then!

allaway:
Hey all, just wanted to follow up on this issue. I think this feature would be very helpful for us in the next month - just wanted to check on what your all's plans are (cc @kthrog ). Thanks!

BrunoGrandePhD:
Thanks for following up! This kept being bumped down my to-do list so I could address blockers in other projects, but things have settled down a bit. I'm setting aside eight hours next week to get this started. I'll aim to have it done by February 5th at the latest, but I'll hopefully open this PR for review by the end of next week.

BrunoGrandePhD:
Since it had been a while, I reviewed the discussion in #312 and updated the first comment on this PR to reflect the changes I'll be implementing. Given the scope of the work, I propose splitting the changes across multiple PRs. 

This first PR will tackle the original feature request of pulling Synapse annotations when generating manifest templates and handling any discrepancies. I invite @milen-sage and @allaway to review my suggested implementation (copied below for convenience). Thanks! 

> ## Pull Request 1
> Leverage existing Synapse annotations when generating manifest templates.
> 
> * Generate a temporary "annotations CSV file" from existing dataset annotations upon template creation.
> * If a CSV manifest doesn't exist, pre-populate the manifest template using this "annotations CSV file", ensuring that any missing attributes (_i.e._ model attributes not among annotations) and extra annotations (_i.e._ annotations not among model attributes) are included in the template.
> * If a CSV manifest already exists, compare it with the "annotations CSV file" using `csv-diff` and warn user about any discrepancies affecting model attributes. If this occurs, include additional columns in manifest template containing the annotation values for attributes affected by discrepancies. The annotation column names would have to be suffixed with something like `FromAnnotation` to avoid clashes in the CSV header. Include additional annotations (_i.e._ whose names don't appear in the data model) in the template.

allaway:
>  Generate a temporary "annotations CSV file" from existing dataset annotations upon template creation.
> If a CSV manifest doesn't exist, pre-populate the manifest template using this "annotations CSV file", ensuring that any missing attributes (i.e. model attributes not among annotations) and extra annotations (i.e. annotations not among model attributes) are included in the template.
>    If a CSV manifest already exists, compare it with the "annotations CSV file" using csv-diff and warn user about any discrepancies affecting model attributes. If this occurs, include additional columns in manifest template containing the annotation values for attributes affected by discrepancies. The annotation column names would have to be suffixed with something like FromAnnotation to avoid clashes in the CSV header. Include additional annotations (i.e. whose names don't appear in the data model) in the template.

I think I am not quite following - Rather than retrieving from a CSV manifest (which may represent be the latest set of annotations), shouldn't schematic retrieve annotations from a fileview? (or perhaps when you say CSV manifest, are you referring to a csv exported from a view?)

BrunoGrandePhD:
I don't think I laid it out very clearly. Here's the ordered sequence of events that I envision will happen every time a manifest template is generated for a given dataset: 

1. Generate a CSV file from the annotations of all files in the given dataset. 
    - This is what I referred to as the "annotations CSV file". This is unrelated to the manifest CSV file that might exist in the dataset folder. This "annotations CSV file" won't ever appear on Synapse; only as a temporary file wherever schematic is running. 
    - This "annotations CSV file" might only contain the entity Synapse IDs and etags if dataset files have no annotations. 
    - This file will be generated by iterating through the entities themselves rather than going through a file view. I'm assuming iterating through a list of entities to extract annotations is relatively quick. If not, I'll look into other approaches. 
2. Check if a CSV manifest file exists in the given dataset folder (_i.e._ `synapse_storage_manifest.csv`). 
    1. If a CSV manifest doesn't exist, pre-populate the template using the "annotations CSV file" generated in step 1 (including the Synapse IDs and etags as frozen/protected columns). 
        - One benefit of generating the "annotations CSV file" in step 1 is that I can leverage the `manifest.generator.populate_manifest_spreadsheet()` function, which takes in a CSV file path as input. 
        - Include any model attributes that aren't not among annotations. 
        - Include any annotations that aren't not among model attributes. 
    2. If a CSV manifest already exists, compare it with the "annotations CSV file" using the `csv-diff` package.
        - Warn user about any discrepancies affecting model attributes. Ignore discrepancies affecting annotations that don't appear in the data model. 
        - Include additional columns in the template for annotations whose values don't agree with the manifest. These annotation columns should be named with a suffix like `FromAnnotation` to avoid clashes in the CSV header.
        - Include additional annotations (i.e. whose names don't appear in the data model) in the template.
3. Provide the template to the user, which will contain the following columns:
    1. All relevant attributes from the data model depending on the data or assay type
    2. All additional annotations whose names aren't among the model attributes
    3. Additional annotations whose names are among the model attributes, but only if at least one of the files has a discrepancy. The user will be informed of the files with discrepancies. 
    4. Annotations whose names are among the model attributes and aren't affected by discrepancies will **not** appear in the template because they are concordant with whatever is in the existing CSV manifest file.

allaway:
>This is what I referred to as the "annotations CSV file". This is unrelated to the manifest CSV file that might exist in the dataset folder. This "annotations CSV file" won't ever appear on Synapse; only as a temporary file wherever schematic is running.

Aaah ok, this was the root of my confusion. Thanks!

>This file will be generated by iterating through the entities themselves rather than going through a file view.

This also makes sense - I think view schemas would probably make this whole process  difficult, something I hadn't considered earlier.

>Check if a CSV manifest file exists in the given dataset folder (i.e. synapse_storage_manifest.csv). 

Just curious, how would this be detected? Filename match or something else? I doubt this would cause issues for us as while we do not use dataset manifests, its unlikely someone would coincidentally upload a csv with this name. 

>Provide the template to the user, which will contain the following columns:
 >   i. All relevant attributes from the data model depending on the data or assay type
 >   ii. All additional annotations whose names aren't among the model attributes
 >   iii. Additional annotations whose names are among the model attributes, but only if at least one of the files has a discrepancy. The user will be informed of the files with discrepancies.
 >   iv. Annotations whose names are among the model attributes and aren't affected by discrepancies will not appear in the template because they are concordant with whatever is in the existing CSV manifest file.

The way I understand these, I think I'd expect the order to be i, iii, ii, iv, since i and iii are relevant to the attributes currently being annotated, but I may just be misunderstanding again.

Thanks for the detailed breakdown, though, this approach makes a lot of sense to me!

BrunoGrandePhD:
**Quick status update**

I have working code that creates a data frame from the Synapse annotations for the files in a given dataset. I also use this code to append these annotations as columns after the formal data model columns in the Google Sheets manifest. Here's an [example](https://docs.google.com/spreadsheets/d/1x3dCZurUAB6pTh7FCrqZdDJ5_wxrkQ0PbzNVEIR6eIA/edit#gid=0). Currently, I only have it working if no manifest already exists and the Google Sheet URL is being requested. For this, I'm using the `additional_metadata` property in `ManifestGenerator` that I happened to stumble upon. 

My next step is handling the case when annotations share the same name as a data model attribute. Again, I'm going to focus on the situation where no manifest exists yet. In this case, the annotations would be used to populate the empty column for the corresponding attribute (instead of being added as a column at the end of the manifest). For this, I'm probably going to use the `update()` [method](https://stackoverflow.com/a/49928647) in Pandas in the middle of a `get_as_df()`/`set_dataframe()` roundtrip with the Google Sheets template.

BrunoGrandePhD:
I now have a working prototype that properly handles annotations whose names match a label in the data model. In this situation, `schematic` converts the label (_i.e._ without spaces) to the display name (_i.e._ with spaces) so that the annotations values can be used instead of blank values in the manifest template, whose column headers use the display names. Here's a manifest [template](https://docs.google.com/spreadsheets/d/1PEFlFbZrpvUACo5KZ6H0kxaIvYqGLwxdSoFEQ75Ki7w/edit?usp=sharing) when you don't use annotations and here's [one](https://docs.google.com/spreadsheets/d/1yG-r49R2Rx5oz_X2-4LOfNcD0yIq-Xcim23bmfLluuw/edit?usp=sharing) when you do use the annotations. Don't miss columns at the very right of the spreadsheets. Also, note that the `Year of Birth` column in the second [sheet](https://docs.google.com/spreadsheets/d/1yG-r49R2Rx5oz_X2-4LOfNcD0yIq-Xcim23bmfLluuw/edit?usp=sharing) contains a value (`1980`), which was taken from an annotation called `YearofBirth`. 

I'm in the process of cleaning up my commits locally. Unfortunately, a bunch of changes affecting the files I've worked on have been committed to `develop`, resulting in several merge conflicts... I'll have to deal with those first. I hope to officially open the PR for review by Monday.

allaway:
This is really awesome! Thanks! I see an incoming tidal wave of file
re-annotating on our side... :)

BrunoGrandePhD:
Alright, I think this PR is ready for review. Please review and merge #396 first though, which will greatly reduce the number of diffs in this PR. I tried to modularize relevant code, especially in `ManifestGenerator.get_manifest`. 

## Test Status

I confirmed that the success for tests marked with `google_credentials_needed`, which cannot currently be tested using GitHub Actions. 

```
> poetry run pytest -m google_credentials_needed
================================================== test session starts ==================================================
platform darwin -- Python 3.9.1, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /Users/bgrande/Library/Caches/pypoetry/virtualenvs/schematicpy-1j0QeG0k-py3.9/bin/python
cachedir: .pytest_cache
rootdir: /Users/bgrande/repos/schematic-dev, configfile: pyproject.toml, testpaths: tests
plugins: cov-2.11.1, mock-3.5.1
collected 26 items / 22 deselected / 4 selected                                                                         

tests/test_manifest.py::TestManifestGenerator::test_get_manifest_first_time[sheet_url-use_annotations] PASSED     [ 25%]
tests/test_manifest.py::TestManifestGenerator::test_get_manifest_first_time[sheet_url-skip_annotations] PASSED    [ 50%]
tests/test_manifest.py::TestManifestGenerator::test_get_manifest_first_time[data_frame-use_annotations] PASSED    [ 75%]
tests/test_manifest.py::TestManifestGenerator::test_get_manifest_first_time[data_frame-skip_annotations] PASSED   [100%]

====================================== 4 passed, 22 deselected in 81.50s (0:01:21) ======================================
```

## Code Coverage

```
Name                                  Stmts   Miss  Cover
---------------------------------------------------------
schematic/__init__.py                     2      0   100%
schematic/__main__.py                    17     17     0%
schematic/configuration.py               71     22    69%
schematic/exceptions.py                  27      9    67%
schematic/loader.py                      41     14    66%
schematic/manifest/__init__.py            1      0   100%
schematic/manifest/commands.py           32     14    56%
schematic/manifest/generator.py         269     41    85%
schematic/models/__init__.py              1      1     0%
schematic/models/commands.py             41     41     0%
schematic/models/metadata.py             87     87     0%
schematic/schemas/__init__.py             3      0   100%
schematic/schemas/curie.py               50     26    48%
schematic/schemas/explorer.py           445    339    24%
schematic/schemas/generator.py          196     40    80%
schematic/schemas/validator.py           79     59    25%
schematic/store/__init__.py               2      0   100%
schematic/store/base.py                   3      0   100%
schematic/store/synapse.py              194    123    37%
schematic/utils/__init__.py               8      0   100%
schematic/utils/cli_utils.py             31      1    97%
schematic/utils/csv_utils.py            404    404     0%
schematic/utils/curie_utils.py           37     31    16%
schematic/utils/df_utils.py              24      8    67%
schematic/utils/general.py               17      7    59%
schematic/utils/google_api_utils.py      44     19    57%
schematic/utils/io_utils.py              23      0   100%
schematic/utils/schema_utils.py         151     64    58%
schematic/utils/validate_utils.py        19     12    37%
schematic/utils/viz_utils.py              8      6    25%
schematic/version.py                      2      2     0%
---------------------------------------------------------
TOTAL                                  2329   1387    40%
```

BrunoGrandePhD:
Now that #396 is approved and merged, I've updated this PR to ignore the changes that were just merged. 

@milen-sage @sujaypatil96: You may now review this PR! We can do that during our meeting tomorrow, but we might want to reserve time for discussion.

milen-sage:
will try to review that in advance

sujaypatil96:
I just resolved a minor merge conflict from a bugfix push we had into `develop`, hope that's okay.

schematic/store/synapse.py
```diff
@@ -476,3 +476,63 @@ def associateMetadataWithFiles(self, metadataManifestPath: str, datasetId: str)
 
         return manifestSynapseFileId
 
+
+    def getFileAnnotations(self, fileId: str) -> dict:
```
sujaypatil96, line 480: Trivial note: We can use `Dict[str, str]` from the `typing` module instead of `dict` here?

tests/test_manifest.py
```diff
@@ -1,26 +1,124 @@
 import logging
 import pytest
+from click.testing import CliRunner
```
sujaypatil96, line 3: Guessing `CliRunner` is not required at the moment?

sujaypatil96 (approved):
The main focus of the PR which were the two methods: `getFileAnnotations` and `getDatasetAnnotations`, and their corresponding tests too, all seem to be in order. We can go over the changes once again during the code review and merge.

This is a fantastic PR @BrunoGrandePhD, thank you.🚀

schematic/store/synapse.py
BrunoGrandePhD: Good catch!

tests/test_manifest.py
BrunoGrandePhD: Very perceptive of you! A previous version of this test suite used the `get_manifest()` CLI function, but I ended up "rebuilding" that function here using a combination of fixtures.

BrunoGrandePhD:
I've addressed @sujaypatil96's comments. Thanks for the review!

BrunoGrandePhD:
So, here's the comparison of the original `update_df()` function and my new `update_manifest()` function. From what I can tell, `update_df()` changes whole columns wholesale, whereas `update_manifest()` can selectively update values in a column (_i.e._ wherever `df_2` has a value other than `NaN`). 

Does this sound like we can replace all instances of `update_df()` with `update_manifest()`? If there are no `NaN` values in `df_2`, then their behaviors are the same, I think. 

Actually, one important difference is that if the index column isn't in `df_1`, `update_df()` just returns the input as is, whereas `update_manifest()` raises an error. Is there a reason for returning the input as is in `update_df()`? 

@milen-sage @sujaypatil96: I want to hear whether you think we should make the switch. 

```
df_1:                           df_2:

numCol entityId strCol          strCol  numCol entityId
     1    syn01    foo             ___     NaN    syn01
     2    syn02    bar             NaN     4.0    syn02



update_manifest:

numCol entityId strCol
   1.0    syn01    ___
   4.0    syn02    bar



update_df:

numCol entityId strCol
   NaN    syn01    ___
   4.0    syn02    NaN
```

BrunoGrandePhD:
@allaway: Do you have a Synapse project that has a lot of files with annotations that you want to validate? I want to test the new feature in that context because I'm afraid it's unreasonably slow due to the iterative retrieval of annotations. 🐌

allaway:
:)

I'm not sure whether this qualifies as "lots," but this project has a
fairly substantial number of files with annotations, at least relative to
most of our projects: syn4984604 <http://synapse.org/cutaneousNF>

milen-sage (approved):
@BrunoGrandePhD the new update df function makes sense. I went through the rest of the changes as well and they reflect what we discussed on Wed. It's ready to merge on my end.

BrunoGrandePhD:
I tried the new `--use_annotations` feature that I implemented in this PR with a large-ish dataset (_i.e._ 3505 files) that @allaway provided ([syn4984604](https://www.synapse.org/#!Synapse:syn4984604)). 

The good news is that the feature seems to work. Here's the [manifest](https://docs.google.com/spreadsheets/d/1BdOLIcTL6hbhk3H2FxQIGPsyvGN8ELl4gSrnMq2MHGQ) that was generated with `--use_annotations`. For reference, I've included the command that I used below. For comparison, here's a [manifest](https://docs.google.com/spreadsheets/d/17qINaRyn83s1p_nlzwxtTC90zT_lhgTJFwFYR8xMOcU) generated with the same command minus `--use_annotations`. 

```
schematic manifest get --dataset_id syn4984604 --jsonld ~/Downloads/NF.jsonld --config tests/data/test_config.yml --sheet_url --data_type GenomicsAssay --use_annotations
```

The bad news, which was anticipated, is the run time. However, the extent of the slowdown associated with retrieving annotations is worse than expected. Generating the Google Sheets URL took 59 seconds without using annotations, but a whopping 20 minutes and 30 seconds when pre-populating with annotations. 

## Action Items

- [ ] @BrunoGrandePhD: I'm going to profile the command to confirm my suspicion that it's the annotation retrieval step that's slowing things down. 
- [ ] @BrunoGrandePhD: I'm going to ask John Hill if we can identify a more efficient way of batch-retrieving annotations for large sets of files. 
- [ ] @allaway: Could you review the pre-populated [manifest](https://docs.google.com/spreadsheets/d/1BdOLIcTL6hbhk3H2FxQIGPsyvGN8ELl4gSrnMq2MHGQ) for any issues?
  - For one, I don't know why the `Filename` column is empty. 
- [ ] @allaway: Do you see a problem with creating these temporary file views in the parent project of each dataset? 
  - The short-term solution for speeding things up might be the creation of a temporary file view for aggregating annotations before download the file view as a table.
  - We could name them using a random string to prevent clashes and delete them as soon as the annotations are retrieved so they don't linger.

BrunoGrandePhD:
I generated a run time [profile](https://github.com/Sage-Bionetworks/schematic/files/6018362/profile.txt). It's my first time generating one of these, so bear that in mind. I also am running into rate limits with Google's API (see #410), but I think I got the information I need. I think it does confirm that the annotation retrieval process is one of the slowest parts. 

I want to highlight lines 33-38 from the [profile](https://github.com/Sage-Bionetworks/schematic/files/6018362/profile.txt). The rows are ordered based on the time cumulatively spent in each function (see `cumtime` column). The total time spent according to the top row is 605 seconds. You can see that the one call to `getDatasetAnnotations()` takes up 527 seconds (first row below `[...]`), which represents the vast majority of the time spent during execution. 

The last row confirms that the bottleneck is indeed with the Synapse API, since `_getRawAnnotations()` is a simple wrapper around `restGET()`. You can see that `restGET()` is also included in the rows below, but we cannot conclude anything from it because that function can be called from anywhere (not necessarily for the annotation retrieval). 

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   2335/1    0.010    0.000  605.419  605.419 {built-in method builtins.exec}
   [...]
        1    0.001    0.001  527.485  527.485 synapse.py:512(getDatasetAnnotations)
     3513    0.030    0.000  489.277    0.139 sessions.py:546(get)
        1    0.033    0.033  487.881  487.881 synapse.py:527(<listcomp>)
     3505    0.131    0.000  487.849    0.139 synapse.py:478(getFileAnnotations)
     3508    0.029    0.000  487.582    0.139 client.py:3766(restGET)
     3505    0.104    0.000  487.298    0.139 client.py:1291(_getRawAnnotations)
```

While I'm looking at the profile, I also noticed that `getFilesInStorageDataset()` also takes up a non-trivial amount of time for a large dataset. On average, it takes 40 seconds to retrieve a list of 3505 files (see `percall` column). 

This function also happens to be called twice (see `ncalls` column), which seems unnecessary. I know that one of these two calls was added in this PR (in `getDatasetAnnotations()`), but I didn't realize this function could be so slow. I think he other call is made in `getDatasetManifest()`, but it seems excessive to compile a list of all of the files in the dataset just to find the manifest (unless the file list is cached). 

We can probably work towards (1) speeding up the function itself (perhaps with the use of a temporary file view) and/or (2) making sure the function is only called once. These optimizations are important for large datasets. 

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   2335/1    0.010    0.000  605.419  605.419 {built-in method builtins.exec}
   [...]
        2    0.008    0.004   79.766   39.883 synapse.py:200(getFilesInStorageDataset)
```

allaway:
> @allaway: Could you review the pre-populated manifest for any issues?
> 
>     For one, I don't know why the Filename column is empty.
> 

Other than the Filename issue, this looks pretty good to me! G-sheets is doing some heavy lifting here and it's not clear whether a dataset of this size could feasibly be annotated interactively with so much data (ie. the sheet is super slow, at least on my computer...maybe it would be more performant in excel?). 

Re: the missing filenames - this is probably not correct, but I'm wondering if it's because the reserved name for filenames on synapse is `name`, and that schematic is looking for an annotation `Filename` to populate this column?

> @allaway: Do you see a problem with creating these temporary file views in the parent project of each dataset?
> 
>     The short-term solution for speeding things up might be the creation of a temporary file view for aggregating annotations before download the file view as a table.
>     We could name them using a random string to prevent clashes and delete them as soon as the annotations are retrieved so they don't linger.

I don't think that would be a problem. Another option, if we are really worried about cluttering up projects or confusing data contributors, would be to add these temporary fileviews to a core DCC project. I agree that a temporary fileview is likely to be faster. 

An additional, alternate, possibility, would be to rely on a pre-existing permanent fileview for each project that contains all possible keys within the schema. I'm not sure if this is realistic or not, and is likely to be more error prone. But in this case, indexing would have already happened, so it would be even faster than generating a new fileview (I suspect, anyway...). 

In NF, we actually have these fileviews per project. It would just be a matter of modifying their schemas to contain all potential keys in our metadata dictionary (I can see how this might make modifying our schema in the future more of a headache...)

BrunoGrandePhD:
@allaway: You can export the Google Sheet to Excel by using the `File` > `Download` > `Microsoft Excel` menu option. Let me know if the performance is better on your computer in Excel. 

As for your comment regarding `Filename`, I'll look into that. Just in case you haven't come across it, there's a [section](https://sagebionetworks.jira.com/wiki/spaces/DATA/pages/1727365372/Setup+and+Usage?focusedCommentId=1912373275#comment-1912373275) in the usage docs indicating that `Filename` does have a special meaning for `schematic`. 

The issue with using pre-existing file views is that `schematic` would have to have to update their schemas to use all user-defined annotations for all entities within their scope. I wouldn't want `schematic` to change anything about existing file view, preferably. Also, I don't know if there's a limit on the number of columns in a file view. I wonder if we would run into that limit if a file view is created with the superset of all possible manifest columns. 🤔 

To minimize the risk of confusion, the temporary file view could be created with local sharing settings where only the current user can view it. This way, the only way for the user to ever see it would be while browsing Synapse when `schematic` is generating a manifest template with `--use_annotations`. This seems unlikely.

allaway:
>@allaway <https://github.com/allaway>: You can export the Google Sheet to
Excel by using the File > Download > Microsoft Excel menu option. Let me
know if the performance is better on your computer in Excel.

Thanks! We've actually used this a bit already and it's worked well, but I
haven't tried with such a large dataset. It's not a big deal, I just wanted
to note it :)

>As for your comment regarding Filename, I'll look into that. Just in case
you haven't come across it, there's a section
<https://sagebionetworks.jira.com/wiki/spaces/DATA/pages/1727365372/Setup+and+Usage?focusedCommentId=1912373275#comment-1912373275>
in
the usage docs indicating that Filename does have a special meaning for
schematic.

Yep! We are using it in (what I understand to be) the intended way - we
want to pull in the synapse filename so that folks know what they are
annotating.

>The issue with using pre-existing file views is that schematic would have
to have to update their schemas to use all user-defined annotations for all
entities within their scope. I wouldn't want schematic to change anything
about existing file view, preferably. Also, I don't know if there's a limit
on the number of columns in a file view. I wonder if we would run into that
limit if a file view is created with the superset of all possible manifest
columns. 🤔

Yes, I agree that this would be an issue. There is a limit on the number of
columns, IIRC 255? (No, that's wrong, it is actually 156
<https://docs.synapse.org/rest/#org.sagebionetworks.repo.web.controller.TableController>).
I don't know that we'd run into that issue _today_ with the NF schema, but
it definitely is a possibility down the road.

>To minimize the risk of confusion, the temporary file view could be
created with local sharing settings where only the current user can view
it. This way, the only way for the user to ever see it would be while
browsing Synapse when schematic is generating a manifest template with
--use_annotations. This seems unlikely.

This seems like the best solution. If doing this, it may also be worth
hard-coding/config-coding the admin team (e.g. say, if we run into an issue
with the app crashing down the road somewhere before fileview deletion, the
admins could tidy things up). But that is a nice-to-have, not that high
priority, I think.

PR #363, opened 2020-12-07
Add schema_explorer argument to SchemaGenerator

BrunoGrandePhD:


schematic/schemas/generator.py
```diff
@@ -38,15 +40,33 @@ def __init__(self,
...
-            return
+
+            # Confirm that given SchemaExplorer instance is valid
+            assert getattr(schema_explorer, "schema_nx") is not None, (
+                "SchemaExplorer instance given to `schema_explorer` "
+                "does not have a `schema_nx` attribute."
+            )
+
+            # User given instance of SchemaExplorer
+            self.se = schema_explorer
```
sujaypatil96, line 69: In addition to checking for the `schema_explorer.schema_nx` attribute/property, we also need to check if the `schema_explorer.schema` property has been provided. Since `load_schema` sets both those properties, it's necessary that the `schema_explorer` object that the user provides have the same structure as that created by `load_schema`.

sujaypatil96 (changes requested):

schematic/schemas/generator.py
BrunoGrandePhD: Good idea! I've addressed this in the faf4062.

sujaypatil96:
The change looks good @BrunoGrandePhD. Feel free to merge.

sujaypatil96 (approved):
The change looks good @BrunoGrandePhD.

PR #364, opened 2020-12-09
CLI for `get_manifest` method

sujaypatil96:
Based on discussion in issue #328 and `click`-example, adding a CLI for the `get_manifest()` method.

To run this command:

`$ python -m schematic manifest get TITLE PATH_TO_JSON_LD DATA_TYPE --DATASET_ID --SHEET_URL --JSON_SCHEMA`


milen-sage (changes requested):
Reflecting our conversation today: 
- provide handling of parameters specified via the config file 
- issue warning when cli positional/keyword parameters overlap with parameters provided in config (i.e. indicating that CLI-provided parameters override the config ones)

BrunoGrandePhD:
I'll review this once the changes that were mentioned in our meeting today are implemented. Ping me once that's ready!

sujaypatil96:
@milen-sage  / @BrunoGrandePhD: The code now uses parameter values from the `config.yml` file. 

With respect to raising warnings using `logging` module, I'll implement it when tackling issue #317. For now, it simply logs to `stdout` instead of `stderr`.

schematic/manifest/commands.py
```diff
@@ -0,0 +1,71 @@
...
+def manifest(): # use as `schematic manifest ...`
+    """
+    Sub-commands with Manifest Generation utilities/methods.
+    """
+    pass
+
+
+# prototype based on getModelManifest() and get_manifest()
+# use as `schematic config get positional_args --optional_args`
+@manifest.command('get', short_help='Prepares the manifest URL based on provided schema.')
```
BrunoGrandePhD, line 20: We might want to better distinguish between mandatory or optional arguments by switching some arguments to being positional.

BrunoGrandePhD (changes requested):
I started doing my review, but then I started implementing some utility functions to minimize code duplication. See #368.

schematic/manifest/generator.py
```diff
@@ -192,7 +193,7 @@ def _get_valid_values_from_jsonschema_property(self, prop:dict) -> List[str]:
             return []
 
 
-    def get_empty_manifest(self, json_schema = None):
+    def get_empty_manifest(self, json_schema_file = None):
```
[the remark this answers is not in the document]
sujaypatil96: I agree. I've made a change to the `get_empty_manifest()` prototype, to change `json_schema_file` argument to `json_schema_filepath`.

schematic/manifest/commands.py
```diff
@@ -0,0 +1,71 @@
...
+# define the optional arguments
+@click.option('-t', '--title', help='Title of generated manifest file.')
+@click.option('-d', '--data_type', help='Data type/component from JSON-LD schema to be used for manifest generation.')
+@click.option('-p', '--path_to_json_ld', help='Path to JSON-LD schema.')
```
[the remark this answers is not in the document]
sujaypatil96: Sounds good. The `schematic manifest get` CLI now uses `--jsonld` as an optional argument.

schematic/manifest/commands.py
sujaypatil96: We agreed that we wouldn't have any positional arguments while defining our CLIs. All arguments will be optional arguments. To enforce certain arguments we would just create "required, optional" arguments.

schematic/utils/cli_utils.py
```diff
@@ -0,0 +1,83 @@
...
+            return None
+        return dictionary.get(key)
+
+    return reduce(extract, keys, dictionary)
+
+
+def fill_in_from_config(
+    arg_name: str,
+    arg_value: Any,
+    config_keys: Sequence[Any]
```
sujaypatil96, line 43: Removed the `config` parameter that was previously being passed to `fill_in_from_config()`. Now it just uses the global `CONFIG` object.

schematic/manifest/commands.py
```diff
@@ -0,0 +1,62 @@
...
+# use as `schematic config get positional_args --optional_args`
+@manifest.command('get', short_help='Prepares the manifest URL based on provided schema.')
+# define the optional arguments
+@click.option('-t', '--title', help='Title of generated manifest file.')
+@click.option('-d', '--data_type', help='Data type/component from JSON-LD schema to be used for manifest generation.')
+@click.option('-p', '--jsonld', help='Path to JSON-LD schema.')
+@click.option('-d', '--dataset_id', help='SynID of existing dataset on Synapse.')
+@click.option('-s', '--sheet_url', type=bool, help='Enable/disable URL generation.')
+@click.option('-j', '--json_schema', help='Path to JSON Schema (validation schema).')
+@click.option('-c', '--config', help='Path to schematic configuration file.', required=True)
```
sujaypatil96, line 31: We can enforce `config` optional argument by setting `required=True`.

sujaypatil96:
This PR is ready for review @BrunoGrandePhD.

Also, I checked the "bug" which was coming up when you were using `syn23643254` as the value for the `dataset_id` argument, and it is because there's no CSV (manifest) file in that folder. So it's a good idea to create a raise an informative exception in that case, and it will be a good idea to that in the `get_manifest()` method. I will save that bugfix for another PR though.

schematic/manifest/commands.py
```diff
@@ -1,7 +1,10 @@
 #!/usr/bin/env python3
 
+import warnings
```
BrunoGrandePhD, line 3: Why was this added? I don't see this module being used anywhere in this file.

BrunoGrandePhD (approved):
If you can remove the unused `warnings` import, go ahead and merge this PR! 

🚀

schematic/manifest/commands.py
sujaypatil96: Thanks, removing `warnings`.

milen-sage (approved):

PR #373, opened 2020-12-21
refactor Synapse-specific code to live in synapse/store module only

sujaypatil96:
Based on a review comment in PR #353, this PR seeks to move Synapse-specific code (being used in [`get_manifest()` ](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/manifest/generator.py#L686)), back into `synapse/store` module. For this, the [`SynapseStorage`](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/synapse/store.py) class constructor had to be modified to be able to read from the `.synapseConfig` file (synapse configuration file as in `config.yml`). I created a separate PR to test it's downstream integration with `get_manifest()` first, before moving onto the method `submit_metadata_manifest()` as in PR #353.

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -685,9 +685,6 @@ class ManifestGenerator(object):
 
             syn_store = SynapseStorage()
 
-            # determine path to the user-specified manifests folder where the manifest should be downloaded to
-            manifests_folder_path = CONFIG["synapse"]["manifest_folder"]
-
             # get manifest file associated with given dataset
             syn_id_and_path = syn_store.getDatasetManifest(datasetId=dataset_id)
 
@@ -695,7 +692,7 @@ class ManifestGenerator(object):
             if syn_id_and_path and sheet_url:
 
                 # get synapse ID manifest associated with dataset
-                manifest_data = syn_store.syn.get(syn_id_and_path[0], downloadLocation=manifests_folder_path, ifcollision="overwrite.local")
+                manifest_data = syn_store.getDatasetManifest(datasetId=dataset_id, downloadFile=True)
 
                 # get URL of an empty manifest file created based on schema component
                 empty_manifest_url = self.get_empty_manifest()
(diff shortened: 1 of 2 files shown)


schematic/synapse/store.py
```diff
@@ -65,12 +62,10 @@ def __init__(self,
...
                 print("Please enter a valid session token.")
                 return
-        elif syn: # if no token, assume a logged in synapseclient instance has been provided
-            if isinstance(syn, synapseclient.Synapse):
-                self.syn = syn
-            else:
-                print("Please make sure 'syn' argument is of type synapseclient.Synapse().")
-                return
+
+        # login using synapse credentials provided by user in .synapseConfig (default) file
```
milen-sage, line 66: Maybe I am misinterpreting the logic here, but wouldn't self.syn be always set to read from Synapse config independent of whether a token was provided (L57) or not?

Is that the desired behavior? Wouldn't we still want to support token-based login when a token is provided?
Shouldn't l67-69 be in an else branch?

sujaypatil96: Oh yes, good catch. I've moved the part that reads from the `.synapseConfig` file into an else-block.

schematic/manifest/generator.py
```diff
@@ -698,7 +695,7 @@ def get_manifest(self, dataset_id: str = None, sheet_url: bool = None, json_sche
             if syn_id_and_path and sheet_url:
 
                 # get synapse ID manifest associated with dataset
-                manifest_data = syn.get(syn_id_and_path[0], downloadLocation=manifests_folder_path, ifcollision="overwrite.local")
+                manifest_data = syn_store.syn.get(syn_id_and_path[0], downloadLocation=manifests_folder_path, ifcollision="overwrite.local")
```
milen-sage, line 698: Let's try minimizing synapse references here and work only with SynapseStorage methods; one approach
- add a boolean-flag argument to SynapseStorage's getDatasetManifest that indicates whether to fetch the manifest data or not
- if flag true, return the manifest data (e.g. instead of the synid) using the call below  
syn_store.syn.get(syn_id_and_path[0], downloadLocation=manifests_folder_path, ifcollision="overwrite.local")
--- by default the flag can be false (for backward compatibility)

That way we'll have all synapse logic encapsulated in SynapseStorage and here we'll only focus on the logic relevant for the manifest population; further that would avoid the repeating lines 698 and 710. You can still check if the manifest data is returned properly by getDatasetManifest similarly to what you do in l695 and l709.

milen-sage (changes requested):
Some comments in line

schematic/manifest/generator.py
sujaypatil96: This sounds like a good idea. I was trying to think of a [`Decorators`](https://book.pythontips.com/en/latest/decorators.html)-based approach to it. To create a decorator (maybe an `@get_manifest_data` decorator) or something, but that would be useful only to this function, not sure if any other methods could/would be decorated this way.

I've implemented the it per your recommendation above @milen-sage, which makes more sense.

milen-sage (approved):
Looks good - thanks @sujaypatil96 !
Will test some on my end.

PR #379, opened 2021-01-15
Added Nuclei Isolation as an option for scRNA-seq level 1 Single Cell…

milen-sage:
… Isolation Method

Requested by HTAPP.


xdoan (approved):
I see nuclei isolation as an option in Single Cell Isolation Method for scRNAseq Lvl 1 after the update.

PR #386, opened 2021-01-25
tests for io_utils functions

sujaypatil96:
This PR adds tests for the utilities in `schematic.utils.io_utils`.

Note: Removing the `.decode()` from `load_json()` function because there's no need to decode an object that has already been decoded (into a `str`).

### DIFF
--- a/schematic/utils/io_utils.py
+++ b/schematic/utils/io_utils.py
@@ -11,7 +11,7 @@ def load_json(file_path):
     """
     if file_path.startswith("http"):
         with urllib.request.urlopen(file_path) as url:
-            data = json.loads(url.read().decode())
+            data = json.loads(url.read())
             return data
     # handle file path
     else:
--- a/schematic/utils/io_utils.py
+++ b/schematic/utils/io_utils.py
@@ -11,7 +11,7 @@ def load_json(file_path):
     """
     if file_path.startswith("http"):
         with urllib.request.urlopen(file_path) as url:
-            data = json.loads(url.read())
+            data = json.loads(url.read().decode('utf-8'))
             return data
     # handle file path
     else:
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -123,7 +123,7 @@ class FakeResponse:
 
 class TestIOUtils:
 
-    def test_json_load(self, tmpdir, mocker):
+    def test_json_load(self, tmpdir):
 
         json_file = tmpdir.join("example.json")
         json_file.write_text(json.dumps([
@@ -134,9 +134,12 @@ class TestIOUtils:
         with open(json_file, encoding='utf-8') as f:
             expected = json.load(f)
 
-        result = io_utils.load_json(str(json_file))
+        local_result = io_utils.load_json(str(json_file))
 
-        assert result == expected
+        assert local_result == expected
+
+    
+    def test_json_load_online(self, mocker):
 
         mock_urlopen = mocker.patch("urllib.request.urlopen",
                                     return_value = FakeResponse(data=json.dumps([
@@ -145,8 +148,8 @@ class TestIOUtils:
                                     ))
                                     )
         
-        res = io_utils.load_json("http://example.com")
-        assert res == [
+        url_result = io_utils.load_json("http://example.com")
+        assert url_result == [
             {'k1': 'v1'},
             {'k2': 'v2'}
         ]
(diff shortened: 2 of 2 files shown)


schematic/utils/io_utils.py
```diff
@@ -11,7 +11,7 @@ def load_json(file_path):
     """
     if file_path.startswith("http"):
         with urllib.request.urlopen(file_path) as url:
-            data = json.loads(url.read().decode())
```
BrunoGrandePhD, line 14: Could you explain this change here for posterity?

tests/test_utils.py
```diff
@@ -94,3 +98,94 @@ def test_fill_in_from_config(self, mocker):
...
+                                    ))
+                                    )
+        
+        res = io_utils.load_json("http://example.com")
+        assert res == [
+            {'k1': 'v1'},
+            {'k2': 'v2'}
+        ]
+
+        assert mock_urlopen.call_count == 1
```
BrunoGrandePhD, line 154: [Nit:](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.) You might want to split this test function into two: `test_json_load_local()` and `test_json_load_online()`. Not required.

tests/test_utils.py
```diff
@@ -94,3 +98,94 @@ def test_fill_in_from_config(self, mocker):
...
+        jsonld_file.write_text(json.dumps(jsonld_str), encoding="utf-8")
+
+        mocker.patch("schematic.LOADER.filename",
+                    return_value=os.path.normpath(jsonld_file))
+
+        expected_default = io_utils.load_default()
+        expected_schemaorg = io_utils.load_schemaorg()
+
+        assert expected_default == jsonld_str
+        assert expected_schemaorg == jsonld_str
```
BrunoGrandePhD, line 191: I don't think we need to patch the loader here. Since it's a local file, we should be able to load the actual BioThings and Schema.org JSON-LD models. Obviously, we can't check whether the entire JSON structure is correct, but I think a few spot checks (_e.g._ number of elements in `@graph`, the list of context keys) should suffice for testing the loading.

BrunoGrandePhD (changes requested):

schematic/utils/io_utils.py
sujaypatil96: My bad, I removed `decode()` initially to fit the test which was giving me an error like "why do you want to decode an object that is already a string?". Later I fixed my test by adding an `encode('utf-8')` on `json.dumps()`.

tests/test_utils.py
sujaypatil96: That makes sense @BrunoGrandePhD, thanks. I've patched the `LOADER.filename()` function since we need some way to overwrite the hardcoded schema/file paths that are already set for each of these methods (`load_default()` and `load_schemaorg()`).

The basic checks I've added are:
- list of context keys
- number of graphs (if multiple)
- number of elements in graph

tests/test_utils.py
```diff
@@ -170,22 +173,61 @@ def test_export_json(self, tmpdir):
         assert expected == json_str
 
 
-    def test_load_schema(self, tmpdir, mocker):
+    def test_load_default(self, mocker):
+        
+        biothings_schema_path = "tests/data/etc/data_models/biothings.model.jsonld"
```
BrunoGrandePhD, line 178: I'm afraid that I wasn't clear in my previous review. My apologies. 

I wasn't thinking of duplicating the JSON-LD files in the `tests/` folder. You should still be able to use the `io_utils.load_default()` and `io_utils.load_schemaorg()` function directly, which should load the JSON-LD files from the `schematic/etc/` folder. You can then perform the same checks as you have here. 

Does that make sense? I just don't think we need to duplicate files in this case.

BrunoGrandePhD (changes requested):
Feel free to `rebase -i` this branch to drop/edit the commit that duplicates the files so that it doesn't linger in the Git history. If you need help with this, let me know!

tests/test_utils.py
sujaypatil96: Right, true, I was going back and forth on whether or not to duplicate the `etc/data_models` folder. I thought I should consolidate all "resources" used by pytest tests within the `tests/data` folder. But you're right, `etc` folder contains package-specific files and will always be present in the core without being modified, so we can use it directly.

Thanks for the review @BrunoGrandePhD, I've fixed it.

sujaypatil96:
Closing this PR since PR #390 already included the intended changes.

Issue #376, opened 2021-01-08
Badly Formatted CSVs with empty column names throw an error upon validation


xdoan:
**Describe the bug**
Jackynn from MSK had this CSV she got from exporting from Excel (can't upload but its on slack if you want it)
<img width="493" alt="Screen Shot 2021-01-08 at 1 14 35 PM" src="https://user-images.githubusercontent.com/15043209/104064831-7a6ea700-51b3-11eb-8920-b527d6e59930.png">
 that has trailing column spaces (all the ,,,,, after the column names) that are incorrectly read as empty columns and given a column name in R, etc. Similar to the error we got in adding extra rows from excel to csv. 

**To Reproduce**
1. download the csv https://htanworkspace.slack.com/archives/G01AUKU8HJT/p1609954274101400?thread_ts=1607713396.074500&cid=G01AUKU8HJT
2. try to to validate it (demographics) on the app or backend
3. On the app for me it hangs and crashes the site until you refresh it a few times. 

**Expected behavior**
The empty columns should be stripped before validation, similar to completely empty rows.

sujaypatil96:
Hi @xdoan. Can you attach a screenshot of the error you're seeing (on the R Shiny app, backend also if tested). I'm seeing the same set of errors on the backend when I run `validateModelManifest()` on the manifest file you've shared, with and without the trailing commas in the column names.

<img width="1112" alt="Screen Shot 2021-01-08 at 2 05 22 PM" src="https://user-images.githubusercontent.com/20239224/104069957-a4779780-51ba-11eb-85c2-ba3a998af4b6.png">

sujaypatil96:
Hi @xdoan. Could you attach a screenshot of the error message you're seeing on the Shiny frontend? I validated (`validateModelManifest()`) the attached manifest against the `Demographics` component, and I'm seeing the same set of errors, both with the trailing commas and without them.

_Note: I did trim out the empty rows after line 58 before running this though._

<img width="1112" alt="Screen Shot 2021-01-08 at 2 05 22 PM" src="https://user-images.githubusercontent.com/20239224/104070309-6fb81000-51bb-11eb-925d-ce612071f463.png">

xdoan:
What you have is the expected output, hopefully you can just trim any completely empty columns so this doesn't happen. 

The server timed out before I got the validation error returned. It just was processing forever (it is 9k rows etc). 

Log attached: 
[...]

PR #389, opened 2021-01-27
trim_commas_df to trim only trailing empty columns

sujaypatil96:
A slight modification was necessary for the `trim_commas_df()`. This is the method that was written to address issue #376. When writing tests for the utilities in `df_utils` I realized, `.dropna(how='all', axis=0)` dropped all empty columns and that wasn't desirable behaviour, since our manifests are allowed to have completely empty columns. The manifest that we used to test from issue #376 didn't have any empty columns besides the last/trailing ones, so this issue didn't come up.

I've made a slight modification to the function `trim_commas_df()` to trim out/drop only the those columns which have the substring "Unnamed" in them. As for empty rows, it shouldn't matter where the empty rows are we can remove them freely.

### DIFF
--- a/schematic/utils/df_utils.py
+++ b/schematic/utils/df_utils.py
@@ -1,4 +1,5 @@
 import pandas as pd
+import numpy as np
 
 def normalize_table(df: pd.DataFrame, primary_key:str) -> pd.DataFrame:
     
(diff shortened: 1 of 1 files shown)


schematic/utils/df_utils.py
```diff
@@ -43,7 +43,7 @@ def update_df(existing_df: pd.DataFrame, new_df: pd.DataFrame, idx_key: str) ->
     # filter to keep only updated values across all columns in the schema
     for col in existing_df.columns:
         if col != idx_key and col != 'ROW_ID' and col != 'ROW_VERSION':
-            existing_col = col + "_existing"
+            # existing_col = col + "_existing"
```
milen-sage, line 46: was this intended?

milen-sage (approved):

PR #390, opened 2021-01-27
tests for df_utils module

sujaypatil96:
Added tests for utilities in the `df_utils` module. 

_Note_: We need to modify the `normalize_table()` function, since it doesn't seem correct semantically. I didn't write any test cases for it since we need to decide what the behaviour needs to be and since it isn't being used anywhere yet.

### DIFF
--- a/schematic/utils/df_utils.py
+++ b/schematic/utils/df_utils.py
@@ -1,4 +1,5 @@
 import pandas as pd
+import numpy as np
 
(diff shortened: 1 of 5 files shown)


milen-sage (approved):

BrunoGrandePhD (approved):
🚀 

I don't think we need to worry about whether the pytest fixtures are conventional or not. If they work, we can always change them later. They seem fine to me.

Issue #335, opened 2020-11-02
Improve the key/field names in the config.yaml file


BrunoGrandePhD:
Once PR #334 is merged, we can think of improving the key/field names in the config file so they're more intuitive. This could entail reorganizing the YAML structure. Here are a few suggestions:

- Reorganize the structure so that everything Synapse-related is under the `synapse` top-level key and everything Google-related is under a `google` top-level key. 
- Add the file types that are expected in the key names. For example, change `location` to `jsonld_location`. 
- Is `file_type` included because the location could be a URL? If so, couldn't that be inferred from the `location` (_i.e._ if prefixed with `http://` or `ftp://`)? I wouldn't be surprised if there was a library that can load files regardless of whether they are local or remote if given a string. 
- Could `synapse_config` be set to `~/.synapseConfig` instead by default? Admittedly, this is more of a usability improvement rather than a clarity improvement. 

**Excerpt from [`config.yaml`](https://github.com/Sage-Bionetworks/schematic/blob/ffab34d4af1af13cfc813359340d26375d5c1fee/config.yml)**
```
# Do not change the 'definitions' section unless you know what you're doing
definitions:
  synapse_config: ".synapseConfig"
  creds_path: "credentials.json"
  token_pickle: "token.pickle"
  service_acct_creds: "quickstart-1560359685924-198a7114b6b5.json"

synapse:
  master_fileview: 'syn20446927'
  manifest_folder: 'manifests'
  manifest_filename: 'data/manifests/synapse_storage_manifest.csv'
  api_creds: 'syn21088684'

model:
  input:
    location: 'data/schema_org_schemas/HTAN.jsonld'
    file_type: 'local'
    validation_schema: 'data/validation_schemas/HTAN_validation_schema_FamilyHistory.json'
    log_location: 'data/json_schema_logs/json_schema_log.json'
  biothings:
    location: 'data/schema_org_schemas/biothings.jsonld'
```

ychae:
@sujaypatil96 can this be closed?

PR #393, opened 2021-02-02
replace print statements with logging - synapse/store

sujaypatil96:
This PR works to replace the `print` statements in the `synapse/store` module with `logging` statements.

_Note_: I'm planning to address issue #317 in a series of multiple PRs, where I'll refactor schematic, module-by-module. This is the first of those PRs.

I've also been meaning to make this addition to the `config.yml` file:

```
manifest:
  synapse:
    title: 'Patient_Manifest'
    data_type: 'Patient'
  local:
    path: '/path/to/local/manifest'
```

Just a note, we can consider this if we want to allow users to configure the path to the local manifest. But I guess we can talk more about it when working on issue #335.

### DIFF
--- a/schematic/exceptions.py
+++ b/schematic/exceptions.py
@@ -53,3 +53,28 @@ class MissingConfigAndArgumentValueError(Exception):
 
     def __str__(self):
         return f"{self.message}"
+
+
+class AccessCredentialsError(Exception):
+    """Exception raised when provided access credentials cannot be resolved.
+    
+    Args:
+        project: Platform/project (e.g., synID of a project)
+        message: custom/pre-defined error message to be returned.
+
+    Returns:
+        message.
+    """
+
+    def __init__(self, project: str, message: str = None) -> str:
+        if message is not None:
+            self.message = message
+        else:
+            self.message = (
+                f"Your access to '{project}'' could not be resolved. "
+                "Please check your credentials and try again."
+            )
+        super().__init__(self.message)
+
+    def __str__(self):
+        return f"{self.message}"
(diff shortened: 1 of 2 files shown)


schematic/exceptions.py
```diff
@@ -53,3 +53,28 @@ def __init__(self, arg_name: str,
...
+        message: custom/pre-defined error message to be returned.
+
+    Returns:
+        message.
+    """
+
+    def __init__(self, project: str, message: str = None) -> str:
+        if message is not None:
+            self.message = message
+        else:
```
BrunoGrandePhD, line 72: You can avoid this else statement by setting the message to the default value before the if statement and then overwrite the message with user-provided value `if message:`.

schematic/exceptions.py
```diff
@@ -53,3 +53,28 @@ def __init__(self, arg_name: str,
...
+            self.message = (
+                f"Your access to '{project}'' could not be resolved. "
+                "Please check your credentials and try again."
+            )
+        super().__init__(self.message)
+
+    def __str__(self):
```
BrunoGrandePhD, line 79: If this pattern (and the pattern in __init__) is going to repeat itself for each exception, you could look into creating a base `SchematicException` class.

BrunoGrandePhD (approved):
I'm happy with the PR as is, but feel free to implement some of my suggested tweaks. 

🚀

schematic/exceptions.py
sujaypatil96: Great. This is cleaner, yes. I've incorporated this change.

schematic/exceptions.py
sujaypatil96: Thanks for the tip Bruno. I'll work on this and incorporate in a future PR.

PR #394, opened 2021-02-04
Tests for utilities in validate_utils module

sujaypatil96:
Building on top of PR #390, adding tests for utilities in `schematic.utils.validate_utils`.


tests/conftest.py
```diff
@@ -24,4 +26,18 @@ def synapse_manifest():
 
 @pytest.fixture()
 def local_manifest():
-    return pd.read_csv("tests/data/mock_manifests/local_manifest.csv")
\ No newline at end of file
+    return pd.read_csv("tests/data/mock_manifests/local_manifest.csv")
```
BrunoGrandePhD, line 29: I suggest you use the new `get_data_frame()` helper function in `helpers` fixture that was just merged.

BrunoGrandePhD: You might have to merge `develop` into your branch first.

tests/conftest.py
```diff
@@ -24,4 +26,18 @@ def synapse_manifest():
...
+
+
+@pytest.fixture()
+def schema_explorer():
```
BrunoGrandePhD, line 33: Could this be repurposed as a helper function like `get_data_frame()`, but something like `get_schema_explorer()`? Because if this is only every used in testing, the JSON-LD path will always be relative to `tests/data/`.

tests/test_utils.py
```diff
@@ -255,3 +258,52 @@ def test_trim_commas_df(self, local_manifest):
         trimmed_df = df_utils.trim_commas_df(df_with_nans)
 
         assert_frame_equal(trimmed_df, local_manifest)
+
+
+class TestValidateUtils:
+
+    def test_validate_schema(self, schema_explorer):
+
+        se_obj = schema_explorer("tests/data/schema_org_schemas/example.jsonld")
```
BrunoGrandePhD, line 267: If you implement `get_schema_explorer()` from my other comment, you can add a fixture in this file called `schema_explorer` that takes `helpers` as an input and produces this specific schema explorer instance.

tests/test_utils.py
```diff
@@ -255,3 +258,52 @@ def test_trim_commas_df(self, local_manifest):
...
+
+        expected = validate_utils.validate_schema(se_obj.schema)
+        
+        actual = None
+
+        assert expected == actual
```
BrunoGrandePhD, line 273: I believe you should be using `assert actual is None` rather than using the `==`. 

Source: https://stackoverflow.com/a/3257957

tests/test_utils.py
```diff
@@ -255,3 +258,52 @@ def test_trim_commas_df(self, local_manifest):
...
         assert_frame_equal(trimmed_df, local_manifest)
+
+
+class TestValidateUtils:
+
+    def test_validate_schema(self, schema_explorer):
+
+        se_obj = schema_explorer("tests/data/schema_org_schemas/example.jsonld")
+
+        expected = validate_utils.validate_schema(se_obj.schema)
```
BrunoGrandePhD, line 269: I think you meant `actual` here and vice versa.

tests/test_utils.py
```diff
@@ -255,3 +258,52 @@ def test_trim_commas_df(self, local_manifest):
...
+        mock_class["@type"] = "rdf:Property"
+        mock_class["@rdfs:comment"] = "This is a mock Patient class"
+        mock_class["@rdfs:label"] = "MockProperty"
+        mock_class["schema:domainIncludes"]["@id"] = "bts:Patient"
+
+        expected = validate_utils.validate_property_schema(mock_class)
+
+        actual = None
+
+        assert expected == actual
```
BrunoGrandePhD, line 309: Same comments about `actual` vs `expected` as well as `assert actual is None`.

BrunoGrandePhD (changes requested):

PR #395, opened 2021-02-05
Removing unneeded calls to Synapse from synapse store

milen-sage:
The calls resulted in considerable delays for datasets containing a large number of files (~10000).
For example, that causes the dataset dropdown menu to take long time to appear in the DC app (which may also slow down the appearance of the metadata template selection menu).


xdoan:
so much faster now!

xdoan (approved):

PR #396, opened 2021-02-05
Various tweaks that were pulled out of my #361 PR

BrunoGrandePhD:
These tweaks either had to be implemented or made it easier to tackle #361 PR. I've pulled out most of these tweaks from #361 to facilitate review because the one PR was getting a bit monolithic. I might not have perfectly teased apart the two sets of chances. This PR should be reviewed and merged before #361. 

## Summary

- Update testing CI workflow
- Reorganize `synapse.store` submodule as `store.synapse`
- Add stub for `BaseStorage` class
- Update dependencies
- Fix bug in `Configuration` class (name clashes between attributes and properties)
- Add pytest fixtures
- Add example data model
- Expand test configuration
- Start `test_storage` test suite
- Modularize `conftest.py`


tests/test_store.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@pytest.fixture
+def synapse_store():
+    access_token = os.getenv("SYNAPSE_ACCESS_TOKEN")
+    if access_token:
+        synapse_store = SynapseStorage(access_token=access_token)
+    else:
+        synapse_store = SynapseStorage()
+    return synapse_store
```
sujaypatil96, line 26: Maybe this should live in `conftest.py`? Since this fixture will be used by other core modules as well. It's okay to leave it here for now, I can move it in a future PR.

.github/workflows/test.yml
```diff
@@ -75,9 +75,11 @@ jobs:
       #              run test suite   
       #----------------------------------------------
       - name: Run tests
+        env:
+          SYNAPSE_ACCESS_TOKEN: ${{ secrets.SYNAPSE_ACCESS_TOKEN }}
```
sujaypatil96, line 79: Quick question, this `SYNAPSE_ACCESS_TOKEN` env variable has to do with the Synapse service account?

sujaypatil96 (approved):
Everything looks good to me. Approving this PR. Feel free to merge anytime.

🚀

tests/test_store.py
BrunoGrandePhD: I would like to avoid overloading the `conftest.py` files with fixtures that are only used in one test script. 

My philosophy is that each fixture should start in the test script where it's used, and as soon as it would be used in another test script, then it can be moved to `conftest.py`. We can discuss this tomorrow.

.github/workflows/test.yml
BrunoGrandePhD: Yup!

BrunoGrandePhD:
Thanks, @sujaypatil96! 🚀

Issue #397, opened 2021-02-05
Allow user to specify whether to use metadata attribute schema label vs. schema display name when storing annotation keys


milen-sage:
**Is your feature request related to a problem? Please describe.**
At present schematic uses schema label (e.g. CamelCase for attributes; camelCase for properties) when storing metadata attributes as annotation keys. This may not always play nicely w/ downstream services like portals (e.g. that always assume camelCase annotation keys).

**Describe the solution you'd like**
Can provide an optional flag argument in the function [here](https://github.com/Sage-Bionetworks/schematic/blob/7de7f7754335881e474402ce58cfa5cf0a9a8682/schematic/synapse/store.py#L364) to let user indicate whether to use attribute schema label or display name. The behavior could be along the lines of 
- if useSchemaLabel is True, use schema label (current behavior; set as default for backward compatibility)
- if useSchemaLabel is False, use display name; note that in this case the attribute display names should not include any characters that are not acceptable by Synapse annotation keys (e.g. space)

**How important is this feature?** Select from the options below:
• Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**When will use cases depending on this become relevant?** Select from the options below:
• Mid-term - 2-4 months

**Additional context**
@allaway does this capture our conversation accurately? Also does the priority above match up w/ your timeline/use expectation?
@ychae could you add to the respective milestone when Robert confirms priority above is accurate?

allaway:
Hi Milen,

Yes, I think this accurately describes our conversation!

This is more imminently a need for us as we'd really like to enable the end
users to apply annotations (considering their time and skillset
constraints); at present we'll probably have to manually intervene in order
to facilitate annotations.

That said, I think this is simple enough that I can tackle (hackle?) on my
end and then PR if it seems like a reasonable-quality solution. I'll
certainly post an update if this ends up being more complicated than I
expect.

milen-sage:
thanks for going through Robert! that would be awesome, please feel free to PR and tag me, Sujay or Bruno.

milen-sage:
so feature is important, urgent and easy to implement - perfect combo! (though I guess important, not urgent and easy to implement would be even better :)

PR #399, opened 2021-02-06
update store.py

allaway:
Re: #397 add useSchemaLabel param to store.associateMetadataWithFiles

This was a simpler change than I anticipated...let me know if I've missed something.  I tested this with both `useSchemaLabel = T` and `useSchemaLabel = F` on the NF staging instance of the curation app, and it worked fine, but would appreciate another test to make sure I didn't do anything weird. 

Discovered during testing: `schematic` uploads nan values for optional blank annotations, which can cause some issues when updating a manifest - will file an issue if it doesn't already exist. Based on my testing, I don't believe this issue is caused by this change.

### DIFF
--- a/schematic/synapse/store.py
+++ b/schematic/synapse/store.py
@@ -356,7 +356,7 @@ class SynapseStorage(object):
         return df, results
 
 
-    def associateMetadataWithFiles(self, metadataManifestPath: str, datasetId: str) -> str:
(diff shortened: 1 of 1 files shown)


milen-sage (approved):
thanks @allaway ! this looks good to me.

allaway:
Hah, I was just editing a bigger description in. Thanks for approving!

milen-sage:
resolves #397

milen-sage:
np - should be good to go for testing if you update the app on your end! 

happy hacking :)

PR #400, opened 2021-02-06
fix typo in store.py

allaway:
Fixing a typo. Oops!


milen-sage (approved):
good catch! :) I sometimes tend to read what I think which are often two different things :D

PR #404, opened 2021-02-09
handler for mismatched component and template type in validator

sujaypatil96:
Replica of PR #403 in `main` instead of `develop`.

### DIFF
--- a/schematic/models/metadata.py
+++ b/schematic/models/metadata.py
@@ -1,8 +1,8 @@
+import json
+import logging
+
 import pandas as pd
 import networkx as nx
-import json
-import re
-
 from jsonschema import Draft7Validator, exceptions, validate, ValidationError
 
 # allows specifying explicit variable types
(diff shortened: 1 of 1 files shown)


xdoan (approved):
Thanks @sujaypatil96 ! From tests the output looks like this in R:

 > annotation_status <- metadata_model$validateModelManifest("files/HTAN Imaging Level 2 (2).csv", "FollowUp")
========================================================================================
JSON schema successfully generated from schema.org schema!
JSON schema file log stored as /home/xdoan/ShinyApps/HTAN_Shiny_dev/schematic/data/json_schema_logs/json_schema_log.json
========================================================================================
ERROR:root:The 'Component' column value(s) ['ImagingLevel2'] do not match the selected template type 'FollowUp'.
> annotation_status
[[1]]
[[1]][[1]]
[1] 2

[[1]][[2]]
[1] "Component"

[[1]][[3]]
[1] "Component value provided is: 'ImagingLevel2', whereas the Template Type is: 'FollowUp'"

[[1]][[4]]
[[1]][[4]][[1]]
[1] "ImagingLevel2"

[[1]][[4]][[2]]
[1] "FollowUp"

which @rrchai can use in his work on the frontend. If there are any thoughts please add them!

sujaypatil96:
Thanks for testing the handler with the DC App @xdoan. It seems everything the DC App needs to read is being sent by the new handler which @rrchai can use. Do let me know if you catch something that I missed. Merging this PR for now.

PR #409, opened 2021-02-17
Tests for methods in utils/validate_utils module

sujaypatil96:
This PR adds test cases for methods in the `validate_utils.py` module.

I closed the original PR (#394), which addressed this issue. Resolving the merge conflicts as a result of multiple merges into `develop` was a bit of a hassle (apologies). I've addressed the review comments from PR #394 here.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,7 @@ import logging
 import pytest
 import pandas as pd
 
+from schematic.schemas.explorer import SchemaExplorer
 from schematic.configuration import CONFIG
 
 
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -40,6 +41,17 @@ def get_data_frame(path, *paths, **kwargs):
         fullpath = os.path.join(DATA_DIR, path, *paths)
         return pd.read_csv(fullpath, **kwargs)
 
+    @staticmethod
+    def get_schema_explorer(path, *paths):
+        fullpath = Helpers.get_data_path(path, *paths)
+        
+        if fullpath is None:
+            return SchemaExplorer()
```
BrunoGrandePhD, line 49: I'm worried that `get_data_path()` doesn't ever return `None`. You might want to use the following approach: 

```suggestion
        if path is None:
            return SchemaExplorer()
            
        fullpath = Helpers.get_data_path(path, *paths)
```

BrunoGrandePhD (approved):
You can merge once you address my one comment. 🚀

tests/conftest.py
BrunoGrandePhD: `path=None` as default?

Issue #407, opened 2021-02-15
Improved design for CSV to JSON-LD parser module


sujaypatil96:
1. Move the current `csv_utils` code to `schematic.schemas.df_parser`
2. Add a method to the `schematic.schemas.df_parser` module called parse_csv() or parse_from_csv() that:
- takes an input filepath or url to a csv and reads to a df
- calls `create_nx_schema_objects()` to parse the df into the relevant python objects
3. Add another module called `schematic.schemas.file_converter` that includes a class `SchemaFileConverter` which takes an input filepath and extension and output filepath and extension
4. `SchemaFileConverter` can include two methods:
- `read_csv()` which would call `parse_from_csv()` from above
- `write_jsonld()` which can call `SchemaExplorer.export_schema()`
5. To the `schematic.schemas` module `commands.py`, add the command convert with function `convert_file` that calls `schematic.schemas.file_converter.SchemaFileConverter` to transform the user-specified input file and format into the user-specified output file and format

E.g.: `schematic schema convert -i <path-to-input-file> -if csv -o <path-to-output-file> -of jsonld`

Note: we could also simplify by auto-detecting input/output formats based on extension…or have an argument like `-t/--transform` for which we provide a defined set of options like `csv_to_jsonld`)

Thanks @jaeddy, for the design suggestion.

PR #415, opened 2021-02-23
Repurpose examples folder

sujaypatil96:
This PR seeks to improve the structure/semantics of some of the folders in `schematic`:

- Renamed the `examples` folder to `scripts`
- Added `rfc_to_data_model` CLI (converted `argparse` utility to `click` utility)
- Removed the "examples" we previously had, since it didn't make sense to repeat some of them (ex. we already have a CLI that generates manifests)

Future work:
-  We can come up with specific use-cases for schema explorer, schema generator, etc. or maybe a combination of multiple core modules, gradually and add them to this `scripts` folder, but for now this should be okay
- If/when we implement issue #407, the existing `rfc_to_data_model.py` script might get eliminated as well (but that's probably in a future release)


milen-sage:
A quick note - people might not know what is rfc. We can discuss a more intuitive name during the code review?

PR #417, opened 2021-03-01
initial draft of documentation on readthedocs

sujaypatil96:
This PR adds documentation (`docs` folder) to the `schematic` repo. The files/commits in this PR replicate the output you see on the [Readthedocs documentation](https://sage-schematic.readthedocs.io/en/develop-readthedocs-branch/).


docs/README.rst
```diff
@@ -0,0 +1,204 @@
+Schematic
```
BrunoGrandePhD, line 1: I'm not a fan that this `README.rst` file contains duplicated content from the main `README.md` file. It's only a matter of time until they diverge in confusing ways. If we're moving to using ReadTheDocs, we should perhaps replace the content in `README.md` with a link to the documentation page.

docs/cli_reference.rst
```diff
@@ -0,0 +1,7 @@
+=============
```
BrunoGrandePhD, line 1: Nice!  🚀

docs/conf.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'Schematic'
+copyright = '2021, Sujay Patil'
+author = 'Sujay Patil'
+
+# The full version, including alpha/beta/rc tags
+release = '1.0.0'
```
BrunoGrandePhD, line 25: Can you look into whether you can update this section using the contents of `pyproject.toml`? It shouldn't be too hard since this is a Python script. There might also be precedent for this online.

BrunoGrandePhD (changes requested):

docs/conf.py
BrunoGrandePhD: This is an optional request.

docs/README.rst
sujaypatil96: Yup, you're right. We can link different sections of the GitHub README to the documentation page(s). We can think/discuss how we need to re-write the README docs.

docs/conf.py
```diff
@@ -0,0 +1,61 @@
...
+import os
+import sys
+sys.path.insert(0, os.path.abspath('..'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'Schematic'
+copyright = '2021, Sujay Patil'
+author = 'Sujay Patil'
```
sujaypatil96, line 22: Update list of authors.

sujaypatil96:
The tests that are not successful were resolved in the latest merges to `develop`, so that shouldn't be a problem.

@BrunoGrandePhD @milen-sage : I've addressed the comment on pulling data from `pyproject.toml` and filling data into `conf.py`.

requirements.txt
```diff
@@ -1,25 +1,27 @@
...
 click-log==0.3.2
-commonmark==0.9.1
-coverage==5.4
+coverage==5.5
 decorator==4.4.2
-Deprecated==1.2.11
+Deprecated==1.2.12
 docutils==0.16
 entrypoints==0.3
-google-api-core==1.25.1
```
milen-sage, line 15: just confirming - not needed?

sujaypatil96: Oh, it's needed. The version was upgraded to `google-api-core==1.26.1`, so you can see it on the right hand side of the diff.

milen-sage: ops - I missed that one :) thanks!

milen-sage (approved):

PR #418, opened 2021-03-01
Minor updates to schema reflecting requests for 1st HTAN data release

milen-sage:
One last update @xdoan ? :)

### DIFF
--- a/data/schema_org_schemas/HTAN.jsonld
+++ b/data/schema_org_schemas/HTAN.jsonld
@@ -2729,6 +2729,18 @@
                 },
                 {
                     "@id": "bts:Cloupe"
+                },
+                {
+                    "@id": "bts:Am"
+                },
+                {
+                    "@id": "bts:Cell"
+                },
+                {
+                    "@id": "bts:Mpg"
+                },
+                {
+                    "@id": "bts:M"
                 }
             ],
             "sms:displayName": "File Format",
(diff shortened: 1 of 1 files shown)


milen-sage:
if not, @sujaypatil96 can incorporate into the DC app after it is transferred to his shiny account.

xdoan:
I'm happy to do it but it'll probably get done by tomorrow as I have a lot of meetings

xdoan:
@milen-sage can you share which templates affects is in so I can make sure I test the right ones?

milen-sage:
The addition of file formats affects all assay data templates, but you can just test with one. The other one is biospecimen. Thanks!

xdoan:
Are you able to generate all the templates? Just spot checking but can't generate scRNAseq level 1 or 2 or minimal metadata due to some "KeyError: 'displayName'" but I wasn't aware of any changes in the display/schema names. @milen-sage

milen-sage:
hmm... thanks for the check; let me see on my end

milen-sage:
ok - got the error; will commit in a bit

milen-sage:
@xdoan the error should be fixed w/ the latest commit - good catch!

Schema label for Other Assay is OtherAssay.

xdoan:
yup looks like the cell am format was conflicting, I was already using the OtherAssay schema name in the config and its all resolved now!

xdoan (approved):

Issue #411, opened 2021-02-21
Speed up annotation retrieval from Synapse for large datasets


BrunoGrandePhD:
## Description

I noticed that the new annotation retrieval feature (implemented in #361) is slow, especially for large datasets. I generated a run time [profile](https://github.com/Sage-Bionetworks/schematic/files/6018362/profile.txt). It's my first time generating one of these, so bear that in mind. I'm also running into rate limits with Google's API (see #410), but I think I got the information I need. It does confirm that the annotation retrieval process is one of the slowest parts. For reference, here's the profiling command I ran:

```
poetry run python -m cProfile -s cumulative -m schematic manifest get --dataset_id syn4984604 --jsonld NF.jsonld --config tests/data/test_config.yml --sheet_url --data_type GenomicsAssay --use_annotations
```

I want to highlight lines 33-38 from the [profile](https://github.com/Sage-Bionetworks/schematic/files/6018362/profile.txt). The rows are ordered based on the time cumulatively spent in each function (see `cumtime` column). The total time spent according to the top row is 605 seconds. You can see that the one call to `getDatasetAnnotations()` takes up 527 seconds (first row below `[...]`), which represents the vast majority of the time spent during execution. 

The last row confirms that the bottleneck is indeed with the Synapse API, since `_getRawAnnotations()` is a simple wrapper around `restGET()`. You can see that `restGET()` is also included in the rows below, but we cannot conclude anything from it because that function can be called from anywhere (not necessarily for the annotation retrieval). 

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   2335/1    0.010    0.000  605.419  605.419 {built-in method builtins.exec}
   [...]
        1    0.001    0.001  527.485  527.485 synapse.py:512(getDatasetAnnotations)
     3513    0.030    0.000  489.277    0.139 sessions.py:546(get)
        1    0.033    0.033  487.881  487.881 synapse.py:527(<listcomp>)
     3505    0.131    0.000  487.849    0.139 synapse.py:478(getFileAnnotations)
     3508    0.029    0.000  487.582    0.139 client.py:3766(restGET)
     3505    0.104    0.000  487.298    0.139 client.py:1291(_getRawAnnotations)
```

While I'm looking at the profile, I also noticed that `getFilesInStorageDataset()` also takes up a non-trivial amount of time for a large dataset. On average, it takes 40 seconds to retrieve a list of 3505 files (see `percall` column). 

[...]

PR #419, opened 2021-03-03
Speed up annotation retrieval (for pre-populating empty manifests)

### LINKED ISSUE #424: annotations supplant valid values when previous annotations exist
**Describe the bug**
When a key/column has annotation values already defined for a file that is within the manifest, the list of valid values in the resulting manifest is populated using the unique list of column annotations, not the valid values from the schema.

**To Reproduce**

1. Use the current `develop` branch. 
2. Generate a manifest retreving previous annotations from the CLI eg: `schematic manifest get --title Soragni-Imaging --dataset_id syn18152772 --jsonld ~/Downloads/NF.jsonld --sheet_url True --data_type ImagingAssay --use_annotations`

**Expected behavior**
I expect to have the previous annotations pulled into the cells, but for the "valid values" dropdown to reflect the valid values in the schema. 

Here's an example of a manifest generated using the previous approach (which works, but doesn't retrieve the files' current annotations): https://docs.google.com/spreadsheets/d/1wdObZxBfKeb1S0_MPlbOP8SGnYsYbl5V4fuhRd0u-l4/edit#gid=0

Here's an example of a manifest generated using new approach (which retrieves the current files' annotation, but doesn't respect the JSON-LD valid values in all columns): https://docs.google.com/spreadsheets/d/1JPlIVRwwmwhmYbTAgx_hjZqTznG0_gHRBf1QaJjZoOE/edit#gid=0

I added @BrunoGrandePhD as the assignee because I already chatted with him about this over Slack, but please reassign if necessary!

BrunoGrandePhD:
I added support for using file views to batch-retrieve annotations for large datasets, which I've tentatively defined as datasets containing over 50 files/folders. Compared to my benchmarking in #411, the code is much faster. Here's an updated [profile.txt](https://github.com/Sage-Bionetworks/schematic/files/6078989/profile.txt). For context, the new "batch mode" is implemented in `getDatasetAnnotationsBatch()` and is called from within `getDatasetAnnotations()`. 

The `getDatasetAnnotations()` function only takes 63 seconds now (instead of 527 seconds using the original approach, or almost 9 minutes). In fact, the annotation retrieval isn't the bottleneck anymore. The `getDatasetAnnotationsBatch()` call only took 15 seconds. From what I can tell, most of the time in `getDatasetAnnotations()` was spent on running `getFilesInStorageDataset()`, which takes 45 seconds on average. This could be cut down significantly by relying on the file view itself to build a list of dataset files, but that'll have to wait for a separate PR. 

Another easy point of improvement is the `getDatasetManifest()` function, which also calls `getFilesInStorageDataset()`. I believe @milen-sage has a more efficient version of `getDatasetManifest()` laying around, so once that's reviewed and merged, that should speed things up by roughly ~40% (once this PR is merged). 

(🌺 Edited by @allaway: Closes #424)


milen-sage:
Yes, thanks @BrunoGrandePhD for the reminder - I'll commit the more efficient getDatasetManifest soon if not today - tomorrow.

BrunoGrandePhD:
The tests all pass locally. I think the failures on GitHub are due to concurrency because each worker is creating/getting the same file view and one of them deletes it before the others can query it. I can readily resolve this by including a random string in the file view name. This works for now since these file views aren't intended to be user-facing and are deleted shortly after being created. 

Not sure if anyone has a better solution to prevent this kind of clash. 

image

BrunoGrandePhD:
Remaining tasks for @BrunoGrandePhD: 

- [ ] Run `cProfile` benchmark using minimal data model and NF's large dataset
- [ ] Resolve pending merge conflicts

Then, this PR can be merged! 🚀

milen-sage:
Thank you for the detailed notes and benchmarking @BrunoGrandePhD !

There is daily and weekly variability in Google's API response time (independent of batch vs. individual requests); however reading through the timing you posted above ~30s to get an empty manifest is a lot: we might want to doublecheck if we are not doing something wrong. Agreed that requires more profiling and a separate PR.

It looks like you can check the box [here](https://github.com/Sage-Bionetworks/schematic/pull/419#issuecomment-806165585) and we can merge during/after today's CR :) ?

milen-sage (approved):
Conceptually the code makes sense; may want to code review in detail w/ @sujaypatil96 

A note to our future selves: we might want to change the names of functions like getStorageDatasetsInProject to something more simple like getProjectDatasets (to keep it less verbose similarly to the function getDatasetProject in this PR)

sujaypatil96:
Thank you for the detailed testing and benchmarking on the speed up @BrunoGrandePhD. I ran the above command on the NF data model as well, and got [similar benchmarking results](https://github.com/Sage-Bionetworks/schematic/files/6274030/profile-1.txt).

Overall runtime was ~106 s.

pyproject.toml
```diff
@@ -52,7 +52,7 @@ pygsheets = "^2.0.4"
 PyYAML = "^5.4.1"
 rdflib = "^5.0.0"
 setuptools = "^52.0.0"
-synapseclient = "2.2.2"
+synapseclient = "^2.3.0"
```
sujaypatil96, line 55: Not sure if we want to make this update just yet? The OAuth integration feature in the Data Curator App is still being [tested](https://github.com/Sage-Bionetworks/data_curator/tree/shiny-server-oauth) and we won't be ready to make this `synapseclient` transition until that's done. The reason is because version 2.3.0 deprecates the `sessionToken` parameter for the `login()` method which is currently the way in which the Data Curator App logs in.

sujaypatil96 (approved):
🚀

BrunoGrandePhD:
I'm not sure what has caused the tests to start failing in my last two commits. I'll investigate tomorrow.

allaway:
🚀 

I checked out this PR and it fixes the issue I was encountering before. As you note, the `Filename` is not populating in this branch but hopefully between this and #425, we'll be able to update our production version in NF. Thanks so much! (insert party parrots and TYSM slackmojis)

allaway:
Also, this branch is blazing fast in manifest generation :)

pyproject.toml
BrunoGrandePhD: @sujaypatil96: I was going to downgrade back to version 2.2, but I realized that I actually need a feature that's been implemented in 2.3. Since session token support has been deprecated but not removed yet, I'll pin the version to `~2.3` ([source](https://python-poetry.org/docs/dependency-specification/)) to avoid updates to a version of `synapseclient` that removes support for session tokens.

PR #421, opened 2021-03-04
Restrict annotation retrieval to files and folders

BrunoGrandePhD:
With the new code that uses folder-bound file views, this update is required to restrict the annotations retrieval to files and folders. Specifically, this would cause issues when running the test suite, which didn't expect a fourth entity (a temporary file view) in the test dataset folder. 

Briefly, I'm using `syn.get(..., downloadFile=False)` instead of `syn._getAnnotations()` so that I can simultaneously retrieve the annotations and the entity type. This way, I can ignore anything that isn't a file or folder. This should make the test suite more robust. 

I also added a permanent folder-bound [file view](https://www.synapse.org/#!Synapse:syn24226514) to our main test dataset to make sure that this new functionality (_i.e._ to ignore them) works.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -488,23 +488,33 @@ class SynapseStorage(BaseStorage):
         Returns:
             dict: Annotations as comma-separated strings.
         """
-        # Retrieve Synapse annotations using _getRawAnnotations()
-        # to get etag, which isn't provided by get_annotations()
-        syn_annotations = self.syn._getRawAnnotations(fileId)
 
-        # Convert all values into comma-separated lists of strings
-        annotations_unordered = from_synapse_annotations(syn_annotations)
-        annotations = OrderedDict()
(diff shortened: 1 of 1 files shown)


BrunoGrandePhD:
@sujaypatil96: While the tests in this PR passed, I ran into errors in #419. I realized that it was possible for the code to fail if a temporary file view is deleted between the moment its Synapse ID is found and when its annotations are retrieved. For now, I'm just skipping them. 

Most of these issues didn't come up locally because I only have one pytest thread. However, on GitHub, we were running into concurrency issues. Thanks for bearing with me! 

It looks like both this PR and #419 passed all tests. This PR is ready for review now.

sujaypatil96 (approved):

sujaypatil96:
Awesome! Thanks for the fix @BrunoGrandePhD. Tests don't seem to be held up on the other PRs now.

PR #422, opened 2021-03-04
Add section to clarify how to update Synapse test resources

BrunoGrandePhD:


BrunoGrandePhD:
I following this approach for #425, where I had to update the annotations again on the test entities on Synapse.

milen-sage (approved):

PR #423, opened 2021-03-04
Fix documentation

allaway:
model:input:log_location has to be a path to a json file, not to a folder. (otherwise you get this error: IsADirectoryError: [Errno 21] Is a directory: 'data/json_schema_logs/')


sujaypatil96 (approved):
Good catch! Thanks for modifying the description here @allaway. The new description is more appropriate.

PR #426, opened 2021-03-05
Order of arguments being passed to ManifestGenerator call

sujaypatil96:
This was the reason why I wasn't able to get the current version of the code in `develop` to work with the Data Curator App. The Data Curator App makes a call to [`getModelManifest()`](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/models/metadata.py#L108), and the order of the arguments being passed to `ManifestGenerator()` was not correct.

The `path_to_jsonld` argument needs to be passed first, if you look at the [`ManifestGenerator()`](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/manifest/generator.py#L21) constructor.

### DIFF
--- a/schematic/manifest/commands.py
+++ b/schematic/manifest/commands.py
@@ -76,8 +76,8 @@ def get_manifest(ctx, title, data_type, jsonld, dataset_id, sheet_url,
 
     # create object of type ManifestGenerator
     manifest_generator = ManifestGenerator(
-        title=title,
         path_to_json_ld=jsonld,
+        title=title,
         root=data_type,
         use_annotations=use_annotations,
     )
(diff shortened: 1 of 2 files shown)


sujaypatil96:
@BrunoGrandePhD, @milen-sage: sorry for requesting a review before fully making sure it was ready. I didn't think to make sure the order of the arguments was corrected in all modules (which are addressed in the commits after I requested for a review). The PR is ready for a review now though.

BrunoGrandePhD (approved):

BrunoGrandePhD:
🚀 

@sujaypatil96: No worries about the premature review request! As long as you post a comment to confirm that it's ready when it does occur, that's fine!

PR #428, opened 2021-03-08
Develop faster manifest fetch

milen-sage:
Reduce overhead of looking up manifest: no longer iterating over all files in a dataset (and walking dataset recursively...). Instead look up manifest for the dataset id argument directly in main fileview.


milen-sage:
@BrunoGrandePhD you'll likely note yourself, but just adding for completion, this PR also fetches the change mentioned in #424 (though not sure if that would be the fix for the issue there).

schematic/store/synapse.py
```diff
@@ -177,7 +177,8 @@ def getStorageDatasetsInProject(self, projectId: str) -> List[str]:
             foldersTable = self.storageFileviewTable[(self.storageFileviewTable["contentType"] == "dataset") & (self.storageFileviewTable["projectId"] == projectId)]
             areDatasets = True
         else:
-            foldersTable = self.storageFileviewTable[(self.storageFileviewTable["type"] == "folder") & (self.storageFileviewTable["projectId"] == projectId)]
+            #foldersTable = self.storageFileviewTable[(self.storageFileviewTable["type"] == "folder") & (self.storageFileviewTable["projectId"] == projectId)]
+            foldersTable = self.storageFileviewTable[(self.storageFileviewTable["type"] == "folder") & (self.storageFileviewTable["parentId"] == projectId)]
```
BrunoGrandePhD, line 181: This is another instance of, "wasn't this the way schematic always worked?" If I'm interpreting this chance correctly, this is now restricting the dataset to only those in the root of a given project ID. I thought this was already the case?

schematic/store/synapse.py
```diff
@@ -250,25 +251,28 @@ def getDatasetManifest(self, datasetId: str, downloadFile: bool = False) -> List
...
-            synapseclient.entity.File: A new Synapse Entity object of the appropriate type.
+            synapseclient.entity.File: A new Synapse Entity object of the appropriate type, if downloadFile is set to True
         """
 
         # get a list of files containing the manifest for this dataset (if any)
-        manifest = self.getFilesInStorageDataset(datasetId, fileNames = [os.path.basename(self.manifest)])
-
-        if not manifest:
+        all_files = self.storageFileviewTable
+        manifest = all_files[(all_files["name"] == os.path.basename(self.manifest)) & (all_files["parentId"] == datasetId)]
```
BrunoGrandePhD, line 259: I'm going to test this function locally, but I wanted to issue this review in the interim.

schematic/manifest/generator.py
```diff
@@ -585,8 +585,8 @@ def get_empty_manifest(self, json_schema_filepath = None):
 
 
             # generating sheet api request to populate a dropdown or a multi selection UI
-            if len(req_vals) > 10 and not "list" in validation_rules:
-                # if more than 10 values in dropdown use ONE_OF_RANGE type of validation since excel and openoffice
+            if len(req_vals) > 0 and not "list" in validation_rules:
+                # if more than 10 values in dropdown use ONE_OF_RANGE type of validation since excel and openoffice
```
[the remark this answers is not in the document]
BrunoGrandePhD: Based on this [comment](https://github.com/Sage-Bionetworks/schematic/issues/424#issuecomment-792964288), I realize that I probably saw this change being merged into `main`, not `develop`. Is this correct?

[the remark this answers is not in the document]
milen-sage: yes, I think it got to main, but I didn't push to develop (thinking develop would merge w/ main eventually).

schematic/store/synapse.py
milen-sage: true projectId is redundant here - let me change that and re-commit

milen-sage: actually looking at the line here and realizing this is in a different function (getStorageDatasetsInProject) and not the manifest retrieval one, scoping by projectId is needed there. 

In the general case, the main/master fileview may span multiple projects and a user might have access to a subset of those; getStorageDatasetsInProject fetches only the datasets in a specific project.

BrunoGrandePhD:
I'm running into numpy compilation issues, which didn't happen previously, and I have to deal with a blocker on HTAN (AWS S3 buckets). I'll revisit this PR early next week.

PR #429, opened 2021-03-10
Integrate CSV to JSON-LD parser into schematic core modules

sujaypatil96:
Based on the approach in issue #407, I decided to skip 3. and 4. since it was making the application unnecessarily verbose.

To test the CLI bundled with this PR:

`poetry run schematic schema --config ~/path/to/config.yml convert --help`

PR #432, opened 2021-03-16
init command added to CLI catalogue

sujaypatil96:
Note: This PR should be reviewed/merged before PR #429.

This PR adds a new command line utility to our existing list of CLI tools: `schematic init`.

The idea to add the `init` command was based on recommendations by @BrunoGrandePhD (Thanks Bruno), in PR #392. The `init` command takes the following arguments:`--auth` and `--config`. Using the `--auth` optional argument you can control what mode of authentication you want to use, i.e., either `OAuth` or `Service Account`, and you can specify how the files are to be configured using `config.yml`. By doing this you would essentially be eliminating the [`Obtain Credentials File(s)`](https://github.com/Sage-Bionetworks/schematic/tree/develop#134-obtain-credentials-files) step from the setup process.

I didn't want to delay pushing this code, so I haven't added tests for this CLI. I will address that in a future PR.

### DIFF
--- a/schematic/help.py
+++ b/schematic/help.py
@@ -102,8 +102,8 @@ init_command = {
             "Initialize mode of authentication for schematic."
         ),
         "auth": (
-            "Specify the mode of authentication you want to use. You can use one of "
-            "either 'token' or 'service_account'. The default mode of authentication "
+            "Specify the mode of authentication you want to use for Google accounts. "
+            "You can use one of either 'token' or 'service_account'. The default mode of authentication "
             "is 'token' which uses OAuth."
         ),
         "config": (
--- a/schematic/help.py
+++ b/schematic/help.py
@@ -93,3 +93,21 @@ schema_commands = {
         }
     }
 }
+
+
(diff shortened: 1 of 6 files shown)


BrunoGrandePhD:
Don't wait on my review for this PR. I'm dealing with a blocker this week, and once that's done, I'll have other stuff to catch up on.

schematic/help.py
```diff
@@ -93,3 +93,21 @@
...
+
+
+# `schematic init` command description
+init_command = {
+    "init": {
+        "short_help": (
+            "Initialize mode of authentication for schematic."
+        ),
+        "auth": (
+            "Specify the mode of authentication you want to use. You can use one of "
```
milen-sage, line 105: Can we specify which service the authentication is for (the user could be confused whether this is auth for Google Sheet API, synapse, or something else).

milen-sage (changes requested):

milen-sage (approved):

PR #437, opened 2021-03-24
schematic readme to point to RTD docs

sujaypatil96:
The `schematic` [README](https://github.com/Sage-Bionetworks/schematic/tree/develop-README) is now a lot smaller and concise, with links and references to the [RTD documentation](https://sage-schematic.readthedocs.io/en/develop-readme/). Per the recommendations on ReadTheDocs I've also added a `.readthedocs.yaml` file to configure how the docs should be built.


sujaypatil96:
Updated the link to [.synapseConfig](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v2.2.2-rc/synapseclient/.synapseConfig) file since the Data Curator App (and consequently `schematic`) are/will be compatible only with `synapseclient==2.2.2` or lower.

sujaypatil96:
@BrunoGrandePhD: So the RTD docs weren't getting built without including [`sphinx-click`](https://github.com/click-contrib/sphinx-click) (a package to automatically document sphinx-based applications) in the `requirements.txt` file. So I cut down the `requirements.txt` file to include only that package. So while building the docs, RTD, includes a bunch of packages/dependencies (refer to this [build](https://readthedocs.org/projects/sage-schematic/builds/13336337/)), but `sphinx-click` is not one of them. So it needs to be passed separately.

The other commit I added had to do with updating the TOC, which is fairly simple.

README.md
```diff
@@ -1,151 +1,54 @@
...
-```bash
-synapse get syn24214983 # download copy of schematic_service_account_creds.json file
-```
-
-Use the `schematic_service_account_creds.json` file for the service account mode of authentication (_for Google services/APIs_).
-
-Note: The `Selection Options` dropdown which allows the user to select multiple values in a cell during manifest annotation [does not work](https://developers.google.com/apps-script/api/concepts) with the service account mode of authentication.
-
-### 1.3.5. Fill in Configuration File(s)
+- [Schematic](#schematic)
```
BrunoGrandePhD, line 5: [Nit:](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.) I would remove this line and de-indent the rest.

BrunoGrandePhD (approved):
One optional comment. Otherwise, 🚀

README.md
sujaypatil96: So this markdown extension I'm using called "Markdown All in One" for VSCode generates the TOC automatically and it recommends including all headers in the TOC, so I've stuck with that.

BrunoGrandePhD: I've used that extension before. You can specify the levels in the VS Code's JSON settings as follows. In this case, you would only be showing level-2 to level-4 headings, effectively skipping the `# Schematic` level-1 heading. This can wait until you next update the README. 

```
"markdown.extension.toc.githubCompatibility": true,
"markdown.extension.toc.levels": "2..4",
```

Issue #410, opened 2021-02-21
Rate limit exceeded with Google's API on large dataset


BrunoGrandePhD:
## Description of the Bug

I tried generating a manifest template for a large NF dataset ([syn4984604](https://www.synapse.org/#!Synapse:syn4984604)). According to `SynapseStorage.getFilesInStorageDataset()`, there were 3505 files in this dataset (_i.e._ `len(file_list)`). However, schematic ran into rate limits imposed by the Google API. This appears related to the recently opened Sage-Bionetworks/data_curator#110. 

I tried running the command using a small dataset ([syn11387814](https://www.synapse.org/#!Synapse:syn11387814)) because the traceback references [line 479](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/manifest/generator.py#L469-L479) and I suspected that I was running into the rate limits due to the number of columns and not the number of rows. On the first try, I ran into an error, but not on subsequent tries. The error on the first try might have been caused by lingering rate limits imposed on my IP address. 

I wonder how much room there is to minimize the number of Google API requests by bundling the changes into fewer but larger changes. For example, can a single JSON request object be built during this [for-loop](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/manifest/generator.py#L469-L479) and then one HTTP request is issued to Google's API? 

Here's the error message (the full output is included further below):

```
"Quota exceeded for quota metric 'Write requests' and limit 'Write requests per minute per user' of service 'sheets.googleapis.com' for consumer 'project_number:269908437164'.". Details: "[{'@type': 'type.googleapis.com/google.rpc.ErrorInfo', 'reason': 'RATE_LIMIT_EXCEEDED', 'domain': 'googleapis.com', 'metadata': {'quota_limit': 'WriteRequestsPerMinutePerUser', 'quota_metric': 'sheets.googleapis.com/write_requests', 'consumer': 'projects/269908437164', 'service': 'sheets.googleapis.com'}}]"
```

### Command

**N.B.** `NF.jsonld` was downloaded from [here](https://github.com/nf-osi/schematic/blob/develop/NF.jsonld). 

```
poetry run python -m schematic manifest get --dataset_id syn4984604 --jsonld NF.jsonld --config tests/data/test_config.yml --sheet_url --use_annotations --data_type GenomicsAssay
```

### Output

```
Starting schematic...
warning: The `use_annotations` option is currently only supported when there is no manifest file for the dataset in question.
[...]

PR #439, opened 2021-03-26
create batch request for Google API to avoid hitting rate limit

sujaypatil96:
This PR fixes issue #410. It repeats the code from within the [for loop](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/manifest/generator.py#L469-L479) but only in order to create a "request body" so we can issue a single [`batchUpdate()`](https://github.com/Sage-Bionetworks/schematic/blob/develop-rate_limit-fix/schematic/manifest/generator.py#L492) API request. We can now execute the command from within the referenced issue to test, and see it works on large datasets.

Benchmarking results: for the `NF.jsonld` data model to generate a metadata manifest template using existing Synapse annotations, takes about ~10 mins.

Tagging @BrunoGrandePhD and @milen-sage for a code review, and @allaway for a feature review.

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -606,8 +606,8 @@ class ManifestGenerator(object):
 
 
             # generating sheet api request to populate a dropdown or a multi selection UI
-            if len(req_vals) > 10 and not "list" in validation_rules:
-                # if more than 10 values in dropdown use ONE_OF_RANGE type of validation since excel and openoffice
+            if len(req_vals) > 0 and not "list" in validation_rules:
+                # if more than 0 values in dropdown use ONE_OF_RANGE type of validation since excel and openoffice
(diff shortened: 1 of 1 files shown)


milen-sage (approved):
Great - we can iterate on refactoring manifest generation to decouple google API requests from other logic throughout the module, but this should work well for now (at least for HTAN; I am not sure what the NF latency was like before the fix - 10 minutes sounds quite a bit still)!

allaway:
@sujaypatil96, do you have more details on this: 

>Benchmarking results: for the NF.jsonld data model to generate a metadata manifest template using existing Synapse annotations, takes about ~10 mins.

Just curious what dataset you generated a manifest for that took that long. I'm guessing it's a pretty "large" dataset compared to the majority of what we work with :)

PR #451, opened 2021-04-14
modified getDatasetManifest() signature

sujaypatil96:
There are three major changes that are addressed in this PR (apologies for not splitting it up into multiple commits):

- Modified the signature of the [`getDatasetManifest()`](https://github.com/Sage-Bionetworks/schematic/blob/develop-dca_compatibility-fix/schematic/store/synapse.py#L245) method to return only the [synapse ID](https://github.com/Sage-Bionetworks/schematic/blob/develop-dca_compatibility-fix/schematic/store/synapse.py#L286) of an existing manifest in a dataset (not a tuple of the synapse ID and file path).

@milen-sage's additions:
- `update_df` needs to be improved to handle more use cases, so removed the piece of code responsible for the update.
- avoid errors when we submit annotation values larger than 500 characters ([code change](https://github.com/Sage-Bionetworks/schematic/blob/develop-dca_compatibility-fix/schematic/store/synapse.py#L462)).

Note: The code formatting has been updated using [yapf](https://github.com/google/yapf) by Google.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -34,6 +34,7 @@ logger = logging.getLogger(__name__)
 class SynapseStorage(BaseStorage):
     """Implementation of Storage interface for datasets/files stored on Synapse.
     Provides utilities to list files in a specific project; update files annotations, create fileviews, etc.
+
     TODO: Need to define the interface and rename and/or refactor some of the methods below.
     """
     def __init__(
@@ -96,10 +97,13 @@ class SynapseStorage(BaseStorage):
         except ValueError:
             raise MissingConfigValueError(("synapse", "master_fileview"))
 
+
     def getPaginatedRestResults(self, currentUserId: str) -> Dict[str, str]:
         """Gets the paginated results of the REST call to Synapse to check what projects the current user has access to.
+
         Args:
             currentUserId: synapse id for the user whose projects we want to get.
+        
         Returns:
             A dictionary with a next page token and the results.
         """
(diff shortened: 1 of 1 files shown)


sujaypatil96:
The recent commit was simply to revert the code formatted using yapf. We can decide on which code formatter we want to use ([yapf](https://github.com/google/yapf) / [black](https://github.com/psf/black)) and then format our files accordingly.

schematic/store/synapse.py
```diff
@@ -440,22 +461,27 @@ def associateMetadataWithFiles(self, metadataManifestPath: str, datasetId: str)
...
 
             #  prepare metadata for Synapse storage (resolve display name into a name that Synapse annotations support (e.g no spaces)
             metadataSyn = {}
             for k, v in row.to_dict().items():
                 keySyn = se.get_class_label_from_display_name(str(k))
 
+                # truncate annotation values to 500 characters if the
+                # size of values is greater than equal to 500 characters
+                if isinstance(v, str) and len(v) >= 500:
+                    v = v[0:499]
```
sujaypatil96, line 483: Truncate the string like `[0:473][truncatedByDataCuratorApp]`.

milen-sage (approved):
just the nit around truncation; ow looks good!

sujaypatil96:
Added an explicit truncated message at the end of truncated annotation cell values in this [commit](https://github.com/Sage-Bionetworks/schematic/pull/451/commits/51cbcdbbd1443aac18a83f9aecb59650c307f12f).

BrunoGrandePhD:
🚀

PR #454, opened 2021-04-16
Configuration options for flake8 and black

sujaypatil96:
I've added a basic `.flake8` configuration file. The error codes you see listed in the `.flake8` file were ignored for compliance with PEP8 in a lot of articles (blog posts) that I read.

Added a `[tool.black]` configuration section for the `black` package as well.

There's an additional step I added (not enabled yet) in the GitHub Actions workflow that we could use? It's a step to check the formatting quality of the code.

CC: @BrunoGrandePhD and @milen-sage

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,8 @@
+[flake8]
+ignore = 
+        # E*** / W*** pycodestyle codes
+        # whitespace before ','
+        E203,
+        # line break before binary operator
+        W503
+max-line-length = 88
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,6 +70,16 @@ jobs:
       #     flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
       #     # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
       #     flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+
+      #----------------------------------------------
+      #             check formatting
+      #----------------------------------------------
+      # Disabled until we agree to turn it on
+      # - name: Code formatting with black
+      #   run: |
+      #     # run black in check mode
+      #     # if files are not formatted correctly, the build will not go through
+      #     black . --check
       
       #----------------------------------------------
       #              run test suite   
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,8 +61,25 @@ pytest = "^6.2.2"
 pytest-cov = "^2.11.1"
 pytest-mock = "^3.5.1"
 flake8 = "^3.8.4"
-black = {version = "^20.8b1", allow-prereleases = true}
 python-dotenv = "^0.15.0"
+black = {version = "^20.8b1", allow-prereleases = true}
+
+
+[tool.black]
+line-length = 88
+target-version = ['py37']
+include = '\.pyi?$'
+exclude = '''
+
+(
+  /(
+      \.eggs
+    | \.git
+    | \.venv
+    | dist
+  )/
+)
+'''
 
 
 [build-system]
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,74 @@
+# Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of experience,
+nationality, personal appearance, race, religion, or sexual identity and
+orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+  address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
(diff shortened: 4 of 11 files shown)


.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
+ignore = E203, E266, E501, W503, F403, F401
+max-line-length = 79
```
BrunoGrandePhD, line 3: This should be 88. [Source](https://black.readthedocs.io/en/stable/compatible_configs.html#flake8).

.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
+ignore = E203, E266, E501, W503, F403, F401
```
BrunoGrandePhD, line 2: I'm seeing E203 and W503 suggested [here](https://black.readthedocs.io/en/stable/compatible_configs.html#flake8). Can you post a reference for the additional errors/warnings that would be ignored? I'm tempted to include this link as a comment in the flake8 config.

.flake8
```diff
@@ -0,0 +1,5 @@
...
+max-line-length = 79
+max-complexity = 18
+select = B,C,E,F,W,T4,B9
```
BrunoGrandePhD, line 5: Don't forget to include empty lines at the end of your files. You can configure VS Code to add these automatically.

.github/workflows/test.yml
```diff
@@ -70,6 +70,16 @@ jobs:
...
+
+      #----------------------------------------------
+      #             check formatting
+      #----------------------------------------------
+      # Disabled until we agree to turn it on
+      # - name: Code formatting with black
+      #   run: |
+      #     # run black in check mode
+      #     # if files are not formatted correctly, the build will not go through
+      #     black . --check
```
BrunoGrandePhD, line 82: Can you confirm that flake8 does anything that `black --check` doesn't already do?

pyproject.toml
```diff
@@ -61,8 +61,31 @@ pytest = "^6.2.2"
...
+    | \.hg
+    | \.mypy_cache
+    | \.tox
+    | \.venv
+    | _build
+    | buck-out
+    | build
+    | dist
+  )/
+)
```
BrunoGrandePhD, line 87: You can remove the directories that aren't relevant to our repo. Build the package with poetry to see what gets generated and should be ignored.

BrunoGrandePhD (changes requested):

.flake8
sujaypatil96: I was trying to follow this [tutorial](https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/) as reference when configuring `black` and `flake8` ensuring their compatibility (and that there are no conflicts).

On there, they recommended configuring `flake8` ([.flake8](https://github.com/Sage-Bionetworks/schematic/blob/develop-black-flake8/.flake8)) in the way as above. After you pointed it out, I realized that the two configurations shouldn't conflict (e.g., both `black` and `flake8` should have line lengths of 79 or 88 or anything else, but they should be the same). So to ensure that it is compatible with `black`, I've changed it back to 88.

.flake8
sujaypatil96: I've posted a link to the reference I used in the comment above.

I realized that the "strictness" that is recommended in the reference is pretty severe. I've removed the `select` key which selects the `pystyleguide`, `PyFlakes`, etc. [error codes](https://flake8.pycqa.org/en/2.6.0/warnings.html) and ignores a select subset of them as above. I wasn't sure how the [McCabe complexity threshold](https://flake8.pycqa.org/en/2.5.5/config.html#settings) (`max-complexity`) is important, so I've removed that requirement as well.

.github/workflows/test.yml
sujaypatil96: From what I've gleaned, `black` is a code formatter and `flake8` is a PEP8 compliance checker. So running `black --check` would check if the code has been formatted per `black` configuration, and `flake8` would lint the code per the [PEP8 specification](https://pep8.readthedocs.io/en/latest/).

When I ran `flake8 schematic` on our core modules there were plenty of linting errors that came up, so it'll probably take a some time to fix the code per the error codes. Maybe it'd be a good idea to turn linting on after fixing the currently returned linting errors? I think we can turn on the black code formatter in the meantime, while we're fixing the linting errors.

pyproject.toml
sujaypatil96: To add `black` to our repo I ran: `poetry add --dev --allow-prereleases black`

Running `poetry build` doesn't add the `[tool.black]` section to the `pyproject.toml` file, so I had to add it manually from [here](https://black.readthedocs.io/en/stable/pyproject_toml.html).

.github/workflows/test.yml
BrunoGrandePhD: I just figured that once code is formatted by `black`, it's automatically PIP8-compliant. But I could be wrong, so having both can't hurt. You have a source describing this workflow, so I won't push on only using `black`. 

I'm guessing that most `flake8` linting errors will disappear once we format the code. A follow-up PR could address the remaining errors.

pyproject.toml
BrunoGrandePhD: I think you missed my point. 

This block contains files/folders that will never appear in our folder (like `.hg`). I was suggesting that you build the package using `poetry build` and see which files appear in the folder (like `dist/`) and only include these in this ignore list.

.flake8
```diff
@@ -1,5 +1,8 @@
...
-max-line-length = 79
-max-complexity = 18
-select = B,C,E,F,W,T4,B9
\ No newline at end of file
+ignore = 
+        # E*** / W*** pycodestyle codes
+        # whitespace before ','
+        E203,
+        # line break before binary operator
+        W503
```
BrunoGrandePhD, line 7: Perfect!

BrunoGrandePhD (approved):
🚀

pyproject.toml
BrunoGrandePhD: I saw that you did do the change that I asked, so ignore my above comment. Looks good!

.github/workflows/test.yml
sujaypatil96: Hm..as mentioned by many people in various [blog posts](https://luminousmen.com/post/my-unpopular-opinion-about-black-code-formatter), black does not fully remove all the PEP8 linting errors that come up when you run your code through a strict linter such as `flake8`. It removes a lot of the warnings and errors, but not all. So we have do one round of refactor even after running it through `black` before we enforce `flake8`. Not sure if I'm making sense, but we can discuss more about this during code review.

I'll mark this as resolved for now.

PR #456, opened 2021-04-20
add use_annotations to getModelManifest

allaway:
Surfacing the `use_annotations` parameter to make it usable in the Shiny app, which calls `getModelManifest`.


allaway:
I'm still struggling to get the annotations to actually populate in the google sheet, but based on the logs I think this is a frontend bug, so I've filed that in the nf_data_curator repository: https://github.com/nf-osi/NF_data_curator/issues/75

schematic/models/metadata.py
```diff
@@ -105,13 +105,14 @@ def getOrderedModelNodes(self, rootNode: str, relationshipType: str) -> List[str
         return ordered_nodes
 
 
-    def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, filenames: list = None) -> str:
+    def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, filenames: list = None, useAnnotations: bool = False) -> str:
```
allaway, line 108:
```suggestion
    def getModelManifest(self, title: str, rootNode: str, datasetId: str = None, jsonSchema: str = None, filenames: list = None, useAnnotations: bool = False) -> str:
```

allaway:
I made a companion PR in the shiny app: https://github.com/Sage-Bionetworks/data_curator/pull/157

sujaypatil96 (changes requested):
Thank you @allaway for the PR. It's definitely important to make sure the Data Curator App has access to the `useAnnotations` parameter. I've suggested a few changes on the PR here, I wanted to know what your thoughts were on that? Let me know if you can't find the time to fix this and I can take it up.

schematic/models/metadata.py
```diff
@@ -126,12 +127,20 @@ def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, fi
...
+                               additional_metadata=additionalMetadata,
+                               use_annotations=useAnnotations)
 
-        if jsonSchema:
-            return mg.get_manifest(json_schema=jsonSchema)
+        if datasetId and jsonSchema:
+            return mg.get_manifest(dataset_id=datasetId, json_schema=jsonSchema, sheet_url = sheetUrl)
+        
+        elif jsonSchema:
+            return mg.get_manifest(json_schema=jsonSchema, sheet_url = sheetUrl)
```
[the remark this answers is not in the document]
allaway: I agree that this makes more sense. I took the format of the code before and just ran with it to avoid breaking stuff, but was wondering whether this was all necessary. Thanks for the review and revisions!

schematic/models/metadata.py
```diff
@@ -126,12 +127,20 @@ def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, fi
         mg = ManifestGenerator(path_to_json_ld=self.inputMModelLocation, 
                                title=title, 
                                root=rootNode, 
-                               additional_metadata=additionalMetadata)
+                               additional_metadata=additionalMetadata,
+                               use_annotations=useAnnotations)
 
-        if jsonSchema:
-            return mg.get_manifest(json_schema=jsonSchema)
+        if datasetId and jsonSchema:
```
allaway, line 133:
```suggestion
        if datasetId:
```

schematic/models/metadata.py
```diff
@@ -126,12 +127,20 @@ def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, fi
...
+            return mg.get_manifest(dataset_id=datasetId, json_schema=jsonSchema, sheet_url = sheetUrl)
+        
+        elif jsonSchema:
```
allaway, line 136:
```suggestion
```

schematic/models/metadata.py
allaway, line 137:
```suggestion
```

schematic/models/metadata.py
```diff
@@ -126,12 +127,20 @@ def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, fi
...
+            return mg.get_manifest(json_schema=jsonSchema, sheet_url = sheetUrl)
+
+        elif datasetId:
```
allaway, line 139:
```suggestion
```

schematic/models/metadata.py
```diff
@@ -126,12 +127,20 @@ def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, fi
...
+            return mg.get_manifest(dataset_id=datasetId, sheet_url = sheetUrl)
```
allaway, line 140:
```suggestion
```

schematic/models/metadata.py
```diff
@@ -126,12 +127,20 @@ def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, fi
...
 
-        return mg.get_manifest()
+        else:
```
allaway, line 142:
```suggestion
```

schematic/models/metadata.py
```diff
@@ -126,12 +127,20 @@ def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, fi
...
+            return mg.get_manifest(sheet_url = sheetUrl)
```
allaway, line 143:
```suggestion
        return mg.get_manifest(sheet_url = sheetUrl)
```

allaway:
@sujaypatil96 I incorporated your suggestions and tested the updated branch on our staging app...seemed to work fine!

sujaypatil96:
Perfect. Thank you for making the changes and testing them @allaway. I missed adding a review comment for another changes, which was, why not make the default value of the `sheetUrl` parameter, `True` instead of `None`?

allaway:
> Perfect. Thank you for making the changes and testing them @allaway. Apologies, I missed adding a review comment for another change, which was, why not make the default value of the `sheetUrl` parameter, `True` instead of `None`?

Only because I wasn't sure if setting `getModelManifest` default to `True` would impact others who might be using getModelManifest already in other contexts. If it is only being used in the shiny app, I assume it would be safe and preferred to set default to `True`. I'll make this change, but happy to revert if you think it could affect others using this function!

schematic/models/metadata.py
```diff
@@ -105,13 +105,14 @@ def getOrderedModelNodes(self, rootNode: str, relationshipType: str) -> List[str
         return ordered_nodes
 
 
-    def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, filenames: list = None) -> str:
+    def getModelManifest(self, title: str, rootNode: str, datasetId: str = None, jsonSchema: str = None, filenames: list = None, useAnnotations: bool = False, sheetUrl: bool = None) -> str:
```
allaway, line 108:
```suggestion
    def getModelManifest(self, title: str, rootNode: str, datasetId: str = None, jsonSchema: str = None, filenames: list = None, useAnnotations: bool = False, sheetUrl: bool = True) -> str:
```

sujaypatil96:
@allaway: would you mind running this file through the [`black`](https://black.readthedocs.io/en/stable/index.html) code formatter? You'd simply need to do this:

`black ~/path/to/metadata.py`

This will resolve the merge conflict that's coming up on this PR.

schematic/models/metadata.py
```diff
@@ -98,20 +104,31 @@ def getOrderedModelNodes(self, rootNode: str, relationshipType: str) -> List[str
...
-    def getModelManifest(self, title: str, rootNode: str, jsonSchema: str = None, filenames: list = None) -> str:
+    def getModelManifest(
+        self,
+        title: str,
+        rootNode: str,
+        datasetId: str = None,
+        jsonSchema: str = None,
+        filenames: list = None,
+        useAnnotations: bool = False,
+        sheetUrl: bool = True,
```
allaway, line 123:
```suggestion
        sheetUrl: bool = True
```

allaway:
What's the deal with the trailing commas that `black` adds? I assume that's valid syntax but it looks so odd to me.

sujaypatil96 (approved):
It's in accordance with the PEP8 [styleguide](https://www.python.org/dev/peps/pep-0008/). See the [When to Use Trailing Commas](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas) section of the guide for more details.

This PR is ready to be merged @allaway.🚀 Approving it now. Feel free to merge it.

sujaypatil96:
> Only because I wasn't sure if setting getModelManifest default to True would impact others who might be using getModelManifest already in other contexts. If it is only being used in the shiny app, I assume it would be safe and preferred to set default to True. I'll make this change, but happy to revert if you think it could affect others using this function!

I forgot to reply to this. Yes, you're right, all the methods in the `metadata_model.py` interface are intended to be used only by the Shiny App.

allaway:
Thanks for the clarification about trailing commas...had never seen that
before, but don't spend much time in python. The style guide rationale
makes sense, but coming from R (where this would not work), it still throws
me a bit to see.

allaway:
And thanks for the detailed review!

BrunoGrandePhD:
@allaway: Building on top of what's said in the [When to Use Trailing Commas](https://www.python.org/dev/peps/pep-0008/#id29) section that @sujaypatil96 linked to, the trailing commas are purely to minimize diffs to a single addition when adding an item to a list as opposed to a deletion and two additions. It just so happens that Python ignores them. I would do the same in R if the language supported them, but R no like trailing commas. 

Here's another [source](https://black.readthedocs.io/en/stable/the_black_code_style.html#trailing-commas).

allaway:
Thanks, that's a helpful explanation. The rationale certainly makes sense,
but I'm not holding my breath for R to support it...(I still feel like no
trailing comma is more readable, but I think that's just b/c it's what I'm
used to)

PR #458, opened 2021-04-22
Escape formulae in annotations (fixes #457)

### LINKED ISSUE #457: Escape retrieved annotations in google sheets
**Is your feature request related to a problem? Please describe.**
More recent versions of schematic appear escape numbers, booleans, and formula-like statements when populating the valid values in a google sheet manifest (eg TRUE in schema becomes 'TRUE in google sheet, -/- becomes '-/- in google sheet). This is super helpful. However, when values that are numbers, booleans, or formula-like values are retrieved from Synapse, these values are not escaped. An example of this can be seen in column T in this manifest: https://docs.google.com/spreadsheets/d/1olSch1oz1arP-D_kBbsvF920GgftPMuMkIbo0e2c9ns/edit#gid=0, where the Synapse annotations of +/- are not escaped.  In the case of formula-like values, the resulting CSV exported from the  google sheet becomes `#ERROR`, meaning that the users need to actually fix these values even though they were initially correct. In the case of numeric/boolean values, the exported value is still correct, but the google sheets formatting suggests they are invalid values. 

It would be helpful if annotations retrieved from synapse were escaped in the same manner that valid values are so that users who are modifying previous annotations don't feel the need to re-assign these already-valid-values.

• low-Medium - can do work without it; but it's important (e.g. to save time or for convenience)

• Short-term - 2-4 weeks
Relevant now, but more of a convenience/quality of life improvement than a blocker. 

**Additional context**

BrunoGrandePhD:
This **might** be one of those easy one-line fixes. 

@allaway: Can you check if this PR addresses the problem you faced in #457? You should be able to install this PR's version of schematic with the following command: 

```
python -m pip install git+https://github.com/Sage-Bionetworks/schematic.git@bgrande/issue-457/escape-annotation-values
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,8 +52,7 @@ pygsheets = "^2.0.4"
 PyYAML = "^5.4.1"
 rdflib = "^5.0.0"
 setuptools = "^52.0.0"
-synapseclient = "2.3.0"  # Update once Data Curator App supports OAuth tokens
-                         # And once schematic handles booleans (see #458)
+synapseclient = "~2.3"  # ~ (not ^) until Data Curator App supports OAuth tokens
 toml = "^0.10.2"
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,8 @@ pygsheets = "^2.0.4"
 PyYAML = "^5.4.1"
 rdflib = "^5.0.0"
 setuptools = "^52.0.0"
-synapseclient = "~2.3"  # ~ (not ^) until Data Curator App supports OAuth tokens
+synapseclient = "2.3.0"  # Update once Data Curator App supports OAuth tokens
+                         # And once schematic handles booleans (see #458)
 toml = "^0.10.2"
 
 
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -748,6 +748,7 @@ class ManifestGenerator(object):
 
         # open google sheets and extract first sheet
         sh = gc.open_by_url(manifest_url)
+        sh.default_parse = False
         wb = sh[0]
 
         # update spreadsheet with given manifest starting at top-left cell
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -748,11 +748,13 @@ class ManifestGenerator(object):
 
         # open google sheets and extract first sheet
         sh = gc.open_by_url(manifest_url)
-        sh.default_parse = False
         wb = sh[0]
 
+        # The following line sets `valueInputOption = "RAW"` in pygsheets
+        sh.default_parse = False
+
         # update spreadsheet with given manifest starting at top-left cell
-        wb.set_dataframe(manifest_df, (1,1), escape_formulae=True)
+        wb.set_dataframe(manifest_df, (1,1))
 
         # set permissions so that anyone with the link can edit
         sh.share("", role = "writer", type = "anyone")
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -751,7 +751,7 @@ class ManifestGenerator(object):
         wb = sh[0]
 
         # update spreadsheet with given manifest starting at top-left cell
-        wb.set_dataframe(manifest_df, (1,1))
+        wb.set_dataframe(manifest_df, (1,1), escape_formulae=True)
 
         # set permissions so that anyone with the link can edit
         sh.share("", role = "writer", type = "anyone")


allaway:
Hi Bruno, this is awesome, thank you! This definitely resolves the largest part of this issue - escaping formulae/formulas: https://docs.google.com/spreadsheets/d/1im1tjVAVO0RIvAY7iWGxwlyGK0gL_wYYtSUsEWSJlXk/edit#gid=0

The TRUE/FALSE values are still being flagged as invalid because they are escaped on the valid-values side of things but not on the annotation side. This is much less of a problem because it's relatively trivial to tidy up and doesn't cause any data loss (like the formula `#ERROR` did). I'm assuming this would be the case for numeric values too (ie. escaped in the valid values, but not the annotation values). 

Do you want me to split that out into a separate ticket? This fix was medium priority from my perspective, the 'TRUE/'FALSE and the boolean/numeric escaping is pretty low priority from my perspective. 

Thanks again!

BrunoGrandePhD:
@allaway: I have one more thing to try to see if we can fix the `TRUE`/`FALSE` issue in this PR. If it doesn't work, I'll revert the commit and we can merge the first commit as is once it's been reviewed by at least one other person. Can you try generating the same manifest? Also, can you post your command so I can try myself if this needs additional testing? Thanks! 

----

@milen-sage: Do you know what would be prepending `TRUE` and `FALSE` with apostrophes? I found the [code](https://github.com/Sage-Bionetworks/schematic/blob/bgrande/issue-457/escape-annotation-values/schematic/manifest/generator.py#L269-L282) that adds the values to the Sheet 2, but I see nothing that's adding the apostrophes. Is this something that happens when you set `valueInputOption = "RAW"`? pygsheets defaults to `USER_ENTERED`, so my second commit here is intended on making it more consistent with the rest of the API calls to Google Sheets (_i.e._ using `RAW`).

allaway:
@BrunoGrandePhD , thanks for looking into this! 

This seems to have worked, at least in part! https://docs.google.com/spreadsheets/d/1QATjJ-8FdiMUVIbtukRHZ4CuBCk-gMEeoq0C59s8dHk/edit#gid=0

Couple of thoughts:
-the formulas are now double-escaped (e.g. `''+/-`) 
-the casing of the booleans is changed, but g-sheets still interprets False as FALSE. I'm not sure if there are any issues with how synapse would interpret these values (would it coerce both False and FALSE to the same thing? I have no idea). 

The command to generate this file is: 

```
schematic manifest --config config.yml get -t CTFcNFWGS -dt ImagingAssay -d syn4984626 -s --use_annotations --oauth -p ./data/schema_org_schemas/NF.jsonld
```

Here's the jsonld I am using: https://raw.githubusercontent.com/nf-osi/schematic/develop/data/schema_org_schemas/NF.jsonld

and here's the config: https://raw.githubusercontent.com/Sage-Bionetworks/schematic/testing/config.yml

allaway:
As I mentioned before, though, this is not a high priority issue for me, so if it turns out to be a rabbit hole, please don't feel pressured to dive in :)

BrunoGrandePhD:
Thanks, @allaway! I think my two fixes weren't compatible with one another. I reverted the first change and kept the second change since it's more in line with how we generate Sheet 2. I did a quick test using your command (thanks for sharing!), and it seems like it works as expected now. I included the Google Sheet that I was able to generate below, but I would have peace of mind if you could also reproduce the successful behavior on your end. 

https://docs.google.com/spreadsheets/d/10FBaJpYM-mpGPBVrdcKlnWAEXhklrpGDEKftvnbHi_E

allaway:
Thanks, @BrunoGrandePhD  - that seems to work on my end too! 

The only remaining comment I have is still about TRUE vs True/FALSE vs False. I'm not sure if this is an issue or not, though. As I mentioned before, it's easy enough to fix if we start getting `True` values and need to correct them to `TRUE` on Synapse.

BrunoGrandePhD:
Can you post the link to your latest Google Sheet manifest? I was getting TRUE and FALSE, as expected. It sounds like we're running the same command but getting different output. 

I'm wondering if this has to do with the new Synapse Python client and it's support for boolean annotations. Can you confirm which version is installed using `pip show synapseclient`?

allaway:
Oops, sorry for not including it. Here you go: https://docs.google.com/spreadsheets/d/1Jzn08FlLdwsc08_IQloGT45JA-_ek_UIvRJ_XUKkKRc

I'm using 2.3.1: 

```
rallaway@ip-10-5-20-187:~$ conda activate data_curator_env_staging
(data_curator_env_staging) rallaway@ip-10-5-20-187:~$ pip show synapseclient
Name: synapseclient
Version: 2.3.1
Summary: A client for Synapse, a collaborative compute space  that allows scientists to share and analyze data together.
Home-page: https://www.synapse.org
Author: The Synapse Engineering Team
Author-email: platform@sagebase.org
License: Apache
Location: /home/rallaway/miniconda3/envs/data_curator_env_staging/lib/python3.9/site-packages
Requires: requests, keyring, deprecated, keyrings.alt
Required-by: schematicpy
(data_curator_env_staging) rallaway@ip-10-5-20-187:~$ 
```

BrunoGrandePhD:
I'm pretty sure the difference we're seeing in how `TRUE`/`FALSE` are handled has to do with the new support for booleans in Synapse in version 2.3.1 ([source](https://github.com/Sage-Bionetworks/synapsePythonClient/releases/tag/v2.3.1)). Case in point, I'm running version 2.3.0 and the capitalization is fine [on my end](https://docs.google.com/spreadsheets/d/10FBaJpYM-mpGPBVrdcKlnWAEXhklrpGDEKftvnbHi_E). 

The clunky short-term solution is to pin the version to 2.3.1 until we figure out how to best handle the new support for booleans in the Python client. 

@allaway: I've pinned the version in this PR, which still needs to be reviewed by @sujaypatil96 and/or @milen-sage. Could you open a new issue to track the boolean issue you experience if you install `synapseclient` 2.3.1? Could you include the annotations schema for one of the files? I can't view it with my permissions. I'm curious if these fields are configured as booleans on Synapse.

sujaypatil96:
Thanks for sharing the command you ran @allaway.

I tested the same command (with this branch of the code) with both versions of the synapseclient (v2.3.0 and v2.3.1) and the boolean values seem to be populated as `TRUE / FALSE` in both cases:

v2.3.0: [manifest](https://docs.google.com/spreadsheets/d/1ZdtPawCpEAzgM5KZ7C74vilgJwf7nYCSXMTWrEfeKpk/edit#gid=0)
v.2.3.1: [manifest](https://docs.google.com/spreadsheets/d/1cd3eLHKdQGbJ8v2gWzb-A9aMVEz7x7IPp3DfwtK-F2w/edit#gid=0)

sujaypatil96 (approved):

BrunoGrandePhD:
I can confirm that running the command with both versions of `synapseclient` yields the same result on my end (see links below). So, maybe the `synapseclient` version is a red herring.

- v2.3.0: https://docs.google.com/spreadsheets/d/18rE49WUZTqmMuyFPj0sGL85I2XeapXEns9ohqk7pTo4/edit
- v2.3.1: https://docs.google.com/spreadsheets/d/1Tn8dqQdzrZDhJl7yPkZElhyCwYhQlSDBtEjIaYHDnWQ/edit

In light of this, I'm going to revert the change that pinned the version and merge the PR. 

@allaway: Please open a new issue so we can figure out why you're getting a different outcome than us. Please include the annotations schema for one of the files so we can see if the annotations are configured as booleans on Synapse. Also include the output of `pip freeze` so we can see if there are any other environment differences between you and us. Thanks!

allaway:
Will do, thanks!

PR #465, opened 2021-04-29
Interpret all values in manifest as str

### LINKED ISSUE #341: Show warning about invalid values (instead of rejecting input)
In metadata template CSV, warn user about invalid value(s) instead of rejecting the invalid input. This would allow the user to override the warning and enter an invalid value. 

_NF use case:_ our schema is always evolving, and we sometimes have new data types that are not yet represented in our schema. We want to allow our contributors to enter these values, even if they're invalid, if need be (after they see a warning).

[Example](https://docs.google.com/spreadsheets/d/1_owmb2_oNGie1qID_RKUc-enbA7IFdY68CMxNRfVMMw/edit?usp=sharing):
image

cc: @allaway

### LINKED ISSUE #463: TRUE/FALSE in csv coerced to True/False in validation step
Perhaps I've just done something wrong here but I have columns with TRUE/FALSE values (https://docs.google.com/spreadsheets/d/1UtYc889bzhUms67bBbHYvyx46eUncztDOSo5KH2xPHM/edit#gid=136843556) However, when I try to validate this data, the validation throws an error like `'False' is not one of ['TRUE', 'FALSE']`, which leads me to believe that the CSV, when getting read in has TRUE/FALSE values changed to True/False. 

Screen Shot 2021-04-27 at 1 37 24 PM

allaway:
This fixes #463. I tested this on the NF Data Curator app staging instance and it works great for solving the TRUE->True and FALSE->False issue. However, I'm not sure if there are other possible issues with applying a boilerplate rule to interpret all values as strings. It _seems_ like it should be safe in this context, but I really don't know. 

edit: I also just added some additional changes to make "strict" validation changeable in the `config.yml` file to resolve #341 .


sujaypatil96 (commented):
@allaway: Hm, it's good that this fix is able to resolve the `TRUE` / `True` ambiguity issue, but it could cause some potential problems.

One edge case I can think of is the `Patient ID` attribute. This attribute is a required column and while not using `dtype=str` flags it as a warning, setting `dtype=str` accepts the empty string as valid input for that column and doesn't flag it during validation.

allaway:
>One edge case I can think of is the Patient ID attribute. This attribute is a required column and while not using dtype=str flags it as a warning, setting dtype=str accepts the empty string as valid input for that column and doesn't flag it during validation.

>In order to test this out, you can leave the Patient ID column blank in your manifest and try both dtype=str and without it.

Sorry, just to clarify, do you know this is the case already/you tested it, or this is just a guess and you want me to test it?

allaway:
I thought that `manifest.convert_dtypes(convert_boolean=False)` might be an easy fix for this, but it causes issues with `.fillna`:

```
Warning: Error in py_call_impl: TypeError: <U1 cannot be converted to an IntegerDtype

Detailed traceback: 
  File "/home/rallaway/miniconda3/envs/data_curator_env_staging/lib/python3.9/site-packages/schematic/models/metadata.py", line 211, in validateModelManifest
    manifest = trim_commas_df(manifest).fillna(
  File "/home/rallaway/miniconda3/envs/data_curator_env_staging/lib/python3.9/site-packages/pandas/core/frame.py", line 4462, in fillna
    return super().fillna(
  File "/home/rallaway/miniconda3/envs/data_curator_env_staging/lib/python3.9/site-packages/pandas/core/generic.py", line 6429, in fillna
    new_data = self._mgr.fillna(
  File "/home/rallaway/miniconda3/envs/data_curator_env_staging/lib/python3.9/site-packages/pandas/core/internals/managers.py", line 621, in fillna
    return self.apply(
  File "/home/rallaway/miniconda3/envs/data_curator_env_staging/lib/python3.9/site-packages/pandas/core/internals/managers.py", line 427, in apply
    applied = getattr(b, f)(**kwargs)
  Fil [... truncated]
  76: <Anonymous>
```

schematic/store/synapse.py
```diff
@@ -493,14 +497,17 @@ def associateMetadataWithFiles(
...
 
+                metadataSyn[keySyn] = v
+                
             # set annotation(s) for the various objects/items in a dataset on Synapse
             annos = self.syn.get_annotations(entityId)
 
             for anno_k, anno_v in metadataSyn.items():
                 annos[anno_k] = metadataSyn[anno_k]
-
+
```
allaway, line 510:
```suggestion
```

PR #466, opened 2021-04-30
update docs for authtoken

allaway:
Which is now preferred over old config files with apikey. Now points towards 2.4.0 .synapseConfig file instead of 2.2.2, which was pre-authentication token days.

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -71,11 +71,9 @@ and
 
 Download a copy of the ``.synapseConfig`` file, open the file in the
 editor of your choice and edit the
-`username <https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v2.2.2-rc/synapseclient/.synapseConfig#L8>`__
-and
-`apikey <https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v2.2.2-rc/synapseclient/.synapseConfig#L9>`__
-attributes under the
-`[authentication] <https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v2.2.2-rc/synapseclient/.synapseConfig#L7>`__
+`authtoken <https://github.com/Sage-Bionetworks/synapsePythonClient/blob/ba42e2d35673d19ce2b5bbd089f4fc1e99aca178/synapseclient/.synapseConfig#L9>`__
+attribute under the
+`[authentication] <https://github.com/Sage-Bionetworks/synapsePythonClient/blob/ba42e2d35673d19ce2b5bbd089f4fc1e99aca178/synapseclient/.synapseConfig#L7>`__
 section.
 
  Description of config.yml attributes


sujaypatil96 (changes requested):
Thanks for the update to the documentation @allaway. I wasn't aware that `authtoken` was being preferred over `apikey` while authenticating applications with Synapse (source: `synapseclient` [docs](url)).

Quick question, any specific reason why you removed the `[username]` section from the previous version of the `README.rst` file? Or was it by mistake?

docs/README.rst
```diff
@@ -71,11 +71,9 @@ and
...
 Download a copy of the ``.synapseConfig`` file, open the file in the
 editor of your choice and edit the
-`username <https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v2.2.2-rc/synapseclient/.synapseConfig#L8>`__
-and
-`apikey <https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v2.2.2-rc/synapseclient/.synapseConfig#L9>`__
-attributes under the
-`[authentication] <https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v2.2.2-rc/synapseclient/.synapseConfig#L7>`__
+`authtoken <https://github.com/Sage-Bionetworks/synapsePythonClient/blob/ba42e2d35673d19ce2b5bbd089f4fc1e99aca178/synapseclient/.synapseConfig#L9>`__
+attribute under the
+`[authentication] <https://github.com/Sage-Bionetworks/synapsePythonClient/blob/ba42e2d35673d19ce2b5bbd089f4fc1e99aca178/synapseclient/.synapseConfig#L7>`__
```
sujaypatil96, line 76: Can we add the `[username]` section again, before the `authtoken` section?

allaway:
> Thanks for the update to the documentation @allaway. I wasn't aware that `authtoken` was being preferred over `apikey` while authenticating applications with Synapse (source: `synapseclient` [docs](url)).

Yep! authtoken is new and preferred b/c you can create a unique one for each place you need authentication (e.g. if you need to delete one, it doesn't affect all of your config files) and they have more granular permissions.  IIRC from demo day the apikey will be phased out at some point, but i might be misremembering. 

> Quick question, any specific reason why you removed the `[username]` section from the previous version of the `README.rst` file? Or was it by mistake?

On purpose - the Synapse PAT does not require a username. You can try this yourself by creating a PAT and logging in with `synapse login -p <authtoken>`

docs/README.rst
allaway: with the authtoken, username is not needed: http://python-docs.synapse.org/build/html/Credentials.html?highlight=authtoken

sujaypatil96:
Perfect, thank you for clarifying that @allaway. Very cool, that the Personal Access Token doesn't require `username` to be specified. I guess it encodes that information within the token.

This PR is ready to be merged.🚀

sujaypatil96 (approved):

PR #468, opened 2021-05-06
Cleanup of data folder for generic develop branch

sujaypatil96:
This PR seeks remove unnecessary data files and folders from the `develop` branch of `schematic`. The project specific (e.g., HTAN specific) data files can be created in an `-htan` release branch, which will be in sync with the `ncihtan` default branch.


sujaypatil96:
Added another [commit](https://github.com/Sage-Bionetworks/schematic/pull/468/commits/dce0102ddfee48842d6707ff597cc59afa6430b1) which does two things:

- Renames `data/schema_org_schemas/example.jsonld` to `data/schema_org_schemas/example.model.jsonld`
- Slight modifications to `config.yml` to make it a little more concise

BrunoGrandePhD (approved):
I love a good streamlining PR! This is good to submit as is, but I have a few questions that might prompt @sujaypatil96 to go further. 🚀 

- Does `data/csv/example.model.csv` provide anything that isn't provided by `tests/data/example.model.csv`? If not, then we can delete it. If so, then we should make sure that `tests/data/example.model.csv` contains those missing features and then we delete `data/csv/example.model.csv`. 
- The only other file in `data/` is `validation_schemas/example_validation_schema.json`. I assume you kept this for a reason. What is that reason? I ask because I wonder if we should generate the validation JSON schema for `tests/data/example.model.jsonld` and include it alongside the JSON-LD file.

sujaypatil96:
Thanks for the review @BrunoGrandePhD.

Great catch, it was my bad. I actually intended to retain the `data/schema_org_schemas/example.model.jsonld` file and delete `data/validation_schemas/`, but I ended up doing the opposite. But along the note that you mentioned, perhaps it'd be a good idea to also include an `example.validation.json` JSON Validation Schema file too (`data/validation_schemas/example.validation.json`).

Also, if you had thoughts on the current folder names, do let me know. Maybe rename `schema_org_schemas` to `data_models` or `schemas`? If we decide to include `validation_schemas` I guess that can remain as is.

I'll look into your comment about contrasting `data/schema_org_schemas/example.model.csv` with `tests/data/example.model.csv` too and add a comment here. Thank you.

sujaypatil96:
@BrunoGrandePhD: nice observation. You're right, in that the JSON-LD you created and put in `tests/data/example.model.jsonld` is more minimal and better suited as an example than `data/schema_org_schemas/example.model.jsonld`. The only difference between the two was that `tests/data/example.model.jsonld` described only one Component (`Patient`), while `data/schema_org_schemas/example.model.jsonld` described two components (`Patient` and `scRNA-seq Assay`).

Note: `scRNA-seq Assay` has been deprecated/removed and there's Assay types for different levels now.

Yup, it's definitely a better idea to remove the `data/` folder altogether. I've made a change in `config.yml` to point to `tests/data/example.model.jsonld`.

BrunoGrandePhD:
@sujaypatil96: Wonderful! My approval still stands. Great job with simplifying the repo!

PR #469, opened 2021-05-13
Test cases for df_parser module

sujaypatil96:
Improving code coverage is part of the current milestone as well. With that in mind, I'm writing tests for some of the modules in our package.

In this PR, I've added two files:
- `tests/data/example_extension.model.csv`: An extension of the `tests/data/example.model.csv` CSV data model, which adds an additional attribute (`Assay`) for testing purposes.
- `tests/test_schemas.py`: file with test cases for testing the `df_parser.py` module.


tests/data/example_extension.model.csv
```diff
@@ -0,0 +1,21 @@
+Attribute,Description,Valid Values,DependsOn,Properties,Required,Parent,DependsOn Component,Source,Validation Rules
```
BrunoGrandePhD, line 1: Could you provide more context on why this additional CSV file is needed? I hesitate to have so much duplicated content that we would have to keep in sync. Would we be able to load the existing CSV data model and update it with code?

tests/test_schemas.py
```diff
@@ -0,0 +1,172 @@
+import os
+import logging
+
+import pandas as pd
+import pytest
+
+from schematic.schemas import df_parser
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
```
BrunoGrandePhD, line 10: **Random thought:** Is there precedent to configuring the logger in every Python file/submodule? Couldn't we define these two lines once in the root `__init__.py` file (one in `schematic/` and another in `tests/`) and import it in submodules? If so, this could happen in another PR.

BrunoGrandePhD (commented):
I'll wait to get more context before reviewing the Python code too closely.

tests/test_schemas.py
sujaypatil96: @BrunoGrandePhD: I think the reason for creating a logger in every (sub)module is so the logging module knows which (sub)module raised the logger warning/error. If the root logger (say in `schematic/__init__.py`) is always imported and used for logging then we'd see warnings like `schematic: logger message` rather than the specific (sub)module (example: `schematic/manifest/generator.py: logger message`).

This [SO answer](https://stackoverflow.com/questions/15727420/using-logging-in-multiple-modules) also recommends defining a logger within every (sub)module.

tests/data/example_extension.model.csv
sujaypatil96: Right, I created this for a couple of unit tests that check if the package is able to create a new extended JSON-LD schema (from `example_extension.model.csv`) from a given base schema JSON-LD (from `example.model.csv`).

BrunoGrandePhD: If possible, I would like to capture those expectations using code instead of a mostly-duplicated CSV data model. For example, if you're adding an attribute, you can look for that attribute in the extended model and confirm some of its keys and values.

tests/test_schemas.py
BrunoGrandePhD: But aren't these log messages intended for the users (and not us, developers)? Why do they have to know which submodule is emitting the log message?

tests/test_schemas.py
```diff
@@ -0,0 +1,172 @@
...
+        # base schema path is passed explicitly
+        biothings_path = df_parser._get_base_schema_path()
+
+        assert os.path.basename(biothings_path) == "biothings.model.jsonld"
+
+    def test_convert_csv_to_data_model(self, helpers):
+
+        csv_path = helpers.get_data_path("example.model.jsonld")
+
+        extended_csv_model_path = helpers.get_data_path("example_extension.model.csv")
```
BrunoGrandePhD, line 160: I see that you do this twice in this particular test suite. In that case, I recommend that you abstract the Pandas code I suggested you write in an earlier comment into a fixture at the top of this file.

BrunoGrandePhD: Then, you would store the file into a temporary CSV file (look into `tempfile` and the `with` statement) before reloading using `_convert_csv_to_data_model()`.

tests/test_schemas.py
```diff
@@ -0,0 +1,172 @@
...
+        # CSV data model as base schema
+        extended_csv_model_se = df_parser._convert_csv_to_data_model(
+            extended_csv_model_path, csv_path
+        )
+
+        # if new attribute can be found in extended_csv_model_se
+        # we know the conversion was successful
+        attribute_present = df_parser.attribute_exists(extended_csv_model_se, "Assay")
+
+        assert attribute_present == True
```
BrunoGrandePhD, line 172:
```suggestion
        assert attribute_present
```

BrunoGrandePhD (changes requested):
I've now taken a closer look at the new tests. I don't expect these changes to be done before @sujaypatil96 resumes work in June.

tests/test_schemas.py
milen-sage: Wearing my user hat on, I'd agree it's easier to define the logger level in one place and have it be in effect across the codebase.

@BrunoGrandePhD are you concerned about having to maintain/change the logger code in multiple places across the codebase (w/ potential for logging levels to diverge)?

@sujaypatil96 do you think the source (sub)module is useful for developers?

tests/test_schemas.py
```diff
@@ -0,0 +1,172 @@
...
+    def test_check_schema_definition(self, helpers):
+
+        data_model_csv_path = helpers.get_data_path("example.model.csv")
+
+        example_model_df = pd.read_csv(data_model_csv_path)
+
+        # when all required headers are provided in the CSV data model
+        actual_df = df_parser.check_schema_definition(example_model_df)
+
+        assert actual_df is None
```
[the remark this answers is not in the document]
sujaypatil96: Good catch noticing the return type discrepancy in `check_schema_definition` @BrunoGrandePhD. Per your comment, you're saying we should return `bool` values rather than returning `None`?

I'll address this in a separate PR once we've decided if we want to stick to `bool` or change the type hints.

tests/test_schemas.py
sujaypatil96: Yup, that makes sense. I've created a module level fixture called `extended_schema_path()` which appends the new `Assay` attribute (for testing) to the example schema.

tests/test_schemas.py
```diff
@@ -0,0 +1,172 @@
...
+
+        extended_csv_model_se = df_parser.create_nx_schema_objects(
+            extended_model_df, se_obj
+        )
+
+        # check if the "Assay" attribute has been added to the new SchemaExplorer 
+        # object with attributes from the extended schema
+        result = df_parser.attribute_exists(extended_csv_model_se, "Assay")
+
+        assert result is True
```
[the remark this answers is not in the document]
sujaypatil96: Ah, great. Thanks for the suggestion Bruno. I'll keep this in mind for future PRs.

BrunoGrandePhD (approved):
🚀

PR #470, opened 2021-05-14
add useSchemaLabel parameter

allaway:
Allows user to configure whether or not to use the schema label vs the native value in the stored annotations.

This change already exists in `main`, but for some reason never got added to develop (oops). Breaking this out of #465 for simplicity.


milen-sage:
Yes, we were wondering what else is going to pop up that we accidentally didn't include. Thanks for the catch @allaway !

sujaypatil96 (approved):
Great catch! Thanks @allaway.🚀

sujaypatil96:
Quick favour here @allaway. Would you mind adding a commit to remove the unnecessary `example.model.jsonld` file from the root of this repository? it seems to have percolated through one of our previous PRs.

Issue #447, opened 2021-04-08
Make config.yml file concise


sujaypatil96:
**Describe the bug**
Currently, the [config.yml](https://github.com/Sage-Bionetworks/schematic/blob/develop/config.yml) file has more keys/attributes than are necessary. We need to improve the `config.yml` so that it is minimal and there are only as many keys that are absolutely necessary. Once we streamline the Synapse credentials specification process (issue #398), we can remove `["definitions"]["synapse_config"]`. Similarly, both `manifest` keys and `["model"]["input"]["validation_schema"]` and `["model"]["input"]["log_location"]` can be removed thereafter.

We need to consider renaming some of the keys so they're more semantic.

The default Googlesheet manifest styles should either be hardcoded into the codebase, and a separate config file perhaps can be used to specify custom styles to the manifest.

**To Reproduce**
1. Open [config.yml](https://github.com/Sage-Bionetworks/schematic/blob/develop/config.yml)
2. Observe changes described above and think about how to fix them

PR #471, opened 2021-05-14
Add strict validation to config

### LINKED ISSUE #341: Show warning about invalid values (instead of rejecting input)

allaway:
Splitting out of #465. 

Closes #341.

This adds the ability to configure the "warn" vs "reject" option in the generated manifest but toggling a boolean in the config file.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -8,7 +8,7 @@ definitions:
 synapse:
   master_fileview: 'syn23643253'
   manifest_folder: 'manifests'
-  manifest_filename: 'synapse_storage_manifest.csv'
+  manifest_filename: 'data/manifests/synapse_storage_manifest.csv'
   token_creds: 'syn23643259'
   service_acct_creds: 'syn25171627'
 
@@ -18,8 +18,10 @@ manifest:
 
 model:
   input:
-    location: 'tests/data/example.model.jsonld'
+    location: 'data/schema_org_schemas/example.jsonld'
     file_type: 'local'
+    validation_schema: 'data/validation_schemas/example_validation_schema.json'
+    log_location: 'data/json_schema_logs/json_schema_log.json'
 
 style:
   google_manifest:
@@ -32,3 +34,4 @@ style:
       green: 1.0
       blue: 0.9019
     master_template_id: '1LYS5qE4nV9jzcYw5sXwCza25slDfRA1CIg3cs-hCdpU'
+    strict_validation: true
--- a/config.yml
+++ b/config.yml
@@ -20,8 +20,6 @@ model:
   input:
     location: 'tests/data/example.model.jsonld'
     file_type: 'local'
-    validation_schema: 'data/validation_schemas/example_validation_schema.json'
-    log_location: 'data/json_schema_logs/json_schema_log.json'
 
 style:
   google_manifest:
--- a/config.yml
+++ b/config.yml
@@ -8,7 +8,7 @@ definitions:
 synapse:
   master_fileview: 'syn23643253'
   manifest_folder: 'manifests'
-  manifest_filename: 'data/manifests/synapse_storage_manifest.csv'
+  manifest_filename: 'synapse_storage_manifest.csv'
   token_creds: 'syn23643259'
   service_acct_creds: 'syn25171627'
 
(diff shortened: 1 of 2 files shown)


config.yml
```diff
@@ -32,3 +34,4 @@ style:
       green: 1.0
       blue: 0.9019
     master_template_id: '1LYS5qE4nV9jzcYw5sXwCza25slDfRA1CIg3cs-hCdpU'
+    strict_validation: true
```
sujaypatil96, line 37: Rather then adding it as a config parameter, what are your thoughts on making it an optional CLI argument? This is once again a suggestion, keeping issue #447 in mind. We'll probably be eliminating this whole Googlesheets customization section from the config file eventually, so I'm not quite sure this is the best place to allow this type of configuration.

sujaypatil96 (commented):
Please let me know what you think of my recommendations @allaway.

CC: @milen-sag

config.yml
```diff
@@ -20,6 +20,8 @@ model:
   input:
     location: 'tests/data/example.model.jsonld'
     file_type: 'local'
+    validation_schema: 'data/validation_schemas/example_validation_schema.json'
+    log_location: 'data/json_schema_logs/json_schema_log.json'
```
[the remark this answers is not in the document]
allaway: This makes sense to me! 

```suggestion
```

config.yml
allaway: Good question. My use-case is not actually the CLI, but for the app, which is why I added it as a config option. IMO, it makes more sense to modify a config file rather than `metadataModelFuns.py` or `app.R`. But we could change this to config in www/config.json in the shiny app instead, maybe? 

Either way, I agree that this should _also_ be a CLI argument.

sujaypatil96: Ah, I see, that makes sense. We can add the option to schematic's config.yml file (the way you have now) and the `www/config.json` file or the Data Curator App. But it probably makes more sense to add it here because schematic reads all configuration information from this config.yml file, and there are cases when users might not choose to use the Data Curator App to leverage schematic's functionalities, but just rely on the CLI.

sujaypatil96 (approved):
I'm approving this PR @allaway.🚀

Quick nit: has this code been `black` formatted? If it hasn't could you format it before merging?

allaway:
> I'm approving this PR @allaway.🚀
> 
> Quick nit: has this code been `black` formatted? If it hasn't could you format it before merging?

Good catch, done now!

config.yml
milen-sage: @sujaypatil96 and @allaway let's keep this as parameter exposed in the config.yml file for now. I agree exposing it as an optional CLI parameter makes a lot of sense as well (and respectively in the corresponding REST API). 

It looks like we may want to introduce a separate manifest config/profile in the future that allows setting style and other manifest parameters. Let's discuss the more general topic of profiles during code review - @sujaypatil96  could you add it to the agenda for next week?

milen-sage: ow I agree this looks good to go!

allaway: 😊

sujaypatil96: @milen-sage: yup, let's expose this as an optional parameter in a future CLI update PR.

> It looks like we may want to introduce a separate manifest config/profile in the future that allows setting style and other manifest parameters. Let's discuss the more general topic of profiles during code review - @sujaypatil96 could you add it to the agenda for next week?

Sounds good, I'll add it to the agenda. Looks like we have quite a few items to discuss during our next code review.

> ow I agree this looks good to go! Thanks @allaway! Btw @sujaypatil96 could you add Robert to the contributor's list on schematic docs? Robert has been driving a bunch of features now, in addition to adding code + testing and providing feedback.

Absolutely! I agree @milen-sage. I'll add @allaway to the contributors section in an upcoming documentation update PR?

PR #480, opened 2021-05-27
Share Synapse login logic

BrunoGrandePhD:
When I tried to initialize schematic to use the Google service account, I ran into an error because schematic was trying to login using the `synapse_config` in the config file, whereas I'm using an environment variable with an access token. In this PR, I'm encapsulating the login logic into a static function so it can be more readily used anywhere in schematic. 

@milen-sage: I'm requesting a review from you since Sujay is out until June.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -64,25 +64,7 @@ class SynapseStorage(BaseStorage):
             syn_store = SynapseStorage()
         """
 
-        # If no token is provided, try retrieving access token from environment
-        if not token and not access_token:
-            access_token = os.getenv("SYNAPSE_ACCESS_TOKEN")
-
-        # login using a token
-        if token:
-            self.syn = synapseclient.Synapse()
-
-            try:
-                self.syn.login(sessionToken=token, silent=True)
-            except synapseclient.core.exceptions.SynapseHTTPError:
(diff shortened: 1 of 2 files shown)


milen-sage (approved):
Encapsulation definitely makes sense. I haven't tested the code, but I trust @BrunoGrandePhD that you've tested (e.g. in the context of updating to OAuth in DC app).

BrunoGrandePhD:
@milen-sage: I haven't tested it in the context of the Data Curator App. I've only tested it locally using some schematic commands. Since this isn't urgent, I'll wait until @sujaypatil96 can confirm that all works as expected.

sujaypatil96 (approved):
I'm still working with Xa to get setup with VPN so I can access the Shiny Server. In the meantime, the changes on schematic's side are working fine, so I think it's okay to merge this PR.🚀

Also, at the moment The Data Curator App is pulling from `schematicpy==0.1.14`, so we'll make the changes to the way the login function is called when we include this feature in the next release.

sujaypatil96:
I tested this code with the Data Curator App deployed on the Shiny Server as [SynapseShinyApp](https://shiny.synapse.org/users/spatil/SynapseShinyApp/), and can confirm that it works as expected. So we are good to merge this.🚀

PR #483, opened 2021-06-02
Retrieving information about manifests associated with all datasets in a project

milen-sage:
Adding a function retrieving all synapse IDs of manifests associated with a project, in addition to listing the data model component associated with each manifest, if applicable.

 @rrchai this may be of use for your work on the metadata dashboard. Note the return list of the function: 
- if no manifest is present in a dataset folder, the manifest ID is empty string
- if a manifest is present, but not Component is specified in the manifest columns, the component name is an empty string. (This is possible for manifests in general, but shouldn't happen given the HTAN data model, for instance.)

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -364,7 +364,7 @@ class SynapseStorage(BaseStorage):
         return manifest_id
 
 
-    def getAllManifests(self, projectId: str) -> List[str]:
+    def getProjectManifests(self, projectId: str) -> List[str]:
         """Gets all metadata manifest files across all datasets in a specified project.
 
         Returns: A list of datasets per project; metadata manifest Synapse ID for each dataset; and the corresponding schema component of the manifest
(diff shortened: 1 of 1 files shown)


schematic/store/synapse.py
```diff
@@ -363,47 +363,67 @@ def updateDatasetManifestFiles(self, datasetId: str) -> str:
 
         return manifest_id
 
-    def getAllManifests(self) -> List[str]:
-        """Gets all metadata manifest files across all datasets in projects a user has access to.
 
-        Returns: A list of projects, datasets per project and metadata manifest Synapse ID for each dataset
+    def getAllManifests(self, projectId: str) -> List[str]:
```
sujaypatil96, line 367: Change method name to something that corresponds to the new functionality, i.e., get all manifests specific to a project. Consider naming it to `getProjectManifests()`.

milen-sage: addressed in commit.

PR #484, opened 2021-06-03
Retrieving information about manifests associated with all datasets in a project

milen-sage:
Opening PR directly from a branch on the upstream to avoid tests failing on PRs from fork.

Adding a function retrieving all synapse IDs of manifests associated with a project, in addition to listing the data model component associated with each manifest, if applicable.

@rrchai this may be of use for your work on the metadata dashboard. Note the return list of the function:

if no manifest is present in a dataset folder, the manifest ID is empty string
if a manifest is present, but not Component is specified in the manifest columns, the component name is an empty string. (This is possible for manifests in general, but shouldn't happen given the HTAN data model, for instance.)

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -364,7 +364,7 @@ class SynapseStorage(BaseStorage):
         return manifest_id
 
 
-    def getAllManifests(self, projectId: str) -> List[str]:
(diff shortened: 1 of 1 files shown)


sujaypatil96 (approved):
🚀

PR #485, opened 2021-06-10
fix use_annotations help

allaway:
`--use_annotations` is now a boolean flag and does not accept T/F input, fixing help message to reflect this.


sujaypatil96 (approved):
Thank you for this update @allaway. Apologies it took so long to get this in.🚀

PR #486, opened 2021-06-10
REST API endpoint that calls get_manifest()

sujaypatil96:
This PR seeks to add the first REST API endpoint for the schematic package.

Brief description of some of the new files that are being added to the repo as part of this PR:
- `api/openapi/api.yaml`: openAPI specification of endpoint in YAML file
- `routes.py`: makes call to the `connexion` library to associate the parameters from the API spec. to the argument of the `ManifestGenerator.__init__()` method as well as the `get_manifest()` method.
- `run_api.py`: contains the driver code to start up the HTTP server

An example endpoint that you can execute to test that the REST API server works:

```
http://localhost:3001/v1/manifest/generate?schema_url=https://raw.githubusercontent.com/ncihtan/data-models/main/HTAN.model.jsonld&data_type=FollowUp&dataset_id=&title=FollowUp_Manifest
```

To view the Swagger interface:

```
http://localhost:3001/v1/ui
```

A few things that still need to be addressed in this PR:

- [x] Add examples and custom responses to API endpoint, i.e., clearly define the structure of the endpoint
- [x] Move schematic config parameters into FlaskApp config object

CC: @milen-sage @BrunoGrandePhD @hweej

### DIFF
--- a/api/__init__.py
+++ b/api/__init__.py
@@ -4,23 +4,18 @@ import connexion
 
 from schematic import CONFIG
 
-# path to `config.yml` file as constant
-CONFIG_PATH = os.path.abspath(os.path.join(__file__, "../../config.yml"))
-
 
 def create_app():
     connexionapp = connexion.FlaskApp(__name__, specification_dir="openapi/")
     connexionapp.add_api("api.yaml")
+
+    # get the underlying Flask app instance
     app = connexionapp.app
 
-    # Configure schematic and do the error handling
-    # check if file exists at the path created, i.e., CONFIG_PATH
-    if os.path.isfile(CONFIG_PATH):
-        CONFIG.load_config(CONFIG_PATH)
-    else:
-        FileNotFoundError(
-            f"No configuration file was found at this path: {CONFIG_PATH}"
-        )
+    # path to config.yml file saved as a Flask config variable
+    app.config["SCHEMATIC_CONFIG"] = os.path.abspath(
+        os.path.join(__file__, "../../config.yml")
+    )
 
     # Configure flask app
     # app.config[] = schematic[]
--- a/api/__init__.py
+++ b/api/__init__.py
@@ -1,0 +1,9 @@
+import connexion
+
+
+def create_app():
+    connexionapp = connexion.FlaskApp(__name__, specification_dir="openapi/")
+    connexionapp.add_api("api.yaml")
+    app = connexionapp.app
+
+    return app
--- a/api/__init__.py
+++ b/api/__init__.py
@@ -1,4 +1,11 @@
+import os
+
 import connexion
+
+from schematic import CONFIG
+
+# path to `config.yml` file as constant
+CONFIG_PATH = os.path.abspath(os.path.join(__file__, "../../config.yml"))
 
 
 def create_app():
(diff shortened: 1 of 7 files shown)


sujaypatil96:
I've addressed both items from the checklist above, and I think this PR is ready for review now. I've tested it on my end and it works.

sujaypatil96:
Hi @hweej. I was was wondering if you could help review this PR?

sujaypatil96:
Made this minor [commit](https://github.com/Sage-Bionetworks/schematic/pull/486/commits/303eacae24dd6906144dd55ccb5d5be7ecb7d051) to change the response code from `200` to `201`. The PR is still good for review.

api/__init__.py
```diff
@@ -0,0 +1,40 @@
+import os
+
+import connexion
+
+from schematic import CONFIG
+
+# path to `config.yml` file as constant
+CONFIG_PATH = os.path.abspath(os.path.join(__file__, "../../config.yml"))
```
hweej, line 8: I would make the CONFIG_PATH `schematic_config.yaml` a flask config variable. Then load the flask config in the create_app() factory function. Then read `flask.app.config['SCHEMATIC_CONFIG']` and use the path to pass to `CONFIG.load_config(CONFIG_PATH)`. It will help keep SCHEMATIC & the api separate down the line.

api/openapi/api.yaml
```diff
@@ -0,0 +1,74 @@
...
+  description: >-
+    This service exposes core functionalities from schematic as REST API
+    endpoints
+
+servers:
+  - url: /v1
+
+paths:
+  /manifest/generate:
+    get:
```
hweej, line 14: Generally, I like to use the `POST` verb here for creating resources on the backend.

api/__init__.py
sujaypatil96: Does this [commit](https://github.com/Sage-Bionetworks/schematic/pull/486/commits/18cdf4e3b3edca6dbc5c5dcc8efa6d3172c0e011) properly address this review comment @hweej?

api/openapi/api.yaml
sujaypatil96: Quick question here. Are you suggesting the use of `POST` over `GET` for security reasons? In that, if the request is `GET` users could easily create resources by hitting endpoints from the browser? I think `POST` has some amount of that risk too?

api/routes.py
```diff
@@ -0,0 +1,51 @@
...
+#     pass
+# def after_request(var1, var2):
+#     # Do stuff after your route executes
+#     pass
+
+# @before_request
+def get_manifest_route(schema_url, title, oauth, use_annotations):
+    # retrieve a JSON-LD via URL and store it in a temporary location
+    with urllib.request.urlopen(schema_url) as response:
+        with tempfile.NamedTemporaryFile(delete=False, suffix=".jsonld") as tmp_file:
```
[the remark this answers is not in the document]
sujaypatil96: Awesome, thanks for the suggestion, this makes sense. Just to clarify what you mean by _local storage_. Do you mean the user's filesystem or the browser's [Web Storage](https://www.w3schools.com/html/html5_webstorage.asp)?

api/openapi/api.yaml
hweej: Yes, there are some other advantages to using `POST` vs `GET`. `GET` requests can be cached by a browser and  ultimately the same request can be executed again with or without a user being present. Perhaps a browser plugin vulnerability.

I googled it, but this table sums it up https://www.diffen.com/difference/GET-vs-POST-HTTP-Requests

sujaypatil96: Thank you for the explanation Jason, this is helpful. For the purposes of this endpoint, which has to do with creation of the manifest, I think it should be fine. The next endpoints to come, validation and submission will be `POST`.

PR #487, opened 2021-06-11
Remove `strict=False` from `_get_column_data_validation_values()` call

sujaypatil96:
This PR seeks to remove an argument that snuck into the `develop` branch when we were working on some other PRs to address the addition of logic to enforce `strict` validation of Googlesheets.

It's necessary that we merge this PR at the earliest, because when we build and install the app with `poetry` and try to run the manifest generation command, it doesn't work.

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -642,7 +642,6 @@ class ManifestGenerator(object):
                     spreadsheet_id,
                     req_vals,
                     i,
-                    strict=False,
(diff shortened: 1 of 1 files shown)


sujaypatil96:
Let's merge this @milen-sage? The CLI utilities don't work as expected with the current codebase.

milen-sage (approved):

milen-sage:
yep - sorry, I meant to review and that slipped

PR #488, opened 2021-06-14
Getting component dependencies as a graph

milen-sage:
get_component_requirements in models/metadata.py now has an argument allowing returning the component requirements graph (i.e. set as_graph to true). Added test to test_metadata.py. 

@rrchai you can call get_component_requirements from MetadataModel w/ input source component (e.g. selected template type in the DCA) and should get a tuple of edges (a directed graph to load in R).

### DIFF
--- a/schematic/models/metadata.py
+++ b/schematic/models/metadata.py
@@ -155,14 +155,16 @@ class MetadataModel(object):
 
         return mg.get_manifest(sheet_url=sheetUrl)
 
-    def get_component_requirements(self, source_component: str, as_graph: bool = False) -> List:
+    def get_component_requirements(
+        self, source_component: str, as_graph: bool = False
+    ) -> List:
         """Given a source model component (see https://w3id.org/biolink/vocab/category for definnition of component), return all components required by it.
         Useful to construct requirement dependencies not only between specific attributes but also between categories/components of attributes;
         Can be utilized to track metadata completion progress across multiple categories of attributes.
 
         Args:
             source_component: an attribute label indicating the source component.
-            as_graph: if False return component requirements as a list; if True return component requirements as a dependency graph (i.e. a DAG) 
+            as_graph: if False return component requirements as a list; if True return component requirements as a dependency graph (i.e. a DAG)
 
         Returns:
             A list of required components associated with the source component.
@@ -173,7 +175,9 @@ class MetadataModel(object):
 
         # retreive components as graph
         if as_graph:
-            req_components_graph = self.sg.get_component_requirements_graph(source_component)
+            req_components_graph = self.sg.get_component_requirements_graph(
+                source_component
+            )
 
             # serialize component dependencies DAG to a edge list of node tuples
             req_components = list(req_components_graph.edges())
@@ -181,7 +185,6 @@ class MetadataModel(object):
             return req_components
 
         return req_components
-
 
     # TODO: abstract validation in its own module
     def validateModelManifest(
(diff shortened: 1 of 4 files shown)


milen-sage:
@rrchai for example if you do 

metadata_model.get_component_requirements("Biospecimen", as_graph = True)

you'd get

[('Biospecimen', 'Patient'), ('Patient', 'Demographics'), ('Patient', 'FamilyHistory'), ('Patient', 'Exposure'), ('Patient', 'FollowUp'), ('Patient', 'Diagnosis'), ('Patient', 'Therapy'), ('Patient', 'MolecularTest')]

and if you do 

metadata_model.get_component_requirements("Biospecimen", as_graph = False)

you'd get 

['Demographics', 'FamilyHistory', 'Exposure', 'FollowUp', 'Diagnosis', 'Therapy', 'MolecularTest', 'Patient', 'Biospecimen']

rrchai:
Cool! It looks great and useful. I will apply it to the dashboard feature once it is merged.

Thank you, Milen.

schematic/models/metadata.py
```diff
@@ -155,26 +155,33 @@ def getModelManifest(
...
         req_components = self.sg.get_component_requirements(source_component)
-        # req_components = get_component_requirements(mm_graph, source_component)
+
+        if not as_graph:
+            return req_components
+        else:
+            req_components_graph = self.sg.get_component_requirements_graph(source_component)
+
+            # serialize component dependencies DAG to a edge list of node tuples
+            req_components = list(req_components_graph.edges())
```
sujaypatil96, line 180: This snippet can be simplified as follows:

```python
if as_graph:
        req_components_graph = self.sg.get_component_requirements_graph(source_component)

        # serialize component dependencies DAG to a edge list of node tuples
        req_components = list(req_components_graph.edges())

        return req_components
```

schematic/schemas/explorer.py
```diff
@@ -93,7 +93,7 @@ def get_descendants_by_edge_type(
         source_node: str,
         relationship: str,
         connected: bool = True,
-        ordered: bool = False,
+        ordered: bool = False
```
sujaypatil96, line 96: Good catch.👍

tests/test_metadata.py
```diff
@@ -0,0 +1,31 @@
...
+            helpers.get_data_path("example.model.jsonld"),
+            "local"
+        )
+
+        yield mm
+
+
+class TestMetadataModel:
+
+    def test_get_component_requirements(self, metadata_model):
```
sujaypatil96, line 24: This test can take `as_graph` equal to both `True` and `False` using the `@pytest.mark.parametrize` syntax.

sujaypatil96 (approved):
The comments I've left are purely recommendations, which you can include if you want. As such, the code works and can be merged.🚀

schematic/models/metadata.py
milen-sage: :) yes this makes a lot of sense - could you amend the code?

tests/test_metadata.py
```diff
@@ -0,0 +1,31 @@
...
+            
+        source_component = "BulkRNA-seqAssay"
+        as_graph = True
+        
+        output = metadata_model.get_component_requirements(source_component, as_graph = as_graph)
+        
+        assert type(output) is list
```
sujaypatil96, line 31: Test this against a list of expected dependencies.

schematic/models/metadata.py
sujaypatil96: I've changed the code to use this simplified logic.

tests/test_metadata.py
sujaypatil96: Used the above syntax to allow the same test `test_get_component_requirements()` to take two parameter values for `as_graph` argument. It takes both `True` and `False`.

Issue #482, opened 2021-06-01
Add Windows-specific instructions to documentation


lopierra:
**Is your feature request related to a problem? Please describe.**
The instructions to activate a Python virtual environment may be slightly different for Windows than for Mac/Linux

**Describe the solution you'd like**
The Mac/Linux command, as currently written in the documentation, is  `source .venv/bin/activate`
On Windows (at least on my machine - Windows 10), I did not have a `bin` folder - instead I used `source .venv/Scripts/activate`

I also found that I need to use Cygwin or other Unix emulator, or use the Git Bash window. The commands did not work in the Windows Command Prompt app. (That might be pretty basic, but as a beginner it took me a while to figure this out.)

**How important is this feature?** Select from the options below:
• Low - it's an enhancement but not crucial for work

**When will use cases depending on this become relevant?** Select from the options below:
• Whenever we have new Windows users

HumanG33k:
You should instead use ubuntu not so hard to install and use. https://ubuntu.com/tutorials/install-ubuntu-desktop#1-overview

sujaypatil96:
> The commands did not work in the Windows Command Prompt app.

@lopierra: I wanted to ask, what command specifically did not work in cmd? Once the virtual environment has been activated in the way you mention, and the package has been installed, the schematic commands for Windows should be the same as the ones we write for bash.

lopierra:
@sujaypatil96: adding the `.\.venv\Scripts\activate    # on Windows` line is helpful, thanks!

More generally, any of the Unix commands will not work in Windows `cmd.exe`. You have to have a specific bash program like git bash, [Cygwin](https://www.cygwin.com/), or [WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10). I think you could just mention it in "2. Installation Requirements and Pre-requisites".

sujaypatil96:
@lopierra: Perfect, thank you for the comment. Does this [commit](https://github.com/Sage-Bionetworks/schematic/pull/492/commits/13683eb3176094cb8e1cc493cf90ee67645e6131) appropriately address what you requested?

Issue #445, opened 2021-04-08
Revise schematic_service_account_creds.json documentation to help users understand what this is and why they might do it


allaway:
**Describe the bug**
Documentation doesn't sufficiently describe how to do this step: "Use the schematic_service_account_creds.json file for the service account mode of authentication (for Google services/APIs)."

**To Reproduce**
Read the docs :) 

**Expected behavior**
I would expect this section to describe how to create or obtain this json. 

**Screenshots**
n/a

**Desktop (if applicable, please complete the following information):**
n/a

**Additional context**
n/a

allaway:
I guess a related thing that I do not understand in the docs: is the OAuth2 credentials/token.pickle method of authentication sufficient, or do I also need to define `schematic_service_account_creds.json`? As far as I can tell, the answer is yes, both need to exist to run schematic.

sujaypatil96:
Can this issue be closed @allaway? Like you mentioned in your followup comment, the relatively new `--oauth` flag allows you to choose between using `OAuth` and the `Service Account`.

allaway:
Hi Sujay, Sorry for not being clearer: I think the issue is primarily that there is not enough context. What is the service account mode of authentication? Why would a user want to do this instead of oauth?

sujaypatil96:
I see, okay. I think there's sufficient documentation on the Google API docs.
1. Authentication overview: https://cloud.google.com/docs/authentication
2. Service Account authentication: https://cloud.google.com/docs/authentication/production

But what you're saying does make sense. I'll add a few more lines of context here ([details.md](https://github.com/Sage-Bionetworks/schematic/blob/develop/docs/md/details.md)). Describing the `service account` and its usage in a little more detail. I'll tackle this in a future documentation update PR.

allaway:
Thanks, I agree that google's docs are probably sufficient for the how (which was the original title of this issue - I'll change that), but the why and how it relates to schematic is something I was also struggling to understand. Thanks!!

PR #492, opened 2021-07-01
Contributor docs update

sujaypatil96:
This PR updates the contributor docs in the `CONTRIBUTION.md` file.

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -126,7 +126,10 @@ refer to the ``Credentials`` section in the
 document.
 
 Use the ``schematic_service_account_creds.json`` file for the service
-account mode of authentication (*for Google services/APIs*).
+account mode of authentication (*for Google services/APIs*). Service accounts 
+are special Google accounts that can be used by applications to access Google APIs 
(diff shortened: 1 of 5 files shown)


sujaypatil96:
This [commit](https://github.com/Sage-Bionetworks/schematic/pull/492/commits/0e7a8d13667a40f0efd0dcbc64b4b17954626ee8) was added to address issue #445.

docs/README.rst
```diff
@@ -52,7 +52,9 @@ credentials files.
 
 .. code:: bash
 
-    source .venv/bin/activate # activate the `venv` virtual environment
+    # activate the `.venv` virtual environment
+    source .venv/bin/activate   # on Linux and MacOS
+    .\.venv\Scripts\activate    # on Windows
```
sujaypatil96, line 57: @lopierra: I added these lines to address issue #482. Do you think this is sufficient? I think most of the other commands work the same way on Windows as they do in a Unix based environment, right?

lopierra:
Could you add git bash as an option as well? It's the easiest if you already use git, since it's installed along with git. Otherwise it looks good!

PR #493, opened 2021-07-06
Update getDatasetManifest with return value and downloadLocation

rrchai:
1. Change `downloadLocation` to its synId folder (e.g `'./syn12345678/synapse_storage_manifest.csv'`). It prevents manifest with the same names from being overwritten. In addition, as downloaded files exist, synapse will check if file needs to be downloaded or not, which increases the downloading speed for multiple datasets.

2. Return `manifest_data` to get the entity information, no matter of `downloadFile ` is set `True` or `False`.
    - Not sure if it will affect other functions since it will not return synIds when `downloadFile` is `False`

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -269,16 +269,18 @@ class SynapseStorage(BaseStorage):
         return file_list
 
     def getDatasetManifest(
-        self, datasetId: str, downloadFile: bool = False
+        self, datasetId: str, downloadFile: bool = False, downloadPath: str = None
     ) -> List[str]:
         """Gets the manifest associated with a given dataset.
 
         Args:
             datasetId: synapse ID of a storage dataset.
             downloadFile: boolean argument indicating if manifest file in dataset should be downloaded or not.
+            downloadPath: target directory to download the manifest file
 
         Returns:
-            manifest_data (synapseclient.entity.File): Synapse entity.
+            manifest_syn_id (String): Synapse ID of exisiting manifest file.
+            manifest_data (synapseclient.entity.File): Synapse entity if downloadFile is True.
             "" (String): No pre-exisiting manifest in dataset.
         """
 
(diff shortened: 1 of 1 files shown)


rrchai:
The second change is unnecessary for the UI-dashboard feature purpose (I was used it to test `etag` and that's why I needed to get `manifest_info` when `downloadFile` set false). If you feel the change is not needed for now, feel free to change it back.

sujaypatil96 (changes requested):
The function that you're modifying in this PR, [getDatasetManifest](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L271), is one of the core library methods, that is called in multiple other places by other modules in its current state. So I'd rather we not modify the core behaviour of the function itself.

The approach that I recommend is the use of Python [decorators](https://realpython.com/primer-on-python-decorators/) to modify the behaviour of `getDatasetManifest()`.

You can implement something along these lines:

In `synapse.py`:

```python
# defining the decorator
def modified_getDatasetManifest(func):
    # wrapper function
    def inner():
         # modified code
         func()
         
    return inner
```

In calling module:

```python
desired_result = modified_getDatasetManifest(getDatasetManifest)
```

> All functions can be named appropriately, I just used random placeholders for them in the above snippet.

rrchai:
> The approach that I recommend is the use of Python [decorators](https://realpython.com/primer-on-python-decorators/) to modify the behaviour of `getDatasetManifest()`.

Thank you for the suggestions. @sujaypatil96 

Please correct me if I misunderstand. Since the `downloadLocation` is also altered and it is not in the argument, I wonder whether `decorators` would be an easy way to approach. 

I made new commits that did the following two things:
1. add path argument and set as `CONFIG["synapse"]["manifest_folder"]` by default
2. reset the second changes on return value, so it will still return `synapseId` if `downloadFile` set `false`

I think in this way, we can minimize the changes that fulfill the front-end needs, but also remain what `getDatasetManifest` did before and not affect other functions.

ychae:
@milen-sage please review and merge if complete

rrchai:
This PR was originally made for data curator dashboard. With the latest updates on dashboard, it  now does
1. not need to download files 
2. not need to specify download path since it can be found in synapseCache
3. use synapse's `get()` directly to replace `getDatasetManifest()`

Close this PR since the changes no longer needed for the dashboard now.

PR #496, opened 2021-07-09
Develop manifest update fix - handle mismatch between schema and existing manifest column-set

milen-sage:
I want to test a bit more before merging in develop, but feel free to review @sujaypatil96 

@BrunoGrandePhD if you are interested (since this code interacts w/  existing set_dataframe_by_url function you write and it may be relevant for other projects like imCORE), feel free to peruse too.

If an existing manifest's column-set doesn't match the latest schema, there are a couple of options:
- either the existing manifest is missing columns that are in the latest schema
- or the latest schema is missing columns that are in the existing manifest

This PR aims to handle generation of google sheet from an existing manifest by
- including all columns from the latest schema in the resulting google sheet
- including all columns from the existing manifest in the resulting google sheet (so that a user can more easily make changes to their data, e.g. to comply with schema requirements)
- sorting columns in the dataframe used to populate the resulting google sheet so that 
---- columns specified in the latest schema are ordered according to the latest schema (this ordering aligns with the column order in the empty google sheet, which we are populating w/ the existing manifest)
---- 'out-of-schema' columns in the existing manifest (if any) are added to the end of the google sheet
---- ensuring any remaining extraneous google sheet validation rules in the sheet range of out-of-schema columns are dropped 

Hopefully this fix handles the most common schema-existing manifest mismatch scenarios.

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -415,6 +415,9 @@ class ManifestGenerator(object):
             ordered_metadata_fields[0]
         )
 
+        logger.info("Empty manifest ordered columns")
+        logger.info(ordered_metadata_fields)
+
         body = {"values": ordered_metadata_fields}
 
         # determining columns range
@@ -752,7 +755,6 @@ class ManifestGenerator(object):
         # print("Manifest successfully generated from schema!")
         # print("URL: " + manifest_url)
         # print("========================================================================================================")
-
         return manifest_url
 
     def set_dataframe_by_url(
@@ -767,6 +769,9 @@ class ManifestGenerator(object):
         Returns:
             ps.Spreadsheet: A Google Sheet object.
         """
+
+        logger.info(manifest_df)
+
         # authorize pygsheets to read from the given URL
         gc = ps.authorize(custom_credentials=self.creds)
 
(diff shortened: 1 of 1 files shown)


sujaypatil96 (commented):
@milen-sage: I did some testing and here is how the code created a manifest based on a different version of the schema.

The command I ran is as follows, for the _Exposure_ data type.
```
$ schematic manifest --config config.yml get --dataset_id syn21973663 --oauth --sheet_url
```

The results are as below:

- The manifest generated using this version of the new [NICHTAN HTAN schema](https://raw.githubusercontent.com/ncihtan/data-models/main/HTAN.model.jsonld): [new manifest](https://docs.google.com/spreadsheets/d/1ZPqwxQxiNOsc4LqrbWnV3O8ropaBTtdFg7w6YjdZcXI/edit#gid=0)
- The manifest generated using an older version of the [HTAN schema](https://github.com/Sage-Bionetworks/schematic/blob/ceda275aefd63b56a20c5f25c670469d9cd958fd/data/schema_org_schemas/HTAN.jsonld): [old manifest](https://docs.google.com/spreadsheets/d/1OyuItmVHOwQdWzLM64m_zSxRrvo-SCLlQZRNz1ZYbFY)

The code does seem to be accommodating support for additional columns (either already existing ones in the manifest file, or new ones from the schema) well.

schematic/manifest/generator.py
```diff
@@ -767,19 +766,54 @@ def set_dataframe_by_url(
...
+        # update existing manifest w/ missing columns, if any
+        if new_columns:
+            manifest_df = manifest_df.assign(**dict(zip(new_columns, len(new_columns)*[""])))
+
+        # sort columns in the updated manifest:
+        # match latest schema order
+        # move obsolete columns at the end
+        manifest_df = manifest_df[self.sort_manifest_fields(manifest_df.columns)]
+        manifest_df = manifest_df[[c for c in manifest_df if c not in out_of_schema_columns] + list(out_of_schema_columns)]
+
```
sujaypatil96, line 805: Do you think it makes sense to wrap this portion of the code in another function, say `_sync_manifest_with_schema()`? It's not essential by any means, and works well without it. I was thinking about it from a functional abstraction point of view.

schematic/manifest/generator.py
```diff
@@ -767,19 +766,54 @@ def set_dataframe_by_url(
...
+        
         # The following line sets `valueInputOption = "RAW"` in pygsheets
         sh.default_parse = False
 
         # update spreadsheet with given manifest starting at top-left cell
         wb.set_dataframe(manifest_df, (1, 1))
 
+        # update validation rules (i.e. no validation rules) for out of schema columns
+        start_col = self._column_to_letter(len(wb_header)) # find start of out of schema columns
+        end_col = self._column_to_letter(len(manifest_df.columns) - 1) # find end of out of schema columns
```
sujaypatil96, line 814: You can reuse `manifest_df_header` here rather than computing `manifest_df.columns` again.

milen-sage: If I am not mistaken, manifest_df_header contains the old column set (i.e. existing manifest prior update basded on latest schema)?

schematic/manifest/generator.py
milen-sage: It could be a useful function yes - could you add a TODO and/or github issue? There is quite  a bit of refactoring needed in this manifest generator class. Let's think about doing it wholesale in the next milestone (in the context of deeper refactor to allow serializing manifest UI configs in other frameworks other than google sheets).

sujaypatil96 (approved):

sujaypatil96 (changes requested):
`EntityId` column appears in the middle of the manifest. We need to ensure that the column is appended at the end.

milen-sage:
I have tried to replicate @sujaypatil96's test results above. Running 

schematic manifest -v INFO --config ./config.yml get --data_type Exposure -d syn21973663 --oauth --sheet_url

results in 

https://docs.google.com/spreadsheets/d/1wcogLI3hOPXfRCURyQHOsxHy9kJANpvfJoeJnzcHKFM/edit#gid=0

which is accurate and as expected. We may need to update the HTAN.model.jsonld schema in [ncihtan](https://raw.githubusercontent.com/ncihtan/data-models/main/HTAN.model.jsonld) if that results in the outputs above.

milen-sage:
> `EntityId` column appears in the middle of the manifest. We need to ensure that the column is appended at the end.

See comment above - I think we are good to merge.

milen-sage:
@milen-sage open a separate PR to update columns of existing manifest csv file to match the latest schema (when not loading existing manifest file in a google sheet).

milen-sage:
@sujaypatil96 Based on the discussion above, I think we are ready to merge?

sujaypatil96 (approved):
@milen-sage: Based on the conversation on [Slack](https://sagebionetworks.slack.com/archives/C026AFVTW4X/p1626919912005400), as confirmed by @mialy-defelice, the code works as expected. We are good to merge 🚀.

PR #497, opened 2021-07-15
Jupyter notebook that illustrates all the steps in the data curation lifecycle

sujaypatil96:
This PR seeks to add a iPython notebook showcasing all the major steps of the schematic data curation lifecycle, which are generation, validation and submission of a manifest.

PR #498, opened 2021-07-20
Jupyter notebook that validates manifest files

sujaypatil96:
This PR logically follows PR #497, and demonstrates the second stage of the data curation lifecycle, i.e., manifest validation. Both the use cases, when the manifest is valid and when the manifest is invalid have been illustrated.


sujaypatil96:
Addressed in PR #497.

PR #510, opened 2021-08-17
Develop add google regex

mialy-defelice:
The purpose of this pull request is to add google sheets support for regex validation, for attributes in which a regex match validation rule has been specified. 

_From in code documentation_

**Purpose:**
- Apply regular expression validaiton rules to google sheets.
- Will only be run if the regex module specified in the validation rules is 'match'.
  - This is because of the limitations of google sheets regex.
- Will do the following:
  - In google sheets user entry text will initially appear red.
  - Upon correct format entry, text will turn black.
  - If incorrect format is entered a validation error will pop up.

**Input:**
- validation_rules: List[str], defines the validation rules applied to a particular column.
- i: int, defines current column.

**Returns:**
- Performs an in function update of the google sheet to comply with conditional formatting and data validation specifications.

**TODO:**
- Add to a different location in script, but might be confusing since other updaters are not in function, and this function is not very general.

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -613,38 +613,71 @@ class ManifestGenerator(object):
                     # Add conditional formatting for regular expressions.
                     # Color the cell green if the user inputs the correct value.
                     regular_expression = split_rules[2]
-                    vr_bg_color = CONFIG["style"]["google_manifest"].get(
-                        "vr_bg_color",
-                        {
-                            "red": 0.778,
-                            "green": 0.978,
-                            "blue": 0.582,
-                        })
-                    requests_vr = {
-                        'requests': [{
-                            'addConditionalFormatRule': {
-                                'rule': {
-                                    "ranges": {
-                                        "startColumnIndex": i,
-                                        "endColumnIndex": i + 1,
-                                    },
-                                    'booleanRule': {
-                                        'condition': {
-                                            'type': 'CUSTOM_FORMULA',
-                                            'values': [{
-                                                'userEnteredValue':
-                                                    '=REGEXMATCH(INDIRECT("RC",FALSE), "{}")'.format(regular_expression)
-                                            }]
+
+                    def update_base_color_request(color = {"red": 1.0}):
+                    
+                        # Change the base font color to red.
+                        # When the condiitonal formatting is applied it will change
+                        # the color to black if a user enters a correctly formatted value.
+                        # this coloring is intended to serve as a visual indicator to the 
+                        # user that the entered values are being validated.
+                        vr_format_body = {
+                            "requests": [
+                                {"repeatCell": {
+                                        "range": {
+                                            "startColumnIndex": i,
+                                            "endColumnIndex": i + 1,
+                                            "startRowIndex": 1,
                                         },
-                                        'format': {
-                                            'backgroundColor': vr_bg_color
-                                        }
+                                        "cell": {
+                                            "userEnteredFormat": {
+                                                "textFormat": {
+                                                    "foregroundColor": color}
+                                                }
+                                        },
+                                        "fields": "userEnteredFormat(textFormat)",
                                     }
-                                },
-                                'index': 0
-                            }
-                        }]
-                    }
+                                }
+                            ]
+                        }
+                        return vr_format_body
+
+                    vr_format_body = update_base_color_request(color={"red": 232./255., "green": 80./255., "blue": 70./255.})
+                    requests_body["requests"].append(vr_format_body["requests"])
+
+                    def make_regex_vr_request(gs_formula, text_color = {'red': 1}):
+                        requests_vr = {
+                            'requests': [
+                                {'addConditionalFormatRule': {
+                                        'rule': {
+                                            "ranges": {
+                                                "startColumnIndex": i,
+                                                "endColumnIndex": i + 1,
+                                                "startRowIndex": 1,
+                                            },
+                                            'booleanRule': {
+                                                'condition': {
+                                                    'type': 'CUSTOM_FORMULA',
+                                                    'values': gs_formula,
+                                                },
+                                                'format': {
+                                                    'textFormat': {
+                                                        "foregroundColor": text_color
+                                                    }
+                                                }
+                                            }
+                                        },
+                                        'index': 0
+                                    }
+                                }
+                            ]
+                        }
+                        return requests_vr
+
+                    text_color = {"red": 0, "green": 0, "blue": 0}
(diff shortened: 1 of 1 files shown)


mialy-defelice:
@allaway Hey Robert, Tagging you here be we are updating Schematic to allow Regex validation and will impact how the config file and schema are used to set strictness levels for google sheet validation. Strictness can be now passed as an argument to each function. Additionally, the config file can allow a default to be set. Validation rules themselves can specify validation strictness for a regex match validation rule. We want to make sure these changes will still be compatible with how you may be using data validation. Please comment if you have any concerns!

allaway:
Hi @mialy-defelice, 

Thanks for looping me in! This sounds fine to me, but I have a couple of questions. 

>Strictness can be now passed as an argument to each function.
I don't quite follow what this means, can you clarify? 

Also, can you clarify what you mean by 'strictness'? Is this the google sheets WARN vs REJECT behavior, or something else? 

thanks!

mialy-defelice:
Hey @allaway,

**To your second question first:** 
Yes, exactly! We call it strictness in the code and config file, but yes, the variable `strict` will define if the sheet rejects user input, or simply issues a warning. 

- `strict = False` will issue a warning
- `strict = True` will reject the user input and supply an error message
- `strict = 'default` will pull from the CONFIG file.

**To the first:**
There are various points in setting up the sheet where data validation is employed. Previously, this all came from the single setting defined in the CONFIG file. In this PR I am allowing each instance of setting up the data validation to set how stringent the data validation should be (so warn or reject). This gives finer control and will allow you to set it depending on what you want to happen. For this PR, I set validation strictness to `strict='default'` but this can be now customized if you want some parts issue a rejection and other parts to issue just a warning. If you need help doing this I can help!

- **For example:** 
-    When adding a `regex match` rule, you can now specify `strict` as well in the rule (at the end) ((i.e. `regex match [0-9]{4}\/[0-9]* strict`)). Doing this would cause a rejection of user input for that specific rule. If the CONFIG file strictness was set to `False` all other `regex match` rules would only issue a warning, if users input an incorrect value. By setting this default to `False`, all other data validation would now only issue warnings. But if you wanted you could go in and convert other data validations to Reject or `strict=True`

[Here](https://docs.google.com/spreadsheets/d/13HgBKmd-C7MMNVJfqnevaF06rQJtRLjggbXyFC1k4gc/edit#gid=0) is an example of what it looks like when `regex match [0-9]{4}\/[0-9]* strict` is set for column AQ.

And [this](https://docs.google.com/spreadsheets/d/1NPZdcDoDD4BB-PTuUt20zrDN4S1q3gHq20JdXdG-aDw/edit#gid=0) is when `regex match [0-9]{4}\/[0-9]*` is set for column AQ.

allaway:
@mialy-defelice, thanks for the explanation and examples! 

Based on what you've said, it sounds like if I left our schema and config untouched but used this branch, that I wouldn't see any difference between this and the current release of schematic.

However, this looks like it could be really useful! I think for now we will stick with our default config (warn) but we'll surely want to use the regex validation and once we start to do that we'll re-assess whether we need to set anything more strict.

schematic/manifest/generator.py
```diff
@@ -248,12 +248,15 @@ def _get_column_data_validation_values(
...
         validation_type="ONE_OF_LIST",
         custom_ui=True,
         input_message="Choose one from dropdown",
-    ):
+    ):  
 
-        strict = CONFIG["style"]["google_manifest"].get("strict_validation", True)
+        # set validation strictness to config file default if indicated.
+        if strict == "default":
+            strict = CONFIG["style"]["google_manifest"].get("strict_validation", True)
```
BrunoGrandePhD, line 259: It's strange that `strict` can be either a string (`"default"`) or a boolean (`True`). Is there a way to avoid this? If `"default"` is simply intended to act as a flag value for "nothing was provided", then I suggest using `None` instead since it's more conventional.

BrunoGrandePhD: Alternatively, you could avoid confusion by using different names for the string and boolean variables.

BrunoGrandePhD (commented):
I'm not super well-versed in the Google Sheets API, so I skimmed over that part and defer to @milen-sage. Instead, I just took a look at the function interface and had a single comment. I also threw in the suggestion to run `black` on the code for consistency. 

Great work! 🚀

schematic/manifest/generator.py
```diff
@@ -283,7 +286,7 @@ def _get_column_data_validation_values(
                 )
                 .execute()
             )
-
+
```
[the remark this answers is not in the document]
mialy-defelice: Done

schematic/manifest/generator.py
mialy-defelice: Changed `'Default'` to `None`

mialy-defelice:
Ready to merge :)

milen-sage:
@mialy-defelice works like a charm!

milen-sage (approved):

allaway:
Super belated review on my part....

@mialy-defelice asked me to validate this function using this json file: `nf_research_tools_gs_vr_test.rdb.model.jsonld` but I am having some trouble (most certainly user error, but I can't find any docs that would help me fix this). Here's what I did: 

1. `pip install git+https://github.com/Sage-Bionetworks/schematic.git@develop-add-google-regex` in a conda environment
2. `schematic manifest --config config.yml get -s  -t test -dt Resource -p 'nf_research_tools_gs_vr_test.rdb.model.jsonld' -oa` 
3. I get this:
```
(data_curator_env) ➜  schematic-upstream git:(develop) ✗ schematic manifest --config config.yml get -s  -t test -dt Resource -p 'nf_research_tools_gs_vr_test.rdb.model.jsonld' -oa
Starting schematic...
The (model > input > validation_schema) argument with value 'data/validation_schemas/example_validation_schema.json' is being read from the config file.
The '--json_schema' argument is being taken from configuration file (model > input > validation_schema), i.e., 'data/validation_schemas/example_validation_schema.json'.
Permission Id: anyoneWithLink
Find the manifest template using this Google Sheet URL:
https://docs.google.com/spreadsheets/d/131VAdy5Yu_7FiY1-2PsV8dL56jrM7mOdxU1bD4jii18
```

The google sheet seems to be taking values from `data/validation_schemas/example_validation_schema.json` via my config file rather than `nf_research_tools_gs_vr_test.rdb.model.jsonld`. Any idea what's going on here?

mialy-defelice:
Hey @allaway,

I don't know what's causing this error, could you make this into an issue? For now I would just update the config file to point to the correct jsonld.

Thank you!

PR #519, opened 2021-09-09
Update logger to allow list validation rule without valid values specified.

mialy-defelice:
Previously there was an error if you worked in the DEBUG level of logging, if you specified a list validation rule without also adding valid values. This was because it was assumed these would be used in conjunction, and the logging message relied on a specific variable. This PR changes the message so its still useful, but does not rely on this variable.


BrunoGrandePhD (approved):
🚀

Issue #522, opened 2021-09-27
`AssertionError` when generating manifest for previously annotated "file-less" component


BrunoGrandePhD:
## Describe the bug

I'm having trouble generating a Google Sheet manifest for a biospecimen dataset (_i.e._ without data files). I first encountered the issue in the Data Curator, and I was able to reproduce it with just schematic on my local machine. In the code block below, you can see that I'm able to create a manifest from a previously annotated `BulkRNAseqLevel1` dataset, but the same command for a previously annotated `Biospecimen` dataset fails with an `AssertionError` in `pygsheets`. 

I'm going to do some more digging, but I wanted to create a ticket to report any findings. 

I'm using the latest commit (b01bd03c1e9a3cd0b5c610c9af5aea2ecc119750) on the `develop` branch. This issue happened on both my local machine (macOS) and on the Shiny Server (Linux)

```
❯ schematic manifest --config ../imcore_data_model/config.yml get --data_type "BulkRNAseqLevel1" --jsonld ../imcore_data_model/imcore.model.jsonld --sheet_url --dataset_id syn26010235 
[...]
Permission Id: anyoneWithLink
Find the manifest template using this Google Sheet URL:
https://docs.google.com/spreadsheets/d/10YWjZy41Aktqh0KiipzTVrxUCXw5WS5cuEUVpbTs1Qc


❯ schematic manifest --config ../imcore_data_model/config.yml get --data_type "Biospecimen" --jsonld ../imcore_data_model/imcore.model.jsonld --sheet_url --dataset_id syn26011024 
[...]
Permission Id: anyoneWithLink
Traceback (most recent call last):
  File "/Users/bgrande/Repos/schematic-dev/.venv/bin/schematic", line 5, in <module>
    main()
  File "/Users/bgrande/Repos/schematic-dev/.venv/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/bgrande/Repos/schematic-dev/.venv/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/bgrande/Repos/schematic-dev/.venv/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/bgrande/Repos/schematic-dev/.venv/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
[...]

PR #523, opened 2021-09-27
Fix google sheet validation rules range calculation and maintains up-to-date list of files in manifest sheet for existing datasets

milen-sage:
Addresses https://github.com/Sage-Bionetworks/data_curator/issues/190

Should also address  #522.
@BrunoGrandePhD could you confirm this resolves the issue?


schematic/manifest/generator.py
```diff
@@ -782,13 +782,16 @@ def set_dataframe_by_url(
...
-        # get headers from existing manifest and sheet
+        # get headers from existing manifest and sheet 
         wb_header = wb.get_row(1)
         manifest_df_header = manifest_df.columns
         
         # find missing columns in existing manifest
         new_columns = set(wb_header) - set(manifest_df_header)
 
+        # clean empty columns if any are present (there should be none)
+        new_columns = new_columns.remove('')
```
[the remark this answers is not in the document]
BrunoGrandePhD: Maybe we can capture this as a `TODO` comment: 

```suggestion
        # TODO: Remove this line once we start preventing empty column names
        new_columns = new_columns.remove('')
```

schematic/manifest/generator.py
```diff
@@ -801,7 +804,8 @@ def set_dataframe_by_url(
         # move obsolete columns at the end
         manifest_df = manifest_df[self.sort_manifest_fields(manifest_df.columns)]
         manifest_df = manifest_df[[c for c in manifest_df if c not in out_of_schema_columns] + list(out_of_schema_columns)]
-        
+       
+        print(manifest_df.columns)
```
BrunoGrandePhD, line 808: Did this sneak through?

```suggestion
```

schematic/manifest/generator.py
```diff
@@ -782,13 +782,16 @@ def set_dataframe_by_url(
         # column-set of the existing manifest so that the user can modify their data if needed 
         # to comply with the latest schema
 
-        # get headers from existing manifest and sheet
+        # get headers from existing manifest and sheet
```
BrunoGrandePhD, line 785: If you run the black formatter on the repo, these errant spaces should be stripped.

BrunoGrandePhD (commented):
I'm leaving my first-pass comments here for now. I'm going to test this branch now. I'll approve the PR once I confirm that this solves the issue I was running into with #522.

BrunoGrandePhD:
I created a [dataset folder](https://www.synapse.org/#!Synapse:syn26251193) and annotated it using the `Patient` component. When I try to generate a [manifest template](https://docs.google.com/spreadsheets/d/1mNUNzTFU1oRMCCwwmnXC1e9Px63kDpYVv5WPcNnJAIY/edit#gid=0), it lacks an entity ID field, which I presume is needed to update the correct entities (instead of creating new ones). Is this expected? 

```
❯ schematic manifest --config ./tests/data/test_config.yml get --data_type "Patient" --jsonld ./tests/data/example.model.jsonld --oauth --sheet_url --dataset_id syn26251193 
[...]
Permission Id: anyoneWithLink
Find the manifest template using this Google Sheet URL:
https://docs.google.com/spreadsheets/d/1nxOdhiOwEV-sEkt-nUHAdL2QkG-GbBmQZX9w3Kavaqo
```

milen-sage:
@BrunoGrandePhD thanks for reviewing! 

The expected behavior when generating a manifest for a dataset of a 'non-file' data type:

- if there is no pre-existing manifest in the dataset folder, then the newly generated manifest does not contain entityIDs; schematic creates the entityIDs on manifest **submission**

- if there is a pre-existing manifest in the dataset folder (i.e. a manifest has already been submitted for this dataset), the generated manifest sheet should contain the corresponding entityIDs (present in the submitted manifest file).

Note: the manifest filename should match the parameter in schematic's config file (e.g. synapse_storage_manifest.csv).

Needless to say, all this would be great to document. We are having a doc-a-thon and will cover it.

schematic/manifest/generator.py
```diff
@@ -782,13 +782,18 @@ def set_dataframe_by_url(
...
+        # get headers from existing manifest and sheet 
         wb_header = wb.get_row(1)
         manifest_df_header = manifest_df.columns
         
         # find missing columns in existing manifest
         new_columns = set(wb_header) - set(manifest_df_header)
 
+        # clean empty columns if any are present (there should be none)
+        # TODO: Remove this line once we start preventing empty column names
+        if '' in new_columns:
```
milen-sage, line 794: This line is getting more and more stinky, but had to add this check as well.

milen-sage:
I'm overloading a bit this PR... addressing issue https://github.com/Sage-Bionetworks/data_curator/issues/193

When calling getModelManifest in MetadatModel (and respectively get_manifest in ManifestGnerator) for an existing dataset, the resulting google sheet will contain an up-to-date list of filenames and entity IDs. 

Previously, the resulting google sheet would be populated only with the content of the existing manifest, if any is present. The existing manifest would be stale if new files were added to the dataset since the existing manifest had been submitted.

The new files should be appended at the end of the manifest. Currently all fields other than filename and entityId are left blank (i.e. empty string).

To test in schematic for the dataset referenced in [this issue](https://github.com/Sage-Bionetworks/data_curator/issues/193) (note this is a HTAN dataset and yo umay need access):

schematic manifest -v INFO --config ./config-htan.yml get --data_type ScRNA-seqLevel1 --title scRNA-seqLevel1 -d syn23520241 --oauth --sheet_url

Output here:
https://docs.google.com/spreadsheets/d/19BH8yUc8AtmVVPaYajdwUDMkVPOR-jmuNPm8WGP4TVs/edit#gid=0

BrunoGrandePhD:
> The expected behavior when generating a manifest for a dataset of a 'non-file' data type:
> 
> * if there is no pre-existing manifest in the dataset folder, then the newly generated manifest does not contain entityIDs; schematic creates the entityIDs on manifest **submission**
> * if there is a pre-existing manifest in the dataset folder (i.e. a manifest has already been submitted for this dataset), the generated manifest sheet should contain the corresponding entityIDs (present in the submitted manifest file).

I'm seeing this behavior when generating a [manifest](https://docs.google.com/spreadsheets/d/1Lza6HLdMlnoQ_QUKXc4N3iqpFpY4rDo253GTgIYVUY0/edit#gid=0) from a previously annotated RNA-seq (_i.e._ file-based) [dataset](https://www.synapse.org/#!Synapse:syn26010235), but it isn't happening with the [template](https://docs.google.com/spreadsheets/d/1EqbVs7Ja33w1i4wOhWQ5G_SNhlORJgRRlEZYSFbgOno/edit#gid=0) for a Biospecimen (_i.e._ non-file-based) [dataset](https://www.synapse.org/#!Synapse:syn26011024). 

> Note: the manifest filename should match the parameter in schematic's config file (e.g. synapse_storage_manifest.csv).

I did change this to just `manifest.csv` in my schematic config, but I'm using this config when running schematic. Also, the fact that schematic is able to pick up the manifest for the RNA-seq dataset indicates that it doesn't have to do with the manifest file name. 

@milen-sage: Once you accept the invitation to join this [Synapse team](https://www.synapse.org/#!Team:3414725), you should be able to see the test dataset folders that I've linked above.

BrunoGrandePhD:
I annotated a [dataset folder](https://www.synapse.org/#!Synapse:syn26251193) using the `Patient` component in our example data model. Here's the resulting [manifest template](https://docs.google.com/spreadsheets/d/1tQzXi90q4AYeEdLCrYZx9aU5Ft-XrNCf3kWTVGJl4mg/edit#gid=0) when I run the following command. As you can see from the output, schematic is picking up on the existing manifest file (_i.e._ `syn26251389`), but it isn't populating the template with its contents. ¯\_(ツ)_/¯

```
❯ schematic manifest --config ./tests/data/test_config.yml get --data_type "Patient" --jsonld ./tests/data/example.model.jsonld --oauth --sheet_url --dataset_id syn26251193
[...]
    MANIFEST:                                                           id                          name
26251389_1_c53eca1b-d091-4b3a-b2b9-be2faed2e9c7  syn26251389  synapse_storage_manifest.csv
[...]
Permission Id: anyoneWithLink
Find the manifest template using this Google Sheet URL:
https://docs.google.com/spreadsheets/d/1vOmSzAawapV_ElZ7E068QtXBfaUDW3fOLJcmc8sCObI
```

milen-sage:
Ok I can reproduce. I can also reproduce with the Therapy component in HTAN's data model, which used to work as expected prior the later commits on this PR.

E.g. 

schematic manifest -v INFO --config ./config-htan.yml get --data_type Therapy --title "Therapy" -d syn24615247 --oauth --sheet_url

previously resulted in 

https://docs.google.com/spreadsheets/d/1Wn3F5pdT1ZR8Osin9ynXHCjpMThMWSAi4seN_jl4v7s/edit#gid=0

and now results in 

https://docs.google.com/spreadsheets/d/1cXebCKF4YF-3UMbsDPlzQTAHkSwyRCsQSO3fpYw8kZQ/edit#gid=0

I think I see what's causing this.

milen-sage:
Ok, I think the commit^ fixes it @BrunoGrandePhD? 

If yes, @mialy-defelice's rule of thumb on <10 line functions would have avoided this error :)

BrunoGrandePhD:
I was able to generate a [manifest template](https://docs.google.com/spreadsheets/d/1Yv_eLO-SKObn8qY47Fb6wtLCE8iso1M8kUHDv26bKsU/edit#gid=0) that was pre-filled with the existing manifest for the non-file-based `Patient` component in the example data model. I was able to get a [similar result](https://docs.google.com/spreadsheets/d/1RxJa78Bzb2VQapiuwMtgWMuBBafugysWLduoQEdBQys/edit) for the non-file-based `Biospecimen` component in the imCORE data model. 

Could the `NaN`s be replaced with empty strings to avoid causing validation errors in Google Sheets? 

With regards to the second feature in this PR, I created a second file in the `Biospecimen` [dataset folder](https://www.synapse.org/#!Synapse:syn26010235) (`sample_b.fastq.gz`) and generated a new [manifest template](https://docs.google.com/spreadsheets/d/1a6cqUZg2dHyfdPem86HkVSAgrxGsWwVaiPdaTwBQi_8/edit). Unfortunately, the new files wasn't appended to the template. Is this expected?

milen-sage:
Good point on the blanks vs NaN. Committed. 

If files are added in a dataset folder for a non-file based type (i.e. a type that does not depend on the Filename attribute, like Biospecimen) the current expected behavior is that these files are *not* added to the manifest.

Another expected behavior in Synapse is that there is delay between the addition of files and their indexing; so sometimes it would take some time before these files can be added to the manifest.

milen-sage:
So the current expected behavior is that files should be added to a manifest only if the data type is 'file-based' and ensuring the file is indexed in Synapse (e.g. allowing for some delay).

BrunoGrandePhD:
Sorry, I just realized that I misspoke in my last paragraph. Here's the corrected version: 

> With regards to the second feature in this PR, I created a second file in the file-based `BulkRNAseqLevel1` dataset folder (sample_b.fastq.gz) and generated a new manifest template. Unfortunately, the new files wasn't appended to the template. Is this expected?

milen-sage:
@BrunoGrandePhD yes, we can add the eTag. But capturing our slack discussion we can do that in a different PR. Could you approve this PR, if it looks alright otherwise?

BrunoGrandePhD (approved):
🚀

milen-sage:
@rrchai changes have been merged into develop branch of schematic.

PR #525, opened 2021-09-27
Remove 'special' characters from column name

milen-sage:
Data manifests may contain columns out of schema. Remove 'special' characters from these column names. For now focus on parenthesis, spaces, etc..

This should address Sage-Bionetworks/data_curator#192


milen-sage:
@rrchai could you take the code from this PR and see if that resolves the submission problem for manifests with columns containing parenthesis (e.g. Sage-Bionetworks/data_curator#192)?

rrchai (approved):
Looks good to me. The manifest with parenthesis in a column name can be submitted to the synapse, and the column name is not altered.

PR #550, opened 2021-12-15
Change validation rules to better work with the data curator app

mialy-defelice:
I updated the validation rule error messaging so it would be more directly usable by the data curator app.

Changes Error output from: 
`['3', 'HTAN Data File ID', 'For the attribute HTAN Data File ID, on row 3, the string is not properly formatted. It should follow the following re.search pattern "HTA_*".', 'Type Error']`

To:
`['3', 'HTAN Data File ID', 'For the attribute HTAN Data File ID, on row 3, the string is not properly formatted. It should follow the following re.search pattern "HTA_*".', 'Hello dear'] `

Gets rid of the Error type specification, and replaces with the invalid value, so the incorrect value can be highlighted within the DCA.

Details behind the request for the change can be found [here](https://github.com/Sage-Bionetworks/data_curator/pull/242)

### DIFF
--- a/schematic/models/validate_attribute.py
+++ b/schematic/models/validate_attribute.py
@@ -19,7 +19,8 @@ logger = logging.getLogger(__name__)
 
 class GenerateError:
     def generate_list_error(
-        list_string: str, row_num: str, attribute_name: str, list_error: str
+        list_string: str, row_num: str, attribute_name: str, list_error: str,
+        invalid_entry:str,
     ) -> List[str]:
         """
             Purpose:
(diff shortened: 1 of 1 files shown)


mialy-defelice:
@rrchai Just to alert you that I have a fix in the works :)

rrchai:
Thank you @mialy-defelice !

I just tested it on DCA. The validation result shows as expected:
[
    "3",
    "HTAN Data File ID",
    "For the attribute HTAN Data File ID, on row 3, the string is not properly formatted. It should follow the following re.search pattern \"HTA_*\".",
    "Hello dear"
]

milen-sage (approved):
Looks good @mialy-defelice! At refactor time, can go over some of the common variables across the various methods in the class and pull them out as class attributes (e.g. self.val_rule, self.error_val, self.error_message). This could simplify method signatures and  make the code more maintainable and extensible.  

Feel free to merge!

mialy-defelice:
@milen-sage Yes! I think we had discussed refactoring this in whole set of functions in the future so it could be expanded more easily :)

PR #554, opened 2022-01-19
Recast column types.

mialy-defelice:
When updating tables on synapse the function `update_df` is called in `df_utils.py`. This function invokes a native pandas function called `df.update`. There is a known issue with pandas where sometimes upon calling `df.update` the column types can sometimes be changed, for example, np.int64 to np.float64. This can cause real issues when uploading to Synapse when the column type is expecting a certain type.

This PR addresses this issue by recasting the column type based on the input table.

### DIFF
--- a/schematic/utils/df_utils.py
+++ b/schematic/utils/df_utils.py
@@ -77,6 +77,10 @@ def update_df(
     input_df_idx.reset_index(inplace=True)
     input_df_idx = input_df_idx[input_df.columns]
 
+     # Sometimes pandas update can change the column datatype, recast
+    for col in input_df_idx.columns:
+        input_df_idx[col] = input_df_idx[col].astype(input_df.dtypes[col])
+
     return input_df_idx
 
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -267,7 +267,7 @@ class TestDfUtils:
         )
         expected_df = pd.DataFrame(
             {
-                "numCol": [1, float(4)],
+                "numCol": [int(1), int(4)],
                 "entityId": ["syn01", "syn02"],
                 "strCol": ["___", "bar"],
             },


schematic/utils/df_utils.py
```diff
@@ -77,6 +77,10 @@ def update_df(
     input_df_idx.reset_index(inplace=True)
     input_df_idx = input_df_idx[input_df.columns]
 
+     # Sometimes pandas update can change the column datatype, recast
+    for col in input_df_idx.columns:
+        input_df_idx[col] = input_df_idx[col].astype(input_df.dtypes[col])
```
milen-sage, line 82: @mialy-defelice just confirming I understand correctly, the dtypes gets the numpy type of data in the given column and astype ensures pandas sets up the type accordingly in the dataframe?

mialy-defelice: @milen-sage It was converting int to float. So yes at the end of the transformations, I just use as type to reset the type to the column types in the input table.

mialy-defelice: I'll go through later and see why all those tests are failing.

milen-sage: got it

milen-sage (approved):

milen-sage (changes requested):
Just a quick change - could you take a look at the [test](https://github.com/Sage-Bionetworks/schematic/runs/4859491698?check_suite_focus=true#step:8:94) for this function it looks like it might be straightforward to update it to accommodate the type cast?

mialy-defelice:
Hey @milen-sage ,
Just to make sure I'm updating the test correct, you are fine with me updating the test to match what we would now expect? It seems like the test is expecting a type change.

milen-sage:
@mialy-defelice yes let's try making the test expect what the updated code outputs.

mialy-defelice:
@milen-sage  I updated the test so now its passing all its checks.

milen-sage (approved):

milen-sage:
@mialy-defelice feel free to merge!

PR #570, opened 2022-02-07
Explicitly defined file encoding as utf-8

GiaJordan:
Defined encoding to use when opening and writing files as 'utf-8' so that the windows default cp-1252.


milen-sage (approved):

milen-sage:
🚀

PR #585, opened 2022-02-25
Generate multiple manifests from API

mialy-defelice:
Similar to PR #584 which allows users to generate multiple manifests through the CLI, by adding multiple `data_type` entries or adding one entry of 'all manifests' in the config file, this PR allows users of the API to make multiple or all manifests in an analogous fashion. 

Within the Swagger UI, users now can see multiple entry fields for `data_type`. Can add or remove additional fields.
Further though `dataset_id` is required, users can now enter 'None' it they do not have a `dataset_id`.

For `title` enter a full title if only making a single manifest and a title prefix only if making multiple manifests.

**Important Note**: 
Output will now be a `list` of URL strings instead of a `str`

Screen Shot 2022-02-25 at 10 01 44 AM

### DIFF
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -32,11 +32,18 @@ paths:
           required: false
         - in: query
           name: data_type
+          style: form
           schema:
-            type: string
+            type: array
+            items: 
+              type: string
             nullable: true
-          description: Data Model Component
-          example: Patient
+          description: >
+            Data Model Component(s). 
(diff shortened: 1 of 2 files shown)


mialy-defelice:
@afwillia @linglp this PR will make changes that may impact your work.

milen-sage (approved):
Cool! You can open a separate issue and PR for the dataset IDs.

PR #592, opened 2022-03-01
Develop contribution documentation

linglp:
Add information related to poetry in our contribution guide.

### DIFF
--- a/tests/data/test_config.yml
+++ b/tests/data/test_config.yml
@@ -1,6 +1,7 @@
 definitions:
   creds_path: "../../credentials.json"
   token_pickle: "token.pickle"
+  synapse_config: "../../.synapseConfig"
 
 synapse:
   master_fileview: "syn23643253"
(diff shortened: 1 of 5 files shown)


mialy-defelice:
Change `pip install` to `poetry add`

linglp:
if you are using token-based authentication, you would need to change the test_config.yml. Tests are failing because this is from a forked branch.

tests/data/test_config.yml
```diff
@@ -1,6 +1,7 @@
 definitions:
   creds_path: "../../credentials.json"
   token_pickle: "token.pickle"
+  synapse_config: "../../.synapseConfig"
```
milen-sage, line 4: @linglp could you add a comment here that this is needed only if you have used the Synapse token authentication approach?

milen-sage:
@linglp #604 handles some of this documentation but not other? Would it make sense to first merge 604 and then open a new PR with any changes here that are outside of #604?

linglp:
@milen-sage most of the changes in contribution.md have already been included in other PR. But we will still need to the update the documentation for API and add comments for test_config.yml. I will open a separate PR for these and close this one for now.

Issue #594, opened 2022-03-03
CLI manifest generation error


milen-sage:
**Describe the bug**
When executing the command schematic manifest -c config-htan.yml get -s -t <title> -dt <data_type>

schematic fails with error stack

Starting schematic...
The (model > input > location) argument with value 'tests/data/HTAN.model.jsonld' is being read from the config file.
The '--jsonld' argument is being taken from configuration file (model > input > location), i.e., 'tests/data/HTAN.model.jsonld'.
Traceback (most recent call last):
  File "/Users/milen-sage/workspace/schematic-dev/schematic/venv-dev/bin/schematic", line 11, in <module>
    sys.exit(main())
  File "/Users/milen-sage/workspace/schematic-dev/schematic/venv-dev/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/milen-sage/workspace/schematic-dev/schematic/venv-dev/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/milen-sage/workspace/schematic-dev/schematic/venv-dev/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/milen-sage/workspace/schematic-dev/schematic/venv-dev/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/milen-sage/workspace/schematic-dev/schematic/venv-dev/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/milen-sage/workspace/schematic-dev/schematic/venv-dev/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/milen-sage/workspace/schematic-dev/schematic/venv-dev/lib/python3.7/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/Users/milen-sage/workspace/schematic-dev/schematic/venv-dev/lib/python3.7/site-packages/schematic/manifest/commands.py", line 171, in get_manifest
    result = create_single_manifest(data_type = dt)
[...]

PR #595, opened 2022-03-03
Change incoming data_type from CLI to list so it is parsed correctly

mialy-defelice:
In response to bug report issued in #594. This fix will take data_type inputs from the command line and convert them to lists so they mimic the config file.


milen-sage (approved):

milen-sage:
Thanks @mialy-defelice! Looks ready to merge.

Issue #590, opened 2022-03-01
Test and enhance schematic REST API submission end point


milen-sage:
Currently the REST API service uses the credentials provided using schematic's installation. However, different service users may have different credentials. Add a parameters to the submission endpoint that allows authentication information to be passed to it.

milen-sage:
@linglp you can chat with @afwillia on that. He may have some useful examples for how to pass authentication information.

afwillia:
@linglp here is a simple example of passing authentication info in a synapse REST API request

`### Get user profile info
# For information on authenticating synapse REST API queries
# https://docs.synapse.org/rest/#org.sagebionetworks.auth.controller.AuthenticationController
# Authentication to Synapse services requires an access token passed in the
# HTTP Authorization header, as per the HTTP bearer authorization standard. 
# https://datatracker.ietf.org/doc/html/rfc6750#section-2.1
#
# https://docs.synapse.org/rest/#org.sagebionetworks.repo.web.controller.UserProfileController
# PAT is the synapse personal access token OR the browser cookie from a logged-in session.
uri <- "https://repo-prod.prod.sagebase.org/repo/v1/userProfile"
req <- GET(uri, add_headers(Authorization=paste0("Bearer ", pat)))
content(req)`

afwillia:
@linglp Functions that data curator uses from [schematic store](https://github.com/Sage-Bionetworks/schematic/blob/f16a10b01c81a07afb6a1e5f3f3f230ad4b014a6/schematic/store/synapse.py#L40)

- [getStorageProjects](https://github.com/Sage-Bionetworks/data_curator/blob/ce9b754f02eb8ec38e81d5e166d2f36721774e15/server.R#L83)
- [getStorageDatasetsInProject](https://github.com/Sage-Bionetworks/data_curator/blob/ce9b754f02eb8ec38e81d5e166d2f36721774e15/server.R#L158)
- [getFilesInStorageDataset](https://github.com/Sage-Bionetworks/data_curator/blob/ce9b754f02eb8ec38e81d5e166d2f36721774e15/server.R#L194)
- [associateMetadataWithFiles](https://github.com/Sage-Bionetworks/data_curator/blob/ce9b754f02eb8ec38e81d5e166d2f36721774e15/server.R#L400)

ychae:
Acceptance Criteria: submission of manifest work through the REST API service
Passing authentication: input field - get the token string and pass that through

afwillia:
[...]

PR #597, opened 2022-03-03
modify the manifest submit endpoint to handle token authentication

linglp:
This PR is related to #590 

Goal: allow users to put in token to submit manifest. Based on Milen's feedbacks, the "manifest/submit" end point will now return the synapse id of the submitted manifest. 

Future items:
1) Figure out a way to handle errors better when the token is not correct. Currently, if the token doesn't work, the system will return "No access to resources. Please make sure that your token is correct". Based on Milen's comments, we might also want to encrypt the token in the future.

### DIFF
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -167,6 +167,13 @@ paths:
             nullable: true
           description: Dataset SynID
           required: true
+        - in: query
+          name: input_token
+          schema:
+            type: string
+            nullable: true
+          description: Token
+          required: true
       operationId: api.routes.submit_manifest_route
       responses:
         "200":
--- a/api/routes.py
+++ b/api/routes.py
@@ -147,12 +147,19 @@ def submit_manifest_route(schema_url):
 
     data_type = connexion.request.args["data_type"]
 
+    input_token = connexion.request.args["input_token"]
+
+    if input_token == 'None':
+        input_token = None
+
+
     metadata_model = MetadataModel(
         inputMModelLocation=jsonld, inputMModelLocationType="local"
     )
-
-    success = metadata_model.submit_metadata_manifest(
-        manifest_path=temp_path, dataset_id=dataset_id, validate_component=data_type,
+    
+    # return id of the manifest 
+    synapse_id = metadata_model.submit_metadata_manifest(
+        manifest_path=temp_path, dataset_id=dataset_id, validate_component=data_type, input_token=input_token
     )
 
-    return success
+    return synapse_id
--- a/schematic/models/metadata.py
+++ b/schematic/models/metadata.py
@@ -280,6 +280,7 @@ class MetadataModel(object):
         validate_component: str = None,
         use_schema_label: bool = True,
         hide_blanks: bool = False,
+        input_token: str = None
     ) -> bool:
         """Wrap methods that are responsible for validation of manifests for a given component, and association of the
         same manifest file with a specified dataset.
@@ -289,6 +290,7 @@ class MetadataModel(object):
             validate_component: Component from the schema.org schema based on which the manifest template has been generated.
         Returns:
             True: If both validation and association were successful.
+            Synapse ID: If no validation is needed.
         Exceptions:
             ValueError: When validate_component is provided, but it cannot be found in the schema.
             ValidationError: If validation against data model was not successful.
@@ -297,7 +299,8 @@ class MetadataModel(object):
         #TODO: avoid explicitly exposing Synapse store functionality
         # just instantiate a Store class and let it decide at runtime/config
         # the store type
-        syn_store = SynapseStorage()
+        syn_store = SynapseStorage(input_token=input_token)
+        result=None
 
         # check if user wants to perform validation or not
         if validate_component is not None:
@@ -328,7 +331,7 @@ class MetadataModel(object):
                 )
 
                 logger.info(f"No validation errors occured during validation.")
-                return True
+                result = True
             else:
                 raise ValidationError(
                     "Manifest could not be validated under provided data model. "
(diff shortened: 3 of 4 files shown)


schematic/store/synapse.py
```diff
@@ -108,6 +110,20 @@ def login(token=None, access_token=None):
 
         return syn
 
+    @staticmethod
```
milen-sage, line 113: consolidate with login() function

api/openapi/api.yaml
```diff
@@ -167,6 +167,13 @@ paths:
             nullable: true
           description: Dataset SynID
           required: true
+        - in: query
+          name: input_token
+          schema:
+            type: string
+            nullable: true
```
milen-sage, line 174: set to false

api/routes.py
```diff
@@ -147,12 +147,19 @@ def submit_manifest_route(schema_url):
 
     data_type = connexion.request.args["data_type"]
 
+    input_token = connexion.request.args["input_token"]
+
+    if input_token == 'None':
+        input_token = None
```
milen-sage, line 153: Try with input_token nullable set to true but w/o this section. If it doesn't work we can note that. 

In any case, after this test  we want to set [nullable to false](https://github.com/Sage-Bionetworks/schematic/pull/597/files#r823104216) in this case

schematic/models/metadata.py
```diff
@@ -328,15 +331,15 @@ def submit_metadata_manifest(
                 )
 
                 logger.info(f"No validation errors occured during validation.")
-                return True
+                result = True
```
milen-sage, line 334: @linglp since you are no longer returning here, python will execute the remainder of the function. It looks like this would  execute  associateMetadataWithFiles twice in some cases? We'd need to avoid that. If the function returns True here, and having the remainder of the code as you have it below, could work?

milen-sage (changes requested):
Only one change request ,and I think we're getting close!

schematic/models/metadata.py
linglp: @milen-sage thanks for pointing that out. I will have the function return a synapse ID here so that associateMetadataWithFiles won't get executed two times. The submit route requires validation of the manifest. If this part returns True (and we don't execute the remaining parts), we can't get synapse ID. I also changed the variable name from "result" to "syn_id" to avoid confusion.

schematic/models/metadata.py
```diff
@@ -280,6 +280,7 @@ def submit_metadata_manifest(
         validate_component: str = None,
         use_schema_label: bool = True,
         hide_blanks: bool = False,
+        input_token: str = None
     ) -> bool:
```
milen-sage, line 284: Change the return type after double checking other functions are not calling this function and expecting a bool return (probably not).

milen-sage (changes requested):
Some comments in code

schematic/models/commands.py
```diff
@@ -91,15 +91,15 @@ def submit_manifest(
     )
 
     try:
-        success = metadata_model.submit_metadata_manifest(
+        syn_id = metadata_model.submit_metadata_manifest(
```
milen-sage, line 94: Let's use manifest_id instead of syn_id

schematic/models/metadata.py
```diff
@@ -323,20 +326,20 @@ def submit_metadata_manifest(
             if not val_errors:
 
                 # upload manifest file from `manifest_path` path to entity with Syn ID `dataset_id`
-                syn_store.associateMetadataWithFiles(
+                syn_id = syn_store.associateMetadataWithFiles(
```
milen-sage, line 329: Let's use manifest_id instead of syn_id. Similarly to other modules outside of store/synapse.py let's try to minimize references to Synapse.

schematic/store/synapse.py
```diff
@@ -85,9 +85,9 @@ def __init__(
             raise MissingConfigValueError(("synapse", "master_fileview"))
 
     @staticmethod
-    def login(token=None, access_token=None):
+    def login(token=None, access_token=None, input_token=None):
```
milen-sage, line 88: @linglp we can proceed with these arguments and logic for now here. However, could you check with Anthony and Rong if input_token coincides with either access_token or token parameter. I should have flagged that earlier, but just noticed we may be repeating logic.

milen-sage (changes requested):

schematic/store/synapse.py
linglp: Thanks Milen! I have double checked with Anthony and the input_token works!

api/openapi/api.yaml
```diff
@@ -167,15 +167,22 @@ paths:
...
+          schema:
+            type: string
+            nullable: false
+          description: Token
+          required: true
       operationId: api.routes.submit_manifest_route
       responses:
         "200":
-          description: Manifest submitted
+          description: Manifest submitted. Synapse ID returned.
```
milen-sage, line 180: Instead of Synapse ID, could we use "Manifest ID (e.g. Synapse ID if your asset management platform is Synapse)".

schematic/models/metadata.py
```diff
@@ -323,20 +326,20 @@ def submit_metadata_manifest(
...
+                return manifest_id
             else:
                 raise ValidationError(
                     "Manifest could not be validated under provided data model. "
                     f"Validation failed with the following errors: {val_errors}"
                 )
 
         # no need to perform validation, just submit/associate the metadata manifest file
-        syn_store.associateMetadataWithFiles(
+        syn_id = syn_store.associateMetadataWithFiles(
```
milen-sage, line 342: for consistency, rename syn_id to manifest_id

milen-sage (approved):
Super close: two very minor changes, but the PR is marked as approved.

Issue #609, opened 2022-03-15
API: Add a config file parameter across all endpoints


milen-sage:
Currently the API service assumes that schematic has been installed with a specific config file. 

Add logic to accept a config file across all endpoints. Use the provided config files instead of the default schematic installation config file.

Some context: the schematic config file contains a couple of constants like synapse fileview ID, name of manifest, etc..  If the config file is fixed in the service, these config constants essentially fix the project for which the service can be deployed. We'd like the service to be flexible and operate across projects.

There are other options to make the service more flexible, when we design the refactor of schematic over the next few months, among them:
- expose the config constants as method arguments across schematic
- create a singleton (or similar) config object and have methods in schematic use that 

One of the above, or a different option, may be better longer term, but would require changing some methods' signatures. We can consider that when refactoring schematic.

milen-sage:
Acceptance criteria:have determined feasibility of using config file passed as a parameter in the REST API endpoints. If feasible has implemented the logic enabling the REST API to utilize this config file.

Issue #603, opened 2022-03-11
Make API endpoints for functions in schematic/store/synapse.py


linglp:
**Is your feature request related to a problem? Please describe.**
Turn the following functions that the data curator uses from 

**Describe the solution you'd like**
Functions that data curator uses from [schematic/store/synapse.py](https://github.com/Sage-Bionetworks/schematic/blob/f16a10b01c81a07afb6a1e5f3f3f230ad4b014a6/schematic/store/synapse.py#L40)

From R: 
- [getStorageProjects](https://github.com/Sage-Bionetworks/data_curator/blob/ce9b754f02eb8ec38e81d5e166d2f36721774e15/server.R#L83)
- [getStorageDatasetsInProject](https://github.com/Sage-Bionetworks/data_curator/blob/ce9b754f02eb8ec38e81d5e166d2f36721774e15/server.R#L158)
- [getFilesInStorageDataset](https://github.com/Sage-Bionetworks/data_curator/blob/ce9b754f02eb8ec38e81d5e166d2f36721774e15/server.R#L194)
- [associateMetadataWithFiles](https://github.com/Sage-Bionetworks/data_curator/blob/ce9b754f02eb8ec38e81d5e166d2f36721774e15/server.R#L400)

Corresponding functions in Python:  
All functions below can be found [here](https://github.com/Sage-Bionetworks/schematic/blob/f16a10b01c81a07afb6a1e5f3f3f230ad4b014a6/schematic/store/synapse.py#:~:text=def%20associateMetadataWithFiles)

- [x]  getStorageProjects
Description: Gets all storage projects the current user has access to, within the scope of the 'storageFileview' attribute.
1) endpoint format: GET /storage/projects
2) Parameters: input_token, synapse ID of master file view, Synapse manifest filename name.
3) response: A list of tuples? 

- [x] getStorageDatasetsInProject
Description: Gets all datasets in folder under a given storage project that the current user has access to.
1) endpoint format: GET /storage/datasets
2) parameters: project ID (synapse ID of a storage project.)
3) response: a list of tuples?

- [x] getFilesInStorageDataset
Description: Gets all files in a given dataset folder.
1) endpoint format: GET /storage/files
2) parameters: 
            * datasetId: synapse ID of a storage dataset.
            * fileNames: get a list of files with particular names; defaults to None in which case all dataset files are returned (except bookkeeping files, e.g.metadata manifests); if fileNames is not None, all files matching the names in the fileNames list are returned if present.
[...]

PR #619, opened 2022-03-21
Develop API endpoints for functions in SynapseStorage

linglp:
This PR is related to the issue about ConfigHandler [here](https://github.com/Sage-Bionetworks/schematic/issues/609) as well as the issue about exposing functions in Synapse Storage [here](https://github.com/Sage-Bionetworks/schematic/issues/603)


linglp:
@milen-sage I also discovered that we might also want to change the **GET manifest/generate** endpoint because currently it takes in the master_fileview from config.yml and users could not modify it from Swagger UI. If we want user to be able to get an existing manifest.csv, do you think we should also modify this endpoint?

api/openapi/api.yaml
```diff
@@ -179,3 +179,146 @@ paths:
...
+        - in: query
+          name: input_token
+          schema:
+            type: string
+            nullable: false
+          description: Token
+          example: Token
+          required: true
+        - in: query
+          name: syn_master_file_view
```
milen-sage, line 197: rename syn_master_file_view to asset_view

api/openapi/api.yaml
```diff
@@ -179,3 +179,146 @@ paths:
...
+          schema:
+            type: string
+            nullable: false
+          description: synapse ID of master file view.
```
milen-sage, line 201: Change "synapse ID of master file view." to "ID of view listing all project data assets. E.g. for Synapse this would be the Synapse ID of the fileview listing all data assets for a given project". We treat Synapse just as an example of an asset management system.

api/openapi/api.yaml
```diff
@@ -179,3 +179,146 @@ paths:
...
+        - in: query
+          name: syn_master_file_view
+          schema:
+            type: string
+            nullable: false
+          description: synapse ID of master file view. 
+          example: syn23643253
+          required: true
+        - in: query
+          name: syn_manifest_file_name
```
milen-sage, line 205: Change syn_manifest_file_name to manifest_file_name

api/openapi/api.yaml
```diff
@@ -179,3 +179,146 @@ paths:
...
+          schema:
+            type: string
+            nullable: false
+          description: Synapse storage manifest file name.
```
milen-sage, line 209: Change "Synapse storage manifest file name" to "Filename of dataset manifest"

api/openapi/api.yaml
milen-sage: For this endpoint specifically, we get all storage projects the current user has access to; this endpoint doesn't require manifest filename. 

Note: if the manifest  filename is not provided, the default manifest name will be used to instantiate the storage object. For this endpoint that's fine since the manifest filename is not relevant.

api/openapi/api.yaml
```diff
@@ -179,3 +179,146 @@ paths:
...
+        - in: query
+          name: input_token
+          schema:
+            type: string
+            nullable: false
+          description: Token
+          example: Token
+          required: true
+        - in: query
+          name: syn_master_file_view
```
milen-sage, line 234: See comment above on avoiding explicit references to Synapse across all endpoints, when possible.

api/openapi/api.yaml
```diff
@@ -179,3 +179,146 @@ paths:
...
+          schema:
+            type: string
+            nullable: false
+          description: synapse ID of master file view. 
+          example: syn23643253
+          required: true
+        - in: query
+          name: syn_manifest_file_name
```
milen-sage, line 242: Does the function routed in this endpoint require manifest filename?

api/routes.py
```diff
@@ -24,17 +26,38 @@
 #     pass
 
 
-def config_handler():
+def config_handler(syn_master_file_view=None, syn_manifest_file_name=None):
```
milen-sage, line 29: Change to Synapse agnostic, similarly to the REST API endpoints.

schematic/configuration.py
```diff
@@ -77,9 +77,24 @@ def load_config(self, config_path=None):
...
+    #     #config = {'synapse': {'master_fileview': 'syn23643253', 'manifest_folder': 'manifests', 'manifest_filename': 'synapse_storage_manifest.csv', 'token_creds': 'syn23643259', 'service_acct_creds': 'syn25171627'}}
+    #     if syn_master_file_view and syn_master_file_name: 
+    #         config['master_file'] = syn_master_file_view
+    #         config['manifest_filename'] = syn_master_file_name
+    #     elif use_default: 
+    #         self.DATA = config
+    #     self.DATA = config
+    #     return self.DATA
```
milen-sage, line 96: Is this block commented out because it's doing the same as the above code segment?

schematic/store/synapse.py
```diff
@@ -64,17 +67,26 @@ def __init__(
...
+        # if syn_master_file_view and syn_master_file_name: 
+        #     self.storageFileview = syn_master_file_view
+        #     self.manifest = syn_master_file_name
+        # else: 
+        #     self.storageFileview = CONFIG["synapse"]["master_fileview"]
+        #     self.manifest = CONFIG["synapse"]["manifest_filename"]
+        # try:
+        #     self.storageFileviewTable = self.syn.tableQuery(
+        #             "SELECT * FROM " + self.storageFileview
+        #         ).asDataFrame()
```
milen-sage, line 81: May want to remove this block - no need to commit commented out code.

schematic/configuration.py
milen-sage, line 96: Is this block commented out because it's doing the same as the above code segment?

milen-sage (changes requested):
Left some comments in code

api/openapi/api.yaml
linglp: Thanks for the suggestion. I changed the description to: ID of view listing all project data assets. For example, for Synapse this would be the Synapse ID of the fileview listing all data assets for a given project.(i.e. master_fileview in config.yml)

api/openapi/api.yaml
linglp: Okay. Got it! I have deleted this parameter for this endpoint.

api/openapi/api.yaml
linglp: Thanks for the suggestion. I changed it to: Dataset manifest filename. For example, for Synapse this would be the manifest_filename in config.yml. 

Note: I used this parameter for /storage/project/datasets endpoint and /storage/dataset/files endpoint. The first endpoint is for getting all datasets in folder under a given storage project that the current user has access to, and the second one is for getting all the files in a given dataset.

api/openapi/api.yaml
linglp: This is a good point. In SynapseStorage class, I could see: 
`        try:
            self.storageFileview = CONFIG["synapse"]["master_fileview"]
            self.manifest = CONFIG["synapse"]["manifest_filename"]
                    "SELECT * FROM " + self.storageFileview
                ).asDataFrame()
`

That means if we don't add a parameter, our endpoint will load manifest file name from `config.yml`. If that's what we want, I could delete the parameter

schematic/configuration.py
linglp: Yeah. This is no longer relevant. I will delete it.

api/openapi/api.yaml
linglp: Note to myself: as we have discussed in the show and tell meeting, we would like to remove `master_fileview` because they are not relevant to the three endpoints that we are adding here.

linglp:
> @milen-sage I also discovered that we might also want to change the **GET manifest/generate** endpoint because currently it takes in the master_fileview from config.yml and users could not modify it from Swagger UI. If we want user to be able to get an existing manifest.csv, do you think we should also modify this endpoint? I just tested it out by adding master_file_view parameter in Swagger UI, which can successfully get an existing manifest. But the api call took quite some time to run (maybe that's because I used HTAN as an example again)

As we have discussed in show and tell, we would like to have this endpoint functions like the GET manifest function in command line.

milen-sage (approved):

milen-sage (changes requested):
Some minor changes and this PR will be ready to merge.

milen-sage (approved):
@linglp looks good! 🚀

PR #622, opened 2022-03-23
Develop optional entities

mialy-defelice:
This PR allows users to specify through the command line or API the way they want their manifest stored on Synapse.

Manifest can be stored as a 'entity', 'table' or 'both'.

 - 'entity' will store the manifest as a csv and create Synapse files for each row in the manifest. (this is the traditional way of storing manifests that will be depreciated in the future)

 - 'table' will store the manifest as a table on Synapse. 

 - 'both' will do both of the options specified above. 
 
Through CLI the default is 'table'.

The API provides a dropdown that lists 'table' as the default selection.


schematic/store/synapse.py
```diff
@@ -429,6 +436,32 @@ def getProjectManifests(self, projectId: str) -> List[str]:
 
         return manifests
 
+    def upload_project_manifests_to_synapse(self, projectId: str) -> List[str]:
+        """Gets all metadata manifest files across all datasets in a specified project.
```
milen-sage, line 440: @mialy-defelice might want to update the docstring here?

schematic/store/synapse.py
```diff
@@ -491,81 +637,52 @@ def associateMetadataWithFiles(
...
+                existingTableId = manifest_synapse_table_id,
+                table_name = manifest['Component'][0].lower() + '_manifest_table',
+                update_col = 'Uuid',
+                specify_schema = False,
+                )
 
-        manifestSynapseFileId = self.syn.store(manifestSynapseFile).id
-
-        return manifestSynapseFileId
+        return manifest_synapse_file_id
```
milen-sage, line 685: @mialy-defelice thank you for making the code format more consistent!

schematic/store/synapse.py
```diff
@@ -429,6 +436,32 @@ def getProjectManifests(self, projectId: str) -> List[str]:
 
         return manifests
 
+    def upload_project_manifests_to_synapse(self, projectId: str) -> List[str]:
```
milen-sage, line 439: Sorry if we covered this during the code review, where was this function called again?

mialy-defelice: This function is just one I wrote when I was scoping the project. I kept it in bc it could be useful for people to use, but isn't used in this PR. I can remove it.

milen-sage: Ah, I remember now we talked about it! Let's keep it in, but perhaps update the docstring to reflect the functionality? Currently it emphasizes a "get" feature 

> Gets all metadata manifest files across all datasets in a specified project

mialy-defelice: Will do!

mialy-defelice:
@milen-sage The doc string has been updated :)

schematic/store/synapse.py
```diff
@@ -441,11 +474,112 @@ def get_synapse_table(self, synapse_id: str) -> Tuple[pd.DataFrame, CsvFileTable
...
         """Associate metadata with files in a storage dataset already on Synapse.
         Upload metadataManifest in the storage dataset folder on Synapse as well. Return synapseId of the uploaded manifest file.
+        
+        If this is a new manifest there could be no Synapse entities associated with the rows of this manifest
+        this may be due to data type (e.g. clinical data) being tabular
+        and not requiring files; to utilize uniform interfaces downstream
+        (i.e. fileviews), a Synapse entity (a folder) is created for each row
+        and an entity column is added to the manifest containing the resulting
+        entity IDs; a table is also created at present as an additional interface
+        for downstream query and interaction with the data.
```
milen-sage, line 582: This now applies only if they have indicated manifest_record_type is 'entity'?

milen-sage (approved):
This looks great! 🚀

schematic/store/synapse.py
mialy-defelice: How does this update sound?
``` 
    If this is a new manifest there could be no Synapse entities associated with the rows of this manifest
    this may be due to data type (e.g. clinical data) being tabular
    and not requiring files; If manifest_record_type is 'entity' or 'both', to utilize uniform interfaces downstream
    (i.e. fileviews), a Synapse entity (a folder) is created for each row
    and an entity column is added to the manifest containing the resulting
    entity IDs; a table is also created at present as an additional interface
    for downstream query and interaction with the data. Otherwise, if
    manifest record type is 'table' a the manifest will be stored as a table, that
    can be joined with fileviews through materialized views. The 'entityId' column will
    contain the synapseId of the generated table in this case.```

Comment (mialy-defelice) [reply schematic/store/synapse.py:589]: How does this update sound?
``` 
    If this is a new manifest there could be no Synapse entities associated with the rows of this manifest
    this may be due to data type (e.g. clinical data) being tabular
    and not requiring files; If manifest_record_type is 'entity' or 'both', to utilize uniform interfaces downstream
    (i.e. fileviews), a Synapse entity (a folder) is created for each row
    and an entity column is added to the manifest containing the resulting
    entity IDs; a table is also created at present as an additional interface
    for downstream query and interaction with the data. Otherwise, if
    manifest record type is 'table' a the manifest will be stored as a table, that
    can be joined with fileviews through materialized views. The 'entityId' column will
    contain the synapseId of the generated table in this case.```

milen-sage:
@mialy-defelice the PR looks ready to merge, but it looks like there may be some conflicts.

PR #647, opened 2022-04-05
Manifest naming convention

GiaJordan:


milen-sage (approved):

milen-sage:
@allaway @rrchai I don't think there's anything in the DCA relying on the manifest naming convention, but just heads up that we are making slight modifications in this PR: 
- dropping the ".csv" extension
- changing the config parameter for manifest filename to basename
- schematic works with manifests based on basename (i.e. the manifest name is no longer fixed)

This allows more flexibility when working with manifests (e.g. we can have multiple manifests per dataset folder).

@ychae

allaway:
Thanks for the heads up!

rrchai:
@milen-sage @allaway  it looks like we do need to change the config file on DCA as well. Otherwise, the `SynapseStorage()` will raise below error and app will freeze at "It looks like you don't have permission of fileview ...":
```
File "/home/rchai/ShinyApps/dc-test/.venv/lib/python3.8/site-packages/schematic/store/synapse.py", line 89, in __init__
    raise AttributeError("storageFileview attribute has not been set.")
AttributeError: storageFileview attribute has not been set
```
As soon as I change ` manifest_filename: 'synapse_storage_manifest.csv' to `manifest_basename: 'synapse_storage_manifest'`, the app works like before.

milen-sage:
Thanks for testing @rrchai ! @GiaJordan let's investigate. We can chat and tackle next sprint?

Issue #648, opened 2022-04-05
Validation on attributes with `list` rule causes error


allaway:
NOTE: Transferred from the DCA repository because it is probably a schematic-level issue I am running into. 

**Describe the bug**
When trying to validate a manifest in the data curator app the app crashes and I get this error:

```
2022-04-05T23:09:41.297404+00:00 shinyapps[5727796]: Warning: Error in validationResult: object 'annotation_status' not found
2022-04-05T23:09:41.297516+00:00 shinyapps[5727796]:   13: fn
2022-04-05T23:09:41.297461+00:00 shinyapps[5727796]:   15: <Anonymous>
2022-04-05T23:09:41.297625+00:00 shinyapps[5727796]:    7: connect$retryingStartServer
2022-04-05T23:09:41.297664+00:00 shinyapps[5727796]:    6: eval
2022-04-05T23:09:41.297580+00:00 shinyapps[5727796]:    8: retry
2022-04-05T23:09:41.297712+00:00 shinyapps[5727796]:    5: eval

````

I tried to generate a new manifest in one of our demo projects and validate that and it worked fine, and I don’t see that other users have had this issue in our logs, so I am guessing it’s something to do with the contents of this manifest or the dataset I’m using…

It appears to happen when include annotations for this column in the manifest: https://github.com/nf-osi/nf-metadata-dictionary/blob/main/NF.csv#L406

**To Reproduce**
Alright, so I was able to come up with something reproducible: it happens whenever I try to annotate something that has a list validation rule. We have three attributes that have this rule: modelSystemName,tumorType and cellType. description of what the bug is.

1. Go to NF Data Curator App
2. Select a dataset, e.g. w/ genomics manifest
3. Fill in manifest, annotating modelSystemName,tumorType or cellType. 
4. Try to validate manifest 

**Expected behavior**
Validation should proceed, instead app crashes. 

**Environment**
This is using the app as built here: https://github.com/nf-osi/NF_data_curator/actions/runs/2001994081/workflow, with the develop version of schematic from 19 days ago, as well as the current version of develop through today (https://github.com/nf-osi/NF_data_curator/actions/runs/2099292364)

**Priority** (select one)
- [ ] Minor ⬇️
- [x] Major 📢
- [ ] Critical 🆘

**Additional context**
slack here: https://sagebionetworks.slack.com/archives/C01ANC02U59/p1649197443845809
Might be related to Sage-Bionetworks/data_curator#309 ?

[...]

PR #650, opened 2022-04-06
Fix error in list validation from Issue #648

mialy-defelice:
This PR fixes the bug raised in Issue #648. The error was caused by how the rows were referenced when validating that the entry was a list. 

@allaway I was able to reproduce your error, and this PR should fix the issue you saw.

Note, that when providing a single value when there is a `list` validation applied, you will need to enter it with a comma after or else you will encounter a Schematic validation error saying the entry is not formatted with a comma. For example `icNF97.2a,`.

### DIFF
--- a/schematic/models/validate_attribute.py
+++ b/schematic/models/validate_attribute.py
@@ -196,13 +196,13 @@ class ValidateAttribute(object):
         errors = []
         manifest_col = manifest_col.astype(str)
         # This will capture any if an entry is not formatted properly.
-        for row_num, list_string in enumerate(manifest_col):
+        for i, list_string in enumerate(manifest_col):
             if "," not in list_string and bool(list_string):
                 list_error = "not_comma_delimited"
                 errors.append(
                     GenerateError.generate_list_error(
                         list_string,
-                        row_num=str(row_num + 2),
+                        row_num=str(i + 2),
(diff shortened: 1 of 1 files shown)


mialy-defelice:
@milen-sage @rrchai The latest develop has the changes to `Synapse/associateMetadataWithFiles` which would require users to enter a `manifest_record_type`. As far as I know this has not yet been tested in the DCA. I am not sure the best way to get this change into DCA without pushing the changes to `Synapse/associateMetadataWithFiles`...

GiaJordan (approved):

allaway:
Thanks for the quick fix!

>you will need to enter it with a comma after or else you will encounter a Schematic validation error saying the entry is not formatted with a comma. For example icNF97.2a,.

Hmm, is it possible to automatically append this comma in the validation function? This shouldn't be a problem for internal users on our NF DCC team, but I imagine that external users might run a this situation a lot where they forget to include a trailing comma.

mialy-defelice:
> Hmm, is it possible to automatically append this comma in the validation function? This shouldn't be a problem for internal users on our NF DCC team, but I imagine that external users might run a this situation a lot where they forget to include a trailing comma.

Yea we could create an issue for this feature!

Issue #569, opened 2022-02-04
schematic manifest get -o does not return a csv


afwillia:
**Describe the bug**
Passing the --output_csv option to schematic manifest get does not return a csv

**To Reproduce**
Set up schematic according to the [README ](https://github.com/Sage-Bionetworks/schematic#manifest)
schematic manifest --config config.yml get --output_csv test.csv

**Expected behavior**
A file to be written with the name of the --output_csv argument.

ychae:
AC: return a csv when using the output csv option

PR #652, opened 2022-04-07
add changes to fix bug related to manifest get in CLI

linglp:
This is related to the bug[ here.](https://github.com/Sage-Bionetworks/schematic/issues/569) This happened because `create_single_manifest` function in `command.py` did not include a flag for output_csv. Also, if we do not put in a dataset_id,  the `get_manifest` function in `generator.py` would trigger  `self.get_empty_manifest` which would then return an empty google sheet URL. 

My initial thought: 
1) Added a sheet_url parameter in `get_empty_manifest` function. To avoid affecting other downstream functions, I set the default to `True`. If this parameter is set to `False`, we would return a pandas data frame. 
2) Add `output_csv` as a parameter for `create_single_manifest` function 

New thought: 
1) Add `output_csv` as a parameter for `create_single_manifest` function 
2) Keep `get_manifest` function (including `get_empty_manifest`) the same 
3) Create two separate function (`export_manifest_csv` and `export_manifest_excel`). These two functions will have `manifest_url` as a parameter and export the result to CSV/Excel


schematic/manifest/commands.py
```diff
@@ -121,7 +126,31 @@ def get_manifest(
         allow_none=True,
     )
 
-    def create_single_manifest(data_type):
+    def export_manifest(file_name, manifest_url, output_excel=None):
```
milen-sage, line 129: do we need this function here given the one in google_api_utils?

schematic/manifest/commands.py
```diff
@@ -6,13 +6,18 @@
 import logging
 import sys
 import pandas as pd
+import requests
```
milen-sage, line 9: we may not need that if we are just using the google_api_utils functions

schematic/manifest/commands.py
linglp: @milen-sage thanks for pointing this out. Where do you think this function should live?  Should it be under `create_single_manifest` or google_api_utils?

linglp:
@milen-sage I have looked into google sheet API. Unfortunately, it does not have an endpoint for export. But the `get` endpoint could get all the metadata of a given spreadsheet, including all the hidden sheets. I have written a separate method under `create_single_manifest` in `commands.py` called `export_manifest_excel` for exporting manifest to Excel using the google sheet API, which still eventually involves using `to_excel` function of pandas

You could check out the official documentation of the `get` endpoint [here](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets.values/get)

schematic/manifest/commands.py
milen-sage: I think google_api_utils is probably better. We can refactor later.

milen-sage:
@linglp you might want to pull the latest version of commands.py to resolve conflicts.

milen-sage:
@linglp you mentioned 

> Added a sheet_url parameter in get_empty_manifest function.

Is this in manifest/generator.py? I don't see the file in the PR.

linglp:
> @linglp you might want to pull the latest version of commands.py to resolve conflicts.

Thanks for the reminder. I have pulled in the latest changes from `develop` and resolved conflicts.

linglp:
> @linglp you mentioned
> 
> > Added a sheet_url parameter in get_empty_manifest function.
> 
> Is this in manifest/generator.py? I don't see the file in the PR.

Let me explain this in more details: 
Initially, we had the following code for `get_manifest` function in `manifest/generator.py`: 
```
        # call get_manifest() on manifest_generator
        result = manifest_generator.get_manifest(
            dataset_id=dataset_id, sheet_url=sheet_url, json_schema=json_schema,
        )

        if sheet_url:
            logger.info("Find the manifest template using this Google Sheet URL:")
            click.echo(result)

        elif isinstance(result, pd.DataFrame):
            if output_csv is None:
                prefix, _ = os.path.splitext(jsonld)
                prefix_root, prefix_ext = os.path.splitext(prefix)
                if prefix_ext == ".model":
                    prefix = prefix_root
                output_csv = f"{prefix}.{data_type}.manifest.csv"

            logger.info(
                f"Find the manifest template using this CSV file path: {output_csv}"
            )

            result.to_csv(output_csv, index=False)
        return result
```
The code above means that the `output_csv` function would only get triggers if `manifest_generator.get_manifest` returns an instance of data frame. But when dataset id is not provided (i.e. when we are doing `schematic manifest --config config.yml get` using CLI), `get_empty_manifest` function would get triggered in `get_manifest`, and this function would only return a URL. Initially, I was thinking about changing the `get_empty_manifest` function so that it could return an empty data frame when dataset_id is not specified.

But since we are using google sheet api, I think we could keep `get_empty_manifest` and `get_manifest` as it is. If `manifest_generator.get_manifest`  returns a url, we would use the google sheet url functionality to export the url to an excel spreadsheet or a CSV. If `manifest_generator.get_manifest`  returns a data frame, we could still use `to_excel` and `to_csv` like we did before.

linglp:
For records, I have tested it out my PR with the following commands from CLI: 
1) Get an existing manifest (and download it as a CSV by default): `schematic manifest --config config.yml get --dataset_id syn28268700`
2) Get an existing manifest and download it as an Excel spreadsheet: `schematic manifest --config config.yml get --output_xlsx 'test.xlsx' --dataset_id syn28268700`
3) Get an existing manifest and download it as a CSV: `poetry run schematic manifest --config config.yml get --output_csv 'test.csv' --dataset_id syn28268700`
4) Get a manifest (without specifying a dataset_id) and download it as a CSV by using our default `config.yml`: `schematic manifest --config config.yml get --output_csv 'test.csv'`
5) Get a manifest (without specifying a dataset_id) and download it as an Excel spreadsheet by using our default `config.yml`: `schematic manifest --config config.yml get --output_xlsx 'test.xlsx'`

milen-sage (approved):
🚀

linglp:
Github issues: 
- [ ] start a Jupyter Notebook for testing and for demo-ing functionalities (also allow people to test more possibilities) 
- [ ] add more tests to `test_cli`

Issue #620, opened 2022-03-21
Write test cases for API endpoints 


linglp:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
There are two kinds of testing: 1) unit testing. The goal of unit testing is to make sure the code that you add don't break other code. 2) functional testing. 

Look into [behave](https://behave.readthedocs.io/en/latest/fixtures.html) and flask  unit testing documentation. 

**How important is this feature?** Select from the options below:
• 🏝 Low - it's an enhancement but not crucial for work

**When will use cases depending on this become relevant?** Select from the options below:
• Mid-term - 2-4 months

**Additional context**

Acceptance criteria: 
Write 1 test case for 1 endpoint.

linglp:
@ychae : I have added two test cases, but I have not yet had them reviewed.

ychae:
@linglp I think we can close this issue as it's met the acceptance criteria. If any problems arise, please feel free to open another issue.

PR #656, opened 2022-04-11
Create test cases for API endpoints

linglp:
This PR is the beginning of creating test cases for API endpoints. (View it [here](https://github.com/Sage-Bionetworks/schematic/issues/620)) I have created two test cases for the following endpoints: **manifest/generate** and **manifest/populate**


thomasyu888:
Lingling, I'm going to close this due to age.  We can re-open if you think we should

Issue #651, opened 2022-04-07
API: get manifest as a JSON doc in addition to returning a manifest csv path


milen-sage:
Add a parameter to the endpoint returning a manifest csv path, to enable returning the content of the csv in JSON format. Note that the conversion of csv to JSON is a handy utility and having that as a function in utils/ would be useful. There is one more place in schematic where we convert the manifest to JSON: [JSONSchema-based validation ](https://github.com/Sage-Bionetworks/schematic/blob/ceb23df5669174df93b7b32d381c1d74c73d62e9/schematic/models/validate_manifest.py#L171).

Issue #632, opened 2022-03-28
Create an API endpoint to download manifests as CSV 


linglp:
**Is your feature request related to a problem? Please describe.**
The DCA needs an endpoint to download an existing manifest as a csv. Currently it uses Synapse's REST API directly to download a csv manifest file. However, we'd like to decouple the DCA from Synapse.

**Describe the solution you'd like**
Create an endpoint for downloading a manifest that [uses this function](https://github.com/Sage-Bionetworks/schematic/blob/8efa7296a6f3381b12e3640e63d3f6b8f1b6a8c7/schematic/store/synapse.py#L278). Currently, the GET manifest endpoint only returns the existing manifest or create a new empty manifest. We would need a new endpoint for the data curator app. 

**How important is this feature?** Select from the options below:
• 🌗 Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**When will use cases depending on this become relevant?** Select from the options below:
• Mid-term - 2-4 months

**Additional context**
Acceptance criteria: A new endpoint gets created and tested.

linglp:
@milen-sage and @afwillia could you let me know what this endpoint should return? I can have the endpoint download an existing manifest (in CSV format) and return local path of the downloaded manifest. Will that be sufficient? Or do we want to return a JSON with content from the CSV?

milen-sage:
Downloading the existing csv and returning a path should work for the DCA use cases, and we can start there? @afwillia would the content of the manifest itself be useful as JSON in any way for the app currently?

afwillia:
@milen-sage @linglp DCA doesn't do much with the content of the manifest so returning JSON at the moment won't have an effect.

linglp:
Thank you @milen-sage and @afwillia ! I will let this endpoint return a local file path then.

lakikowolfe:
Just popping in with some thoughts on this

Lingling and I spoke briefly today about the existing API endpoints and what will be needed for the Data Flow Coordinator App (AKA Release Administrator). My initial thought is that the content of the manifest would be most useful for this app as JSON rather than returning a path to a downloaded csv. Just wanted to log that here for now! Happy to talk further as the Data Flow Coordinator develops.

milen-sage:
@lakikowolfe that's great to know! We can think about adding an argument to the endpoint to indicate whether to return the JSON representation of the csv file.

PR #662, opened 2022-04-12
modify download/manifest endpoint to make it return a json

linglp:
This PR is related to the issue mentioned [here](https://github.com/Sage-Bionetworks/schematic/issues/651) and [here.](https://github.com/Sage-Bionetworks/schematic/issues/632) Based on my discussion with @lakikowolfe , for the release administrator app, returning a local CSV file path might not be sufficient. We need to add an additional parameter for **download/manifest endpoint** to return the manifest in JSON format.

### DIFF
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -115,6 +115,13 @@ paths:
           description: this dataset_id should be the parent ID of the manifest.
           example: syn28268700
           required: true
+        - in: query
+          name: as_json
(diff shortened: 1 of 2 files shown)


milen-sage (approved):

linglp:
@lakikowolfe  If you have time, please feel free to test this out and let me know if it works!

lakikowolfe:
@linglp will do!

Issue #599, opened 2022-03-07
Download a file from manifest get instead of gsheet url


afwillia:
**Is your feature request related to a problem? Please describe.**
Related to #569. Posting xlsx versions of manifests as templates on Synapse is common. Having `schematic manifest get` return an xlsx file downloaded from the google sheet would eliminate needing to manually access the URL.

**Describe the solution you'd like**
A flag for xlsx, csv, possibly other output in the [google drive api](https://developers.google.com/drive/api/v3/ref-export-formats)
`schematic manifest --config config.yml get --output_csv test.csv` but with an xlsx output type.

**How important is this feature?** Select from the options below:
• 🌗 Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**When will use cases depending on this become relevant?** Select from the options below:
• Mid-term - 2-4 months

ychae:
AC: implement downloading a manifest in xlsx format

PR #663, opened 2022-04-12
Modify get manifest CLI to return manifest in xlsx format

linglp:
This PR is related to the issue mentioned [here](https://github.com/Sage-Bionetworks/schematic/issues/599). 

I have made sure that the following commands have the right behaviors: 
- `schematic manifest --config config.yml get` would download the manifest in CSV format with a standard file name
- `schematic manifest --config config.yml get --output_csv test.csv` would download the manifest in CSV format with the file name `test`
- `schematic manifest --config config.yml get --output_xlsx test.xlsx` would download the manifest in Excel format with the file name `test`
- if we change the data_type to "all_manifests" in `config.yml`, all manifests would be downloaded in CSV format by using  `schematic manifest --config config.yml get`
- `schematic manifest --config config.yml get --help` would return all the correct help message for `output_xlsx`


linglp:
*Notes*: This should work for `schematic manifest --config config.yml get --output_xlsx test.xls` too. Also, the documentation exporting google sheet using API could be found [here](https://developers.google.com/drive/api/v3/reference/files/export).

linglp:
*Notes*: After discussing with Milen, it turned out that we should also be able to export the hidden sheet when we export manifest to excel. Google drive api does not have capacity to export hidden sheet, so we have to use google sheet api. However, google sheet api only has `get` endpoint for getting all metadata of google sheet and does not have an endpoint for export. Check out the official documentation [here](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets.values/get).

Issue #664, opened 2022-04-13
Integers are uploaded to synapse as doubles


allaway:
**Describe the bug**
Uploading a integer-like string, (e.g. `1`) as an annotation to Synapse gets coerced to a double (`1.0`). This causes two issues:
 - it takes a valid value (string `1`) and converts it to an invalid value (double `1.0`) so the resulting data are no longer compliant, these have to be fixed manually after the fact
 - if _not_ fixed after the fact, when a user goes to re-annotate files, the data that they had originally input that were valid have been altered and no longer pass validation. 

**To Reproduce**
1. Go to DC app. 
2. Create a manifest. 
3. Fill in a column with all integer-like values. 
4. Submit to synapse. 
5. See annotations, which should have doubles (See screenshot for example) 

**Expected behavior**
I would expect schematic to not alter the user's input. 

**Priority** (select one)
- [ ] Minor ⬇️
- [ X ] Major 📢
- [ ] Critical 🆘

**Screenshots**
<img width="298" alt="Screen Shot 2022-04-13 at 10 38 30 AM" src="https://user-images.githubusercontent.com/20408008/163238538-eaf44cf4-a0ad-47b9-a945-03e6a8c7585b.png">

**Additional context**
Assume this related to #600 ?
I believe this will affect the HTAN data model too. For example, readPair and readLength in the NF Data Model (ReadIndicator and ReadLength in HTAN) are both converted to doubles.

allaway:
Doing some more digging I'm guessing this is why this happens: https://github.com/Sage-Bionetworks/schematic/blob/4b077c73272fd653c3479d7d1d7e49697b827e39/schematic/utils/df_utils.py#L33

IMO this causes two issues: for strings that are integer-like (e.g. read pair = 1) this makes it a double (1.0), and for strings that could theoretically be doubles (e.g. age = 50) this adds false precision (50.0) when we don't necessarily have that accurate of an age.

allaway:
@GiaJordan commented in #664:

>It should address https://github.com/Sage-Bionetworks/schematic/issues/664 but from what I've gathered it think there's also something happening before schematic when the manifest is created that influences the number value read. The manifests that I've created and uploaded during my testing process for this did not have the issue you described, but this fix was intended for something slightly different so there may be more at play for that specifically.

[...]

Issue #600, opened 2022-03-08
associateMetadataWithFiles alters boolean values


allaway:
**Describe the bug**
When a string like `TRUE` or `FALSE` is added to manifest, `validateModelManifest` treats these values as strings (which is our desired behavior), but then, I assume, `associateMetadataWithFiles` interprets these as Python booleans. If TRUE is in the manifest, the resulting annotation in Synapse is `true` and in the manifest is `True`. Then, if someone retrieves a manifest to update these files, the resulting manifest no longer passes validation.

**To Reproduce**
1. Go to the data curator app.
2. Generate a manifest. 
3. Provide TRUE or FALSE as values in one of the columns.
4. Submit the manifest for validation
5. Submit the manifest to Synapse. 
6. Regenerate the manifest as in step 2. 
7. Validate (will fail). 

**Expected behavior**
The TRUE value should persist in synapse annotations and csv manifest. 

**Priority** (select one)
- [x] Minor ⬇️
- [ ] Major 📢 - ~(any manifests uploaded at the moment will need to have True/False values corrected, and this will be a real pain to track down and correct between annotations and many csv manifests).~
- [ ] Critical 🆘 
edit: downgraded to minor: we are going to switch to yes/no values in our schema. This issue will no longer affect us. 

**Screenshots**
![](https://user-images.githubusercontent.com/20408008/157135933-cf114c9e-48f2-441a-a7ce-8b12270c4a77.png)

**Other comments**
the issue title, "associateMetadataWithFiles alters boolean values" is a bit of a guess.

allaway:
Some slack discussion indicated this might be a synapseclient versioning issue. I tested 2.5.1, 2.5.0, and 2.4.0 (which is the version we have used on the shiny server before switching to shinyapps.io without issues). Unfortunately, this behavior persists across all three of these versions.

allaway:
It seems like the same thing is happening with integers, they get converted to doubles, and then when re-annotating they don't pass validation when the schema expects integer values.

allaway:
@GiaJordan commented in #665:

>I'm not sure about https://github.com/Sage-Bionetworks/schematic/issues/600, so I'll likely need to investigate a fix for that specifically at a later date because it seems like those values have been read as strings for a while and still are here.

I can confirm that this issue persists in the latest version of develop (35fa1eb) and was not resolved by #665. Just wanted to close the loop on that!

allaway:
[...]

PR #665, opened 2022-04-13
Develop csv import

GiaJordan:


schematic/models/metadata.py
```diff
@@ -22,7 +22,7 @@
 # we shouldn't need to expose Synapse functionality explicitly
 from schematic.store.synapse import SynapseStorage
 
-from schematic.utils.df_utils import trim_commas_df
+from schematic.utils.df_utils import trim_commas_df, load_df
```
milen-sage, line 25: Do we still need trim_commas_df?

milen-sage (approved):
@GiaJordan this looks great! Just one minor  question in code

schematic/models/metadata.py
GiaJordan: I think we'll be able to do without it now!

GiaJordan:
@allaway As of the merge of this PR the CSV import for schematic is standardized. To make the errors generated from the validation suite more meaningful, changes were made to the datatypes of columns and cells of imported manifest CSVs. 

Previously, each cell was imported as string type but now numerical cells will be the appropriate type: either an `int` or `float` type instead of `string`. All non-numerical text will still be type `string`. I know that previously you updated the [csv read function](https://github.com/Sage-Bionetworks/schematic/commit/e7ce4ba58f0a38af4a02fa65e13f21a4dd75ea07) to import everything as type string, so I wanted to make you aware of this change.

Can you let @milen-sage and I know how this new import method works with your workflow, and if you're able to adapt your workflow to work with the new standard should there be disparities?

allaway:
Hi @GiaJordan!

The main reason I had implemented that change back in #539 was because `schematic` didn't have any more sophisticated logic at the time to ensure that the data was preserved. I'm not too worried about preserving the type as string - it was just a lazy fix to ensure that the data itself wasn't changed to something different (e.g. TRUE->True, 1->1.0) etc. 

It sounds like this PR will fix both my issues #664 and #600. I can pull these changes into our staging DCA and close them if so!

GiaJordan:
It should address #664 but from what I've gathered it think there's also something happening before schematic when the manifest is created that influences the number value read. The manifests that I've created and uploaded during my testing process for this did not have the issue you described, but this fix was intended for something slightly different so there may be more at play for that specifically.

I'm not sure about #600, so I'll likely need to investigate a fix for that specifically at a later date because it seems like those values have been read as strings for a while and still are here.

We'll be merging more associated changes within the next day with PR #682 that includes some more logic for data model CSVs and manifests when they're not being read for validation purposes.
If you'd like to test it though you can pull this branch now! Otherwise I can put those issues on my backlog as appropriate.
@allaway

allaway:
Hey @GiaJordan - thanks for the comments, that's really helpful.
I updated the DCA including the #682 PR. I am running into an issue where integers with *defined* values are being flagged as invalid (probably because the defined values in the schema are treated as strings, but `schematic`is reading in the CSV and interpreting these as integers). Do you think this is an issue on my end (maybe we need to define this in the schema differently to have schematic treat the schema values as integers?) or is this a bug in how schematic is interpreting the schema values? Unfortunately we can't simply set`validation` to be integers, because there are a limited number of valid values, it's not all integers. 

Screenshot example: 
Screen Shot 2022-04-27 at 7 51 15 PM

allaway:
If this seems like a `schematic` bug and not just something we need to fix in the schema, I'll file this as a proper issue :) let me know what you think!

GiaJordan:
@allaway addressed in #686!
**cc** @milen-sage

PR #666, opened 2022-04-13
Add annotations to manifest tables and CSVs

mialy-defelice:
This PR adds annotations to uploaded manifest tables and csv files.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -608,7 +608,9 @@ class SynapseStorage(BaseStorage):
 
             # Add component to metadata
             metadata['Component'] = component[0]
+        
         elif is_table:
+            # Get table metadata
             metadata = self.getTableAnnotations(manifest_synapse_id)
         
         # Get annotations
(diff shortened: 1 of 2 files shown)


mialy-defelice:
@GiaJordan

schematic/store/synapse.py
```diff
@@ -577,6 +581,47 @@ def format_annotations(self, se, row, entityId, useSchemaLabel, hideBlanks):
...
+        annos = self.syn.get_annotations(manifest_synapse_id)
+
+        # Add metadata to the annotations
+        for annos_k, annos_v in metadata.items():
+            annos[annos_k] = annos_v
+
+        return annos
+
     def associateMetadataWithFiles(
         self, metadataManifestPath: str, datasetId: str, manifest_record_type: str,
```
milen-sage, line 626: @mialy-defelice could you add a default option for manifest_record_type? We can go with "both" for now. This would avoid problems with functions that [call](https://github.com/Sage-Bionetworks/schematic/blob/a9301999d407b10caa7370478ae96cff32f32472/schematic/store/synapse.py#L393) this function but don't provide this record type parameter.

milen-sage (approved):

milen-sage:
@mialy-defelice looks ready!

Issue #690, opened 2022-04-29
`test_invalid_manifest` is failing


linglp:
**Describe the bug**
`test_invalid_manifest` in `test_validation.py` is failing GitHub actions. More specifically, we could start with this test: 
`

        assert GenerateError.generate_cross_error(
            val_rule = 'matchExactlyOne',
            attribute_name='checkMatchExactly',
            matching_manifests = ['syn27600102', 'syn27648165']
            ) in errors

`

After some digging, I found out the "expected error message" (the error message from `metadataModel.validateModelManifest`) is: `Manifests ['syn29862066', 'syn27648165'] match the values in the source attribute."`, and our output error message is: `Manifests ['syn27600102', 'syn27648165'] match the values in the source attribute."`

1. Go to develop branch
2. Run `poetry run pytest tests/test_validation.py -s`

**Expected behavior**

**Priority** (select one)
- [ ] Minor ⬇️
- [X] Major 📢
- [ ] Critical 🆘

**Screenshots**

**Desktop (if applicable, please complete the following information):**

**Additional context**

PR #691, opened 2022-04-29
change syn id of matching manifest to allow test to past

linglp:
This PR is related to the bug mentioned [here](https://github.com/Sage-Bionetworks/schematic/issues/690) Please let me know if the change makes sense.


milen-sage (approved):
Thanks @linglp !

PR #701, opened 2022-05-03
Cross manifest validation project scope option

GiaJordan:
Add option to specify project scope for cross manifest validation

### DIFF
--- a/schematic/models/metadata.py
+++ b/schematic/models/metadata.py
@@ -300,7 +300,7 @@ class MetadataModel(object):
         use_schema_label: bool = True,
         hide_blanks: bool = False,
         input_token: str = None,
-        project_scope = None,
+        project_scope: List = None,
     ) -> string:
         """Wrap methods that are responsible for validation of manifests for a given component, and association of the
         same manifest file with a specified dataset.
--- a/schematic/models/metadata.py
+++ b/schematic/models/metadata.py
@@ -187,7 +187,7 @@ class MetadataModel(object):
 
     # TODO: abstract validation in its own module
     def validateModelManifest(
-        self, manifestPath: str, rootNode: str, restrict_rules: bool = False, jsonSchema: str = None, 
+        self, manifestPath: str, rootNode: str, restrict_rules: bool = False, jsonSchema: str = None, project_scope = None,
     ) -> List[str]:
         """Check if provided annotations manifest dataframe satisfies all model requirements.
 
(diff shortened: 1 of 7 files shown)


schematic/models/metadata.py
```diff
@@ -187,7 +202,7 @@ def get_component_requirements(
 
     # TODO: abstract validation in its own module
     def validateModelManifest(
-        self, manifestPath: str, rootNode: str, restrict_rules: bool = False, jsonSchema: str = None, 
+        self, manifestPath: str, rootNode: str, restrict_rules: bool = False, jsonSchema: str = None, project_scope: str = None,
```
milen-sage, line 205: Let's have project_scope as a list here (for consistency with other calling functions). Perhaps parsing the CLI project_scope arg into a list can be done in the commands.py

milen-sage (changes requested):

schematic/models/metadata.py
```diff
@@ -74,6 +74,21 @@ def __init__(self, inputMModelLocation: str, inputMModelLocationType: str,) -> N
                 f"The type '{inputMModelLocationType}' is currently not supported."
             )
 
+    def parse_project_scope(self,project_scope):
```
milen-sage, line 77: Start a cmd_utils.py and add the parser function there as parse_synid_list() (or a different better name? :)

schematic/models/validate_attribute.py
```diff
@@ -356,7 +356,7 @@ class ValidateAttribute(object):
         - Add string length validator
     """
 
-    def get_target_manifests(target_component):
+    def get_target_manifests(target_component, project_scope: List):
```
milen-sage, line 359: @GiaJordan are we guaranteed we have a project_scope set here? I.e. do we need a default value?

milen-sage (commented):
@GiaJordan could you confirm all the places that need default project_scope have one specified. Otherwise, I think we are good to go.

schematic/models/validate_attribute.py
GiaJordan: As it stands, yes. This is a new method I made that currently is only called from the `cross_validation` method as part of the validation process where `project_scope `has either been set to a list or initialized to `None`. However, if we think that this method has utility outside cross manifest validation and want to move it outside the `ValidateAttribute` object so it can be used elsewhere then I should initialize the scope to `none `again here

milen-sage: Thanks - great! Then we can keep it strictly required for now.

milen-sage (approved):

Issue #703, opened 2022-05-04
not of type \'array\'' affects /model/submission endpoint


linglp:
**Describe the bug**
When I tried to use `model/submit` endpoint for submitting manifest, I came across an error: 
```
jsonschema.exceptions.ValidationError: Manifest could not be validated under provided data model. Validation failed with the following errors: [[2, 'Family History', '"[\'Breast\', \'Colorectal\']" is not of type \'array\'', "['Breast', 'Colorectal']"]]
```

For context, this is how my manifest looks like: 

Patient ID | Sex | Year of Birth | Diagnosis | Component | Cancer Type | Family History
-- | -- | -- | -- | -- | -- | --
123 | Female |   | Healthy | Patient | Breast | Breast,Colorectal

And the submission would fail too if I update the `family history` column to contain only one string. Something like below: 

Patient ID | Sex | Year of Birth | Diagnosis | Component | Cancer Type | Family History
-- | -- | -- | -- | -- | -- | --
123 | Female |   | Healthy | Patient | Breast | Breast

Error message: 
```
jsonschema.exceptions.ValidationError: Manifest could not be validated under provided data model. Validation failed with the following errors: [['2', 'Family History', "For attribute Family History in row 2 it does not appear as if you provided a comma delimited string. Please check your entry ('Breast'') and try again.", 'Breast'], [2, 'Family History', '"[\'Breast\']" is not of type \'array\'', "['Breast']"]]
```

**To Reproduce**
1. Checkout branch: `develop-fix-submit-manifest-endpoint`
2. Run API endpoint `model/submit` by using the manifest that I mentioned above.

**Expected behavior**

**Priority** (select one)
- [ ] Minor ⬇️
- [X] Major 📢
- [ ] Critical 🆘

**Screenshots**

**Desktop (if applicable, please complete the following information):**

**Additional context**
Might be related to this issue: https://github.com/Sage-Bionetworks/schematic/issues/699

GiaJordan:
@milen-sage related to #699

anngvu:
Since NF is experiencing this, I also combed through the issues -- this is also related to https://github.com/Sage-Bionetworks/schematic/issues/653.

linglp:
@GiaJordan  Thanks! I tested it out with "family history" column has the following values and it all worked: 
*empty 
*with only one string
*with two strings

Thank you!

PR #704, opened 2022-05-04
add restrict_rule and manifest_record_type parameter to submit endpoint

linglp:
This is related to a bug reported by @lakikowolfe After updating the ` submit_metadata_manifest` function, we would also need to update the API /model/submit route. But currently, this is also blocked by the bug here: #703

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -160,12 +160,12 @@ def submit_manifest_route(schema_url, manifest_record_type=None):
     input_token = connexion.request.args["input_token"]
 
     if data_type == 'None':
-        manifest_id = metadata_model.submit_metadata_manifest(
-            manifest_path=temp_path, dataset_id=dataset_id, validate_component=None, input_token=input_token, manifest_record_type = manifest_record_type, restrict_rules = restrict_rules
-        )
-    else: 
-        manifest_id = metadata_model.submit_metadata_manifest(
-        manifest_path=temp_path, dataset_id=dataset_id, validate_component=data_type, input_token=input_token, manifest_record_type = manifest_record_type, restrict_rules = restrict_rules)
+        validate_component = None
+    else:
+        validate_component = data_type
+
+    manifest_id = metadata_model.submit_metadata_manifest(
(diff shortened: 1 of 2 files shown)


api/routes.py
```diff
@@ -155,17 +155,19 @@ def submit_manifest_route(schema_url):
...
+    restrict_rules = connexion.request.args["restrict_rules"]
+
     metadata_model = initalize_metadata_model(schema_url)
 
     input_token = connexion.request.args["input_token"]
 
     if data_type == 'None':
         manifest_id = metadata_model.submit_metadata_manifest(
-            manifest_path=temp_path, dataset_id=dataset_id, validate_component=None, input_token=input_token
+            manifest_path=temp_path, dataset_id=dataset_id, validate_component=None, input_token=input_token, manifest_record_type = manifest_record_type, restrict_rules = restrict_rules
```
milen-sage, line 166: nitpick:

In the if/else block, just set the variables validate_component to its respective value (i.e. set it to None if data_type is None, else set it to data_type); then call the function 

metadata_model.submit_metadata_manifest(...) after the if/else statement block.

milen-sage (changes requested):
See comment in code.

api/routes.py
linglp: Thanks for pointing it out Milen! I think we initially designed it this way just in case someone enters the string "None" for dataset_id in the UI. If we do something like below, the code will break when someone enters the string "None": 
```
def submit_manifest_route(schema_url, data_type=None):
    # call config_handler()
    config_handler()

    # Get path to temp file where manifest file contents will be saved
    temp_path = csv_path_handler()

    dataset_id = connexion.request.args["dataset_id"]

    manifest_record_type = connexion.request.args["manifest_record_type"]

    restrict_rules = connexion.request.args["restrict_rules"]

    metadata_model = initalize_metadata_model(schema_url)

    input_token = connexion.request.args["input_token"]

    manifest_id = metadata_model.submit_metadata_manifest(
            manifest_path=temp_path, dataset_id=dataset_id, validate_component=data_type, input_token=input_token, manifest_record_type = manifest_record_type, restrict_rules = restrict_rules
        )
```

milen-sage (approved):

PR #736, opened 2022-05-19
Add Manifest download to the CLI

mialy-defelice:
To match function in the API, add a manifest download option to the CLI along with the new functionality to change the manifest name upon download.


mialy-defelice:
@mialy-defelice bump to check if this needs to still needs to be pushed.

mialy-defelice:
@AmyHeiser do we want to push this?

AmyHeiser:
It would be nice to add this option, but it's an old PR can you merge without any issues @mialy-defelice ?

mialy-defelice:
@AmyHeiser I was able to resolve the merge conflicts, and it can be pushed.

Issue #734, opened 2022-05-18
Schema conversion fails "IndexError: string index out of range"


allaway:
**Describe the bug**
With the current develop branch, this schema (https://github.com/nf-osi/nf-metadata-dictionary/blob/97fc0254d736e02907971d07bda2405ab38c14f0/NF.csv) fails to build with `schematic schema convert`. With the `pip` version of schematic, it builds fine. 

**To Reproduce**
Using https://github.com/nf-osi/nf-metadata-dictionary/blob/97fc0254d736e02907971d07bda2405ab38c14f0/NF.csv, follow the steps in th code chunk below.

**Expected behavior**
I expected `schematic schema convert to work fine` 

**Priority** (select one)
- [ ] Minor ⬇️
- [X] Major 📢
- [ ] Critical 🆘

**Additional context**

pip version: 
```
(data_curator_env) ➜  Downloads pip3 install schematicpy                  
Collecting schematicpy
....
Successfully installed schematicpy-1.0.0 synapseclient-2.3.1

(data_curator_env) ➜  Downloads schematic schema convert -v "DEBUG" NF.csv
Starting schematic...
Done adding requirements and value ranges to attributes
By default, the JSON-LD output will be stored alongside the first input CSV file. In this case, it will appear here: 'NF.jsonld'. You can use the `--output_jsonld` argument to specify another file path.
The Data Model was created and saved to 'NF.jsonld' location.

(data_curator_env) ➜  Downloads pip uninstall schematicpy                 
Found existing installation: schematicpy 1.0.0
Uninstalling schematicpy-1.0.0:
  Would remove:
    /opt/miniconda3/envs/data_curator_env/bin/schematic
    /opt/miniconda3/envs/data_curator_env/lib/python3.9/site-packages/schematic/*
    /opt/miniconda3/envs/data_curator_env/lib/python3.9/site-packages/schematicpy-1.0.0.dist-info/*
Proceed (y/n)? y
  Successfully uninstalled schematicpy-1.0.0
```

github develop version:
```
(data_curator_env) ➜  Downloads pip install git+https://github.com/Sage-Bionetworks/schematic@develop
Collecting git+https://github.com/Sage-Bionetworks/schematic@develop
...
Successfully installed schematicpy-1.0.0 synapseclient-2.5.1

(data_curator_env) ➜  Downloads schematic schema convert -v "DEBUG" NF.csv             
Starting schematic...
Traceback (most recent call last):
[...]

PR #737, opened 2022-05-20
df_parser: Replace calls for pd.isnull to look for empty string

mialy-defelice:
This PR corrects the bug reported by @allaway in issue  #734.

The reason why this bug was being hit was the df_parser was relying on `pd.isnull()` to determine if a string was empty. Instead `pd.isnull('')` was returning `False`. This did not present as an error in our test cases because our examples were never passing an empty string into these fields. But for another issue I was trying to debug, and for Robert they ran into this test case. Normally this error would be bypassed silently for other fields that are not passed by the user but in the case of `Properties` the property name string was attempted to be accessed later. Since there was no value, an error was being thrown.

In the fix, I updated _all_ calls to `pd.isnull(string)` to `string == ""` to more explicitly catch empty strings.

### DIFF
--- a/schematic/schemas/df_parser.py
+++ b/schematic/schemas/df_parser.py
@@ -89,6 +89,9 @@ def get_class(
     # determine parent class of element and add subclass relationship to schema - required by biothings
     # if no subclass is provided, set a default to schema.org Thing
     if subclass_of:
+        if type(subclass_of[0]) == float:
+            if np.isnan(subclass_of[0]):
+                subclass_of[0] = ""
         if len(subclass_of) == 1 and subclass_of[0] == "":
             parent = {"rdfs:subClassOf": [{"@id": "schema:Thing"}]}
         else:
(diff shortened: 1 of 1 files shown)


mialy-defelice:
@ychae

milen-sage (approved):
✈️

mialy-defelice:
@allaway this is a temporary fix that will enable you to make your JSONLD model, but we will be changing some more things to fix this issue 'properly'.

allaway:
Thanks Mialy! I should have clarified in the issue: I am able to make the JSONLD using the pip version of schematic, so this is not a blocker for me, but I figured I should mention it before the current develop is released to pip :)

mialy-defelice:
> Thanks Mialy! I should have clarified in the issue: I am able to make the JSONLD using the pip version of schematic, so this is not a blocker for me, but I figured I should mention it before the current develop is released to pip :)

Its' no worry! I ran into the bug as well when I was trying to work on another issue. In getting a solution, we realized there is a broader problem in how nulls are being processed when the models are being downloaded. I had this solution, so it should work for now, but we aren't going to push it to develop, but will now prioritize pushing in a better solution.

mialy-defelice:
closing due to refactor.

PR #746, opened 2022-05-23
Flag loading data model in df_utils.load_df

mialy-defelice:
Adding a flag to `df_utils.load_df` that a data model is being loaded, so that it is processed correctly. Remove portion that looks for `model` in the file path as an indicator that it is a model.

This will fix a bug reported by @allaway in #734. The bug was happening when a model was being used that did not contain `model` in the name. This caused the model to be imported in a way that caused breaks in `df_parser`.


milen-sage (approved):
🚀

PR #760, opened 2022-05-25
Bug Fix: Spaces Removed from Manifest Column Names

GiaJordan:


GiaJordan:
It seems like there is already a boolean flag `--use_schema_label/--use_display_label` that have similar functionality, renaming the manifest attributes, but for annotation purposes.
I can extend the functionality of this flag instead of adding a new one if it wouldn't be too confusing, but currently the default behavior is to change the names, which would conflict with the desired default behavior to not change the manifest column names.

@milen-sage @mialy-defelice

PR #762, opened 2022-05-26
Manifest gather bug fix, type hints

GiaJordan:


GiaJordan:
This PR fixes a bug where manifests without synIDs would be returned and would cause an error when downloading and trying to access a nonexistent path.
It also adds type hints so method definitions are the same as in the UML diagram

mialy-defelice (approved):
Looks good!

PR #766, opened 2022-06-01
Develop rule combination

GiaJordan:
Includes the architecture for validating multiple rules for a single attribute, and the first set of valid rule combinations

pairs:
- `int::inRange`
- `float::inRange`
- `num::inRange`
- `protectAges::inRange`

as well as updating the `list::regex` combination for the new format


milen-sage (approved):
🚀 
Screen Shot 2022-07-26 at 11 42 44 AM

PR #781, opened 2022-06-08
reformat string list into list of strings for annotations

GiaJordan:


milen-sage:
@andrewelamb this may be useful for you.

PR #783, opened 2022-06-10
Bug Fix for DCA Crashing

GiaJordan:


GiaJordan:
Per @afwillia 

> it works successfully on https://sagebio.shinyapps.io/data_curator-staging/ for HTAN BU archive scRNA-seq Level 1 and Bulk RNAseq level1

mialy-defelice (approved):
@milen-sage  A change to the name of the inserted `Filename` column so it is now `File_name`. This circumvents overlap with the HTAN Data Model where they have an attribute called `Filename`. Please comment if you think this will break things down the line. For now it seems fine to overcome their bug.

Issue #787, opened 2022-06-14
Error surfaces after creating a manifest.


mialy-defelice:
**Describe the bug**
In the newest version of `Develop` a bug is surfacing after making a manifest. The manifest is still successfully made and a URL is provided to the google sheet but following this the following error code pops up:

```
(schematicpy-gze8NI9f-py3.9) sh-3.2# schematic manifest --config ../schematic_package_docs/config.yml get -s -oauth
Starting schematic...
The (manifest > data_type) argument with value '['ToolView']' is being read from the config file.
The '--data_type' argument is being taken from configuration file (manifest > data_type), i.e., '['ToolView']'.
The (model > input > location) argument with value '../test_files/692/csbc_data_model_modified.jsonld' is being read from the config file.
The '--jsonld' argument is being taken from configuration file (model > input > location), i.e., '../test_files/692/csbc_data_model_modified.jsonld'.
The (manifest > title) argument with value 'csbc_data_model_tool' is being read from the config file.
The '--title' argument is being taken from configuration file (manifest > title), i.e., 'csbc_data_model_tool'.
JSON schema successfully generated from schema.org schema!
JSON schema file log stored as ../test_files/692/csbc_data_model_modified.ToolView.schema.json
Permission Id: anyoneWithLink
Find the manifest template using this Google Sheet URL:
https://docs.google.com/spreadsheets/d/1dTvd4xRB3Gz4btOtO68olmNMCfUJuDGHOMuKrcZvPIA
Traceback (most recent call last):
  File "/Users/mialydefelice/Library/Caches/pypoetry/virtualenvs/schematicpy-gze8NI9f-py3.9/lib/python3.9/site-packages/pandas/io/excel/_base.py", line 827, in __new__
    engine = config.get_option(f"io.excel.{ext}.writer", silent=True)
  File "/Users/mialydefelice/Library/Caches/pypoetry/virtualenvs/schematicpy-gze8NI9f-py3.9/lib/python3.9/site-packages/pandas/_config/config.py", line 243, in __call__
    return self.__func__(*args, **kwds)
  File "/Users/mialydefelice/Library/Caches/pypoetry/virtualenvs/schematicpy-gze8NI9f-py3.9/lib/python3.9/site-packages/pandas/_config/config.py", line 115, in _get_option
    key = _get_single_key(pat, silent)
[...]

PR #792, opened 2022-06-20
change shorthand of output_xlsx

linglp:
This is related to the issue mentioned by @mialy-defelice [here](https://github.com/Sage-Bionetworks/schematic/issues/787). 

This issue happened because a short-hand of flag `--oauth` is supposed to be `oa` instead of `-oauth`. Since `-oauth` was used, the code took `auth` as a file name for an excel document. File name `auth` does not have an extension `.xlsx` so the code breaks. 
> Note: In summary, the code read command `schematic manifest --config config.yml get -s -oauth` as `schematic manifest --config config.yml get -s -o auth`

I also noticed that the short-hand of `--output_csv` and `--output_xlsx` were both `-o`. I have updated them to be something different. In the future, I will also remind myself of testing the short-hand commands when I test CLI.

### DIFF
--- a/schematic/manifest/commands.py
+++ b/schematic/manifest/commands.py
@@ -77,7 +77,7 @@ def manifest(ctx, config):  # use as `schematic manifest ...`
     help=query_dict(manifest_commands, ("manifest", "get", "output_csv")),
 )
 @click.option(
-    "-o",
+    "-oxlsx",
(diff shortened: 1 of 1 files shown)


linglp:
To do: 
- change the helper function to reflect the changes

milen-sage (approved):

PR #813, opened 2022-07-06
develop into develop-rdb

andrewelamb:
@mialy-defelice @milen-sage 
This PR has quite a few conflicts, especially to the schematic python code.


andrewelamb:
@mialy-defelice I haven't tried due to the merge conflicts.

mialy-defelice:
> @mialy-defelice I haven't tried due to the merge conflicts.

Ah okay, somehow I missed that. So the first step will to be to resolve all those conflicts, then we can test.

andrewelamb:
@mialy-defelice I regenerated the poetry lock file but now some tests are failing

andrewelamb:
@mialy-defelice @milen-sage 
I've made the changes:
` get_property_label_from_display_name(strict_camel_case = False)`
But that doesn't seem to fix the issue in the notebook, the column is still howtoAquire. I tested RDB.py.get_property_labels_from_table_attrs() and that method is working, so I'm not sure what the issue is now.

mialy-defelice:
> But that doesn't seem to fix the issue in the notebook, the column is still howtoAquire.
@andrewelamb 
Can you elaborate a bit more? 
- What does the JSONLD have the name as? 
-  What does the local version of the DB have the name as? 
-  I'm assuming you removed all the hard coded bits of code that would change the name right?

andrewelamb:
jsonld:

        {
            "@id": "bts:howtoAcquire",
            "@type": "rdf:Property",
            "rdfs:comment": "How to acquire a particular resource. ",
            "rdfs:label": "howtoAcquire",
            "schema:domainIncludes": {
                "@id": "bts:Resource"
            },
            "schema:isPartOf": {
                "@id": "http://schema.biothings.io"
            },
            "sms:displayName": "How to Acquire",
            "sms:required": "sms:true",
            "sms:validationRules": []
        },

local database is "howtoAcquire"

sql_query_nf.py still has:

COL_NAME_UPDATES = {'modelofManifestation': 'modelOfManifestation',
                    'animalModelofManifestation': 'animalModelOfManifestation'}

mialy-defelice:
So you regenerated the JSONLD and subsequently the database calling a strict camelcase but its not actually making a strict camelcase name, though if you call it inline it does?

andrewelamb:
Oh, you know I did not regenerate the JSONLD. Could that be the issue?

Yes to: " though if you call it inline it does?"

Issue #676, opened 2022-04-25
Optional Rule Arguments


GiaJordan:
**Is your feature request related to a problem? Please describe.**
Related to #675
Would like to include optional rule arguments to specify severity of rule violation, eg error vs warning levels, that do not raise errors during model conversion if absent.

**Describe the solution you'd like**
Allow for the specification of optional validation rule arguments, so that errors are only raised when required args are missing.

**How important is this feature?** Select from the options below:
• 🌗 Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks / Mid-term - 2-4 months

**Additional context**
Low priority because Refactor #614 (Involving content validation rules `unique`, `protectAges`, and `inRange`) is yet to be merged (so rules are not in use as of yet), and users can be instructed to treat the message level argument as required for the time being.

GiaJordan:
@milen-sage

GiaJordan:
Raising the priority on this ticket; when #814 is merged this feature will then apply to the `list` rule which brings the total affected rules to:

- `list`
- `protectAges`
- `unique`
- `inRange`

Currently scheduled for sprint 13.
cc: @ychae @milen-sage

PR #814, opened 2022-07-06
Behavior Definition for Lists of ength 1

GiaJordan:


milen-sage (approved):
🚀

GiaJordan:
@milen-sage @ychae 
Confluence documentation updated accordingly

allaway:
🥳 this is great, thanks! (cc @anngvu )

anngvu:
Hey @GiaJordan, are we supposed to use rule "list" or "list like" for non-strict lists? @allaway and I are confused about the semantics; also see this comment: https://github.com/Sage-Bionetworks/schematic/blob/c8484d19940337d8089bc1e843d4d101836fe3fe/schematic/models/validate_attribute.py#L433

Asking because I've kept just "list" in the model, and it's not able to validate with this test file: https://github.com/nf-osi/nf-metadata-dictionary/pull/172/commits/103b3a1dd0a63c1807aa6d0943892452adf430ab

GiaJordan:
Hi @anngvu and @allaway, you should be using `list like` as the validation rule. In the future, when optional arguments are implemented in #676, `list` without a specified level of robustness will default to the old `list` behavior (what is now invoked with `list strict`). That is also what is happening now without specifying a level, but for now specifying a level of robustness for the lists is required

anngvu:
OK, thanks

milen-sage:
@anngvu let us know if `list like` works as expected.

allaway:
@milen-sage , I don't believe that this works as expected. With single-value lists we are getting an error that a valid value is not valid. e.g. 

`[6, 'cellType', "['schwann'] is not one of ['iPSC-derived telencephalic organoids', 'NeuN+', 'epithelial-like', 'fibroblast', 'epithelial', 'Embryonic stem cells', 'cultured Muller glia', 'round', 'oligodendrocyte', 'iPSC-derived glia', 'iPSC-derived astrocytes', 'schwann', 'macrophages', 'monocyte-derived microglia', 'CNON', 'iPSC', 'GLUtamatergic neurons', 'GABAergic neurons', 'Meningioma', 'SH-SY5Y', 'iPSC-derived neuron', 'iPSC-derived neuronal progenitor cell', 'B-lymphocytes', 'lymphoblast', 'Teratoma', 'S", ['schwann']]]`

I'm also getting this error when we have a column that's only partially filled: 
```
[[4, 'age', 'On row 4 the attribute age does not contain the proper value type num.', ''], [5, 'age', 'On row 5 the attribute age does not contain the proper value type num.', ''], [6, 'age', 'On row 6 the attribute age does not contain the proper value type num.', ''],

```

this is not a list-like validation column so I don't think this is possibly a different issue

`

Issue #807, opened 2022-07-05
expose load_df function as an API endpoint 


linglp:
**Is your feature request related to a problem? Please describe.**
This issue is related to the feature request mentioned [here](https://github.com/Sage-Bionetworks/schematic/issues/723).  Based on our discussion during code review, we will use the load_df function [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/utils/df_utils.py#:~:text=def-,load_df,-(file_path%2C))
 to return a data frame with attribute names and data types of attributes. 

**Describe the solution you'd like**

**How important is this feature?** Select from the options below:
• 🌗 Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

**Additional context**

linglp:
Also explore the functionality of synapse client and see if we could use some functions to determine the data types of columns.

linglp:
@ychae this is almost done. I am waiting for Gianna to double check one line of code since that would change a function she wrote. Also, I think we could decrease the priority level of this issue since Andrew has paused his work on project live.

PR #815, opened 2022-07-08
Develop an endpoint to return data types of columns in a manifest 

linglp:
This PR is related to the issue [here](https://github.com/Sage-Bionetworks/schematic/issues/807). Here are the tests that I did: 
- Test manifest [here](https://www.synapse.org/#!Synapse:syn32602107)
Output: 
```
{
  "Component": "string",
  "File Format": "string",
  "Filename": "string",
  "Genome Build": "string",
  "Genome FASTA": "string",
  "Sample ID": "Int64",
  "Test": "Float64",
  "Test2": "string",
  "Test3": "string",
  "Test4": "object",
  "Test5": "string",
  "entityId": "string"
}
```

- Test manifest [here](https://www.synapse.org/#!Synapse:syn32631266)
Output: 
```
{
  "Check Ages": "Int64",
  "Check Float": "Float64",
  "Check Int": "Float64",
  "Check List": "string",
  "Check Match Exactly": "Int64",
  "Check Match at Least": "Int64",
  "Check Num": "object",
  "Check Range": "Int64",
  "Check Recommended": "string",
  "Check Regex List": "string",
  "Check Regex Single": "string",
  "Check String": "object",
  "Check URL": "string",
  "Check Unique": "string",
  "Component": "string"
}
```

Notes: if two rows under the same column have different data types (i.e. an integer and a string), then it will return as type `object`. Currently, there's also no type `list`. `list` like object (see test 1, column `Test 2` and `Check Regex List` column in test 2)  get returned as "string"

### DIFF
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -350,6 +350,43 @@ paths:
           description: Check schematic log
       tags:
         - Manifest Operations
+  /get/datatype/manifest:
+    get:
+      summary: Retrieve asset view table as a dataframe.
+      description: Retrieve asset view table as a dataframe.
+      operationId: api.routes.get_manifest_datatype
+      parameters:
+        - in: query
+          name: input_token
+          schema:
+            type: string
+            nullable: false
+          description: Token
+          example: Token
+          required: true
+        - in: query
+          name: asset_view
+          schema:
+            type: string
+            nullable: false
+          description: ID of view listing all project data assets. For example, for Synapse this would be the Synapse ID of the fileview listing all data assets for a given project.(i.e. master_fileview in config.yml)
+          example: syn23643253
+          required: true
+        - in: query
+          name: manifest_id
+          schema:
+            type: string
+            nullable: false
+          description: Manifest ID
+          example: syn23643253
+          required: true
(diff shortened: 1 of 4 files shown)


linglp:
@GiaJordan Just like we discussed in Slack, if I do not add `org_df = org_df.astype(str)` in `load_df` function, I would get error: `TypeError: descriptor 'isdigit' for 'str' objects doesn't apply to a 'float' object`. Could you check if I could do that? I am not sure if that would affect anything downstream. Thanks

linglp:
@ychae this one is almost done. I am just waiting for Gianna come back and double check a line of code.

GiaJordan:
@linglp it seems like the line you added is changing the behavior of cell values left empty and filling them in with something, and that's what's causing the `Check Recommended` assertion in the tests to fail

linglp:
> @linglp it seems like the line you added is changing the behavior of cell values left empty and filling them in with something, and that's what's causing the `Check Recommended` assertion in the tests to fail

Thank you for the hint! I have figured out how to resolve the error in testing :0

milen-sage (approved):

PR #818, opened 2022-07-12
alphabetize in getModelManifest

allaway:
Surfacing this parameter for the DCA. This sets alphabetize_valid_values to True as a default.


allaway:
I am not sure what I was thinking but alphabetize_valid_values is not a boolean!!

milen-sage:
But just to confirm @allaway, you'd like the default behavior to be alphabetized valid values right?  Agreed that this makes a lot of sense!

PR #836, opened 2022-07-25
BugFix: Import dataType from CLI as list not string

GiaJordan:
Bug outlined in #839


milen-sage (approved):

milen-sage:
@mialy-defelice merging now since a collaborator may be bumping into this bug. Let's see if this resolves it.

PR #837, opened 2022-07-25
BugFix: Manifest Import - Mask null cells as empty strings instead of NaNs

GiaJordan:
Bug outlined in #838


linglp (approved):

linglp:
@GiaJordan thanks. I have pulled your branch and test it out to make sure that endpoint `get/datatype/manifest` is still working as expected

Issue #819, opened 2022-07-13
List like validation not working as expected


allaway:
**Describe the bug**
There are two list validation issues we are encountering with the latest version of develop. With a 'list like' validation for modelSystemName and cellType, we get the following errors when we supply valid values from this test manifest:

`
[...]

PR #840, opened 2022-07-26
BugFix: list like behavior

GiaJordan:
The bug described in #819 arose because arguments had been added to the `list` validation rule that then made all the logic that checked if `list` was a specified rule for an attribute invalid. 
ie. performing `if 'list' in validation_rules` no longer worked as expected when the rule specification changed to `list like` or `list strict`
While this happened with the list rule something similar could happen with other rules if the number of arguments change, so a more general solution was necessary.

A function was added to `validate_utils.py` that will be used to check if a rule type is present in a given list of validation rules. This is more applicable now that rule combinations have expanded with the merge of #766, and a single attribute can have multiple rules associated with it. The function performs a regex search for the rule type within the list of rules provided and returns a `match` object if a match is found, or `NULL` if no match is found.

### DIFF
--- a/schematic/models/validate_manifest.py
+++ b/schematic/models/validate_manifest.py
@@ -34,7 +34,7 @@ class ValidateManifest(object):
         self.jsonSchema = jsonSchema       
 
     def get_multiple_types_error(
-        validation_rules: list, attribute_name: str, error_type: str
+        self, validation_rules: list, attribute_name: str, error_type: str
     ) -> List[str]:
         """
             Generate error message for errors when trying to specify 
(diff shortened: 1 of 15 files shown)


schematic/models/validate_manifest.py
```diff
@@ -21,14 +21,7 @@
 from schematic.schemas.generator import SchemaGenerator
 from schematic.store.synapse import SynapseStorage
 from schematic.models.GE_Helpers import GreatExpectationsHelpers
-
-from ruamel import yaml
```
milen-sage, line 25: Were the imports below unused?

GiaJordan: That's correct. They were leftovers from before `GE_Helpers` was added, and were only used in those functions. Since it's been moved to its own file they're no longer necessary here, but they are present where needed in the other file.

milen-sage (approved):

Issue #828, opened 2022-07-19
Implement data model graph class


milen-sage:

mialy-defelice:
Create a skeleton for the following:
- [ ] schema/data_model_graph.py
- [ ] Class DataModelGraph:
    - [ ] Following a [singleton class design](https://refactoring.guru/design-patterns/singleton)
    - [ ] init():
    - [ ] generate_data_model_graph()
    - [ ] convert data_model_parser output into a networkx schema.
    - [ ] Analogous to function that converts JSONLD to networkx schema (load_schema_into_networkx)
    - [ ] Iteratively load nodes and edges into networkx by calling on node and edge classes.
    - [ ] Add attributes to nodes analogous to previous function.
    - [ ] Returns data_model_graph

- [ ] Class DataModelGraphExplorer:
Add all explorer functions

mialy-defelice:
Created skeleton classes.

Issue #827, opened 2022-07-19
Implement data model parser class


milen-sage:

mialy-defelice:
Create skeleton for the following:

- [ ] schema/data_model_parser.py

- [ ] Class DataModelParser:
Get the relationships and attributes from the user submitted CSV or JSONLD.
    - [ ] init():
    - [ ] get_model_type():
    - [ ] parse_model():

- [ ] Class DataModelCSVParser:
Take and input csv and gather all relationships and their attributes
Return in a form that can then be used to load into networkx graph: https://networkx.org/documentation/latest/tutorial.html
A lot of the general logic exists in the current df_parser/create_nx_schema_object/, but its loaded in a way that is specific to JSONLD… decouple this logic.
Will mean that many of the current helper functions will be swapped out.
Return attributes and relationships

- [ ] Class DataModelJSONLDParser:
Get JSONLD and gather all relationships and attributes.
A lot of this already exists in utils/schema_utils.py but decouple from networks for now.

mialy-defelice:
Created skeleton files and classes.

Issue #826, opened 2022-07-19
Implement new JSON-LD object, property and class classes


milen-sage:

mialy-defelice:
Create a skeleton for the following:

- [ ] Create `create_json_ld.py`
- [ ] Add `Class JSONLD_object`
    - [ ] [Decorator class design](https://refactoring.guru/design-patterns/decorator)
    - [ ]   __init_()
    - [ ] self.to_template() -> returns a dictionary or a string thats a JSON-LD object
    - [ ] Returns jsonld_class_template or jsonld_property_template
    - [ ] Call on JSONLD_project and JSONLD_class
- [ ] Add `Class JSONLD_property`
    - [ ] [Decorator class design](https://refactoring.guru/design-patterns/decorator)
    - [ ] __init_()
    - [ ] self.to_template() 
    - [ ] explore_property: 
    - [ ] edit_property(self, property_info)
- [ ] Add `Class JSONLD_class` 
    - [ ] [Decorator class design](https://refactoring.guru/design-patterns/decorator)
    - [ ] __init_()
    - [ ] self.to_template() 
    - [ ] explore_class: 
    - [ ] edit_class(self, property_info)

mialy-defelice:
Created skeleton for classes

Issue #825, opened 2022-07-19
Implement new data mode graph node and edge classes


milen-sage:

mialy-defelice:
Create a skeleton for the following:

- [ ] schema/data_model_nodes.py
The point of these classes is to take in relationships and attribute data to create the nodes and edges for the network graph.
- [ ] Class DataModelNodes:
    - [ ] init
    - [ ] generate_node():
    - [ ] edit_node()

- [ ] schema/data_model_edges.py
- [ ] Class DataModelEdges:
    - [ ] init
    - [ ] generate_edge()
    - [ ] edit_edge()

mialy-defelice:
Created skeleton classes.

PR #841, opened 2022-07-26
Refactor Schemas

mialy-defelice:
This PR is a WIP to address all the work needed to refactor the Schemas portion of Schematic.

Related to the following Issues:

- [x] #825 
- [x] #826 
- [x] #827 
- [x] #828

@milen-sage


mialy-defelice:
Starting with #827.

schematic/schemas/data_model_json_schema.py
```diff
@@ -0,0 +1,356 @@
...
+				"items": {"enum": node_range + [""] if blank else node_range},
+				"maxItems": len(node_range),
+			}
+		}
+
+		return schema_node_range_array
+
+	def get_non_blank_schema(
+		self, node_name: str
+	) -> Dict:  # can't define heterogenous Dict generic types
```
andrewelamb, line 49: You can define heterogenous dict types in a few ways such as:
`dict[Any, Any]` : literally anything for both keys and values
`dict[str, Any]` : The keys are always strings, but the value could be anything
`dict[str, Union[str, int]]` They keys are always strings, but the values vould be a string or an int

For this particular type you could use:
`dict[str, dict[str, Any]]`

But I think you might want to refer to the above comment about dataclasses.

schematic/schemas/data_model_json_schema.py
```diff
@@ -0,0 +1,356 @@
...
+			JSON object with nodes.
+		"""
+		if blank:
+			schema_node_range = {node_name: {"enum": node_range + [""]}}
+		else:
+			schema_node_range = {node_name: {"enum": node_range}}
+
+		return schema_node_range
+
+	def get_json_validation_schema(self, source_node: str, schema_name: str) -> Dict:
```
andrewelamb, line 83: See above comment about dataclasses

schematic/schemas/data_model_json_schema.py
```diff
@@ -0,0 +1,356 @@
...
+		# if root_dependencies is empty it means that a class with name 'source_node' exists
+		# in the schema, but it is not a valid component
+		if not root_dependencies:
+			raise ValueError(f"'{source_node}' is not a valid component in the schema.")
+
+		nodes_to_process += root_dependencies
+
+		process_node = nodes_to_process.pop(0)
+
+		while process_node:
```
andrewelamb, line 137: This while loop is a bit out of control? If I had to come in and work on it , I wouldn't know where to start. The comments are helpful, but there is no way I could keep track of how this whole loop worked in my head. I strongly suggest breaking this up into more managable chunks.

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
...
+                            "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+                            "schema": "http://schema.org/",
+                            "xsd": "http://www.w3.org/2001/XMLSchema#",
+                            },
+                         "@graph": [],
+                         "@id": "http://schema.biothings.io/#0.1",
+                        }
+        return base_template
+
+    def create_object(self, template:dict, node:str)->dict:
```
andrewelamb, line 43: I think the name of this method could be more descriptive. It sounds like its processing the template object and returning it? 

I'm also thinking that the template object would benefit from being its own dataclass.

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
...
+        Args:
+            template, dict: empty class or property template to be filled with information for the given node.
+            node, str: target node to fill the template out for.
+        Returns:
+            template, dict: filled class or property template, that has been processed and cleaned up.
+        """
+        data_model_relationships = self.dmr.relationships_dictionary
+
+        # For each field in template fill out with information from the graph
+        for rel, rel_vals in data_model_relationships.items():
```
andrewelamb, line 54: I suggest breaking this for loop into several methods.

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
...
+            template, dict: JSONLD template that has been filled up to the current node, with information
+        Returns:
+            template, dict: JSONLD template where contexts have been added back to certain values.
+        Note: This will likely need to be modified when Contexts are truly added to the model
+        """
+        for jsonld_key, entry in template.items():
+            try:
+                # Retrieve the relationships key using the jsonld_key
+                key= [k for k, v in self.rel_dict.items() if jsonld_key == v['jsonld_key']][0]
+            except:
```
andrewelamb, line 136: I don't think you want to catch all exceptions here?

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
+from functools import wraps
+from typing import Any, Dict, Optional, Text, List
+import networkx as nx
+
+from schematic.schemas.data_model_graph import DataModelGraphExplorer
+from schematic.schemas.data_model_relationships import DataModelRelationships
+from schematic.utils.schema_utils import get_label_from_display_name, convert_bool_to_str
+
+
+class DataModelJsonLD(object):
```
andrewelamb, line 10: I think the template object that gets referred to by several methods in this class should be its own class.

schematic/schemas/data_model_nodes.py
```diff
@@ -0,0 +1,232 @@
...
+class DataModelNodes():
+    def __init__(self, attribute_relationships_dict):
+        self.namespaces = dict(rdf=Namespace("http://www.w3.org/1999/02/22-rdf-syntax-ns#"))
+        self.data_model_relationships = DataModelRelationships()
+        self.value_relationships = self.data_model_relationships.define_value_relationships()
+        self.edge_relationships_dictionary = self.data_model_relationships.define_edge_relationships()
+        self.properties = self.get_data_model_properties(attr_rel_dict=attribute_relationships_dict)
+
+    '''    
+    def node_present(self, G, node_name):
```
andrewelamb, line 24: This should be removed or uncommented

schematic/schemas/data_model_nodes.py
```diff
@@ -0,0 +1,232 @@
...
+        """Take in a tuple containing attriute name and relationship dictionary, and find all nodes defined in attribute information.
+        Args:
+            attr_info, tuple: (Display Name, Relationships Dictionary portion of attribute_relationships dictionary)
+        Returns:
+            nodes, list: nodes related to the given node (specified in attr_info).
+        Note:
+            Extracting nodes in this fashion ensures order is preserved.
+        """
+        # retrieve a list of relationship types that will produce nodes.
+        self.node_relationships =list(self.edge_relationships_dictionary.values())
```
andrewelamb, line 40: This should first be defined in the init method like `self.node_relationships = None`

schematic/schemas/data_model_nodes.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+        all_nodes = []
+        for attr_info in attr_rel_dict.items():
+            nodes = self.gather_nodes(attr_info=attr_info)
+            all_nodes.extend(nodes)
+        # Remove any duplicates preserving order
+        all_nodes = list(dict.fromkeys(all_nodes).keys())
+        return all_nodes
+
+    def get_rel_node_dict_info(self, relationship: str) -> tuple:
```
andrewelamb, line 75: This method is kind of problematic in that it says it returns a tuple, but it only returns one if both nested if statements are true. Should be `Optional[tuple[str, dict]]`

andrewelamb:
@mialy-defelice I left a bunch of suggestions. I'm still going through the code though.

mialy-defelice:
> @mialy-defelice I left a bunch of suggestions. I'm still going through the code though.

@andrewelamb Thank you!

schematic/schemas/data_model_edges.py
```diff
@@ -0,0 +1,64 @@
...
+
+class DataModelEdges():
+    def __init__(self):
+        self.dmr = DataModelRelationships()
+        self.data_model_relationships = self.dmr.relationships_dictionary
+
+    def generate_edge(self, G: nx.MultiDiGraph, node: str, all_node_dict: dict, attr_rel_dict: dict, edge_relationships: dict) -> nx.MultiDiGraph:
+        """Generate an edge between a target node and relevant other nodes the data model. In short, does this current node belong to a recorded relationship in the attribute, relationshps dictionary. Go through each attribute and relationship to find where the node may be.
+        Args:
+            G, nx.MultiDiGraph: networkx graph representation of the data model, that is in the process of being fully built.
```
GiaJordan, line 15: I would add to the description to say where it is in the process of being built: ex "A graph representation of the data model that has all nodes added but none/some edges"

GiaJordan (commented):
I've gone through some of the files so far, I'll add more comments as I get through more

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+        '''Parses the path to the data model to extract the extension and determine the data model type.
+        Args:
+            path_to_data_model, str: path to data model
+        Returns:
+            str: uppercase, data model file extension.
+        Note: Consider moving this to Utils.
+        '''
+        return pathlib.Path(path_to_data_model).suffix.replace('.', '').upper()
+
+    def parse_base_model(self):
```
andrewelamb, line 60: The output here could be typed as `dict` at least. Maybe basemodel should be its own dataclass?

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+
+        # Determine base schema path
+        base_model_path = self._get_base_schema_path(self.base_schema_path)
+
+        # Parse
+        jsonld_parser = DataModelJSONLDParser()
+        base_model = jsonld_parser.parse_jsonld_model(base_model_path)
+        return base_model
+
+    def parse_model(self):
```
andrewelamb, line 78: Same as comment about parse base model

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+            elif "Requires" in list(model_df.columns) or "Requires Component" in list(
+                model_df.columns
+            ):
+                raise ValueError(
+                    "The input CSV schema file contains the 'Requires' and/or the 'Requires "
+                    "Component' column headers. These columns were renamed to 'DependsOn' and "
+                    "'DependsOn Component', respectively. Switch to the new column names."
+                )
+            logger.debug("Schema definition csv ready for processing!")
+        except:
```
andrewelamb, line 132: This looks like it should be expecting a specific error. You generally don't want to except all errors.

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+                {Attribute Display Name: {
+                    Relationships: {
+                                    CSV Header: Value}}}
+        '''
+        # Check csv schema follows expectations.
+        self.check_schema_definition(model_df)
+
+        # Get the type for each value that needs to be submitted.
+        # using csv_headers as keys to match required_headers/relationship_types
+        self.rel_val_types = {v['csv_header']:v['type']for k, v in self.rel_dict.items() if 'type' in v.keys()}
```
andrewelamb, line 155: If you're going to set an attribute in a method, it should probably be set in the `__init__` method as well, even if it just as `None`

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+        # Build attribute/relationship dictionary
+        relationship_types = self.required_headers
+        attr_rel_dictionary = {}
+
+        for attr in attributes:
+            # Add attribute to dictionary        
+            attr_rel_dictionary.update({attr['Attribute']: {'Relationships': {},
+                                                            },
+                                        }
+                                       )
```
andrewelamb, line 169: Could we break this for block up into several methods?

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+            - Right now, here we are stripping contexts, will need to track them in the future.
+        '''
+        
+        # Retrieve relevant JSONLD keys.
+        jsonld_keys_to_extract = ['label', 'subClassOf', 'id', 'displayName']
+        label_jsonld_key, subclassof_jsonld_key, id_jsonld_key, dn_jsonld_key = [self.rel_dict[key]['jsonld_key']
+                                                    for key in jsonld_keys_to_extract ]
+        
+        # Gather all labels from the model.
+        model_labels = [v[label_jsonld_key] for v in model_jsonld]
```
andrewelamb, line 246: model_labels doesn't seem to be used anywhere

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+
+        attr_rel_dictionary = {}
+        # Move through each entry in the jsonld model
+        for entry in model_jsonld:
+            # Get the label of the entry
+            try:
+                # Get the entry display name (if recorded)
+                entry_name = entry[dn_jsonld_key]
+            except:
```
andrewelamb, line 255: Should we be excepting a specific error here?

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+                entry_name = entry[label_jsonld_key]
+
+            # If the entry is an attribute that has not already been added to the dictionary, add it.
+            if entry_name not in attr_rel_dictionary.keys():
+                attr_rel_dictionary.update({entry_name: {'Relationships': {}}})
+            
+            # Add relationships for each attribute
+            # 
+            # Go through each defined relationship type (key) and its attributes (val)
+            for key, val in self.rel_dict.items():
```
andrewelamb, line 266: Can we break this for block up? It's very complicated.

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+                        # Strip context from string and convert true/false to bool
+                        elif type(rel_entry) == str:
+                            # Remove contexts and treat strings as appropriate.
+                            if ':' in rel_entry and 'http:' not in rel_entry:
+                                parsed_rel_entry = rel_entry.split(':')[1]
+                                # Convert true/false strings to boolean
+                                if parsed_rel_entry.lower() =='true':
+                                    parsed_rel_entry = True
+                                elif parsed_rel_entry.lower == 'false':
+                                    parsed_rel_entry == False
```
andrewelamb, line 288: Should this be `parsed_rel_entry = False`?

andrewelamb:
@mialy-defelice There are some general themes:

- Could you run black on all the files? There are some with weird spacing issues (ie. data_model_validator.py).
- Some methods don't have type hints.
- It would be great if some of the larger methods with lots of if/else logic got broken up into smaller methods. This wouyld make it easier to do unit testing and for making the code easier to understand.
- Some of the docstrings don't match the arguments names. I'm guessing these were the old docstrings that need to be updated?
- There are several cases where certain inputs and outputs are very complex, and would benefit from being turned into their own classes (I think dataclasses to be specific). This would make typing much easier and could lead to less bugs as you could enforce contstraints on the classes.

schematic/schemas/data_model_relationships.py
```diff
@@ -0,0 +1,222 @@
...
+                'required': {
+                    'jsonld_key': 'sms:required',
+                    'csv_header': 'Required',
+                    'node_label': 'required',
+                    'type': bool,
+                    'jsonld_default': 'sms:false',
+                    'edge_rel': False,
+                    'required_header': True,
+                    'node_attr_dict':{'default': False,
+                                      'standard': convert_bool_to_str,
```
mialy-defelice, line 114: The way this is actually being used now is to convert str to bool or maybe more accurately extract a boolean value (bc will just pull bool if its recorded as a bool). Update the function name here, update schema utils and update data_model_nodes. Can pull the function directly from data_model_nodes.run_rel_functions"

schematic/schemas/data_model_nodes.py
```diff
@@ -0,0 +1,232 @@
...
+                    return True
+                elif attr_relationships[csv_header].lower() == 'false':
+                    return False
+        
+            elif type(attr_relationships[csv_header]) == bool:
+                return attr_relationships[csv_header]
+        
+        else:
+            # Raise Error if the rel_func provided is not captured.
+            raise KeyError(f"The function provided ({rel_func}) to define the relationship {key} is not captured in the function run_rel_functions, please update.")
```
mialy-defelice, line 159: This should be a value error instead.

GiaJordan (commented):
@mialy-defelice still working my way thorough the files, more comments to come

schematic/schemas/data_model_graph.py
```diff
@@ -0,0 +1,705 @@
...
+            node_dict = self.dmn.generate_node_dict(node, self.attribute_relationships_dict)
+
+            # Add each node to the all_node_dict to be used for generating edges
+            all_node_dict[node] = node_dict
+
+            # Generate node and attach information
+            G = self.dmn.generate_node(G, node_dict)
+
+        ## Connect nodes via edges
+        for node in all_nodes:
```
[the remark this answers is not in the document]
mialy-defelice: Yep @GiaJordan  thats exactly it!

schematic/schemas/data_model_parser.py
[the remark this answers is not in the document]
mialy-defelice: @GiaJordan the current refactor functionality does not allow for any base schema. I hard coded in biothings here, bc we are not even passing a parameter if passed, but I should perhaps remove that option from the CLI. We have tickets to allow base schema in the future.

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+        # Load required csv headers
+        self.required_headers = self.dmr.define_required_csv_headers()
+
+    def check_schema_definition(self, model_df: pd.DataFrame) -> bool:
+        """Checks if a schema definition data frame contains the right required headers.
+        Args:
+            schema_definition: a pandas dataframe containing schema definition; see example here: https://docs.google.com/spreadsheets/d/1J2brhqO4kpeHIkNytzlqrdIiRanXDr6KD2hqjOTC9hs/edit#gid=0
+        Raises: Exception
+        """
+        try:
```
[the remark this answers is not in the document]
mialy-defelice: This was actually just a copy paste from the current schematic ver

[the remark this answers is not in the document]
GiaJordan: oh, that's interesting

schematic/schemas/data_model_relationships.py
```diff
@@ -0,0 +1,222 @@
...
+            Returns:
+                edge_relationships: dict, key: csv_header if the key represents an edge relationship.
+        """
+        edge_relationships = {}
+        for k, v in self.relationships_dictionary.items():
+            try:
+                if v['edge_rel']:
+                    edge_relationships.update({k:v['csv_header']})
+            except KeyError:
+                print(f"Did not provide a 'edge_rel' key, value pair for the nested dictionary {k} : {key}")
```
linglp, line 203: here it seems like there is a typo. `key` is not defined right? It should be v or k?

schematic/schemas/data_model_relationships.py
```diff
@@ -0,0 +1,222 @@
...
+
+        return required_headers
+
+    def define_edge_relationships(self):
+        """Helper function to retrieve CSV headers for edge relationships.
+            Returns:
+                edge_relationships: dict, key: csv_header if the key represents an edge relationship.
+        """
+        edge_relationships = {}
+        for k, v in self.relationships_dictionary.items():
```
linglp, line 198: should we avoid using single character as variable name here? I think `k` could be changed to "relationship_name" and `v` could be changed to `relationship_info`?

schematic/schemas/data_model_nodes.py
```diff
@@ -0,0 +1,232 @@
...
+        nodes = []
+        if attribute not in nodes:
+            nodes.append(attribute)
+        for rel in self.node_relationships:
+            if rel in relationships.keys():
+                nodes.extend([node.strip()
+                                for node in relationships[rel]])
+        return nodes
+
+    def gather_all_nodes(self, attr_rel_dict: dict):
```
linglp, line 55: Nit: I think the function name "gather_all_nodes" is too similar to the name "gather_nodes".. and it seems like "gather_all_nodes" function is mainly for removing duplicates after calling "gather_nodes".. maybe this two can be consolidated or the function could be renamed? not a big deal but just my two cents..

schematic/schemas/data_model_relationships.py
```diff
@@ -0,0 +1,222 @@
...
+            try:
+                if v['edge_rel']:
+                    edge_relationships.update({k:v['csv_header']})
+            except KeyError:
+                print(f"Did not provide a 'edge_rel' key, value pair for the nested dictionary {k} : {key}")
+
+        return edge_relationships
+
+    def define_value_relationships(self):
```
linglp, line 207: this function is very similar to "define_edge_relationships".. can we possibly combine the two?

schematic/schemas/data_model_nodes.py
```diff
@@ -0,0 +1,232 @@
...
+                elif attr_relationships[csv_header].lower() == 'false':
+                    return False
+        
+            elif type(attr_relationships[csv_header]) == bool:
+                return attr_relationships[csv_header]
+        
+        else:
+            # Raise Error if the rel_func provided is not captured.
+            raise KeyError(f"The function provided ({rel_func}) to define the relationship {key} is not captured in the function run_rel_functions, please update.")
+            return
```
linglp, line 160: Nit: this return should be removed right? Because you already have "raise"?

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
...
+                    # if possible pull standard function to get node information
+                    rel_func = self.rel_dict[key]['node_attr_dict']['standard']
+                except:
+                    # if not pull default function to get node information
+                    rel_func = self.rel_dict[key]['node_attr_dict']['default']
+
+                # Add appropritae contexts that have been removed in previous steps (for JSONLD) or did not exist to begin with (csv)
+                if key == 'id' and rel_func == get_label_from_display_name:
+                    template[jsonld_key] = 'bts:' + template[jsonld_key]
+                elif key == 'required' and rel_func == convert_bool_to_str:
```
mialy-defelice, line 150: If the template value is left, then SMS would not have been stripped. Need to add a check to see if `sms` is already part of the str before trying to add it back, same for above with `bts` (in this case its an extra safety)

linglp (commented):
There are some old functions that are not being used but probably still need to be updated. 

1. In `data_model_graph.py`, `get_nodes_descendants` function takes unused parameter `node_label` and references variable "component" that is not defined. 
2. In `data_model_graph.py`, `find_adjacent_child_classes` function takes unused parameter node_label and node_display_name while references `schema_class` that is not defined.
3. In `data_model_graph.py`, `find_node_range` function takes unused parameter `node_display_name` and references `display_name` that is not defined.
4. In `data_model_graph.py`,  `get_edges_by_relationship` function takes unused parameter `class_label` and references `node` that is not defined
5. There are a lot of unused imports probably also need to get cleaned up there

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
...
+        '''
+        for rels in data_model_relationships.values():
+            if rels['jsonld_key'] in template.keys() and not template[rels['jsonld_key']]:
+                if 'jsonld_default' in rels.keys():
+                    template[rels['jsonld_key']] = rels['jsonld_default']
+                else:
+                    del template[rels['jsonld_key']]
+        return template
+
+    def strip_context(self, context_value: str) -> tuple[str]:
```
mialy-defelice, line 170: can move this to a util function

schematic/schemas/data_model_json_schema.py
```diff
@@ -0,0 +1,357 @@
...
+			self.jsonld_path_root, jsonld_ext = os.path.splitext(self.jsonld_path)
+			prefix = self.jsonld_path_root
+			prefix_root, prefix_ext = os.path.splitext(prefix)
+			if prefix_ext == ".model":
+				prefix = prefix_root
+			json_schema_log_file = f"{prefix}.{source_node}.schema.json"
+
+		logger.info(
+			"The JSON schema file can be inspected by setting the following "
+			"nested key in the configuration: (model > input > log_location)."
```
linglp, line 352: This should be "model > location" because we updated the config

schematic/manifest/generator.py
```diff
@@ -35,6 +37,7 @@ class ManifestGenerator(object):
     def __init__(
         self,
         path_to_json_ld: str,  # JSON-LD file to be used for generating the manifest
+        graph: nx.MultiDiGraph,
```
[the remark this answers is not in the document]
mialy-defelice: At this point its fully formed and is not super different than passing the sg in past times, which gave access to the graph.
But I can make that clear, and update it through different parts of the code.

schematic/models/metadata.py
```diff
@@ -53,17 +53,24 @@ def __init__(self, inputMModelLocation: str, inputMModelLocationType: str,) -> N
...
-            raise TypeError(
-                f"Please make sure {inputMModelLocation} is a .jsonld file."
-            )
+        logger.debug(
+            f"Initializing DataModelGraphExplorer object from {inputMModelLocation} schema."
+        )
+
+        self.inputMModelLocation = inputMModelLocation
+
+        data_model_parser = DataModelParser(path_to_data_model = self.inputMModelLocation)
```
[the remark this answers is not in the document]
mialy-defelice: @GiaJordan Do you think this would belong in schema_utils?

schematic/schemas/commands.py
```diff
@@ -32,32 +38,71 @@ def schema():  # use as `schematic model ...`
...
+    data_model_parser = DataModelParser(schema)
+
+    #Parse Model
+    logger.info("Parsing data model.")
+    parsed_data_model = data_model_parser.parse_model()
 
-    # output JSON-LD file alongside CSV file by default
+    # Convert parsed model to graph
+    # Instantiate DataModelGraph
+    data_model_grapher = DataModelGraph(parsed_data_model)
```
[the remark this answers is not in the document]
mialy-defelice: I agree... which naming do you prefer?

schematic/schemas/data_model_edges.py
```diff
@@ -0,0 +1,64 @@
...
+
+        Returns:
+            G, nx.MultiDiGraph: networkx graph representation of the data model, that has had new edges attached.
+        """
+        # For each attribute in the model.
+        for attribute_display_name, relationship in attr_rel_dict.items():
+            # Get the relationships associated with the current attribute
+            relationships = relationship['Relationships']
+            # Add edge relationships one at a time
+            for key, csv_header in edge_relationships.items():
```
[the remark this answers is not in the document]
mialy-defelice: Updated! They are the keys in the relationships dictionary.

schematic/schemas/data_model_json_schema.py
```diff
@@ -0,0 +1,356 @@
...
+				 ):
+		self.jsonld_path = jsonld_path
+		self.graph = graph
+		self.DME = DataModelGraphExplorer(self.graph)
+		self.dmr = DataModelRelationships()
+		self.rel_dict = self.dmr.relationships_dictionary
+			
+	def get_array_schema(
+		self, node_range: List[str], node_name: str, blank=False
+	) -> Dict[str, Dict[str, List[str]]]:
```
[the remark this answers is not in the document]
mialy-defelice: @andrewelamb the JSON Schema part of the code base was actually not touched in this refactor. @linglp  just opened a ticket to address it. For now we just reorganized where it was located (moved it from the previous location and into its own class).

schematic/schemas/data_model_json_schema.py
mialy-defelice: @linglp

schematic/schemas/data_model_json_schema.py
mialy-defelice: @andrewelamb Yes, we decided to leave this outside of the scope of this refactor.

schematic/schemas/data_model_json_schema.py
mialy-defelice: @linglp updated, but will actually will comment out the logging for now, since the change to the config i think actually broke the saving of the json schema file. There is already a ticket to address this in the backlog.

mialy-defelice: The logging can be un-commented when/if its fixed.

schematic/schemas/data_model_nodes.py
mialy-defelice: Renamed the function to `gather_all_nodes_in_model`.
Its actually gathering all the nodes calling gather_nodes per attribute within, rather than just removing duplicates after all the nodes have been found. I think in general we should be keep functions as small as possible, so I hesitate to join them.

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+        self.rel_dict = self.dmr.define_data_model_relationships()
+        # Get edge relationships
+        self.edge_relationships_dictionary = self.dmr.define_edge_relationships()
+        # Load required csv headers
+        self.required_headers = self.dmr.define_required_csv_headers()
+
+    def check_schema_definition(self, model_df: pd.DataFrame) -> bool:
+        """Checks if a schema definition data frame contains the right required headers.
+        Args:
+            schema_definition: a pandas dataframe containing schema definition; see example here: https://docs.google.com/spreadsheets/d/1J2brhqO4kpeHIkNytzlqrdIiRanXDr6KD2hqjOTC9hs/edit#gid=0
```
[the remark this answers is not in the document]
mialy-defelice: Yes thanks for catching this.

schematic/schemas/data_model_parser.py
mialy-defelice: given the time constraints of the refactor, i dont think I can convert to a dataclass for the initial release, but will open a ticket.

mialy-defelice: https://sagebionetworks.jira.com/browse/FDS-1211

schematic/schemas/data_model_parser.py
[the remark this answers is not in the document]
mialy-defelice: I was able to get rid of the try with an additional elif.

schematic/schemas/data_model_parser.py
mialy-defelice: I was able to get rid of the try with an additional elif.

schematic/schemas/data_model_parser.py
mialy-defelice: I'll move it up to init

schematic/schemas/data_model_parser.py
```diff
@@ -0,0 +1,319 @@
...
+        self.rel_val_types = {v['csv_header']:v['type']for k, v in self.rel_dict.items() if 'type' in v.keys()}
+        
+        # get attributes from Attribute column
+        attributes = model_df[list(self.required_headers)].to_dict("records")
+        
+        # Build attribute/relationship dictionary
+        relationship_types = self.required_headers
+        attr_rel_dictionary = {}
+
+        for attr in attributes:
```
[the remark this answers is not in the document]
mialy-defelice: @GiaJordan Do you still want an explaination?

schematic/schemas/data_model_parser.py
mialy-defelice: no, it was just a way of dealing with if there wasnt a displayname recorded, but I can use if:elif

schematic/schemas/data_model_parser.py
mialy-defelice: Thanks for catching this, I removed my references to it at some point.

schematic/schemas/data_model_parser.py
mialy-defelice: Extracted parsing to its own function so it reads a bit nicer.

schematic/schemas/data_model_parser.py
mialy-defelice: Thank you fo catching this!

schematic/schemas/data_model_jsonld.py
mialy-defelice: Can you elaborate a bit more?

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
...
+    def clean_template(self, template: dict, data_model_relationships: dict) -> dict:
+        '''Get rid of empty k:v pairs. Fill with a default if specified in the relationships dictionary.
+        Args:
+            template, dict: JSONLD template for a single entry, keys specified in property and class templates.
+            data_model_relationships, dict: dictionary containing information for each relationship type supported.
+        Returns:
+            template: JSONLD template where unfilled entries have been removed, or filled with default depending on specifications in the relationships dictionary.
+        '''
+        for rels in data_model_relationships.values():
+            if rels['jsonld_key'] in template.keys() and not template[rels['jsonld_key']]:
```
linglp, line 163: here, I found `template[rels['jsonld_key']]` difficult to understand. Could you do something like: 
```
XXX = template[rels['jsonld_key']]
```
and clarify what is  `template[rels['jsonld_key']]` in a comment? 

Thank you!

schematic/schemas/data_model_jsonld.py
andrewelamb: There's a template that gets referred to several times in this class, for example a base template gets created in `base_jsonld_template`. It might make sense for that template to be its own class `Template` and when it's created with nor parameters it could have the same values as the one created by `base_jsonld_template`

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
...
+                # Order edges
+                sorted_edges = self.DME.get_ordered_entry(key=key, source_node_label=template_label)
+                edge_weights_dict={edge:i for i, edge in enumerate(sorted_edges)}
+                ordered_edges = [0]*len(edge_weights_dict.keys())
+                for k,v in edge_weights_dict.items():
+                    ordered_edges[v] = {'@id': 'bts:' + k}
+                
+                # Throw an error if ordered_edges does not get fully filled as expected.
+                if 0 in ordered_edges:
+                    logger.error("There was an issue getting values to match order specified in the data model, please submit a help request.")
```
linglp, line 214: I am seeing an error related to `logger.error` in my vs code. I think you might forget to import `logger`

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
...
+        Returns:
+            template, dict: list entries re-ordered to match user supplied order.
+        Note:
+            User order only matters for nodes that are also attributes
+        '''
+        template_label = template['rdfs:label']
+
+        for jsonld_key, entry in template.items():
+            # Make sure dealing with an edge relationship:
+            is_edge = ['True' for k, v in self.rel_dict.items() if v['jsonld_key']==jsonld_key if v['edge_rel'] == True]
```
linglp, line 198: Nit: could you also update the `k` and `v` here so that they are clearer?

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
...
+            
+            #if the entry is of type list and theres more than one value in the list attempt to reorder
+            if is_edge and isinstance(entry, list) and len(entry)>1:
+                # Get edge key from data_model_relationships using the jsonld_key:
+                key, edge_key = [(k, v['edge_key']) for k, v in self.rel_dict.items() if jsonld_key == v['jsonld_key']][0]
```
linglp, line 203: also similarly, could you update `key` here?

schematic/schemas/data_model_jsonld.py
```diff
@@ -0,0 +1,284 @@
...
+                
+                # Order edges
+                sorted_edges = self.DME.get_ordered_entry(key=key, source_node_label=template_label)
+                edge_weights_dict={edge:i for i, edge in enumerate(sorted_edges)}
+                ordered_edges = [0]*len(edge_weights_dict.keys())
+                for k,v in edge_weights_dict.items():
```
linglp, line 209: and also here, the `k` and `v` :) thank you!!

schematic/schemas/data_model_jsonld.py
linglp, line 198: Here, `is_edge` returns either [True] or `[]` right? so does that mean `if if_edge` will always be True? I don't think that's what you intended?

tests/test_schemas.py
```diff
@@ -1,294 +1,645 @@
...
+        # Get Graph
+        graph_data_model = generate_graph_data_model(helpers, data_model_name=data_model)
+
+        # Instantiate DataModelJsonLD
+        data_model_jsonld = DataModelJsonLD(Graph=graph_data_model)
+
+        # Test that __init__ is being set up properly 
+        assert type(data_model_jsonld.graph) == nx.MultiDiGraph
+        assert type(data_model_jsonld.rel_dict) == dict
+        assert 'required' in data_model_jsonld.rel_dict
```
linglp, line 415: why do we need to check if 'required' is in data_model_jsonld.rel_dict ?
--Migrated from: https://github.com/Sage-Bionetworks/schematic/pull/1313

tests/test_schemas.py
```diff
@@ -1,294 +1,645 @@
...
+
+            # Check contents of validation schema
+            assert 'Diagnosis' in json_validation_schema['properties']
+            assert 'Cancer' in json_validation_schema['properties']['Diagnosis']['enum']
+        except:
+            # Should only fail if no source node is provided.
+            assert source_node == ''
+
+class TestDataModelJsonLd:
+    @pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
```
linglp, line 404: since all the tests require `@pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))`, why not move it to the class level? so something like: 
```
@pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
class TestDataModelJsonLd:
```

tests/test_schemas.py
```diff
@@ -1,294 +1,645 @@
...
+    def test_create_object(self, helpers, data_model, template_type, node):
+        # Get Graph
+        graph_data_model = generate_graph_data_model(helpers, data_model_name=data_model)
+
+        # Instantiate DataModelJsonLD
+        data_model_jsonld = DataModelJsonLD(Graph=graph_data_model)
+
+        # Get empty template
+        if template_type == 'property':
+            template = data_model_jsonld.property_template()
```
linglp, line 447: Is checking all the expected keys got returned sufficient? Or should we also check the values of the key to make sure that they are expected too?

Migrated from: https://github.com/Sage-Bionetworks/schematic/pull/1313

tests/test_schemas.py
```diff
@@ -1,294 +1,645 @@
...
+            actual_keys = list(object_template.keys())
+            if template_type == 'property':
+                expected_keys = ['@id', '@type', 'rdfs:comment', 'rdfs:label', 'schema:isPartOf', 'sms:displayName', 'sms:required', 'sms:validationRules']
+            elif template_type == 'class':
+                expected_keys = ['@id', '@type', 'rdfs:comment', 'rdfs:label', 'rdfs:subClassOf', 'schema:isPartOf', 'sms:displayName', 'sms:required', 'sms:requiresDependency', 'sms:validationRules']
+            assert (set(actual_keys) - set(expected_keys)) == (set(expected_keys) - set(actual_keys))
+
+    @pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
+    @pytest.mark.parametrize("template_type", ['property', 'class'], ids=['property', 'class'])
+    def test_add_contexts_to_entries(self, helpers, data_model, template_type):
```
linglp, line 470: why this test is for testing `add_contexts_to_entries`, but `add_contexts_to_entries` didn't get called?

tests/test_schemas.py
```diff
@@ -1,294 +1,645 @@
...
+        # TODO: This will need to change with contexts bc they are hard coded here.
+        # Get Graph
+        graph_data_model = generate_graph_data_model(helpers, data_model_name=data_model)
+
+        # Instantiate DataModelJsonLD
+        data_model_jsonld = DataModelJsonLD(Graph=graph_data_model)
+
+        # Get empty template
+        template = data_model_jsonld.class_template()
+        assert 'sms:requiresDependency' in template
```
linglp, line 503: Nit: When I was first reading the test, i was a bit confused why here only `sms:requiresDependency` got tested in assert, but it turned out that other keys such as `sms:requiresComponent` and `schema:rangeIncludes` got deleted as well. Maybe it will be worth to clarify this in a comment. 

But in general, i like how this test was written. I think it captures everything that needs to be tested in the `clean_template` function 

--Migrated from https://github.com/Sage-Bionetworks/schematic/pull/1313

tests/test_schemas.py
```diff
@@ -1,294 +1,645 @@
...
+        assert type(data_model_jsonld.DME) == DataModelGraphExplorer
+        assert data_model_jsonld.output_path == ''
+
+    @pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
+    def test_base_jsonld_template(self, helpers, data_model):
+        # Get Graph
+        graph_data_model = generate_graph_data_model(helpers, data_model_name=data_model)
+
+        # Instantiate DataModelJsonLD
+        data_model_jsonld = DataModelJsonLD(Graph=graph_data_model)
```
linglp, line 425: I saw this line also got called multiple times in the tests. Could you organize it as a fixture instead. Something like: 
```
class TestDataModelJsonLd:
    @pytest.fixture
    def data_model_jsonld(self, helpers, data_model) -> XX:
        """some comments"""
        graph_data_model = generate_graph_data_model(helpers, data_model_name=data_model)
        data_model_jsonld = DataModelJsonLD(Graph=graph_data_model)
        return data_model_jsonld
```

tests/test_schemas.py
```diff
@@ -1,294 +1,645 @@
...
+
+        # Get empty template
+        template = data_model_jsonld.class_template()
+
+        # Fill out template with some 'Sex' information
+        template['@id'] = 'Sex'
```
linglp, line 552: I think this is a good test too. But at the same time, I also wonder what if the 'id' that you enter here is not in the schema, then what would be the expected response. Should we consider adding that in the test too?

schematic/schemas/data_model_jsonld.py
mialy-defelice: Okay, I think the move to data classes might have addressed this, but let me know if you were thinking of something different.

schematic/schemas/data_model_jsonld.py
mialy-defelice: Since template is mutable, I don't want to make another copy when not necessary, but I added additional docstrings, and converted `rels['jsonld_key']` to a variable that is more descriptive.

schematic/schemas/data_model_graph.py
```diff
@@ -0,0 +1,705 @@
...
+
+    def find_adjacent_child_classes(self, node_label: Optional[str]=None, node_display_name: Optional[str]=None)->List[str]:
+        '''Find child classes of a given node.
+        Args:
+            node_display_name: Display name of the node to look up.
+            node_label: Label of the node to look up.
+        Returns:
+            List of nodes that are adjacent to the given node, by SubclassOf relationship.
+        '''
+        return self.get_adjacent_nodes_by_relationship(node_label = schema_class, relationship = self.rel_dict['subClassOf']['edge_key'])
```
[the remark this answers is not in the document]
mialy-defelice: A lot of these discrepancies actually come from me updating old names from things like schema_class, class_label etc to node_label but forgetting to update other parts :( they arent tested and often arent used so they slipped through

schematic/schemas/data_model_graph.py
```diff
@@ -0,0 +1,705 @@
...
+
+    def find_child_classes(self, schema_class: str) -> list:
+        """Find schema classes that inherit from the given class
+        Args:
+            schema_class: node label for the class to from which to look for children.
+        Returns:
+            list of children to the schema_class.
+        """
+        return unlist(list(self.graph.successors(schema_class)))
```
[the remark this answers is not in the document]
mialy-defelice: Its imported at the top

schematic/schemas/data_model_graph.py
```diff
@@ -0,0 +1,705 @@
...
+
+    def is_class_in_schema(self, node_label: str) -> bool:
+        """Determine if provided node_label is in the schema graph/data model.
+        Args:
+            class_label: label of node to search for in the
+        Returns:
+            True, if node is in the graph schema
+            False, if node is not in graph schema
+        """
+        if node_label in self.graph.nodes():
```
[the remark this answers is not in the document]
mialy-defelice: no bc if node labels isnt in the graph itll throw an error.

schematic/schemas/data_model_graph.py
```diff
@@ -0,0 +1,705 @@
...
+                    edges.append((_path[i], _path[i + 1]))
+            return visualize(edges, size=size)
+        elif direction == "both":
+            paths = self.find_parent_classes(source)
+            edges = list(nx.edge_bfs(self.graph, [source]))
+            for _path in paths:
+                _path.append(source)
+                for i in range(0, len(_path) - 1):
+                    edges.append((_path[i], _path[i + 1]))
+            return visualize(edges, size=size)
```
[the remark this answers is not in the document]
mialy-defelice: I actually dont know who uses this. I'll put optional for now. I just copied this function over.

schematic/schemas/data_model_nodes.py
```diff
@@ -0,0 +1,273 @@
...
+            rel_node_dict, dict: node_attr_dict, from relationships dictionary for a given relationship
+        TODO: Move to data_model_relationships.
+        """
+        for k, v in self.data_model_relationships.relationships_dictionary.items():
+            if k == relationship:
+                if "node_attr_dict" in v.keys():
+                    rel_key = v["node_label"]
+                    rel_node_dict = v["node_attr_dict"]
+                    return rel_key, rel_node_dict
+
```
andrewelamb, line 93: I'm a little confused about this method. Is 
```
for k, v in self.data_model_relationships.relationships_dictionary.items():
            if k == relationship:
```
the same as `self.data_model_relationships.relationships_dictionary[relationship]`, I'm wondering why you need to iterate through the whole dict instead of treating using the key to get the value out of the dict as normal?

andrewelamb: I'd write this as 

```
value =  self.data_model_relationships.relationships_dictionary.get(relationship)
if value:
    value2 = value.get("node_attr_dict")
    if value2:
         rel_key = v["node_label"]
         rel_node_dict = v["node_attr_dict"]
         return rel_key, rel_node_dict
```

Unless I'm misunderstanding what the method is doing

mialy-defelice: I think that could probably work. I tend to get a little nervous about key errors so add additional checks but in this case for making a node_dict I think we can assume those keys are present.

andrewelamb: Yep, that's why I'm using `dict.get(relationship)`. This safely gets the value in the dict associated with the key. If they key exists it returns the value, if it doesn't then it returns None. Therefore there's no Key error ever.

https://www.w3schools.com/python/ref_dictionary_get.asp

mialy-defelice:
Running tests, saw the same errors as on Dev.
https://github.com/Sage-Bionetworks/schematic/pull/1317#issuecomment-1791323052
```
FAILED tests/test_api.py::TestSynapseStorage::test_get_dataset_files[None-False] - AssertionError: assert (['syn25705259', 'Boolean Test'] and ['syn23667202', 'Dat...
FAILED tests/test_api.py::TestManifestOperation::test_generate_new_manifest[data_type1-None] - assert 500 == 200
FAILED tests/test_api.py::TestSchemaVisualization::test_visualize_component[Patient-Component,Component,TBD,False,,,,Patient] - assert 'Component,Component,TBD,False,,,,Patient' in 'Attribute,Label,Descriptio...
FAILED tests/test_api.py::TestSchemaVisualization::test_visualize_component[BulkRNA-seqAssay-Component,Component,TBD,False,,,,BulkRNA-seqAssay] - assert 'Component,Component,TBD,False,,,,BulkRNA-seqAssay' in 'Attribute,Label,D...
FAILED tests/test_api.py::TestValidationBenchmark::test_validation_performance[Check String] - assert 5.464691235916689 < 5.0
FAILED tests/test_api.py::TestValidationBenchmark::test_validation_performance[Check URL] - assert 61.93822409585118 < 5.0
FAILED tests/test_api.py::TestValidationBenchmark::test_validation_performance[Check Match at Least] - assert 11.73978321091272 < 5.0
FAILED tests/test_api.py::TestValidationBenchmark::test_validation_performance[Check Match at Least values] - assert 10.318522033980116 < 5.0
FAILED tests/test_api.py::TestValidationBenchmark::test_validation_performance[Check Match Exactly] - assert 10.6404055331368 < 5.0
FAILED tests/test_api.py::TestValidationBenchmark::test_validation_performance[Check Match Exactly values] - assert 10.578840872971341 < 5.0
```

mialy-defelice:
@mieko @AmyHeiser All API and CLI checks are complete.  Nothing new was observed that was not already noted for v23.11.1. I created tickets.

mialy-defelice:
Closing this PR as the refactor schemas branch was officially pushed into dev with PR #1350

PR #842, opened 2022-07-26
Develop NA string behavior

GiaJordan:
Change `pd.read_csv` parameters for manifests to not cast strings to Null values when possible. Fixes string `NA` values from being uploaded to synapse as blanks and type `double`.
Behavior for data models is unchanged


GiaJordan:
@milen-sage, @mialy-defelice and I wanted your input, we were concerned about other groups entering `NA` values and expecting them to convert to blanks. Should we make the old behavior the default and only allow this as an option?

milen-sage (approved):

Issue #675, opened 2022-04-25
Single Rule Argument Validation Error


GiaJordan:
**Describe the bug**
Validation rules that take a single argument are not validated correctly during data model conversion to JSONLD. If the one argument is not included in the data model rule definition, an error will not be raised like it would if there were two arguments for the rule.

**To Reproduce**

1. Create a data model and specify a rule that only requires one argument, but do not include the argument. eg. `protectAges` instead of  `protectAges error`
2. Convert the data model
3. The model is successfully converted

**Expected behavior**
An `incorrect_num_args` error should be raised in the process of data model conversion to JSONLD and the model should not convert with errors

**Priority** (select one)
- [x] Minor ⬇️
- [ ] Major 📢
- [ ] Critical 🆘

**Screenshots**

**Desktop (if applicable, please complete the following information):**
 - OS: Windows 10
 - Python: 3.9.0

**Additional context**

PR #843, opened 2022-07-28
Develop optional rule args

GiaJordan:
With the introduction of rules: `list`, `protectAges`, `unique`, and `inRange`, there are now multiple rules that allow arguments but do not require all of them to be specified every time because default behavior is defined. 

The process of declaring novel rules available to the suite has been simplified. The three dictionaries `valid_rule_combinations`, `validation_types` ( in the `validate_rules_utils.py`), and the separate `validation_types` ( in the `validate_manifest.py`) have been combined into the `validation_rule_info` dict that holds a dict of information for each rule containing arguments, type, and complementary rules available for use. New rules can now be declared in a single place, and rule spec updates are contained to one function in one file.

This PR also closes the bug in #675 where rules that required 1 argument and were missing it were not flagged as invalid.

### DIFF
--- a/schematic/models/GE_Helpers.py
+++ b/schematic/models/GE_Helpers.py
@@ -404,7 +404,7 @@ class GreatExpectationsHelpers(object):
                         values.append(item) if isinstance(item,type_dict[observed_type]) else values
 
                 #call functions to generate error messages and add to error list
-                if validation_types[rule.split(" ")[0]]=='type_validation':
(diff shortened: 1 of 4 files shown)


milen-sage (approved):
Tentatively approved. But let's wait until the next release of schematic before merging.

GiaJordan:
@milen-sage will merge after release this week

GiaJordan:
Merging now that release 22.8.1 is live

Issue #25, opened 2019-08-23
Export requirement dependencies and controlled vocabulary value constraints to JSONSchema


milen-sage:
Serialize dependency requirements graph to JSONSchema, so that off-the-shel JSONSchema validator libraries can be used to validate metadata.

#roadmap

PR #844, opened 2022-07-29
Check type of 'url' entry when raising errors

GiaJordan:
Errors arose when non-`str` type entries were input for attributes with the `url` validation rule, because numerical type values could not be parsed as strings were. Ex: [#25](https://github.com/ncihtan/HTAN-data-curator/issues/25) in the [HTAN-data-curator](https://github.com/ncihtan/HTAN-data-curator) repo
Updated the error generation criteria to check type of entries when generating errors instead of assuming type `str`.


mialy-defelice (approved):
Looks good!

Issue #846, opened 2022-07-29
Fix example data model list validation rule


linglp:
**Describe the bug**
When generating a manifest by using the example data model, "familyHistory" which is defined as a "list" in the data model should allow people to enter multiple values. But it is currently showing as "dropdown" and only allows one entry. This might because in the data model, we specify "familyHistory" as a "list", and we should change "list" to either "list strict" or "list like" (since validation rules have now changed)

Similarly for "Check List" attribute in the data model.

In addition, we need to add valid value in the example data model for "Check List" and see if they get pop out when creating an empty manifest

**To Reproduce**

**Expected behavior**

**Priority** (select one)
- [x] Minor ⬇️
- [ ] Major 📢
- [ ] Critical 🆘

**Screenshots**

**Desktop (if applicable, please complete the following information):**

**Additional context**

linglp:
@GiaJordan for your awareness too

GiaJordan:
related to [#76](https://github.com/ncihtan/data-models/issues/76)

linglp:
This is also related to the issue that @afwillia mentioned today for endpoint /validate/manifest. We noticed that the endpoint on `develop` branch is not working properly (an example of error message: "['Breast', 'Lung'] is not one of ['Lung', 'Prostate', 'Colorectal', 'Breast', 'Skin', '']")

PR #847, opened 2022-07-29
Develop fix manifest generator

linglp:
I updated the following: 
* For "Check List", changed it from "list" to "list strict" 
* "check list" now takes the following valid values: `ab, cd, ef, gh`
* For attribute "Check Regex List", change "list::regex match [a-f]" to "list strict::regex match [a-f]" 
* Add an additional attribute in data model: Check Regex Format
* Update invalid_test_manifest.csv and valid_test_manifest.csv
* Update example.model.jsonld
 
In addition, I also updated the ManifestGenerator class so that something like `elif "list" in validation_rules and values` gets changed to `elif ("list strict" in validation_rules) or ("list like" in validation_rules) and valid_values`

When I created the manifest based on the example data model and opened it in Firefox (see an example here: https://docs.google.com/spreadsheets/d/1Ax_fb-TdQK67i9B60rncjAffVYJ8_hNKpguXBuIEH3k/edit#gid=0), I could now see a list of valid values for "Check List" in sheet 2. 

> But when I tried to enter a comma separated list under column "Check List" in sheet 1, I still got an error message related to the data validation rules. Should we update the "Data validation" rule as "On invalid data: show warning" instead of "Reject input" so that a comma separate list could show up?

^^ Note: In my code, for list in validation rule that allows multi-selection, I set "strict" to "False" so that users could enter multiple values. See here: https://github.com/Sage-Bionetworks/schematic/pull/847/files#:~:text=%3DNone%2C-,strict%3DFalse%2C,-custom_ui%3DFalse

> If we can't change all the "list" to "list like"/"list strict" in all existing data models, should we still maintain the logic related to "list"?

^^ Note: all the old logic related to "list" in schemaGenerator was handled by branch `develop-listlike-behavior`. After merging that branch with this branch here, there seems to be no more test errors.

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -250,7 +250,16 @@ class TestManifestValidation:
             attribute_name = 'Check Regex Single',
             module_to_call = 'search',
             invalid_entry = 'q'
-            ) in errors   
+            ) in errors 
+
+        assert GenerateError.generate_regex_error(
+            val_rule = 'regex',
+            reg_expression = '[a-f]',
+            row_num = '3',
+            attribute_name = 'Check Regex Format',
+            module_to_call = 'match',
+            invalid_entry = 'm'
+            ) in errors     
 
         assert GenerateError.generate_url_error(
             url = 'http://googlef.com/',
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -104,6 +104,15 @@ class TestManifestValidation:
             val_rule = 'regex',
             reg_expression = '[a-f]',
             row_num = '3',
+            attribute_name = 'Check Regex Format',
+            module_to_call = 'match',
+            invalid_entry = 'm'
+            ) in errors   
+
+        assert GenerateError.generate_regex_error(
+            val_rule = 'regex',
+            reg_expression = '[a-f]',
+            row_num = '3',
             attribute_name = 'Check Regex Single',
             module_to_call = 'search',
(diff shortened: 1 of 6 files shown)


linglp:
For testing, here's my test manifest: https://docs.google.com/spreadsheets/d/1DwWA_f6kYwPkLlM0jpHuTAyltk8CuWS9nyxrGBcq8oY/edit#gid=0

*Note*: 
* "Check List" could now accept a list of values (separate by commas)
* If the value that you enter in "Check Regex Format" does not match [a-f], it will reject your entry. Otherwise, you would see the text turns red and then turn black (which is expected based on the formatting rules set in manifest generator)

linglp:
This PR is related to #846

GiaJordan:
@GiaJordan I'll update the documentation for google sheets limitations with regex matching in combo with list rule

tests/test_validation.py
```diff
@@ -100,6 +100,15 @@ def test_invalid_manifest(self,helpers,sg,metadataModel):
             invalid_entry = 'ab cd ef'
             ) in errors
 
+        assert GenerateError.generate_regex_error(
```
GiaJordan, line 103: Can you add this assertion to `test_in_house_validation` as well?

GiaJordan (changes requested):

GiaJordan (approved):
🚀

Issue #733, opened 2022-05-17
Blank values are uploaded as annotations


allaway:
**Describe the bug**
This may be related to #402 / #579 . Blank values in manifests are being uploaded to synapse annotations (see screenshot) even when hideBlanks = T. 

**To Reproduce**
1. Uploade a file to https://www.synapse.org/#!Synapse:syn22410515
2. Go to DCA (https://sagebio.shinyapps.io/NF_data_curator-staging/)
3. Select "Neurofibromatosis Demo" project
4. Select "RNAseq" folder
5. Generate a manifest of any type
6. Minimally annotate with Component, dataType, resourceType
7. Validate, submit
8. View annotations (see screenshot)

**Expected behavior**
when hideBlanks=T (https://github.com/nf-osi/NF_data_curator/blob/104c5cf1d392e897a2cda7ffa5f10bf0d63e1f66/server.R#L447), these annotations should not be uploaded. 

**Priority** (select one)
- [ ] Minor ⬇️
- [X] Major 📢
- [ ] Critical 🆘

**Screenshots**

<img width="326" alt="Screen Shot 2022-05-17 at 12 40 21 PM" src="https://user-images.githubusercontent.com/20408008/168896466-091260ad-1828-47dc-be14-cd3acb292d51.png">

**Additional context**

milen-sage:
It looks like we can have a parameter flag that determines whether blanks are uploaded as annotations or not. In some cases the expectation is that blanks will be uploaded as annotations (e.g empty strings); and some where blank annotations should not be submitted. We can expose that as an option with a backward compatible default (i.e. blank annotations are uploaded).

Issue #710, opened 2022-05-06
list validation values not being uploaded to synapse as stringList values


allaway:
**Describe the bug**
Values for attributes with list validation requirements that are uploaded to synapse are uploaded as strings, instead of split and uploaded as stringlists. 

**To Reproduce**
1. Install schematic (5eca846740696ec5b882edeb70cf5d29bd20be49). 
2. Create a manifest (e.g. https://docs.google.com/spreadsheets/d/1uMT1kRafUS7WS3-X_vqyxJH9qDXgknWEqw35DLUMLKU/edit#gid=0) that has comma separated values for a `list` validation attribute, such as cellType, here. 
3. Upload to DCA. 
4. Validate (should pass)
5. Submit to synapse. 

**Expected behavior**
The cellType column should be uploaded as stringlisted values/arrays and not as plain strings (see screenshot).

**Priority** (select one)
- [ ] Minor ⬇️
- [ ] Major 📢
- [X] Critical 🆘 This is pretty critical for us as it is blocking our use of the DCA and we are getting manifests emailed to us... :) 

**Screenshots**
<img width="1722" alt="Screen Shot 2022-05-06 at 2 56 13 PM" src="https://user-images.githubusercontent.com/20408008/167221939-f88f2f4f-09c2-4420-8dcc-9843a8273681.png">

ychae:
@allaway Would you be able to let me know if starting this ticket at our next sprint (Tuesday May 10) would be quick enough or does it need to be addressed sooner?

allaway:
I don't imagine we'll get a ton of use this weekend. If it would be possible and it's simple enough of an issue to fix it on Monday that would be great (recognizing that I'm reporting this at 3pm on a friday PT... 😆 ), but if it's a complicated issue starting work on tuesday makes sense to me!

Issue #402, opened 2021-02-06
Do not upload "nan" for blank values in manifest


allaway:
**Describe the bug**
This is related to #401 but distinct - blank values are currently uploaded as `nan`. This can cause issues with adding annotations in the future because synapse interprets nan as a `DOUBLE`. 

**To Reproduce**
1. Leave a field blank that is not required
2. Validate and upload manifest
3. Annotation on synapse is `nan` for the blank field. 

**Expected behavior**
I would expect no annotation to be uploaded for a blank field.

allaway:
I was just wondering if there was an updated timeline for this? I tried to tackle this a couple weeks back but failed miserably. 

At the moment, I'm manually cleaning up all of the `nan`s as they come in, which is a bit of a headache.

ychae:
Subset of #441

milen-sage:
@GiaJordan another related issue! We'd close a lot by resolving #441 :)

PR #862, opened 2022-08-12
Develop blank annotation behavior

GiaJordan:
Addresses #733, which is related to #402 that was previously addressed by #579. 
This PR extends the functionality of #579, which was previously contained to the handling of `NA` or `NULL` values.
Now, `hide_blanks` will prevent both empty string values and `NULL` values from being uploaded as annotations.

Also, out of necessity and in response to [#173](https://github.com/nf-osi/nf-metadata-dictionary/issues/173), an additional test was added that submits a manifest to annotate a dataset and verifies that the annotations for the matching entity on synapse are correct. Currently the test assertions are written to cover issues raised with `int` annotations (#664), `list` annotations (#710), and empty string / `NULL` annotations (#733 / #402)

Also fixes a bug with cross manifest validation when trying to join a manifest with only one element.

### DIFF
--- a/tests/test_store.py
+++ b/tests/test_store.py
@@ -3,6 +3,8 @@ import os
 import math
 import logging
 import pytest
+import platform
+import shutil
 
 import pandas as pd
 from synapseclient import EntityViewSchema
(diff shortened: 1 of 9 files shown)


milen-sage (approved):

tests/test_store.py
```diff
@@ -76,6 +82,47 @@ def test_getFileAnnotations(self, synapse_store):
 
         assert expected_dict == actual_dict
 
+    def test_annotation_submission(self, synapse_store, helpers, config):
+        # Duplicate base file to avoid conflicts
+        version=platform.python_version()
+        manifest_path = helpers.get_data_path("mock_manifests/annotations_test_manifest.csv")
+        temp_manifest_path = manifest_path.replace('.csv',version+'.csv')
+        shutil.copyfile(manifest_path,temp_manifest_path)
+
```
milen-sage, line 91: This may be useful for all test functions that write to manifests.

PR #864, opened 2022-08-17
update documentation for /manifest/download endpoint

linglp:
add documentation related to how to use /manifest/download endpoint in swagger UI. 

The current (new) behavior requires users to also update `manifest_basename` key in `config.yml` with the name of the manifest that users want to download. 

Question for discussion: what if user want to download different manifests with different names? Should we also allow users to enter multiple names of manifest?


linglp:
No longer relevant. We will need to investigate the naming convention of manifest.

Issue #874, opened 2022-08-23
Component not present error when trying to submit manifest 


linglp:
**Describe the bug**

**To Reproduce**
1. Run the following command to validate manifest: `schematic model -c config.yml validate -mp /Users/lpeng/Documents/test-schematic/CA184897_1.csv -dt DatasetView `
I could see that the message here: Your manifest has been validated successfully. There are no errors in your manifest, and it can be submitted without any modifications.

2. Try to submit the manifest by using the following command: `schematic model -c config.yml submit -mp /Users/lpeng/Documents/test-schematic/CA184897_1.csv -d syn33691763 -vc DatasetView -mrt table` Here, I could see error message: error: Component 'DatasetView' is not present in 'tests/data/CSBCDataModel_new.jsonld', or is invalid.

A few notes: 
1) The error message here is confusing. It should raise a different error message to indicate a different error. Please note that this error was not triggered by `self.sg.se.is_class_in_schema(validate_component)` in [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/models/metadata.py#:~:text=self.sg.se.is_class_in_schema(validate_component))
2) I think the error was raised because `getDatasetProject` function from `upload_format_manifest_table` [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#:~:text=schema%20%3D%20Schema(name%3Dtable_name%2C%20columns%3Dcol_schema%2C%20parent%3Dself.getDatasetProject(datasetId))) is trying to see if the dataset ID that we are feeding is a Project. Since `syn33691763` is only a folder, `self.getDatasetProject(datasetId)` is returning nothing. (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#:~:text=already%20project%20itself-,try%3A,return%20datasetId,-except%20SynapseHTTPError%3A)) We might need to add an additional parameter to let users feed project ID if they want to submit a table. 

**Expected behavior**

**Priority** (select one)
- [ ] Minor ⬇️
- [X] Major 📢
- [ ] Critical 🆘

**Screenshots**

**Desktop (if applicable, please complete the following information):**

**Additional context**

linglp:
[...]

PR #875, opened 2022-08-23
add project id parameter for user to submit manifest as a table

linglp:
This PR is related to the issue [here](https://github.com/Sage-Bionetworks/schematic/issues/874). I added a parameter called `Project id`. When submitting manifest as a table, users would need to specify that to successfully upload a manifest.


linglp:
Note: `upload_project_manifests_to_synapse` function in synapse.py also uses `upload_format_manifest_table` function. But `upload_project_manifests_to_synapse` is not called else where in the code.

linglp:
No longer relevant.

PR #879, opened 2022-08-25
Develop manifest submission error handle

linglp:
This issue is related to the issue #874  . Here are the few changes: 

* When a dataset ID is not present in the master file view, I changed it to raise `AttributeError` instead of `ValueError`. 
* If user has no permission to upload it to a given folder, I changed the error type from "ValueError" to "PermissionError"
* I also updated `command.py` to catch AttributeError being raised 
* In addition, I added "asset view" parameter to swagger UI. Users could update their asset view/file view when using the submit endpoint as well.

### DIFF
--- a/schematic/models/commands.py
+++ b/schematic/models/commands.py
@@ -147,6 +147,10 @@ def submit_manifest(
         logger.error(
             f"Validation errors resulted while validating with '{validate_component}'."
         )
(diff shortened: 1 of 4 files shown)


GiaJordan (approved):

GiaJordan:
🚀

Issue #865, opened 2022-08-17
Issue with intermediate JSON schema file storage


PozhidayevaDarya:
**Describe the bug**
The issue occurs when generating a google sheet template which points schematic to a json-ld model. 
This may be related to improper intermediate JSON schema file storage as the error dead-ends in FileNotFoundError: [Errno 2] No such file or directory.' More detail is shown below.

**To Reproduce**
1. Follow the installation guide [here](https://github.com/Sage-Bionetworks/schematic#installation-guide-for-developerscontributors).
2. export the HTAN model google sheet to csv
3. run [this schematic command](https://sage-schematic.readthedocs.io/en/develop/cli_reference.html#schematic-schema-convert) to generate a json-ld file from the csv
4. run [this schematic command](https://sage-schematic.readthedocs.io/en/develop/cli_reference.html#schematic-manifest-get) to generate a manifest: e.g. schematic manifest -c config-htan.yml get -t BiospecimenTest -dt Biospecimen --jsonld 'HTAN schema.jsonld' -s -oa 
5.  See error

`(synapse_env) (base) darya@tigerlily synapse_env % schematic manifest -c load_metadata/config-htan.yml get -t BiospecimenTest -dt Biospecimen --jsonld 'HTAN schema v22.02 - latest.jsonld' -s -oa  
Starting schematic...
URL being requested: POST https://www.googleapis.com/drive/v3/files/1LYS5qE4nV9jzcYw5sXwCza25slDfRA1CIg3cs-hCdpU/copy?alt=json
JSON schema successfully generated from schema.org schema!
Traceback (most recent call last):
  File "/Users/darya/Documents/synapse_env/bin/schematic", line 8, in <module>
    sys.exit(main())
  File "/Users/darya/Documents/synapse_env/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/darya/Documents/synapse_env/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/darya/Documents/synapse_env/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/darya/Documents/synapse_env/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #881, opened 2022-08-25
do not create dir if located in cwd

GiaJordan:
Issue #865 arose when a .jsonld file was used from the users current directory. A safeguard in place created the directory specified in the path to the .jsonld, but when it was in the current directory, this field was `''`. 
This PR no longer tries to create the directory containing the .jsonld if it is located in the current directory.

### DIFF
--- a/schematic/schemas/generator.py
+++ b/schematic/schemas/generator.py
@@ -708,7 +708,9 @@ class SchemaGenerator(object):
                 "nested key in the configuration: (model > input > log_location)."
             )
         else:
-            os.makedirs(os.path.dirname(json_schema_log_file), exist_ok=True)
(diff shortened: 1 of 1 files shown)


linglp:
Note: I could reproduce the issue by doing: `schematic % schematic manifest -c config.yml get -t Biospecimen -dt Biospecimen --jsonld 'example.model.jsonld' -s -oa`. The error message: FileNotFoundError: [Errno 2] No such file or directory: ''

linglp (approved):
Thanks. I have tested it out and made sure that it works for the example data model by running: `schematic manifest -c config.yml get -t Biospecimen -dt Biospecimen --jsonld 'example.model.jsonld' -s -oa`

PR #887, opened 2022-08-31
Add support for Schematic API to Dockerfile

BrunoGrandePhD:
Since Schematic has a REST API, it makes sense to make that accessible via the container. In this PR, I tweak the container's entrypoint to allow the `run_api.py` script to be run. I had to tweak the file paths and host address to make this work. I also make a change to the dependency installation of the Dockerfile that will speed things up significantly by leveraging cached layers. Lastly, I added a docker-compose YAML file so it's easy to launch Schematic using the current configuration in your repository (including your credentials). You just need to run the following command and visit http://127.0.0.1:3001/v1/ui/. 

```console
docker compose up
```

The above command is equivalent to (although, I still omitted some unimportant bits for brevity):

```console
docker build -t schematic .
docker run --rm -p 3001:3001 -v $(pwd):/work -w /work --name schematic -e SCHEMATIC_CONFIG=/work/config.yml schematic python /usr/src/app/run_api.py
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,9 +17,12 @@ RUN apt-get update -qqy \
 
 RUN pip install --no-cache-dir "poetry==$POETRY_VERSION"
 
+COPY pyproject.toml poetry.lock ./
+
+RUN poetry config virtualenvs.create false
+
+RUN poetry install --no-interaction --no-ansi --no-root
+
 COPY . ./
 
-RUN poetry config virtualenvs.create false \
-  && poetry install --no-interaction --no-ansi
-
-ENTRYPOINT [ "schematic" ]
+RUN poetry install --no-interaction --no-ansi --only-root
--- a/api/__init__.py
+++ b/api/__init__.py
@@ -13,9 +13,9 @@ def create_app():
     app = connexionapp.app
 
     # path to config.yml file saved as a Flask config variable
-    app.config["SCHEMATIC_CONFIG"] = os.path.abspath(
-        os.path.join(__file__, "../../config.yml")
-    )
+    default_config = os.path.abspath(os.path.join(__file__, "../../config.yml"))
+    schematic_config = os.environ.get("SCHEMATIC_CONFIG", default_config)
+    app.config["SCHEMATIC_CONFIG"] = schematic_config
 
     # Configure flask app
     # app.config[] = schematic[]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,13 +3,14 @@ services:
   schematic:
     build: .
     container_name: schematic
-    entrypoint: python run_api.py
+    entrypoint: python /usr/src/app/run_api.py
     ports:
       - "3001:3001"
     volumes:
-      - .:/work
-    working_dir: /work
+      - .:/schematic
+    working_dir: /schematic
     environment:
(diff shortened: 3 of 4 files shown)


afwillia:
@BrunoGrandePhD awesome, I was able to get this running with an existing Data Curator/Schematic configuration with the following modification 
```
docker run --rm -p 3001:3001 \
  -v $(pwd):/schematic -w /schematic --name schematic \
  -e SCHEMATIC_CONFIG=/schematic/schematic_config.yml \
  sagebionetworks/schematic:commit-d833dd3 \
  python /usr/src/app/run_api.py
```

However, I encountered the following error trying to validate [this Biospecimen manifest](https://github.com/Sage-Bionetworks/schematic/files/9463173/Example.Biospecimen.Tier.1.2.-.Sheet1.csv) with the [example data model](https://raw.githubusercontent.com/Sage-Bionetworks/schematic/develop/tests/data/example.model.jsonld) using the Swagger UI for /model/validate

```
great_expectations.exceptions.exceptions.ConfigNotFoundError: Error: No great_expectations directory was found here!
    - Please check that you are in the correct directory or have specified the correct directory.
    - If you have never run Great Expectations in this project, please run `great_expectations init` to get started.
```

GiaJordan:
@BrunoGrandePhD would changing the entrypoint also change the directory structure? The directory referenced in @afwillia's error is the [`great_expectations`](https://github.com/Sage-Bionetworks/schematic/tree/develop/great_expectations) directory in schematic's main directory and it just seems like it's in a different relative path now. Alternatively, adding the command `great_expectations init` after the poetry install might be another way to resolve the issue

BrunoGrandePhD:
The entrypoint isn't expected to change the directory structure, but we are changing the working directory (either with `-w` in `docker run` or `working_dir` in the docker-compose file). We could avoid changing the working directory, but that limits the flexibility of the container. 

I expected `great_expectations` to have an option to configure where to look for its configuration. Indeed, I found one in the [source code](https://github.com/great-expectations/great_expectations/blob/341a9a5f238dc2c05be7c2b3c3cc4e42440230c5/great_expectations/data_context/data_context.py#L4255), which involves setting an environment variable called `GE_HOME`. Elsewhere in the [source code](https://github.com/great-expectations/great_expectations/blob/341a9a5f238dc2c05be7c2b3c3cc4e42440230c5/great_expectations/data_context/data_context.py#L4117), they define this as "the 'context root directory' - this is the parent of 'great_expectations' dir". In the case of this container, I believe that would be `/usr/src/app/great_expectations/`. 

@afwillia: Could you try the following command? I've updated the docker-compose file to include `GE_HOME` and to use your preferred working directory name (_i.e._ `/schematic`). 

```console
docker run --rm -p 3001:3001 \
  -v $(pwd):/schematic -w /schematic --name schematic \
  -e SCHEMATIC_CONFIG=/schematic/schematic_config.yml \
  -e GE_HOME=/usr/src/app/great_expectations/ \
  sagebionetworks/schematic:commit-d833dd3 \
  python /usr/src/app/run_api.py
```

BrunoGrandePhD:
Before this diverges too much from `develop`, could I get a review so we can merge? Thanks!

afwillia (approved):
This looks good to me

Issue #870, opened 2022-08-18
AttributeError when trying to submit mock manifest with columns that have data type float and data type integer


linglp:
**Describe the bug**
Error message: 
```
  File "/Users/lpeng/Documents/schematic-git2/schematic/api/routes.py", line 204, in submit_manifest_route
    manifest_id = metadata_model.submit_metadata_manifest(
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/models/metadata.py", line 347, in submit_metadata_manifest
    manifest_id = syn_store.associateMetadataWithFiles(
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/store/synapse.py", line 841, in associateMetadataWithFiles
    manifest_synapse_table_id, manifest, table_manifest = self.upload_format_manifest_table(
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/store/synapse.py", line 638, in upload_format_manifest_table
    col_schema = as_table_columns(table_manifest)
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.9/lib/python3.9/site-packages/synapseclient/table.py", line 429, in as_table_columns
    maxStrLen = df[col].str.len().max()
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.9/lib/python3.9/site-packages/pandas/core/generic.py", line 5478, in __getattr__
    return object.__getattribute__(self, name)
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.9/lib/python3.9/site-packages/pandas/core/accessor.py", line 181, in __get__
    accessor_obj = self._accessor(obj)
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.9/lib/python3.9/site-packages/pandas/core/strings/accessor.py", line 168, in __init__
    self._inferred_dtype = self._validate(data)
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.9/lib/python3.9/site-packages/pandas/core/strings/accessor.py", line 225, in _validate
    raise AttributeError("Can only use .str accessor with string values!")
AttributeError: Can only use .str accessor with string values!
```

**To Reproduce**
1. Go to 'develop'
[...]

PR #902, opened 2022-09-13
HOTFIX: Mixed type column manifest table upload

GiaJordan:
Patch for #870. The types of manifest entries are not longer inferred when being read in before file annotation. This avoids the error raised in Synapse and pandas when mixed-type columns are incorrectly treated as strings.

This PR also includes a fix for error conflicts. When the `AttributeError` was raised within Pandas, it was caught in `commands.py` and an error message referring to not being able to find the datasetID within the fileview was displayed. The catch statement, and exception it is intended to catch have been changed to avoid the conflation of the two exceptions raised in Pandas and Schematic

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -994,7 +994,7 @@ class SynapseStorage(BaseStorage):
             load_args={
                 "dtype":"string",
             }
-            manifest = load_df(metadataManifestPath, preserve_raw_input=False, **load_args)
+            manifest = load_df(metadataManifestPath, preserve_raw_input = True, **load_args) #HOTFIX: set preserve_raw_input to true to allow mixed type cols in table uploads schematic#870
         except FileNotFoundError as err:
             raise FileNotFoundError(
                 f"No manifest file was found at this path: {metadataManifestPath}"
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1277,7 +1277,7 @@ class SynapseStorage(BaseStorage):
             )
 
         # If not, then assume dataset not in file view
-        raise AttributeError (
+        raise LookupError (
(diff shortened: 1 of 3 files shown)


andrewelamb (approved):

schematic/store/synapse.py
```diff
@@ -994,7 +994,7 @@ def associateMetadataWithFiles(
             load_args={
                 "dtype":"string",
             }
-            manifest = load_df(metadataManifestPath, preserve_raw_input=False, **load_args)
+            manifest = load_df(metadataManifestPath, preserve_raw_input = True, **load_args) #HOTFIX: set preserve_raw_input to true to allow mixed type cols in table uploads schematic#870
```
milen-sage, line 997: @GiaJordan in the past we've had this set to False, if I remember right, because setting it to True could potentially cause errors in other parts of the code? Do I remember correctly? Or perhaps we decided to expose the parameter as an option, and we just thought False would be a good backward compatible value, but True is a good working option as well.

linglp: @GiaJordan I am wondering the same as Milen. Also could you add a comment in function `load_df` to clarify parameter `preserve_raw_input`? Thank you!

milen-sage (approved):

schematic/store/synapse.py
milen-sage: Thanks @GiaJordan!

linglp (approved):
Thanks.

schematic/store/synapse.py
GiaJordan: After doing manual tests, this fix doesn't appear to break `int` and `list` annotation uploads @milen-sage

milen-sage:
🚀

Issue #856, opened 2022-08-09
In JIRA: schematic: expose default schema load in config


milen-sage:
Provide a config option for the default schema to load; we can still use biothings as a default for now, but can also move to schema.org or a different schema in the future.

AC: 
- allow empty schema 
- allow a specific schema

milen-sage:
This is a follow up on #708

mialy-defelice:
@milen-sage @MiekoHash what is the priority for this issue for Sprint 3?

milen-sage:
This would be lowest priority for this sprint unless you think it would inform refactor implementation.

PR #903, opened 2022-09-13
Allow schema.org schema to be loaded and used

GiaJordan:
The `schema_org.model.jsonld` schema from schema.org included in schematic was out of date. Trying to load and use the schema raised errors due to the structure of the schema not being supported in the existing code. This PR updates the included schema and updates the method that loads schemas into networkx so that both the `BioThings` and `schema.org` schemas can be loaded and used.

This work will be continued as part of #856 to allow for users to configure which base schema to load.


schematic/utils/schema_utils.py
```diff
@@ -17,6 +17,8 @@ def load_schema_into_networkx(schema):
         # adding nodes to the graph
         node = {}
         for (k, value) in record.items():
+            if isinstance(value,dict) and "@language" in value.keys():
```
milen-sage, line 20: @GiaJordan could you add a short in-code comment on why this line is needed to load schema.org? Otherwise the PR looks good!

milen-sage (approved):

schematic/utils/schema_utils.py
GiaJordan: @milen-sage sure thing!

milen-sage: Perfect - thanks! 🚀

PR #904, opened 2022-09-14
added get_node_range endpoint

### LINKED ISSUE #851: schematic: endpoint for SchemaGenerator.get_node_range()
Endpoint for SchemaGenerator.get_node_range()

• Short-term - 2-4 weeks

andrewelamb:
fixes #851

### DIFF
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -699,7 +699,7 @@ paths:
             type: string
             nullable: false
           description: Node / term for which you need to retrieve the range.
-          example: MolecularEntity
+          example: FamilyHistory
           required: true
         - in: query
           name: return_display_names
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -704,8 +704,10 @@ paths:
         - in: query
           name: return_display_names
           schema:
-            type: bool
+            type: boolean
+            nullable: false
           description: If true returns the display names of the nodes.
+          required: true
       responses:
         "200":
           description: A list of nodes.
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -678,3 +678,38 @@ paths:
           description: Check schematic log. 
       tags:
         - Schema Operation
+
+  /explorer/find_class_specific_properties:
+    get:
+      summary: Find properties specifically associated with a given class
+      description: Find properties specifically associated with a given class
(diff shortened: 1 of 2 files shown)


api/openapi/api.yaml
```diff
@@ -676,5 +676,42 @@ paths:
                     ]
         "500":
           description: Check schematic log. 
+      tags:
+        - Schema Operation
+
+  /explorer/get_node_ranges:
```
linglp, line 682: You are calling `get_node_range` function in `routes.py`? Here it should be `get_node_range`

api/openapi/api.yaml
```diff
@@ -676,5 +676,42 @@ paths:
...
+    get:
+      summary: Get all the valid values that are associated with a node label.
+      description: Get all the valid values that are associated with a node label.
+      operationId: api.routes.get_node_ranges
```
linglp, line 686: Same as above. This one should be: `get_node_range`

api/openapi/api.yaml
```diff
@@ -676,5 +676,42 @@ paths:
...
+            type: string
+            nullable: false
+          description: Node / term for which you need to retrieve the range.
+          example: MolecularEntity
+          required: true
+        - in: query
+          name: return_display_names
+          schema:
+            type: boolean
+            nullable: false
```
linglp, line 708: since `return_display_names` is an optional parameter (and the default is set to True), let's set "required" to False (and `nullable` to true): 

```
        - in: query
          name: return_display_names
          schema:
            type: boolean
          description: If true returns the display names of the nodes.
          required: false
```

linglp (changes requested):

api/openapi/api.yaml
```diff
@@ -676,5 +676,42 @@ paths:
...
+          example: >-
+            https://raw.githubusercontent.com/Sage-Bionetworks/schematic/develop/tests/data/example.model.jsonld
+          required: true
+        - in: query
+          name: node_label
+          schema:
+            type: string
+            nullable: false
+          description: Node / term for which you need to retrieve the range.
+          example: MolecularEntity
```
linglp, line 702: Can we change the example here to `FamilyHistory`? Setting it to `MolecularEntity` does not return anything, and I think `FamilyHistory` is probably easier for user to understand the purpose of this endpoint.

linglp (approved):

PR #905, opened 2022-09-14
added get_property_label endpoint

### LINKED ISSUE #850: schematic: endpoint for SchemaExplorer.get_property_label_from_display_name()
Endpoint for SchemaExplorer.get_property_label_from_display_name()

• Short-term - 2-4 weeks

andrewelamb:
fixes #850

### DIFF
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -678,3 +678,39 @@ paths:
           description: Check schematic log. 
       tags:
         - Schema Operation
+
+  /explorer/get_property_label_from_display_name:
+    get:
+      summary: Converts a given display name string into a proper property label string
+      description: Converts a given display name string into a proper property label string
+      operationId: api.routes.get_property_label_from_display_name
+      parameters:
+        - in: query
+          name: schema_url
+          schema:
+            type: string
+          description: Data Model URL
+          example: >-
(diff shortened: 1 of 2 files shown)


api/openapi/api.yaml
```diff
@@ -676,5 +676,41 @@ paths:
...
+          schema:
+            type: string
+            nullable: false
+          description: The display name to be converted
+          example: MolecularEntity
+          required: true
+        - in: query
+          name: strict_camel_case
+          schema:
+            type: bool
```
linglp, line 707: This `type` here has to be set to `boolean`.  Please Make sure that you run `python3 run_api.py` after adding an endpoint. If you can't see swagger UI, that means there's probably a typo in `api.yaml`

api/openapi/api.yaml
```diff
@@ -676,5 +676,41 @@ paths:
...
+            nullable: false
+          description: If true the more strict way of
+            converting to camel case is used.
+      responses:
+        "200":
```
linglp, line 712: Please change this part to something like below: 
```
      responses:
        "200":
          description: The property label of the display name.
        "500":
          description: Check schematic log. 
```

linglp (approved):

PR #906, opened 2022-09-14
added node_dependencies endpoint

### LINKED ISSUE #855: schematic: endpoint for SchemaGenerator.get_node_dependencies()
Endpoint for SchemaGenerator.get_node_dependencies()

• Short-term - 2-4 weeks

andrewelamb:
fixes #855


andrewelamb:
@linglp What would be a good example for this one?

linglp (changes requested):

linglp (approved):

Issue #889, opened 2022-08-31
model/submit is not using manifest_basename to name manifest in synapse


lakikowolfe:
**Describe the bug**
Even though I have specified `"manifest_basename": "data_flow_status_manifest"` the `model/submit` endpoint names the manifest using `synapse_storage_manifest_<component>`.

**To Reproduce**
1. Modify schematic config `manifest_basename` to anything other than "synapse_storage_manifest"
2. Submit manifest
3. View manifest in Synapse

**Expected behavior**
I would expect that the model/submit endpoint defaults to using the `manifest_basename` parameter to name the submitted manifest.

**Priority** (select one)
- [X] Minor ⬇️
- [ ] Major 📢
- [ ] Critical 🆘

**Screenshots**

Below you can see that before I updated my schematic repo the data flow status manifest was uploaded using the `manifest_basename` specified in the config ("data_flow_status_manifest.csv"). After update the uploaded manifest was named following the `synapse_storage_manifest_<component>.csv` structure regardless of what is in my config.

<img width="1229" alt="Screen Shot 2022-08-31 at 4 23 05 PM" src="https://user-images.githubusercontent.com/30699094/187801868-6227f82e-0416-4a70-a14c-e6a7e4dff273.png">

**Desktop (if applicable, please complete the following information):**
 - OS: iOS
 - Browser chrome
 - Dev branch

**Additional context**

linglp:
@lakikowolfe This is actually not a bug. We talked about this in the back-end code review meeting, and the decision is that we want to standardized manifest names by using the base name parameter in config. Users shouldn't change the base name parameter (and even if they do), we would change the name to `synapse_storage_manifest_<component name>`.

linglp:
@lakikowolfe Let me double check then. Thanks for bringing this up.

PR #907, opened 2022-09-15
update upload_manifest_file function to use manifest basename in config

linglp:
This PR is related to the issue [here](https://github.com/Sage-Bionetworks/schematic/issues/889) and also the PR [here](https://github.com/Sage-Bionetworks/schematic/pull/869/files)

Made the change so that we won't override the base name that users defined in `config.yml`

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -762,7 +762,7 @@ class SynapseStorage(BaseStorage):
         # update file name
         file_name_full = metadataManifestPath.split('/')[-1]
         file_extension = file_name_full.split('.')[-1]
-        file_name_new = CONFIG["synapse"]["manifest_basename"] + component_name + '.' + file_extension
+        file_name_new = os.path.basename(CONFIG["synapse"]["manifest_basename"]) + component_name + '.' + file_extension
 
         manifestSynapseFile = File(
             metadataManifestPath,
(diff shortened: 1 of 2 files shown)


schematic/store/synapse.py
```diff
@@ -762,7 +762,7 @@ def uplodad_manifest_file(self, manifest, metadataManifestPath, datasetId, restr
         # update file name
         file_name_full = metadataManifestPath.split('/')[-1]
         file_extension = file_name_full.split('.')[-1]
-        file_name_new = 'synapse_storage_manifest_' + component_name + '.' + file_extension
+        file_name_new = CONFIG["synapse"]["manifest_basename"] + component_name + '.' + file_extension
```
GiaJordan, line 765: Can we add update the reading from the CONFIG to `os.path.basename(CONFIG["synapse"]["manifest_basename"])`? I believe in some cases users could want to store the manifest(s) in a directory other than their working directory and this would allow for that without errors while also matching previous style: #647 (and before). I see now though that my parameter naming in the config could have been more clear to avoid this confusion, apologies

GiaJordan (changes requested):

GiaJordan (approved):

GiaJordan:
:rocket:

linglp:
Close this PR without merging for now. The issue has been address [here](https://github.com/Sage-Bionetworks/schematic/pull/931/files)

Issue #899, opened 2022-09-12
schematic: allow passing manifests as JSON data to the manifest validation REST API endpoint


milen-sage:
Currently schematic requires a csv manifest file to be passed as a parameter for validation. 

In addition, we can enable passing the manifest as JSON, which is then converted to csv. This is a dependency for https://github.com/Sage-Bionetworks/data_curator/issues/399

AC: REST API manifest validation call supports passing the manifest as JSON data as an additional and alternative option to passing the manifest as a csv.

PR #910, opened 2022-09-15
allow users to upload csv or json file when validating manifest using API endpoints

linglp:
This PR is related to the issue mentioned [here](https://github.com/Sage-Bionetworks/schematic/issues/899)


afwillia (approved):
Looks good and works with DCA. Thanks!

Issue #863, opened 2022-08-16
schematic: update asset view table endpoint to allow returning the asset view table as a dataframe or JSON


afwillia:
**Is your feature request related to a problem? Please describe.**
When running the schematic REST API from a docker container, the storage/assets/tables endpoint returns the filepath to a csv that the user doesn't have access to. For example, in data curator app and accessing schematic hosted on a remote server, I get the following error
`Warning: Error in : '/home/app/schematic/tests/data/file_view_table.csv' does not exist.`

**Describe the solution you'd like**
Either return the table in the response or save it to a location the user has access to; or both. We can start with the easiest.

**How important is this feature?** Select from the options below:
• 🌗 Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

afwillia:
This branch adds an option to storage/assets/tables to export a json instead of saving the file and returning a filepath https://github.com/Sage-Bionetworks/schematic/tree/develop-asset-view-table-csv

milen-sage:
Getting this as a dataframe/json would be best. @afwillia could you link here to the code of the "storage/assets/tables" endpoint?

milen-sage:
Is this endpoint used in the DCA standards compliance dashboard or the DCA core modules currently released?

afwillia:
@milen-sage It's used in the compliance dashboard

milen-sage:
Ok thanks -I updated the description a bit. I think this can wait until next sprint, but we should tackle then.

milen-sage:
@afwillia looks like @linglp can tackle this sprint.

PR #912, opened 2022-09-20
update asset view table endpoint to allow returning a json 

linglp:
This PR is related to the issue [here](https://github.com/Sage-Bionetworks/schematic/issues/863)


afwillia (approved):
Looks good to me. Works well with the DCA. Thank you!

Issue #911, opened 2022-09-16
Data disappears from Synapse Table when manifest submitted through DCA


lopierra:
**Describe the bug**
I have an INCLUDE project with 3 folders for 3 different cohorts. I attempted to use the INCLUDE Data Curator App to upload a Study metadata manifest in each folder (all named synapse_storage_manifest_study.csv) and then be able to see the 3 manifests combined in a single Table. 

The Table is automatically created and looks correct when I validate & submit the first manifest, but when I submit subsequent manifests, the StudyName field for the previous manifests disappears from the Table and is only kept for the latest manifest. The CSV files for the manifests all have the correct values.

(Not sure if this is relevant, but StudyName is the only required field that is a string - the other two required fields (StudyCode and Program) are enums, and dbGaP is optional & blank.)

**To Reproduce**
1. Create [Synapse Project](https://www.synapse.org/#!Synapse:syn31538613/files/) with 3 folders for 3 different cohorts  ([ABCDS](https://www.synapse.org/#!Synapse:syn36358145), [DSC](https://www.synapse.org/#!Synapse:syn36358756), and [BRI-DSR](https://www.synapse.org/#!Synapse:syn31538629)) [I added @GiaJordan to the team, but please let me know if someone else needs access]
2. Use [INCLUDE Data Curator App](https://www.synapse.org/#!Synapse:syn30109400) to validate & submit a Study manifest to each folder
3. Manifests will be correct in CSV form ([ABCDS](https://www.synapse.org/#!Synapse:syn36358705), [DSC](https://www.synapse.org/#!Synapse:syn36358959), [BRI-DSR](https://www.synapse.org/#!Synapse:syn36358405))
4. StudyName values disappear from the common Table ([study_synapse_storage_manifest_table](https://www.synapse.org/#!Synapse:syn36358372/tables/)) for all but the most recently submitted manifest

**Expected behavior**
A new row with all values present should be added to the Table after each manifest is submitted.

**Priority** (select one)
- [ ] Minor ⬇️
- [ ] Major 📢
- [X] Critical 🆘

**Screenshots**
![image](https://user-images.githubusercontent.com/34149564/190528758-793b72be-cc6c-4e57-8e86-7e66c40529d0.png)

Manifests were submitted in this order: BRI-DSR, DSC, ABC-DS. Each time, StudyName only appears for the most recently submitted manifest, and disappears from the previous rows.

[...]

Issue #834, opened 2022-07-21
Weird Table Manifest Annotations Uploading


brynnz22:
**Describe the bug**
When attempting to use the table feature, instead of uploading manifests as Synapse annotations, the table leaves empty metadata values that were included in the csv. When uploading another manifest, it may, or may not, correctly fill in the table. When reuploading the same csv (with missing metadata in the table) it will update the metadata correctly, but also append the same rows onto the table, creating duplicates. It will then cause other empty metadata values for manifests that were originally uploaded correctly. Please see the version history of the table [here](https://www.synapse.org/#!Synapse:syn33342647.3/tables/).

**To Reproduce**
1. Enter the command `schematic model -c /path/to/config.yml submit -mp /path/to/manifest/csv -d syn33340846 -mrt table`
2. See how the manifest is reflected in the corresponding table.
3. Upload another manifest using the same command
4. See how the manifest is reflected in the corresponding table.
5. Upload a third manifest using the same command.
6. See how the manifest is reflected in the corresponding table.

**Expected behavior**
The table will accurately capture all of the metadata reflected in the manifest, leaving no empty values. It will also overwrite existing metadata when re-submitting a manifest (and not append more rows onto the table.). It will not impact other metadata that was originally correctly uploaded.

**Priority** (select one)
- [x ] Critical 🆘

**Screenshots**

**Desktop (if applicable, please complete the following information):**
 - OS: iOS
 - Browser Chrome

**Additional context**
The two attached manifests can be used to testing purposes. The CSBC data model csv can be found [here](https://docs.google.com/spreadsheets/d/1hbG1vdi8a0gc-6psn5VgsBgyMl6IOhDSbHWDuC_8EXo/edit#gid=59750547)
[CA184898.csv](https://github.com/Sage-Bionetworks/schematic/files/9162572/CA184898.csv)
[CA193313.csv](https://github.com/Sage-Bionetworks/schematic/files/9162573/CA193313.csv)

mialy-defelice:
@brynnz22 can you give me access to your table?

Naively, I think the problem is, the manifests will all be called the same thing, since they are all referencing the same component (the name is generated from the component name). This will cause synapse to keep adding to the same table on Synapse. Schematic in this case is anticipating that per project you would have one version of each manifest. If you're loading into the same dataset this could also cause overwriting of the manifest that is loaded as a csv. For this we might need to create a new feature to address this unique use case. 

[...]

PR #917, opened 2022-09-21
Develop Table Uploads: Replace

GiaJordan:
When uploading a manifest to synapse and storing as a Table, if the table exists and has metadata from another manifest already, the new manifest's metadata will be appended to the existing table, and some of the existing metadata will be replaced with dashes "-".  This was half of the requests from multiple issues #834 #911.
The current intended behavior for schematic in this scenario is to replace the existing data entirely, and only store the metadata from the new manifest. The [`replace_table`](https://github.com/Sage-Bionetworks/schematic_db/blob/develop/synapse/synapse.py#L241) method from schematic_db(@andrewelamb) has been adapted and introduced to the schematic synapse store's `make_table` method.

Further PRs will address feature requests from the two earlier mentioned issues, as well as splitting table behavior for censored and uncensored manifests.


lopierra:
@GiaJordan I'm not sure how to test this - is there something specific @brynnz22 and I should do?

GiaJordan:
@lopierra I'm currently releasing a containerized version of this branch on docker that you can use schematic in to test that the manifest table uploads work as expected! You, @brynnz22, and I can also work to test it together if need be!

GiaJordan:
@lopierra and @brynnz22 
The [container](https://hub.docker.com/layers/sagebionetworks/schematic/commit-6da9c93/images/sha256-7790ee4f5ebd6fce244c5be456effade286fc41fdae3ab8f0a2087468b92eceb?context=explore) is now up for you to download and use to test!

brynnz22:
@GiaJordan can we test this together? Perhaps on Thursday or Friday? I think both @lopierra and I are not super stars with Docker and it would be helpful to go through this.

schematic/store/synapse.py
```diff
@@ -1317,20 +1350,73 @@ def getDatasetAnnotationsBatch(
...
+            # locate its position in the old and new table.
+            if manipulation == 'update':
+                table_to_load = update_df(existing_table, table_to_load, update_col)
+            
+            elif manipulation == 'replace':
+                # remove rows
+                self.syn.delete(existing_results)
+
+                # wait for row deletion to finish on synapse before getting empty table
+                sleep(1)
```
milen-sage, line 1401: @GiaJordan @andrewelamb do you think the sleep interval is always enough? It seems a bit risky?

milen-sage (commented):
Please see comment below. Otherwise it looks good.

schematic/store/synapse.py
GiaJordan: In testing I had decreased it from 30 seconds in intervals to arrive at the 1 second. But after second thought we could increase it to be on the safe side for larger tables that might take more time?
@milen-sage @andrewelamb

andrewelamb (approved):

schematic/store/synapse.py
GiaJordan: @milen-sage  increased to 5sec

GiaJordan:
@brynnz22 and @lopierra thanks for your patience as we sorted out the short term future of this branch. It's not been merged to `develop` and will also be included in the next release, so you'll also be able to use this feature by pulling the current latest `develop` or by installing the newest schematic version after #934 is merged. 
If you only wanted assistance with using docker then feel free to go ahead and use the `develop` branch or install the upcoming release to use this feature; otherwise we can still find another time to meet about the actual use of the table feature since the working session time doesn't work for all of us.

@/afwillia and I will also be working this sprint to put together a set of instructions for installing and using docker to test features: #924

brynnz22:
Sounds good. Thanks so much @GiaJordan!

Issue #922, opened 2022-09-22
Unintentional Manifest Override


GiaJordan:
**Describe the bug**
When two manifests are uploaded to the same dataset, ie in the case of uploading a censored and uncensored version of the same manifest. The manifests are uploaded with the same name and the 2nd manifest overrides the values of the first.

**To Reproduce**
1. Validate manifest containing sensitive ages (under 18yo or over 90yo) as a value for an attribute with the `protectAges` rule
2. submit manifest to synapse dataset
3. Only one manifest is present after completion of command instead of two

**Expected behavior**
Two manifests should be present, with `_censored` appended to the name as appropriate

**Priority** (select one)
- [ ] Minor ⬇️
- [ x] Major 📢
- [ ] Critical 🆘

**Screenshots**

**Desktop (if applicable, please complete the following information):**
 - Schematic on develop as of 9/21/2022
 - Windows 10

**Additional context**

GiaJordan:
cc @milen-sage

milen-sage:
@GiaJordan this only applies for censored and uncensored manifests, but doesn't apply to manifests from different components stored in the same folder correct? If that's the case @linglp may be best positioned to resolve - I think this could be related to the change around using the base config manifest name and how components names are appended.

GiaJordan:
@milen-sage I believe so, it's because they're the same component so the names for the two manifests at upload are identical now. And I agree

PR #931, opened 2022-09-26
change manifest name for censored manifest when upload

linglp:
This PR is related to the issue mentioned [here](https://github.com/Sage-Bionetworks/schematic/issues/922). 

I have tested it out by using `schematic model -c config.yml submit --manifest_path /Users/lpeng/Documents/schematic-git2/schematic/tests/data/mock_manifests/Rule_Combo_Manifest.csv --dataset_id syn27221721` . This would upload both `Rule_Combo_Manifest.csv` and `Rule_Combo_Manifest_censored.csv` to synapse dataset folder. 
These two manifests will be renamed as: `synapse_storage_manifest_mockcomponent_censored.csv` and `synapse_storage_manifest_mockcomponent.csv`

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -762,8 +762,10 @@ class SynapseStorage(BaseStorage):
         # update file name
         file_name_full = metadataManifestPath.split('/')[-1]
         file_extension = file_name_full.split('.')[-1]
-        if restrict_manifest: 
-            file_name_new = os.path.basename(CONFIG["synapse"]["manifest_basename"]) + "_" + component_name + '_censored' + '.' + file_extension
+
+        # Differentiate "censored" and "uncensored" manifest
+        if "censored" in file_name_full: 
+            file_name_new = os.path.basename(CONFIG["synapse"]["manifest_basename"]) + "_" + component_name + "_censored" + '.' + file_extension
         else: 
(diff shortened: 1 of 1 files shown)


GiaJordan (approved):

schematic/store/synapse.py
```diff
@@ -762,7 +762,10 @@ def uplodad_manifest_file(self, manifest, metadataManifestPath, datasetId, restr
         # update file name
         file_name_full = metadataManifestPath.split('/')[-1]
         file_extension = file_name_full.split('.')[-1]
-        file_name_new = 'synapse_storage_manifest_' + component_name + '.' + file_extension
+        if restrict_manifest:
```
GiaJordan, line 765: The logic for naming should be updated. Currently, the censored manifest does not have the `censored` appended, the uncensored manifest does. Restrictions are correctly generated for the uncensored manifest, naming is just off.

GiaJordan (changes requested):

schematic/store/synapse.py
linglp: I think it is a bit confusing to use `restrict_manifest` to infer if a manifest is a censored manifest or not. Let me switch it to use the file name.

GiaJordan: Yep! It's based off the file name elsewhere as well

GiaJordan (approved):

PR #935, opened 2022-09-26
Add Restriction Flag to Make Synapse Table Method

GiaJordan:
Adds restriction flag to other table upload method. Technically redundant since it is updating a table that already exists and has been uploaded with correct permissions, but including to be overly safe


linglp (approved):
Looks good to me.

Issue #932, opened 2022-09-26
Generate manifest by populating excel spreadsheet


linglp:
**Is your feature request related to a problem? Please describe.**
Currently, we sent metadata to Google when populating a manifest. But for security reasons, we should not be sending metadata to Google APIs. A new workflow for populating a manifest would look something like below: 
1) Generate an empty google sheet (with new formatting) 
2) Export the empty google sheet as an Excel spreadsheet
3) Put our metadata to Excel
4) Generate a file path that links to Excel 

**Describe the solution you'd like**

**How important is this feature?** Select from the options below:
• 🌗 Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

**Additional context**

PR #940, opened 2022-09-28
populate manifest as an excel spreadsheet without sending data to Google api

linglp:
This PR is related to the issue mentioned [here](https://github.com/Sage-Bionetworks/schematic/issues/932). I added a parameter for endpoint '/manifest/populate' to control if it is returning a google sheet or an Excel spreadsheet. I have done the following testing: 

- [x] in swagger UI, set return_excel as True, and it would return an Excel workbook with hidden sheet. The formatting of google sheet could still be preserved in Excel and new columns get added. 
- [x] in swagger UI, set return_excel as False, and it would return a google sheet as usual 
- [x] Have the endpoint return Excel file path

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -1469,11 +1469,11 @@ class ManifestGenerator(object):
         
         return output_excel_file_path
 
-    def handle_output_format_logic(self, output_form: str = None, output_path: str = None, sheet_url: bool = None, manifest_url: str = None, empty_manifest_url: str = None, dataframe: pd.DataFrame = None):
+    def _handle_output_format_logic(self, output_format: str = None, output_path: str = None, sheet_url: bool = None, manifest_url: str = None, empty_manifest_url: str = None, dataframe: pd.DataFrame = None):
         """
-        handle the logic between sheet_url parameter and output_form parameter to determine the type of output to return
+        handle the logic between sheet_url parameter and output_format parameter to determine the type of output to return
         Args: 
-            output_form: Determines if Google sheet URL, pandas dataframe, or Excel spreadsheet gets returned.
+            output_format: Determines if Google sheet URL, pandas dataframe, or Excel spreadsheet gets returned.
             sheet_url (Will be deprecated): a boolean ; determine if a pandas dataframe or a google sheet url gets return
             manifest_url: Google sheet URL populated with metadata
             empty_manifest_url: Google sheet URL that leads to an empty manifest 
@@ -1482,8 +1482,8 @@ class ManifestGenerator(object):
         return: a pandas dataframe, file path of an excel spreadsheet, or a google sheet URL 
         """
 
-        # check if output_form parameter gets set. If not, check the sheet_url parameter
-        if not output_form: 
+        # check if output_format parameter gets set. If not, check the sheet_url parameter
+        if not output_format: 
             # if the sheet_url parameter gets set to True, return a google sheet url 
             if sheet_url: 
                 return manifest_url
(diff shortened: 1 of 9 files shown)


linglp:
[Example.xlsx](https://github.com/Sage-Bionetworks/schematic/files/9668575/Example.xlsx)

^ example output

linglp:
> @linglp question - currently DCA only uses the manifest/generate endpoint, not manifest/populate. It will create a filled manifest or empty one, depending on what's in Synapse. Is it possible to have both endpoints return an excel sheet?

I think for the /get/manifest endpoint, if there's no existing manifest, the problem of "sending metadata to Google sheet API" doesn't exist because we are not asking the users to upload any manifests. For getting an existing manifest, currently we are creating an empty spreadsheet first, and then populate that spreadsheet with content from an existing manifest. I think we could change that flow and make it work like the populate endpoint. What do you think @milen-sage ?

milen-sage:
> Is it possible to have both endpoints return an excel sheet?

@afwillia this is a good point. If the DCA is configured to expose an excel spreadsheet, we'd like that behavior to be consistent for both empty and pre-populated spreadsheets. 

@linglp could we adjust the end point logic so that it's possible to get an excel sheet in both these cases?

linglp:
@milen-sage Yes. I could do that. Currently, the `get_manifest` function has a bunch of parameters: `dataset_id: str = None, sheet_url: bool = None, json_schema: str = None`.  Should I add one called `return_excel` and set the default to False, or should I re-name `sheet_url` parameter? If we keep both `sheet_url` and `return_excel` parameters, it could potentially be confusing. 

In addition, if a dataset ID is not provided, currently we just let users return a sheet url (regardless of sheet_url parameter). should we also let users return an Excel spreadsheet without dataset ID? Also, if there is an existing manifest when dataset ID is provided, the expected behavior is just downloading the Excel spreadsheet directly? Or should we download an empty one after getting an empty manifest url and then populate it based on the existing manifest?

linglp:
@milen-sage  and @GiaJordan Please view a summary of  the logic behind `get_manifest` function here:

Current behaviors of `get_manifest` after adding `return_excel` parameter: 
- return_excel parameter is default to False
-  if dataset_id does not exist, the code would check `sheet_url` parameter. If `sheet_url = True or sheet_url = None`, the code would return an Excel spreadsheet. If explicitly set `sheet_url = False` and `return_excel = True`, then an Excel spreadsheet would return (Note: before adding `return_excel` parameter, when the dataset id does not exist,  the code would always return a sheet url regardless of sheet_url parameter. Hopefully the logic here aligns with the previous behavior)
- In swagger UI, return_excel is an optional parameter. If it is set to true, an Excel spreadsheet would return. If it is set to False or None, a google sheet URL would be returned. 
- If dataset_id exists and `sheet_url = True`, a google sheet url would get returned 
- If dataset_id exists and `sheet_url = None or False`, and if `return_excel = True`, an empty excel spreadsheet would first get downloaded. To avoid sending metadata to google, we would populate the excel spreadsheet by using the manifest record from the existing dataset_id
- If dataset_id exists and `sheet_url = None or False`, and if `return_excel = False`, a dataframe would get returned 

Let me know if anything doesn't make sense!

Questions: 
* when writing tests, I added  `dataset_id = "syn27600056"`. I am not sure if that's best practice. What do you think @GiaJordan 
* Should we rename `sheet_url` to a different parameter to avoid confusion? For example, `output = dataframe, google_sheet_url, and excel`? 
* How should we handle when users make a mistake by setting `sheet_url` and `return_excel` both to True? (The current behavior is that `sheet_url` gets checked first, so in this case, we would still return a google sheet)

schematic/manifest/generator.py
```diff
@@ -1586,10 +1586,10 @@ def get_manifest(
                 manifest_url, manifest_df = self.get_manifest_with_annotations(annotations)
```
milen-sage, line 1586: Note that here, we are assuming annotations can't contain protected, e.g. PHI, contributors' data. Otherwise, we'd be communicating that with Google.

milen-sage: I think this is fine for now; something to highlight with ISO and see if it'd be a problem.

milen-sage (approved):
Looks good @linglp ! I left a couple of comments but those can be resolved later

PR #945, opened 2022-10-03
Fixing parsing of table schema parameters on table replace

milen-sage:
@GiaJordan this should resolve table schema issues like max character limit.

PR #946, opened 2022-10-03
BugFIx: Table Schema Replace

GiaJordan:
Fixes the error raised that `UUID` strings are too long when trying to upload a manifest as a table to replace an existing table

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1399,6 +1399,9 @@ class SynapseStorage(BaseStorage):
             if manipulation == 'update':
                 table_to_load = update_df(existing_table, table_to_load, update_col)
             
+
+                # store table with existing etag data and impose restrictions as appropriate
+                self.syn.store(Table(existingTableId, table_to_load, etag = existing_results.etag), isRestricted = restrict)
             elif manipulation == 'replace':
                 # remove rows
                 self.syn.delete(existing_results)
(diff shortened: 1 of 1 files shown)


schematic/store/synapse.py
```diff
@@ -1409,10 +1418,38 @@ def make_synapse_table(self,
...
+
+                datasetParentProject = self.getDatasetProject(dataset_id)
+                if specify_schema:
+                    if column_type_dictionary == {}:
+                        logger.error("Did not provide a column_type_dictionary.")
+                    #create list of columns:
+                    
+                    cols = []
+                    for col in table_to_load.columns:
+                        if col in column_type_dictionary:
```
milen-sage: @GiaJordan based on  https://github.com/Sage-Bionetworks/schematic/pull/945/files#r986183078

this section of the code wouldn't work.

GiaJordan: So I had made a new dictionary in [`upload_format_manifest_table`](https://github.com/Sage-Bionetworks/schematic/blob/8b4b1615c4d36dc6e069bff59e46c1eaf9f913b6/schematic/store/synapse.py#L763) that created the dictionary as specified, and that led to nested `SynapseHTTPError`s that I couldn't resolve. Right now, the `else` operations are being executed that sets the columns to type `string` and the `maximumSize` to 500 and allows the table to be uploaded.  So far I also haven't found another place in the code where `make_synapse_table` is used to gain more insight on how this dictionary should be structured to work properly.
@milen-sage

milen-sage: Another issue to be aware of is that Synapse has a limit on the number of characters overall allowed in a column/table. By setting the limit to 500 on each column, for some data types (e.g. followup), we bump into that limit. This [line ](https://github.com/Sage-Bionetworks/schematic/pull/945/files#r986255073) tries to, for now, to avoid that.

milen-sage: Yes, `make_syn..._table` is called in a few locations.

> So I had made a new dictionary in [upload_format_manifest_table](https://github.com/Sage-Bionetworks/schematic/blob/8b4b1615c4d36dc6e069bff59e46c1eaf9f913b6/schematic/store/synapse.py#L763) ...

Looking at the link above, col_schema is still [passed](https://github.com/Sage-Bionetworks/schematic/blob/8b4b1615c4d36dc6e069bff59e46c1eaf9f913b6/schematic/store/synapse.py#L775) as [returned by](https://github.com/Sage-Bionetworks/schematic/blob/8b4b1615c4d36dc6e069bff59e46c1eaf9f913b6/schematic/store/synapse.py#L761)  the `as_table_columns` method. Did you mean linking a different line?

GiaJordan: Yes, I had created a new dictionary with the values that were to be parsed in `make_synapse_table`, but that led to the errors so it was never pushed. However, I looked at your new method and think I understand it better now. I think all that is necessary now is also performing the column setting for the `replace` method and linking to the existing table

milen-sage: Looking at the `[process_schema_cols`](https://github.com/Sage-Bionetworks/schematic/blob/efdb3ea02e50d71a90cb59d9879893761fb42300/schematic/store/synapse.py#L1458), it has a similar issue: it assumes that the column_type_dictionary is indeed a dictionary; but `as_table_columns` 

- does not return a dictionary; it's an array of dictionaries
- and each dictionary doesn't have keys 'column_type', 'maximum_size', 'maximum_list_length'; it has keys `columnType`, `maximumSize`, among others.

The `_get_table_schema_by_cname` in #945 processes that. @GiaJordan it sounds like you're making the changes to incorporate this logic.

I haven't checked why column_type_dictionary format is so different now than what had been expected by these other functions;  perhaps the client changed how the table cols specification is returned by `as_table_columns`. We can figure that out later. For now would be happy to take a look when you update your PR.

GiaJordan:
Closed, addressed in #947

PR #952, opened 2022-10-06
add default max size for columns missing this key

GiaJordan:
Fixes an issue where columns that have a specified type but no `maximumSize` key and value would cause the program to crash when trying to set the max size. Now defaults to 100 for columns missing this key.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1437,7 +1437,7 @@ class SynapseStorage(BaseStorage):
                         
                         if col in table_schema_by_cname:
                             col_type = table_schema_by_cname[col]['columnType']
-                            max_size = table_schema_by_cname[col]['maximumSize']
+                            max_size = table_schema_by_cname[col]['maximumSize'] if 'maximumSize' in table_schema_by_cname[col].keys() else 100
                             max_list_len = 250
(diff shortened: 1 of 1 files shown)


schematic/store/synapse.py
```diff
@@ -1487,7 +1487,7 @@ def make_synapse_table(self,
                 for col in table_to_load.columns:
                     if col in table_schema_by_cname:
                         col_type = table_schema_by_cname[col]['columnType']
-                        max_size = table_schema_by_cname[col]['maximumSize']
+                        max_size = table_schema_by_cname[col]['maximumSize'] if 'maximumSize' in table_schema_by_cname[col].keys() else 100
```
milen-sage, line 1490: @GiaJordan do you think it'd make sense to have a single function that returns this information; e.g. a tuple of col_type, max_size, etc.? Just to avoid repeating this in the code?

GiaJordan: @milen-sage yes, I was planning on restructuring this and the related code as part of the table up-sert feature. It's just been  delayed because of the previous issues we had to solve with the table uploads and entity migration. But I plan to separate this method into smaller ones that are more specific in their purpose

milen-sage (approved):

Issue #894, opened 2022-09-07
Use pdoc to auto-generate reference doc for Schematic methods


linglp:
**Is your feature request related to a problem? Please describe.**
Use pdoc to generate reference doc for API methods. 

See the example workflow [here](https://github.com/Sage-Bionetworks/schematic_db/blob/develop/.github/workflows/pdoc.yml)

**Describe the solution you'd like**

**How important is this feature?** Select from the options below:
• 🏝 Low - it's an enhancement but not crucial for work

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

**Additional context**

linglp:
The documentation is now live here: https://sage-bionetworks.github.io/schematic/api/routes.html

But I am not sure how useful is this page. `sage-bionetworks.github.io/schematic` page would probably be more helpful for other documentation.

milen-sage:
@linglp yes, let's use pdoc to generate the reference API for all schematic's methods, not the REST API; we can host that gh-pages.

milen-sage:
The pdoc docs don't look bad at all! I'm happy it captured quite a bit of our docstrings. It's a good base and we can collectively build on it: update docstrings; organize the pdoc pages, if possible, to be a bit more user-friendly, etc.. But this is a very nice start - thanks @linglp !

PR #957, opened 2022-10-10
Develop pdoc

linglp:
This is related to issue #894.  I used Github action and workflow to create a documentation page for API endpoints.


linglp:
@milen-sage I have updated the github page [here](https://sage-bionetworks.github.io/schematic/index.html) to reflect the functions in schematic instead of functions for API endpoints. But github pages server can be down sometimes, so using github pages for documentation might not be ideal.

GiaJordan (approved):

Issue #973, opened 2022-10-19
schematic docker container doesn't work for APIs


linglp:
**Describe the bug**
The following doesn't work: 

```
docker run --rm -p 3001:3001 \
-v $(pwd):/schematic -w /schematic --name schematic2 \
-e APP_HOST=0.0.0.0 \
-e APP_PORT=3001 \
-e SCHEMATIC_CONFIG=/schematic/config.yml \
-e GE_HOME=/usr/src/app/great_expectations/ \
sagebionetworks/schematic:latest \
python /usr/src/app/run_api.py
```

When actually opening up `http://localhost:3001/v1/ui/`, the error message: this site can't be reached

But `sagebionetworks/schematic:v22.10.1` works 

**To Reproduce**

**Expected behavior**

**Priority** (select one)
- [X] Major 📢

**Screenshots**

**Desktop (if applicable, please complete the following information):**

**Additional context**

PR #974, opened 2022-10-20
fix api endpoint for docker container

linglp:
This PR is related to the issue [here](https://github.com/Sage-Bionetworks/schematic/issues/973) This change is important for using the API through docker container.

### DIFF
--- a/run_api.py
+++ b/run_api.py
@@ -15,4 +15,6 @@ if __name__ == "__main__":
     # Launch app
     app = create_app()
     CORS(app, resources={r"*": {"origins": "*"}})
-    app.run(port=3001, debug=True)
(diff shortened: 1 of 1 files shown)


linglp:
> @linglp I think this is backward compatible right? Or do docs for how to run the REST API need to be updated, too?

This is actually the old line of code that we had. I was using an old branch for schema visualization so I accidentally overwrote it. The original line could be found [here](https://github.com/Sage-Bionetworks/schematic/pull/962/files#diff-2c29786c4bd8cb9a215208ab3cac78ac134838e0aed57c3fbedc8369ec14e12d:~:text=app.-,run,-(host%3D)

And no, we don't need to change how we run API.

GiaJordan (approved):

Issue #960, opened 2022-10-11
schematic: allow passing config and google credentials information as env variables to docker container


milen-sage:
Currently schematic relies on config and credential parameters stored as local files at runtime. 

schematic assumes these files will be locally available when a REST API is run.

To enable deploying schematic in a Fargate-based architecture, based on IT specifications, we'd like to avoid storing these files locally. Instead they can be passed as environmental variables when a docker image is created.

The schematic docker container (usage here #924 ) currently doesn't support env variables passing. We need to enable that.

AC:

- [ ]  change docker container to support passing environmental variables, similarly to https://github.com/Sage-Bionetworks/data_curator/issues/400

- [ ]  add logic to schematic's REST API to allow reading config and credential options as env variables passed when the docker image is created. E.g. running the REST API can check if these parameters are present as env variables and read them in, avoiding in this case to look for config and credential files.

linglp:
related to #975

PR #975, opened 2022-10-20
Pass environment variable to docker container 

linglp:
This PR is related to the issue [here](https://github.com/Sage-Bionetworks/schematic/issues/960)

This PR allows users to pass an environment variable to load the content of config file (instead of using path of config)

### DIFF
--- a/api/__init__.py
+++ b/api/__init__.py
@@ -15,9 +15,9 @@ def create_app():
     # path to config.yml file saved as a Flask config variable
     default_config = os.path.abspath(os.path.join(__file__, "../../config.yml"))
     schematic_config = os.environ.get("SCHEMATIC_CONFIG", default_config)
+    schematic_config_content = os.environ.get("SCHEMATIC_CONFIG_CONTENT")
 
     app.config["SCHEMATIC_CONFIG"] = schematic_config
-    schematic_config_content = os.environ.get("SCHEMATIC_CONFIG_CONTENT")
     app.config["SCHEMATIC_CONFIG_CONTENT"] = schematic_config_content
 
     # Configure flask app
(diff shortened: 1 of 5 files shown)


GiaJordan (approved):
Looks good to me, thanks for adding documentation as well!

linglp:
@afwillia  could you also try testing it out too? Thank you!

afwillia:
@linglp This worked, I just had to make sure the $pwd that is mounted contains the credential files.

afwillia (approved):

linglp:
@milen-sage Based on @afwillia 's testing, this PR would still require running the container from a directory that has the service account creds. Maybe we should also turn `service account creds` into a environment variable and pass it to docker container too. 

And this might relate to the issue [here](https://github.com/Sage-Bionetworks/schematic/issues/986)

milen-sage:
> Maybe we should also turn service_account_creds.json into an environment variable and pass it to docker container too.

Yes - this would be ideal. I think we can then deprecate the need for other google credentials (those are only needed for the multi-select appscript widget, which we can deprecate soon too).

milen-sage:
> Maybe we should also turn service_account_creds.json into an environment variable and pass it to docker container too.

Yes - this would be ideal. I think we can then deprecate the need for other google credentials (those are only needed for the multi-select appscript widget, which we can deprecate soon too).

milen-sage:
> Maybe we should also turn service_account_creds.json into an environment variable and pass it to docker container too.

Yes - this would be ideal. I think we can then deprecate the need for other google credentials (those are only needed for the multi-select appscript widget, which we can deprecate soon too).

PR #989, opened 2022-10-26
Remove Manifest Table Submission Dependency on Project Scope when not Validating

GiaJordan:
The project synID is necessary for table replacements. currently it is being taken from the `--project_scope` argument when calling the manifest submission command but it can cause errors as this was not previously required to facilitate the command. This pr removes the requirement of including the project scope in the command and gathers the information with a different method.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -717,16 +717,16 @@ class SynapseStorage(BaseStorage):
 
         return df, results
 
-    def _get_tables(self) -> List[Table]:
-        project = self.syn.get(self.project_scope[0])
+    def _get_tables(self, datasetId: str = None) -> List[Table]:
+        project = self.syn.get(self.getDatasetProject(datasetId))
         return list(self.syn.getChildren(project, includeTypes=["table"]))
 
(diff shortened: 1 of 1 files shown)


linglp (approved):
Looks good to me. I have tested it out by using swagger UI as well as docker container. 

Tests: 

1. In swagger UI, I was able to submit a manifest file 
2. In swagger UI, I was also able to submit a json str
3. The following works for me too: 
```
docker run --rm -p 3001:3001 \               
  -v $(pwd):/schematic -w /schematic --name schematic-test \
  -e SCHEMATIC_CONFIG=/schematic/config.yml \
  -e GE_HOME=/usr/src/app/great_expectations/ \
sagebionetworks/schematic:commit-f3192cd \
schematic model \
-c config.yml submit -mp tests/data/mock_manifests/Valid_Test_Manifest.csv -d syn23643250 -vc MockComponent -mrt table
```

milen-sage (commented):
@GiaJordan I'm looking at the code and wondering how did it work to the extent that it did prior to the change :)

PR #995, opened 2022-10-31
add API tests 

linglp:
add tests for all the API endpoints in schematic

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -3,6 +3,7 @@ import pytest
 from api import create_app
 import configparser
 import json
+from flask import jsonify
 
 '''
 To run the tests, you have to keep API running locally first by doing `python3 run_api.py`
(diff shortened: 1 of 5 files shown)


tests/test_api.py
```diff
@@ -0,0 +1,473 @@
...
+import configparser
+import json
+import os
+
+
+'''
+To run the tests, you have to keep API running locally first by doing `python3 run_api.py`
+'''
+
+@pytest.fixture
```
andrewelamb, line 13: Have you considered using [fixture scope](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) so that you don't have to create the app for each test individually?

andrewelamb (approved):

tests/test_api.py
linglp: Thanks @andrewelamb ! I tried to add scope but the tests with "scope = module" ran slower (total time of 58 tests: 444.02 seconds) and tests with default scope function took 394.02 seconds.

andrewelamb: Weird!

milen-sage (approved):

PR #997, opened 2022-10-31
update error messages in synapseStore when `manifest_basename` key is missing in config.yml 

linglp:
It seems like our `SynapseStorage` class mixes the error of missing `master_fileview` key in the config.yml and missing `manifest_basename` key. 

Currently, if the `manifest_basename` key is missing in the `config.yml`, it would raise the error of missing `master_fileview` key

This PR is for separating the error messages for these two.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -84,13 +84,13 @@ class SynapseStorage(BaseStorage):
         # check if "master_fileview" has been set
         try: 
             self.storageFileview = CONFIG["synapse"]["master_fileview"]
-        except: 
+        except KeyError: 
             raise MissingConfigValueError(("synapse", "master_fileview"))
 
         # check if "master_basename" has been set
         try: 
             self.manifest = CONFIG["synapse"]["manifest_basename"]
-        except: 
+        except KeyError: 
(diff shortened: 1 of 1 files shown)


linglp:
@milen-sage this is driven by the conversation I had with someone from HTAN: https://htanworkspace.slack.com/archives/GUGUTPA06/p1666894352973189

schematic/store/synapse.py
```diff
@@ -79,8 +79,23 @@ def __init__(
 
         self.syn = self.login(token, access_token, input_token)
         self.project_scope = project_scope
+
+
+        # check if "master_fileview" has been set
+        try: 
+            self.storageFileview = CONFIG["synapse"]["master_fileview"]
+        except:
```
GiaJordan, line 87: can we specify that we're catching a `KeyError` for the try except blocks?

GiaJordan (changes requested):

schematic/store/synapse.py
linglp: Yes. I could fix that.

GiaJordan (approved):

PR #1006, opened 2022-11-11
Fix DCA disconnect on submission, cast UUIDs to type string

GiaJordan:
DCA disconnect at submission occurred when adding new rows to a manifest that had already been submitted. The error occurred because the UUID for the new rows was being added to the manifest as type `uuid` and not type `str`. This fix casts the UUIDs as type `str` before adding to the manifest.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1055,7 +1055,7 @@ class SynapseStorage(BaseStorage):
 
         for idx,row in manifest.iterrows():
             if not row["Uuid"]:
-                gen_uuid = uuid.uuid4()
+                gen_uuid = str(uuid.uuid4())
                 row["Uuid"] = gen_uuid
                 manifest.loc[idx, 'Uuid'] = gen_uuid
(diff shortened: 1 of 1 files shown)


milen-sage (approved):

milen-sage:
@GiaJordan the change looks good as code (and it'd be great to test as with DCA as part of a bug bash we were chatting), but just curious, were you able to repro the disconnect even when just using schematic's CLI?

GiaJordan:
@milen-sage not as the disconnect per-se, but I was able to reproduce the [error Adam included](https://github.com/ncihtan/HTAN-data-curator/issues/38#issuecomment-1306194975) from the shiny logs that caused the disconnect. In the CLI this was being caught unintentionally and displaying that it was caused by a different issue. 

I think it might benefit us longer term to take another look at how we handle errors throughout the codebase, and/or creating schematic specific exceptions to make it easier to notice issues like this one and avoid confusing different issues.

Issue #984, opened 2022-10-25
Migrate Orphaned Entities in HTAN projects and flag manifests that were overwritten during entities migration


GiaJordan:
Develop functionality to migrate entities no longer denoted in metadata manifests (e.g. checking for entities in manifest version history).

Also, in some HTAN projects,  prior the conversion of entities to tables there were datasets that contained both manifests named synapse_storage_manifest.csv and synapse_storage_manifests_<component_name>.csv

In those cases, the conversion script took synapse_storage_manifest.csv and reuploaded it as synapse_storage_manifests_<component_name>.csv overwriting the information in the already existing synapse_storage_manifests_<component_name>.csv 

Let's flag all manifests across all HTAN projects for which this happened (e.g. by checking if the previous version of a synapse_storage_manifests_<component_name>.csv was uploaded by a user other than the one running the conversion script). Let's leave only the synapse_storage_manifests_<component_name>.csv manifest and move synapse_storage_manifests.csv to the corresponding folder in archive.

Since this would comb through manifests, let's ensure that the "download as" attribute of manifest files matches the name of the manifest.

AC: 

- [ ] all entities in previous manifest versions are migrated to the HTAN archive project
- [x] all manifests that were overwritten during conversion, as described above, are flagged 
- [x] all old synapse_storage_manifest.csv files are migrated to the HTAN archive project, ~~pending confirm from @adamjtaylor ~~ as confirmed by @adamjtaylor
- [x] #1023

GiaJordan:
`downloadAs` Addressed in #1017, will be used for final HTAN Migrartion

adamjtaylor:
@GiaJordan I think so..... but we would want to make sure we knew where they originally came from - either as a folder structure in the archive project or with an annotation.
An alternative would be to lock the manfest by removing access to all but the HTAN_DCC team.

Would like to hear @milen-sage, @elv-sb or @aclayton555's thoughts before proceding.

GiaJordan:
@adamjtaylor yes, I was planning to sort them into the folder hierarchy just as the entities were, and they would be located in the appropriate Project and Dataset folders corresponding to where they were previously located, along with the migrated entities

GiaJordan:
@milen-sage updating status to "backlog" to reflect earlier discussion on timeline

GiaJordan:
With the majority of the entities migrated and the manifests with the old naming convention also migrated. Work for HTAN is unblocked.
[...]

PR #1017, opened 2022-11-17
Fixed a bug where in some instances manifests would be uploaded with different display and `downloadAs` names

GiaJordan:
Prerequisite for #984. This PR updates the `downloadAs` manifest metadata field to match the manifest name whenever a manifest is uploaded.


milen-sage (approved):
Looks good 👍

PR #1020, opened 2022-11-17
Strip hyphens from Synapse annotation keys

### LINKED ISSUE #1019: Support hyphens in data model attribute names
**Is your feature request related to a problem? Please describe.**
Currently, the synapse client annotation keys do not allow hyphens as characters in key names. schematic already strips other not-allowed special characters in synapse annotations. We may as well strip hyphens too when storing synapse annotations; that would allow attribute names to contain hyphens in data models.

• Short-term - 2-4 weeks

milen-sage:
Blacklist hyphen in annotation key names to avoid a Synapse error when setting an annotation key containing a hyphen. Should resolve #1019.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -831,7 +831,7 @@ class SynapseStorage(BaseStorage):
         # this could create a divergence between manifest column and annotations. this should be ok for most use cases.
         # columns with special characters are outside of the schema
         metadataSyn = {}
-        blacklist_chars = ['(', ')', '.', ' ']
+        blacklist_chars = ['(', ')', '.', ' ', '-']
         
         for k, v in row.to_dict().items():
 


schematic/store/synapse.py
```diff
@@ -831,7 +831,7 @@ def format_row_annotations(self, se, sg, row, entityId, useSchemaLabel, hideBlan
         # this could create a divergence between manifest column and annotations. this should be ok for most use cases.
         # columns with special characters are outside of the schema
         metadataSyn = {}
-        blacklist_chars = ['(', ')', '.', ' ']
+        blacklist_chars = ['(', ')', '.', ' ', '-']
```
GiaJordan, line 834: Do we also need to make this change for [table column names](https://github.com/Sage-Bionetworks/schematic/blob/49a949137386f027daebea19c3288dbc820fc2dd/schematic/store/synapse.py#L748) as well?

milen-sage: Yes - nice catch! For consistency, let's do that. I just pushed a commit. 

In the refactor we'd want to have a single list of invalid characters for Synapse annotation key names and Synapse table column names. 

(Note to self: manifest column names need to remain w/o changes, which is what the current behavior does.)

GiaJordan: I agree, we could put in in the utils, or create like a `store_utils` that could house other things we find specific to synapse that might change in the future

GiaJordan (approved):

Issue #981, opened 2022-10-25
TT1 schematic: modify CDK AWS deployment script


milen-sage:
To deploy in Fargate-based architecture we need to modify the CDK [script ](https://sagebionetworks.jira.com/wiki/spaces/IT/pages/2720759990/Deploying+a+Web+Application+to+AWS#Create-a-deployment-script%2C-saved-to-a-GitHub-repository)and add it to our CI/CD github workflow.

AC: script is modified as needed and added to GH actions on its own branch, using the docker container developed in #960.

milen-sage:
@afwillia this should be pretty much the same as https://github.com/Sage-Bionetworks/data_curator/issues/425 but this time for schematic given the latest version of the docker container @linglp developed.

linglp:
Follow up after #986

linglp:
@milen-sage I checked out the documentation, and here are some of my questions: 

Questions: 
- the documentation here mentioned about `Create and containerize the application`. Currently, `schematic` can be run by using docker, and we have a workflow `docker.yml` for deployment. Will that be sufficient? 
- The documentation also mentions about scanning vulnerabilities of the docker container. I used `docker scan sagebionetworks/schematic`, and found out that a lot of vulnerabilities are related to our base image (see below). What should we do about it? I could potentially experiment changing the base image if that's what we want. 
```
✗ Critical severity vulnerability found in aom/libaom0
  Description: Buffer Overflow
  Info: https://snyk.io/vuln/SNYK-DEBIAN11-AOM-1300249
  Introduced through: imagemagick@8:6.9.11.60+dfsg-1.3
  From: imagemagick@8:6.9.11.60+dfsg-1.3 > imagemagick/imagemagick-6.q16@8:6.9.11.60+dfsg-1.3 > imagemagick/libmagickcore-6.q16-6@8:6.9.11.60+dfsg-1.3 > libheif/libheif1@1.11.0-1 > aom/libaom0@1.0.0.errata1-3
  Image layer: Introduced by your base image (python:3.10.6-bullseye)
```

> Anthony mentioned that anything "high" or "critical" needs to be resolved before deployment. 

- "activating Dependabot on the GitHub repository and monitoring and remediating its alerts" also mentioned by the documentation. I then enabled "Dependabot alerts" in schematic, but I am not entirely sure if that would be sufficient. 

Note: 
- schematic-infra has already been created. See here: https://github.com/Sage-Bionetworks/schematic-infra
[...]

PR #1037, opened 2022-11-28
Changed base docker image and removed package for security reasons 

linglp:
related to issue #981 

For CDK deployment, we will need to create a docker image without high/critical level of security risks.


milen-sage (approved):
This looks good for now. Next sprint, on the tiger team, we can review the container again to confirm this is a good way to resolve vulnerabilities. 👍

Issue #1029, opened 2022-11-21
Deprecate functionalities related to using `token.pickle` and `credentials.json` in schematic 


linglp:
**Is your feature request related to a problem? Please describe.**
After #1010 , we would want to use `service_account_creds` for google api authentication. There are couple of changes that need to happen for schematic: 
- Deprecate `schematic init --auth`. The `--auth` flag currently could let users choose `OOB` authentication or `service_account_creds`. In the future, users would only be using `service_account_creds`
- Some where in schema generator, if a schema is a list, that triggers app script, which would need to be deprecated
- there's also template micro that is hidden in the `master_template` (refer to `master_template_id`). We would need to remove that too 

**Describe the solution you'd like**

**How important is this feature?** Select from the options below:
• 🌗 Medium - can do work without it; but it's important (e.g. to save time or for convenience)

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

**Additional context**

linglp:
checking how app script authentication works (figure out an alternative way of authentication)

linglp:
@milen-sage one more clarification question: if we are deprecating multi-select, can I assume that the column that previously requires multi-select would turn into a free text column that doesn't have any validation rules associated with it? For example, in the spreadsheet generated by our example data model [here](https://docs.google.com/spreadsheets/d/1edlBA41uHLeEGcEr6BxMfXnTjDYaBTdfA6HfGpX1Mfo/edit#gid=0), "FamilyHistory" would look like "Year of Birth"?

milen-sage:
Yes - the attributes with multi-select would become free text. Perhaps we can change the tooltip to say something along the lines of "Please enter applicable comma-separated items selected from the set of allowable terms for this attribute. See our data standards for allowable terms."

PR #1040, opened 2022-11-30
Deprecated token pickle and credentials.json

linglp:
related to #1029 

For testing, I have: 
- generated manifest manually by using swagger UI
- ran api tests related to manifestOperation
- ran command: `schematic init config.yml` and checked the help function 
- ran command to get an existing manifest: `schematic manifest --config ~/path/to/config.yml get --dataset_id syn28268700 --data_type BulkRNA-seqAssay`
- ran command to get an empty manifest `schematic manifest --config ~/path/to/config.yml get --sheet_url`

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -7,7 +7,6 @@ synapse:
   master_fileview: 'syn23643253'
   manifest_folder: 'manifests'
   manifest_basename: 'synapse_storage_manifest' 
-  token_creds: 'syn23643259'
   service_acct_creds: 'syn25171627'
 
 manifest:
--- a/config.yml
+++ b/config.yml
@@ -1,8 +1,6 @@
 # Do not change the 'definitions' section unless you know what you're doing
 definitions:
   synapse_config: ".synapseConfig"
-  creds_path: "credentials.json"
-  token_pickle: "token.pickle"
   service_acct_creds: "schematic_service_account_creds.json"
 
 synapse:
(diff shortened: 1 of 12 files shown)


config.yml
```diff
@@ -1,8 +1,6 @@
 # Do not change the 'definitions' section unless you know what you're doing
 definitions:
   synapse_config: ".synapseConfig"
-  creds_path: "credentials.json"
-  token_pickle: "token.pickle"
   service_acct_creds: "schematic_service_account_creds.json"
 
 synapse:
```
GiaJordan, line 6: remove token synId key value pair as well

GiaJordan (changes requested):

linglp:
@GiaJordan and @milen-sage after the code review yesterday, I noticed that the valid values for attribute `family history` that requires `list` were missing in sheet 2. I then noticed that in the code [here](https://github.com/Sage-Bionetworks/schematic/blob/b081276cda09b487e78b75814d6a2acb2c3d447c/schematic/manifest/generator.py#L937), the original `_request_dropdown_or_multi` function not only sets the data validation in manifest, but also populate valid values in sheet 2. To make sure that valid values still get populated in sheet 2 when multi-select is required, I wrapped the part related to populating valid values in a function called `def _store_valid_values_as_data_dictionary` (see [here](https://github.com/Sage-Bionetworks/schematic/pull/1040/files#:~:text=def%20_store_valid_values_as_data_dictionary)). Everything looks correct to me now.

GiaJordan (approved):

Issue #1038, opened 2022-11-29
Publish readthedocs documentation with github service account 


linglp:
**Is your feature request related to a problem? Please describe.**
Since #1027 is not possible, we would need to publish readthedocs by using `sphinx-click` with github service account. 

**Describe the solution you'd like**

**How important is this feature?** Select from the options below:
• 🏝 Low - it's an enhancement but not crucial for work

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

**Additional context**

PR #1047, opened 2022-12-05
Automatically updated CLI documentation on Readthedocs

linglp:
Related to #1038 

@milen-sage the workflow that Tom has (see [here](https://github.com/Sage-Bionetworks/challengeutils/blob/develop/.github/workflows/sphinx-ghpages.yaml)) could only publish documentation to github pages. But I noticed that readthedocs could let you publish documentation to readthedocs if you manually import the package and configure the webhook (without linking the project to your own github account). 

More relevant documentation could be found here: 
- About using readthedocs.yml: https://docs.readthedocs.io/en/stable/config-file/v2.html
- About having readthedocs to use `poetry` for installing dependencies: https://github.com/readthedocs/readthedocs.org/issues/4912
- About importing project to readthedocs without linking account (checkout the second point): https://sphinx-rtd-tutorial.readthedocs.io/en/latest/read-the-docs.html#:~:text=master%24%20git%20push-,Importing%20into%20ReadTheDocs,-Once%20you%20have
- About manually configure webhook on GitHub: https://docs.readthedocs.io/en/latest/integrations.html#:~:text=GitLab%2C%20or%20Bitbucket.-,Integration%20Creation,-%C2%B6


linglp:
For testing purpose, currently the docs would get automatically updated whenever there's a push request to `develop-docs` branch. But we could change the default branch on `readthedocs` to look at `develop` branch once `develop-docs` branch gets merged.

GiaJordan (approved):

Issue #1054, opened 2022-12-13
TT1 schematic: create a docker container for AWS deployment


milen-sage:
https://sagebionetworks.jira.com/browse/DCA-10

Issue #1053, opened 2022-12-13
TT1 schematic: GitHub workflow for building and registering schematic REST API docker container


milen-sage:
https://sagebionetworks.jira.com/browse/DCA-13

PR #1062, opened 2023-01-03
Added workflow to deploy schematic docker container in Github container registry

linglp:
For schematic deployment on AWS, this PR added a github workflow to deploy schematic docker container in Github container registry and also added logic in schematic to handle AWS credentials. 

schematic endpoints deployed on AWS could be found here: https://schematic.dnt-dev.sagebase.org/v1/ui/

related to #1054 #981 #1053


linglp:
@milen-sage For AWS deployment, we will have to add entrypoint or command specified in Dockerfile (see Bruce's comment [here](https://sagebionetworks.jira.com/browse/DCA-12?focusedCommentId=161271) in Jira). But I am not sure if that would fit our general use case since some users might want to use schematic docker container (without necessarily use it to run APIs). Should I add entry point in Dockerfile or should I create a separate docker file for AWS deployment?

GiaJordan (approved):

PR #1078, opened 2023-01-19
Added an endpoint to check entity type on Synapse and an endpoint to check if an entity is in the asset view

linglp:
This should in general help with debugging


mialy-defelice:
User tested and the endpoints are working as expected for me.

mialy-defelice (approved):
Looks good!

Issue #1065, opened 2023-01-06
`storage/project/manifests` incorrectly returns a data type to folders with no manifest


lakikowolfe:
**Describe the bug**
The `storage/project/manifests` endpoint returns a data type and manifest csv for folders that contain no manifest. The datatype and csv returned appear to be the same as the most recently returned manifest. When an empty folder is the first returned in the `storage/project/manifests` output it displays data type and manifest csv correctly.

**To Reproduce**
1. Call `storage/project/manifests` endpoint for HTAN BU (syn22124336)
2. Look at the returned list ([link to output as csv](https://github.com/Sage-Bionetworks/schematic/files/10356404/storage_project_manifests_bu.1.csv))

**Expected behavior**
The expected behavior is that folders that contain no manifest would have "" for dataset type and manifest csv

**Priority** (select one)
- [ ] Minor ⬇️
- [X] Major 📢
- [ ] Critical 🆘

**Screenshots**

**Desktop (if applicable, please complete the following information):**
 - OS: iOS
 - Browser: chrome
 - Version 

**Additional context**

PR #1084, opened 2023-01-19
Fixed `getProjectManifests` function in synapse storage 

linglp:
Related to the issue Loren reported #1065 

All the rows marked as yellow are rows that are incorrect. When there's no manifest (check the row with dataset_name `h_and_e`, column `dataset` and column `manifest_csv` should be empty. But as you could see from the screen shot, the code wrongly took the value from previous rows and put it there. 

Screen Shot 2023-01-19 at 5 37 43 PM

I think this is because we didn't reset "manifest_name" and "component" when running loop `for (datasetId, datasetName) in datasets`

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -571,9 +571,10 @@ class SynapseStorage(BaseStorage):
                             logging.warning(
                             f"Manifest {manifestId} is composed of multiple components. Schematic does not support mulit-component manifests at this time."
                             "Behavior of manifests with multiple components is undefined"
-                            )              
-
-            #save manifest list with applicable informaiton
+                            )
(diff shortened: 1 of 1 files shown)


mialy-defelice (changes requested):
Hey @linglp, this looks great so far. 

Did you mean to leave the print statements in `schematic/store/synapse.py` at line 598?

linglp:
> Hey @linglp, this looks great so far.
> 
> Did you mean to leave the print statements in `schematic/store/synapse.py` at line 598?

Thank you @mialy-defelice ! I have removed the print statement

mialy-defelice (approved):
Thanks @linglp!

Issue #1082, opened 2023-01-19
API endpoint for get_nodes_display_names


andrewelamb:
**Is your feature request related to a problem? Please describe.**
For schematic_db, we want to be able to be able to use the display name of components and attributes instead of the generated camel case name in the schema. 

**Describe the solution you'd like**
Expose SchemaGenorator.get_nodes_display_names() as endpoint

**How important is this feature?** Select from the options below:
• 🌋 Important - it's a blocker and can't do work without it

**When will use cases depending on this become relevant?** Select from the options below:
• Short-term - 2-4 weeks

**Additional context**

PR #1094, opened 2023-01-27
Develop api node display names

mialy-defelice:
Created an endpoint to expose the `schema/generator` function `get_nodes_display_names`.

Addresses issue #1082.


andrewelamb (approved):

linglp (approved):
@mialy-defelice I could confirm that this works locally, and I also added the test. Thank you!

PR #1101, opened 2023-01-31
Update `synapseClient` dependency and api for manifest table uploads

GiaJordan:
Update minimum required version of the synapse python client to `v2.7.0` for permanent fix to #870, revert temporary solve introduced in #902

PR #1092, opened 2023-01-24
Update schematic dependencies

GiaJordan:
Update dependencies in the `pyproject.toml` to the latest major or minor release to be current and make compatible with `schematic-db`. Also add `schematic-db` to dependencies

PR #1115, opened 2023-02-03
Set default of "table_manipulation" as "replace" in API endpoint when users enter None and updated tests

linglp:
I noticed that the `table_manipulation` is using `replace` as the default option. ~~If that's the case, table_manipulation parameter should be set as "required" so that users can't set an empty string or None to `table_manipulation` parameter on swagger UI~~.  To match with the downstream submit function, I set "table_manipulation" as an optional parameter, and if users don't set anything, we would pass "replace" by default. I also updated the description of the parameter for clarify. 

Context:  I discovered this issue when running tests for #1101

### DIFF
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -303,6 +303,7 @@ paths:
           schema:
             type: string
             enum: ["replace", "upsert"]
+          required: true
       operationId: api.routes.submit_manifest_route
       responses:
         "200":
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -300,10 +300,11 @@ paths:
           }]'
         - in: query
           name: table_manipulation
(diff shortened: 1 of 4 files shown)


api/openapi/api.yaml
```diff
@@ -303,6 +303,7 @@ paths:
           schema:
             type: string
             enum: ["replace", "upsert"]
+          required: true
```
linglp, line 306: If not setting this as "required", users could pass in None for "table_manipulation" which could end up breaking the downstream function.

mialy-defelice (changes requested):
Is the change to generator related to a different issue? Even though its so tiny maybe we should move it to a different PR for clarity?

linglp:
> Is the change to generator related to a different issue? Even though its so tiny maybe we should move it to a different PR for clarity?

Thanks for checking Mialy! The change to the generator is related to version updates of pandas. I was running tests for #1101, #1092 and realized that the pandas version has changed in the `develop` branch at one point (and I was still using an older version of pandas locally). After #1092 gets merged and `develop-synclient-branch` pulled from the new develop branch, this line won't show as a "change"

mialy-defelice (approved):
Looks great!

GiaJordan (approved):
Thanks @linglp! Just confirming, the necessary merge order is #1092, this one, and then #1101?

GiaJordan:
Also, @linglp from the title it seems like you're making `--table_manipulation` required, but from the code it seems like you're making it optional. Should we update the title?

PR #1124, opened 2023-02-07
Bugfix cryptography error

### LINKED ISSUE #1117: Unable to use schematic when updating packages to current lock file
**Describe the bug**
Issue and fix are described [here.](https://stackoverflow.com/questions/74981558/error-updating-python3-pip-attributeerror-module-lib-has-no-attribute-openss)

-  incompatibility between cryptography==39.0.0 and pyopenssl
-  fix is to downgrade to cryptography==38.0.4
-  to get it to work locally:

1.   -  `poetry add cryptography==38.0.4` 
2.   -  `poetry install`

-  may only happen on Ubuntu

**Priority** (select one)
- [ ] Minor ⬇️
- [x] Major 📢
- [ ] Critical 🆘

andrewelamb:
fixes #1117

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,7 @@ openpyxl = "^3.0.9"
 Flask-Cors = "^3.0.10"
 pdoc = "^12.2.0"
 dateparser = "^1.1.4"
+cryptography = "38.0.4"
 
 
 [tool.poetry.dev-dependencies]
--- a/poetry.lock
+++ b/poetry.lock
@@ -310,7 +310,7 @@ toml = ["tomli"]
 
 [[package]]
 name = "cryptography"
-version = "39.0.0"
+version = "38.0.4"
 description = "cryptography is a package which provides cryptographic recipes and primitives to Python developers."
 category = "main"
 optional = false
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -68,6 +68,7 @@ openpyxl = "^3.0.9"
 Flask-Cors = "^3.0.10"
 pdoc = "^12.2.0"
 dateparser = "^1.1.4"
+cryptography = "38.0.4"
```
GiaJordan, line 71: Do other minor releases of version `38.x.x` work or only this specific one? Could we open it up to `^38.0.4` or would that break things?

andrewelamb: Not sure! My guess is anything before 38.0.4 would work.

Here's another possible [fix](https://github.com/pyca/cryptography/issues/7959)

GiaJordan: I noticed that possibility in the SO thread too. It might be better to try leaving `cryptography` as the latest major version and adding `pyopenssl` to our .toml with the indicated version and above. I noticed in our .lock for the `pyopenssl` extra its `>=20.0.0` so that could be where the issue is coming from.

PR #1129, opened 2023-02-08
Add `use_schema_label` parameter to manifest submission endpoint, separate manifest submission and table upsert tests

GiaJordan:
Add `use_schema_label` paramter to manifest submission endpoint as a boolean with default value `true`, preserving previous behavior.
Also add new function `parse_bool` to parse inputs from api that are type `string` as type `bool` so that they can be used correctly by downstream functions.

Manifest submission api tests are passing, ~~excluding the upsert tests because it has not been implemented on this branch~~. Also manually tested by uploading manifests to synapse through swagger.

This PR also separates out the normal manifest submission tests from the upsert submission tests because of the requirements to only use `--use_display_label` and a different component from the other submission tests.

### DIFF
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -311,6 +311,13 @@ paths:
             type: string
             enum: ["replace", "upsert"]
           required: false
+        - in: query
+          name: use_schema_label
+          description: "Store attributes using the schema label (true, default) or store attributes using the display label (false). Attribute display names in the schema must not only include characters that are not accepted by Synapse. Annotation names may only contain: letters, numbers, '_' and '.'"
+          schema:
+            type: boolean
+            default: true
+          required: true
       operationId: api.routes.submit_manifest_route
       responses:
         "200":
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -317,7 +317,7 @@ paths:
           schema:
             type: boolean
             default: true
-          required: true
+          required: false
       operationId: api.routes.submit_manifest_route
       responses:
         "200":
--- a/api/routes.py
+++ b/api/routes.py
@@ -379,7 +379,12 @@ def submit_manifest_route(schema_url, asset_view=None, manifest_record_type=None
 
     input_token = connexion.request.args["input_token"]
 
-    use_schema_label = parse_bool(connexion.request.args["use_schema_label"])
+
+    use_schema_label = connexion.request.args["use_schema_label"]
+    if use_schema_label == 'None':
+        use_schema_label = True
+    else:
+        use_schema_label = parse_bool(use_schema_label)
 
     if not table_manipulation: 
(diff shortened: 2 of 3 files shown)


api/openapi/api.yaml
```diff
@@ -311,6 +311,13 @@ paths:
             type: string
             enum: ["replace", "upsert"]
           required: false
+        - in: query
+          name: use_schema_label
+          description: "Store attributes using the schema label (true, default) or store attributes using the display label (false). Attribute display names in the schema must not only include characters that are not accepted by Synapse. Annotation names may only contain: letters, numbers, '_' and '.'"
+          schema:
+            type: boolean
+            default: true
+          required: true
```
milen-sage, line 320: Per @linglp comment set to optional

api/routes.py
```diff
@@ -161,7 +161,15 @@ def convert_json_file_to_csv(self, file_key):
             temp_path = save_file(file_key='file_name')
             return temp_path
         
-
+def parse_bool(str_bool):
```
milen-sage, line 164: Maybe try casting

GiaJordan: @milen-sage Per discussion on slack with @andrewelamb, suggestions were to try converting the string entry parameter by using the `.astype()` method in [this way](https://github.com/Sage-Bionetworks/schematic_db/blob/develop/schematic_db/rdb_updater/rdb_updater.py#L144) or in [this way](http://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.astype.html#). We decided that neither of these were optimal as they are methods for `Dataframe` and `Series` objects, and that `string` type objects do not have such a built in method.

We also considered the use of [this method](https://www.educative.io/answers/what-is-the-numpycharchararrayastype-method-in-python), but decided that it was also not the best fit as it was for `numpy.array` objects and not `string`.

Previously, I had tried using `bool('true')` or `bool('false')` to convert the string parameter entry to boolean and found this ineffective as non-empty strings always resolve to `True` and only empty strings resolve to `False`. We tried to see if this behavior was case sensitive and trialed `bool('True')` and `bool("False")`, and found the behavior to be the same as above.

The last option would be to do mapping like in [this example](https://statisticsglobe.com/convert-string-boolean-pandas-dataframe-column-python). We could adapt it to use a dictionary mapping for the one string, but the currently implemented function has the benefit of being accessible to other endpoints that would need booleans parsed, like if we had to add the `restrict_rules` parameter to the `/model/validate` endpoint.

milen-sage: Yes, given all, the way you have it works - thank you for looking into this!

milen-sage (approved):

GiaJordan:
@afwillia this adds a new parameter to the api, but it is optional and the old behavior is preserved as default so you shouldn't need to update anything on your end unless you want to make this configurable

PR #1130, opened 2023-02-09
Remove `try: catch:` block from manifest submission command function

GiaJordan:
Remove `try: catch:` block in `commands.submit_manifest` and verify that errors are handled elsewhere in the codebase at the appropriate function.

### DIFF
--- a/schematic/models/commands.py
+++ b/schematic/models/commands.py
@@ -117,35 +117,35 @@ def submit_manifest(
         inputMModelLocation=jsonld, inputMModelLocationType=model_file_type
     )
 
-    try:
-        manifest_id = metadata_model.submit_metadata_manifest(
-            path_to_json_ld = jsonld,
-            manifest_path=manifest_path,
-            dataset_id=dataset_id,
-            validate_component=validate_component,
-            manifest_record_type=manifest_record_type,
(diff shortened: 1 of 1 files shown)


GiaJordan:
@afwillia I wanted to check with you on the format of one of the errors raised and how it interacts with the DCA. The `ValidationError` that gets raised when there are validation errors in the manifest has a message but also includes the complete list of errors, in their individual list forms. This is an example message for one error in the manifest being validated:
```
Manifest could not be validated under provided data model. Validation failed with the following errors: [[2, 'MockRDB_id', 'On row 2 the attribute MockRDB_id does not contain the proper value type int.', 'str']
```
In cases with many errors, this could overwhelm users who get it depending on how the error is displayed. I'm not sure currently how this error is displayed DCA, but if it's necessary I can remove the error list from the error raised for the submission method.

afwillia:
@GiaJordan DCA will show each error on a new line but if it's problematic we can limit the number of errors it shows. This seems okay to me. Thanks for checking.

GiaJordan:
@afwillia perfect. Thanks for reviewing!

mialy-defelice (approved):
Looks great!

PR #1131, opened 2023-02-10
Display exceptions raised during validation with Great expectations, allow exclusion of upper bound OR lower bound for `inRange` rule

GiaJordan:
- Raise `GreatExpectationsError` if an exception is raised during the GE expectation suite validation
- Allow comparison across types for `inRange` rule
  - _note: while comparisons across types are now permitted, non-numerical values will raise validation errors/warnings_
- Allow exclusion of upper bound OR lower bound for `inRange` rule
- Re-order logic for message level determination so that user spec overrides default behavior and whether attribute is required or not

### DIFF
--- a/schematic/models/GE_Helpers.py
+++ b/schematic/models/GE_Helpers.py
@@ -272,8 +272,8 @@ class GreatExpectationsHelpers(object):
                     
                     elif base_rule==("inRange"):
                         args["mostly"]=1.0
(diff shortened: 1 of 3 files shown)


andrewelamb (approved):

GiaJordan:
> @GiaJordan for the in range rule if I wanted to have a min of 0 and not a max it would look like:
> 
> `inRange 0 None error`

Yep!

PR #1139, opened 2023-02-15
Update Documentation - python/package versions and POCs

GiaJordan:
Update documentation to reflect current changes and developers


milen-sage (approved):
Thanks @GiaJordan! We can revisit compressing the docs later - this is a lot of information; and hopefully we'll streamline the install process so that details like that rarely needed by users. But for now it looks good.

PR #1100, opened 2023-01-30
Parallelize operations in `load_df`

GiaJordan:
Noticed slower behavior on large manfiests during the development of #1073 with an iAtlas manifest. This PR parallelizes the operations of `load_df` across the available CPU cores on a machine, also using available memory to transfer information.

With a manifest containing ~11100 rows and three columns, performance for loading and parsing a manifest on a local machine with 4 CPU cores and 16GB memory was as follows:
<table>
<thead>
<tr>
<th></th>
<th>develop</th>
<th>develop-df-parallelization</th>
</tr>
</thead>
<tbody>
<tr>
<td>With other programs running</td>
<td>435.88s</td>
<td>228.22s</td>
</tr>
<tr>
<td>Without other programs running</td>
<td>303.97s</td>
<td>175.64s</td>
</tr>
</tbody>
</table>

This behavior could be further improved by using a larger EC2 instance with more CPU cores and RAM, so would have greater impacts for the single instance of the DCA and API running remotely

PR #1142, opened 2023-02-16
Merge AWS deployment branch to develop

linglp:
Highlights of changes:
* file structure of APIs (see `api.yaml` and the api folder). The structure was adopted from Thomas's mono repo: https://github.com/Sage-Bionetworks/sage-monorepo/tree/main/apps/schematic/api
* calculate cache on aws (see df_utils.py and general.py)
* new Dockerfile for deploying to AWS
* purging cache on AWS (see synapse.py)
* Added uWSGI
* Modified documentation related to running APIs
* Added parallelize operations for manifest validation: https://github.com/Sage-Bionetworks/schematic/pull/1100

All the PRs that got merged to this branch: 
* https://github.com/Sage-Bionetworks/schematic/pull/1144
* https://github.com/Sage-Bionetworks/schematic/pull/1168
* https://github.com/Sage-Bionetworks/schematic/pull/1178
* https://github.com/Sage-Bionetworks/schematic/pull/1190
* https://github.com/Sage-Bionetworks/schematic/pull/1192


linglp:
@GiaJordan and @mialy-defelice For yesterday's demo, I realized that the reason that it didn't work was because that I updated my .env file for a different PR, and I forgot to change it back. If you have some time, could you follow the instructions [here](https://github.com/Sage-Bionetworks/schematic/blob/develop-add-uWSGI-two/schematic_api/api/README.md) and see if you could have the containers up and running and use the APIs? If not, let me know. Thank you.

linglp:
Here are the steps to make sure that the code works by utilizing github code space:
* open github code space by using branch `develop-add-uwsgi-two`
* launch virtual env by using `poetry shell`
* install dependencies by doing: `poetry install`
* create a copy of .env-example and rename it as .env. Replace empty string in SERVICE_ACCOUNT_CREDS by using content of service_account_creds.json
* running `docker compose up` (this step will take quite some time). For me, it took 700-800s. 
* After the image got exported, github would give you a hint open the API in a separate browser. For me, the url looks like: https://linglp-stunning-system-757g4xx9vx5frrrp-7080.preview.app.github.dev. I could then added `v1/ui` to the end of the url and visit: `https://linglp-stunning-system-757g4xx9vx5frrrp-7080.preview.app.github.dev/v1/ui/` for uwsgi set up or '`https://linglp-stunning-system-757g4xx9vx5frrrp-3001.preview.app.github.dev/v1/ui/`' for normal set up. Please note that your browser might try to redirect you to `localhost:7080` or `localhost:3001`, but that won't work because the APIs are not running locally. 
* When I then ran `manifest/generate` endpoint, it worked as expected

GiaJordan:
I was able to get a container running and get to the api. I'll try using it today or tomorrow, and leave comments about the process

schematic/store/synapse.py
```diff
@@ -1,10 +1,12 @@
+from datetime import datetime, timedelta
 from copy import deepcopy
 import os
 import uuid  # used to generate unique names for entities
 import json
 import atexit
 import logging
 import secrets
+from dataclasses import dataclass
```
GiaJordan, line 9: Do you think it would be useful in the future to use pydantic's `dataclass` for the validation aspect?

schematic/store/synapse.py
```diff
@@ -1724,7 +1818,6 @@ def getDatasetProject(self, datasetId: str) -> str:
             str: The Synapse ID for the parent project.
         """
 
-        self._query_fileview()
```
GiaJordan, line 1727: Removing this will potentially cause issues for table operations when in use, and will likely cause issues with tests covering the function of table operations

GiaJordan (changes requested):

mialy-defelice:
> * When I then ran `manifest/generate` endpoint, it worked as expected
Okay I was able to get the expected output as well for the docker image.

env.example
```diff
@@ -0,0 +1,5 @@
+# API server
+SERVER_PROTOCOL=http://
+SERVER_DOMAIN=localhost
+SERVER_PORT=7080
+SERVICE_ACCOUNT_CREDS='Provide service account creds'
```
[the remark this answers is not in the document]
mialy-defelice: An additional note for me is it required a single ' instead of " for some reason

GiaJordan:
I'm able to use most of the endpoints I tried, but had issues validating a manifest due to credentials

env.example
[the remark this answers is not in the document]
linglp: Thanks for the feedbacks. I will update the ReadMe.

schematic/store/synapse.py
linglp: I have not yet looked into the details of dydantic's dataclass. But that's something we could consider for sure.

pyproject.toml
```diff
@@ -69,8 +69,10 @@ openpyxl = "^3.0.9"
 Flask-Cors = "^3.0.10"
 pdoc = "^12.2.0"
 dateparser = "^1.1.4"
-schematic-db = {version = "0.0.6", extras = ["synapse"]}
+pandarallel = "^1.6.4"
+schematic-db = {version = "^0.0.6", extras = ["synapse"]}
 pyopenssl = "^23.0.0"
+uWSGI = "^2.0.21"
```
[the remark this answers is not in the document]
linglp: Thanks. I have changed it to be optional.

schematic_api/api/routes.py
```diff
@@ -1,39 +1,37 @@
 from email import generator
```
mialy-defelice, line 1: This might be improved when using Black later, but in general we should be keeping imports in alphabetical order. They don't all need to be changed now, but when adding new imports try to keep this in mind.

schematic/store/synapse.py
```diff
@@ -50,8 +52,100 @@
 from schematic.exceptions import MissingConfigValueError, AccessCredentialsError
 
 from schematic import CONFIG
+from schematic.utils.general import entity_type_mapping
```
mialy-defelice, line 55: Please move this to be next to other `schematic.utils` imports.

schematic/store/synapse.py
mialy-defelice, line 1: In the future I think changes like this should be done in a separate PR and pushed to develop then pulled into your AWS dev branch since its touching core Schematic features. This will make sure things do not get out of sync and the PRS on this branch to push into dev don't get too big.

schematic/store/synapse.py
```diff
@@ -100,42 +194,55 @@ def __init__(
...
-            return "project"
-        else: 
-            return type_entity
+        # try clearing the cache
+        # scan a directory and check size of files
+        cache = self.syn.cache
+        if os.path.exists(root_dir):
+            # set maximum synapse cache allowed
+            # set maximum ephemeral stroage allowed on AWS
+            maximum_storage_allowed_cache_gb = 7
```
mialy-defelice, line 212: I think these storage maximums should be passed as args to the function, with the defaults being set as you have them here. That way in the future its easier for people to update the allowed cache sizes if they want to.

mialy-defelice (changes requested):

schematic/utils/general.py
```diff
@@ -26,3 +37,138 @@ def unlist(_list):
...
+
+def get_dir_size(path: str):
+    """calculate total size of a directory
+    Args:
+        path: path to a folder
+    return: total size of a directory
+    """
+    total = 0
+    # Example usage of os.scandir could be found here: https://docs.python.org/3/library/os.html#os.scandir
+    # Technically, scandir.close() is called automatically. But it is still advisable to call it explicitly or use the with statement.
```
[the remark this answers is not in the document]
linglp: Thanks for pointing it out Mialy. What about I changed it to something like: 
```
def get_dir_size(path: str):
    """Recursively descend the directory tree rooted at the top and call .st_size function to calculate size of files in bytes. 
    Args:
        path: path to a folder
    return: total size of all the files in a given directory in bytes. 
    """
    total = 0
    # Recursively scan directory to find entries
    with os.scandir(path) as it:
        for entry in it:
            if entry.is_file():
                total += entry.stat().st_size
            elif entry.is_dir():
                total += get_dir_size(entry.path)
    return total
```

schematic/utils/general.py
```diff
@@ -26,3 +37,138 @@ def unlist(_list):
...
+
+
+def convert_gb_to_bytes(gb: int):
+    """convert gb to bytes
+    Args:
+        gb: number of gb
+    return: total number of bytes
+    """
+    bytes_to_return = gb * 1024 * 1024 * 1024
+    return bytes_to_return
```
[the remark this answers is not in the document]
linglp: Yeah that works. Let me make that change.

schematic_api/api/README.md
```diff
@@ -1,11 +1,25 @@
...
 
 ```bash
 source .venv/bin/activate
 python3 run_api.py
 ```
+If you define `APP_PORT` as `3001` in `docker-compose.yml`, you should be able to see the application running when visiting `http://localhost:3001/v1/ui/`
+
+To start a docker container that runs flask application with uWSGI:
+1) Comment out line 4 to line 19 in `docker-compose.yml`. 
+2) Create .env file based on env.example. Fill in `service_account_creds.json`
```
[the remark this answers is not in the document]
linglp: Sounds good thanks Mialy. Let me update the instructions to make it more detailed.

schematic/store/synapse.py
linglp: Yeah sure. I could do that.

schematic/utils/df_utils.py
```diff
@@ -40,7 +42,12 @@ def load_df(file_path, preserve_raw_input=True, data_model=False, **load_args):
         #Cast the columns in dataframe to string while preserving NaN
         null_cells = org_df.isnull() 
         org_df = org_df.astype(str).mask(null_cells, '')
-        ints = org_df.applymap(lambda x: np.int64(x) if str.isdigit(x) else False, na_action='ignore').fillna(False)
+
+        if org_df.size < 1000:
```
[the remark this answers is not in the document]
GiaJordan: @linglp @mialy-defelice I'll update that PR and merge it first. I'll add comments, but don't think we should make the "large manifest" size cutoff configurable, since below a certain size performance is actually worse with parallelization, and I wouldn't expect users to know what the cutoff would be and don't want to have them to have to use trial and error to find something that works.

schematic_api/api/__main__.py
```diff
@@ -0,0 +1,18 @@
+import os
+import connexion
+from schematic import CONFIG
+from flask_cors import CORS
+from schematic_api.api import app
```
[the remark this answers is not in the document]
linglp: Thanks for the comment Mialy. I just cleaned it up. It turns out that we don't need the config and then CORS here any more.

schematic/utils/df_utils.py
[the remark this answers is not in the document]
mialy-defelice: @GiaJordan I misspoke, I meant add it as a variable not an arg. When numbers are hard coded like this I think it makes it more clear what's happening.

schematic_api/api/routes.py
```diff
@@ -170,7 +168,19 @@ def parse_bool(str_bool):
...
         )
-        
+
+def return_as_json(manifest_local_file_path):
+    manifest_csv = pd.read_csv(manifest_local_file_path)
+    try:
+        manifest_json = manifest_csv.to_dict(orient="records")
+        return manifest_json
+    except JSONDecodeError as e:
+        logger.error("Json data is not formatted correctly", e)
```
GiaJordan, line 178: Is this logging the original error and your message to whatever logger was initialized before?
It might be more helpful to [wrap the caught errors](https://peps.python.org/pep-3134/#explicit-exception-chaining) with the useful messages you've added so they're included in the stack trace instead of logging them to the logger where they might be missed, like how you've done in other parts of the code. Like:
```
except JSONDecodeError as e:
    SomeSpecificFormatError("Json data is not formatted correctly") from e
```
Apologizes if I've misunderstood the output of the log and it already does something similar!

schematic_api/api/routes.py
```diff
@@ -546,8 +561,47 @@ def get_viz_tangled_tree_layers(schema_url, figure_type):
...
+    """
+    # call config_handler()
+    config_handler()
+
+    # use Synapse Storage
+    store = SynapseStorage(input_token=input_token)
+    # try logging in to asset store
+    try:
+        syn = store.login(input_token=input_token)
+    except SynapseAuthenticationError as e:
```
GiaJordan, line 584: do we need to catch these errors if they only just get raised after being caught?

schematic/__init__.py
```diff
@@ -20,6 +20,7 @@
 # Suppress INFO-level logging from some dependencies
 logging.getLogger("keyring").setLevel(logging.ERROR)
 logging.getLogger("rdflib").setLevel(logging.ERROR)
+logging.getLogger("Synapse storage").setLevel(logging.INFO)
```
GiaJordan, line 23: Does the synapse Store need to be hardcoded to the `INFO` log level?

schematic/store/synapse.py
```diff
@@ -50,8 +52,100 @@
...
+        except(SynapseUnmetAccessRestrictions, SynapseAuthenticationError):
+            # if there's an error getting an uncensored manifest, try getting the censored manifest
+            if not manifest_df.empty:
+                censored_regex=re.compile('.*censored.*')
+                censored = manifest_df['name'].str.contains(censored_regex)
+                new_manifest_id=manifest_df[censored]["id"][0]
+                self.manifest_id = new_manifest_id
+                try: 
+                    manifest_data = self._download_manifest_to_folder()
+                except (SynapseUnmetAccessRestrictions, SynapseAuthenticationError) as e:
```
GiaJordan, line 128: Do these errors also need to be caught?

schematic_api/api/README.md
```diff
@@ -1,11 +1,31 @@
+## Installation
+For unix OSs users, simply run the following poetry command to install uWSGI: 
+```bash
+poetry install -E uwsgi
+```
+
 ## Setup
+There are two ways to run schematic APIs: 1) start a flask server and run your application locally (preferred for external users); 2) build a docker image that allows you to run a container with flask application with uWSGI (only works for unix OSs users or windows user with WSL that could install uWSGI)
```
GiaJordan, line 8: Thanks for adding in the note for Windows users!

schematic/utils/df_utils.py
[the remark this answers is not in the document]
GiaJordan: Ahh I see, yeah that would be better!

schematic/store/synapse.py
linglp: Yeah. That's a good idea. I will update the function. Thanks Mialy!

schematic_api/api/routes.py
linglp: Yeah I think here maybe I don't even need the try and except block. The error message that I wrote didn't really add much values

linglp: Let me do a light refactor of routes.py later this sprint and try clearing up the code.

schematic/utils/general.py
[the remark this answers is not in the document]
mialy-defelice: @linglp yep that looks great!

schematic/__init__.py
linglp: Hmm.. so I wanted to be able to log the part of purging AWS cache on AWS cloudwatch console: 
```
logger.info(f'{num_of_deleted_files} number of files have been deleted from {root_dir}')
```
and
```
logger.info(f'Estimated {remaining_space} bytes (which is approximately {converted_space}) remained in ephemeral storage after calculating size of .synapseCache excluding OS')
```
But I could also change `logger.info` to `logger.debug` if that's better.

schematic/store/synapse.py
linglp: hmm my logic here is: 
in `getDatasetManifest` function, first [`_get_manifest_id`](https://github.com/Sage-Bionetworks/schematic/blob/develop-add-uWSGI-two/schematic/store/synapse.py#L463) function would first try to figure out what manifest that we wants to download. 
* If censored and uncensored manifests are both available, we try downloading the uncensored manifest first. 
* If there's an error downloading the uncensored manifest because of access restricted (that's why I have the first try and except block [here](https://github.com/Sage-Bionetworks/schematic/blob/develop-add-uWSGI-two/schematic/store/synapse.py#L117)), try downloading the uncensored manifest first
* if there's still an error downloading the censored manifest, the error of not having access gets raised again, which is why I have the code here. 

I think the general logic is what we have in the old code. If I don't catch the errors here, those errors might be raised by python client. But I think here maybe I could emphasize that users don't have access to both censored and uncensored manifest? So something like: 
```
                except (SynapseUnmetAccessRestrictions, SynapseAuthenticationError) as e:
                    raise PermissionError("You don't have access to censored and uncensored manifests in this dataset.") from e
```

schematic_api/api/routes.py
linglp: hmm my understanding is that import should be organized based on the order: general standard libraries comes first, then 3rd party, then local packages. This is mentioned here: https://peps.python.org/pep-0008/#imports:~:text=Imports%20should%20be%20grouped%20in%20the%20following%20order%3A I am sure there are third party tools that could organize imports if black is not sufficient. We could discuss this in the code review

linglp: `from emails import generator` seems a bit strange. I removed it because it doesn't seem like this is being used

schematic/store/synapse.py
linglp: After I merged the latest develop to this branch, I then tested by running TestTableOperations locally. Even though the first two times, I got a strange lookup error, the error seemed to be resolved when I tried the second and third time.

schematic/store/synapse.py
GiaJordan: Ahh I see, that makes sense. Thanks for explaining!

schematic/__init__.py
GiaJordan: That makes sense that you'd want to log the cache purging. I was just concerned about hard coding the logging level for the whole store when schematic is initialized. Is there a way to set the logger when you're running it on AWS so that it isn't changed for other contexts or users?

linglp: hmm I don't think there is such a way especially that we are going to use `develop` for deployment

mialy-defelice (approved):
Thanks @linglp

GiaJordan (approved):

PR #1143, opened 2023-02-17
Redesigned API endpoints 

linglp:
This is related to the issue here: https://sagebionetworks.jira.com/browse/DCA-59

### DIFF
--- a/schematic_api/api/openapi/api.yaml
+++ b/schematic_api/api/openapi/api.yaml
@@ -10,10 +10,10 @@ servers:
   - url: /v1
 
 paths:
-  /manifest/generate:
+  /get/manifest/generate:
     get:
       summary: Endpoint to facilitate manifest generation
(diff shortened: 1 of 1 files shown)


linglp:
@tschaffter Any recommendation on designing file paths for API endpoints? I tried modifying the paths based on the documentation that I read here: https://oai.github.io/Documentation/best-practices.html

There are some endpoints that I am not sure if I name them correctly. For example: 
* I named an endpoint that for "converting a given display name string into a proper property label string" as `/get/schema/property_label/display_name`
* I named an endpoint that I used for getting a subgraph containing all edges of a given type as `/get/schema/graph/edge_type`

tschaffter:
@linglp Let's schedule a 30 min call to review the current design and discuss best practices.

linglp:
Close this PR for now since we are renaming the API endpoints in mono-repo

Issue #1148, opened 2023-02-21
Column headers are missing when using `manifest/generate` endpoint to return an existing manifest


linglp:
**Describe the bug**
The issue was first reported by @afwillia 
Use the following parameters: 
* dataset_id = syn36003517
* asset_view = syn33715412
* output_format = excel

**Expected behavior**

**Priority** (select one)
- [ ] Minor ⬇️
- [ ] Major 📢
- [x] Critical 🆘

**Screenshots**

**Desktop (if applicable, please complete the following information):**

**Additional context**

afwillia:
![no-column-names-patient](https://user-images.githubusercontent.com/8195343/220459231-d13d822f-dcca-4b51-9e1b-c44b4a0a098a.png)

afwillia:
also, the second sheet with the column specifications does not show up.

milen-sage:
@afwillia and @linglp I updated this issue to critical, since we'd have to resolve before rolling out.

@afwillia is this a new behavior (i.e. this worked as expected before and only recently w/ updates from schematic it stopped working)? I remember discussing how the "Google"-free way of generating Excel spreadsheets was reproducing the functionality of "Google-based" sheets.

I'm not so concerned about the 2nd tab with the dictionary, but the headers, descriptions, formatting, dropdowns, etc.. all seem to be missing. This looks like a csv version of the data, except column names are missing as well.

PR #1164, opened 2023-02-23
Fixed column headers problem when generating an Excel spreadsheet for getting an existing manifest 

linglp:
This is related to the issue #1148 

It turned out that this issue occurred because at one point we switched to a newer version of pandas, and our old syntax of populating an existing manifest using `pd.ExcelWriter ` no longer works. To modify an existing excel spreadsheet, we have to set `if_sheet_exists` parameter. 

Also, I noticed that in our old logic, if we generate an Excel spreadsheet. we have something like: 
```
        # Populate empty template with existing manifest
        if manifest_record:

            # TODO: Update or remove the warning in self.__init__() if
            # you change the behavior here based on self.use_annotations
                            
            # populate empty manifest with content from downloaded/existing manifest
            manifest_sh = self.set_dataframe_by_url(empty_manifest_url, manifest_record[1])

            # determine the format of manifest
            result = self._handle_output_format_logic(output_format = output_format, output_path = output_path, sheet_url = sheet_url, manifest_url = manifest_sh.url, empty_manifest_url=empty_manifest_url, dataframe = manifest_record[1])
            return result
```

This means that when a manifest record exists, we would first populate an empty manifest URL with content from an existing manifest before exporting the manifest to Excel. This shouldn't happen. To avoid sending metadata to Google, I changed the logic so that `set_dataframe_by_url` only gets used when the output format is set to Google sheet

To test, I did the following: 
* Tested it out manually on swagger UI: 
   - example data model
   - existing dataset id: syn51200926
   - data type: Patient
   - output format: excel
   - asset view: syn51200927

* Using CLI to get an existing manifest: `schematic manifest --config ~/path/to/config.yml get --output_xlsx 'test.xlsx' --dataset_id syn51200926`
* Ran tests related to generating a new manifest and getting an existing manifest in test_api.py

To do: 
* Add tests in test_api.py that check if Excel spreadsheet has column headers.


afwillia (approved):

linglp:
Note: I have tested it in uWSGI+flask by using my branch `develop-add-uWSGI-two` too and it worked.

schematic/store/synapse.py
```diff
@@ -378,6 +378,9 @@ def getDatasetManifest(
                 not_censored=~censored
                 if any(not_censored):
                     manifest_syn_id=manifest[not_censored]["id"][0]
+                # If only censored manifests are available, use the first censored manifest
+                else:
+                    manifest_syn_id=manifest["id"][0]
```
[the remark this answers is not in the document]
mialy-defelice: @linglp Yes lets create a new bug report and a new PR to isolate it away.

mialy-defelice (approved):
Looks good @linglp!

mialy-defelice:
> Could you review the code again and let me know if there's a better way of handling it? I plan to add some unit tests too.

I do not see any glaring issues so I trust that this is the best method of handling the issue. Can you possibly move this PR back to draft so that we can make sure to go through the whole review process again when you add the unit tests or maybe add the unit tests in a different PR?

linglp:
> > Could you review the code again and let me know if there's a better way of handling it? I plan to add some unit tests too.
> 
> I do not see any glaring issues so I trust that this is the best method of handling the issue. Can you possibly move this PR back to draft so that we can make sure to go through the whole review process again when you add the unit tests or maybe add the unit tests in a different PR?

Thanks Mialy. Let me convert it to a draft and let more tests.

mialy-defelice:
Hey Lingling, 
I am experiencing an odd issue. I am trying to grab the manifest that you point to in your example, the censored patient manifest. I tested in the API and CLI. From both I can get the filled google sheet. From the API I can also return a filled DF. But from neither can I get a filled excel file. I tried several times. I recently pulled as well.

schematic/manifest/generator.py
```diff
@@ -1554,12 +1559,9 @@ def get_manifest(
 
             # TODO: Update or remove the warning in self.__init__() if
             # you change the behavior here based on self.use_annotations
-                            
-            # populate empty manifest with content from downloaded/existing manifest
-            manifest_sh = self.set_dataframe_by_url(empty_manifest_url, manifest_record[1])
 
             # determine the format of manifest
-            result = self._handle_output_format_logic(output_format = output_format, output_path = output_path, sheet_url = sheet_url, manifest_url = manifest_sh.url, empty_manifest_url=empty_manifest_url, dataframe = manifest_record[1])
+            result = self._handle_output_format_logic(output_format = output_format, output_path = output_path, sheet_url = sheet_url, empty_manifest_url=empty_manifest_url, dataframe = manifest_record[1])
```
thomasyu888, line 1564: Nit: Personally, I think the logic can be improved which would simplify things a bit. 

1. Separate out the concerns
    1. If you want to return the dataframe -> Don't call this function at all
    1. You have 2 output formats: "excel", "sheet_url", one returns the Excel spreadsheet and one returns the sheet URL
1. Have 3 output formats: "excel", "dataframe", "sheet_url"

What I like about option 1 is...

1. You simplify the `_handle_output_format_logic`. It seems odd to pass in output_format as "dataframe" AND the dataframe only to just return the dataframe.
2. You simplify your function and the test you have to write for the function by just returning on return type (str)
3. You can remove `sheet_url`

What I like about option 2...
1. It might be the least change to the existing code
2. You can remove `sheet_url`

schematic/manifest/generator.py
```diff
@@ -1605,22 +1609,20 @@ def populate_existing_excel_spreadsheet(self, existing_excel_path: str = None, a
...
-            for i, col_name in enumerate(out_of_schema_columns_lst):
-                col_index = len(workbook_headers) + 1 + i
-                worksheet.cell(row=1, column=col_index).value = col_name
-        
-        # save and close
-        writer.save()
-        writer.close()
+            # if there are new columns, add them to the end of spreadsheet
+            # Note: previously, we tried looping through the out of schema columns and use worksheet.cell to modify a particular cell. But that functionality is no longer working. 
+            if len(out_of_schema_columns_lst) > 0:
```
thomasyu888, line 1619: Nit: Python has implicit booleanness.  [stackoverflow](https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty)

That said, I think it is fine to explicitly call this the `len() > 0`.  I just personally like to leverage the implicit booleanness which is "more Pythonic" `if out_of_schema_columns_lst`

schematic/manifest/generator.py
thomasyu888, line 1: Just wanted to say: You'll often see me use "Nit" across all the PR's I review.  What does this mean?

> Reviewers should always feel free to leave comments expressing that something could be better, but if it’s not very important, prefix it with something like "Nit:" to let the author know that it’s just a point of polish that they could choose to ignore.

> Code review can have an important function of teaching developers something new about a language, a framework, or general software design principles. Sharing knowledge is part of improving the code health of a system over time. Just keep in mind that if your comment is purely educational, but not critical to meeting the standards described in this document, prefix it with “Nit: “ or otherwise indicate that it’s not mandatory for the author to resolve it in this CL.

schematic/manifest/generator.py
linglp: Yeah. I agree with that technically, sheet_url is redundant. But Milen said that for keeping the code backward compatible, we should keep the option. I think maybe i could open a separate jira issue to clear this task.

schematic/manifest/generator.py
linglp: Thanks Tom. I like your approach. I will update the code to reflect it.

linglp:
> * --dataset_id syn51200926

Thank you for testing Mialy. I noticed that I messed up with indentation. If you don't mind, could you try testing it again? When testing with CLI using the example that I listed, don't forget to update the "master_fileview" in `config.yml`.

thomasyu888 (approved):
🔥 🔥 🔥  Thanks for including me on PR review!

PR #1168, opened 2023-02-24
Allowed CORS on given routes instead of all routes

linglp:
Based on the documentation [here](https://flask-cors.readthedocs.io/en/latest/#resource-specific-cors), this should allow cors on given routes. I tested this out by using `main` branch of schema visualization with HTAN and NF schema, and this seemed to work. 

Also related to GitHub issue here: https://sagebionetworks.jira.com/browse/DCA-67

### DIFF
--- a/schematic_api/api/routes.py
+++ b/schematic_api/api/routes.py
@@ -559,7 +559,7 @@ def get_viz_tangled_tree_text(schema_url, figure_type, text_format):
     
     return text_df
 
-@cross_origin(["http://localhost:8000/main.html", "https://sage-bionetworks.github.io/schema_visualization/"])
+@cross_origin(["http://localhost", "https://sage-bionetworks.github.io"])
 def get_viz_tangled_tree_layers(schema_url, figure_type):
 
     # call config_handler()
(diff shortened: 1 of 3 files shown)


schematic_api/api/routes.py
```diff
@@ -556,6 +559,7 @@ def get_viz_tangled_tree_text(schema_url, figure_type, text_format):
     
     return text_df
 
+@cross_origin(["http://localhost:8000/main.html", "https://sage-bionetworks.github.io/schema_visualization/"])
```
tschaffter, line 562: These should be domain names, not full URL to resources.

```suggestion
@cross_origin(["http://localhost", "https://sage-bionetworks.github.io"])
```

tschaffter (changes requested):

tschaffter:
@linglp Is the web client (DCA) going to be hosted on https://sage-bionetworks.github.io?

tschaffter:
Sounds good. We should only add extra origins to the CORS header when needed.

tschaffter (approved):

PR #1175, opened 2023-03-06
Develop add file only manifest submission option

mialy-defelice:
This PR serves to add an additional manifest submission option to Schematic. This option is to allow the user to submit a csv file only (no added entities). This option serves user needs and should also help speed up loading for the multi-tenant deployment. Option present in CLI and API.

Addition of this option on its own would have made the function `associateMetadataWithFiles` crowded, so this PR also includes a light refactor. I also changed the options for `manifest_record_type` so it is more clear what is happening (@afwillia) so users will likely have to update their language as well.

`manifest_record_type` options are:

- `table`: manifest uploaded to Synapse as a table and csv file. Manifests in both locations  contain an `entityId` column that is the synId of the table (name did not change)
- `file`: Manifest is uploaded to synapse as a file (CSV) (option did not previously exist)
- `file_w_entities`: Manifest is uploaded to Synapse as a file, with entities (option previously called `entity`)
-  `combo`: combines `file_w_entities` and `table` (option previously called `both`)

An example submission (just add a real datasetId).
`schematic model -c config.yml submit -mp tests/data/mock_manifests/Valid_Test_Manifest.csv -d synID -mrt file_w_entities`


mialy-defelice:
Currently troubleshooting an issue with table upserts.

schematic/store/synapse.py
```diff
@@ -1162,70 +1138,383 @@ def associateMetadataWithFiles(
...
+            component_name (str): Name of the manifest component (if applicable)
+        """
         # Create table name here.
         if 'Component' in manifest.columns:
             component_name = manifest['Component'][0].lower()
             table_name = component_name + '_synapse_storage_manifest_table'
         else:
             component_name = ''
-            table_name = 'synapse_storage_manifest_table'
+            tabsle_name = 'synapse_storage_manifest_table'
```
GiaJordan, line 1157: Was this meant to be chagned?

schematic/models/commands.py
```diff
@@ -79,8 +79,8 @@ def model(ctx, config):  # use as `schematic model ...`
 @click.option(
     "--manifest_record_type",
     "-mrt",
-    default='both',
-    type=click.Choice(['table', 'entity', 'both'], case_sensitive=True),
+    default='combo',
+    type=click.Choice(['table', 'file', 'file_w_entities', 'combo'], case_sensitive=True),
```
GiaJordan, line 83: Might be worth having similar options: `file`, `file_w_entity`, `file_w_table`, `file_w_both`
Also we should consider renaming `table` to `DB` or something referencing a database in the future to match naming in synapse store and future asset store refactor and by synapse agnostic

schematic/store/synapse.py
```diff
@@ -1162,70 +1137,383 @@ def associateMetadataWithFiles(
...
+        # Upload manifest to synapse based on user input (manifest_record_type)
+        if manifest_record_type == "table":
+            manifest_synapse_file_id = self.upload_manifest_as_table(
+                                        se,
+                                        schemaGenerator,
+                                        manifest,
+                                        metadataManifestPath,
+                                        datasetId,
+                                        table_name,
+                                        component_name,
```
GiaJordan, line 1466: Ref to discussion about switching to kwargs to aid future development

GiaJordan (changes requested):

linglp:
@mialy-defelice Do you think if you could update the API test here: https://github.com/Sage-Bionetworks/schematic/blob/develop/tests/test_api.py#:~:text=def%20test_submit_manifest(self%2C%20client%2C%20syn_token%2C%20data_model_jsonld%2C%20json_str%2C%20test_manifest_csv%2C%20use_schema_label)%3A

GiaJordan:
@GiaJordan will work with @mialy-defelice to update docs for asset store refactor
@MiekoHash

mialy-defelice:
> @mialy-defelice Do you think if you could update the API test here: https://github.com/Sage-Bionetworks/schematic/blob/develop/tests/test_api.py#:~:text=def%20test_submit_manifest(self%2C%20client%2C%20syn_token%2C%20data_model_jsonld%2C%20json_str%2C%20test_manifest_csv%2C%20use_schema_label)%3A

Like adding additional tests?

schematic/models/commands.py
mialy-defelice: I like the options you laid out, making it more clear that files are always present.
I very much dislike `both` and `combo`...what if it was like: `file`, `file_and_entites`, `file_and_table`, `file_table_and_entities`. Or do we always assume file...then have `manifest_record_addition` instead of `manifest_record_type` and just accept `entities`, `table`, `entities_and_table`? @milen-sage

GiaJordan: I'd prefer the 2nd option, that seems more similar to our old convention of delineating only what is added in addition to the file. It could then be `entities`, `table`, `entities_and_table`, and then something to the effect of `None` for the file only upload. 
But if we prefer the first option then `file_table_and_entities` is better than `both`! I feel being more explicit is better in that case.

mialy-defelice: I'll wait on @milen-sage 's input before making any changes.

linglp:
> > @mialy-defelice Do you think if you could update the API test here: https://github.com/Sage-Bionetworks/schematic/blob/develop/tests/test_api.py#:~:text=def%20test_submit_manifest(self%2C%20client%2C%20syn_token%2C%20data_model_jsonld%2C%20json_str%2C%20test_manifest_csv%2C%20use_schema_label)%3A
> 
> Like adding additional tests? The submit tests in the api use `mrt=table` so it didn't need to be changed.

Yeah could you add additional tests for  at least the `file` option? I want to make sure that tests are update-to-date with the latest parameter (especially all the common parameters!)

api/openapi/api.yaml
```diff
@@ -263,7 +263,7 @@ paths:
           name: manifest_record_type
           schema:
             type: string
-            enum: [ "table", "entity", "both"]
+            enum: ["table", "file", "file_w_entities", "combo"]
           description: Manifest storage type.
```
linglp, line 267: Can you add here in the description about what the default is for user? In swagger UI, user could still select an empty field since this is not a required parameter.

linglp: You might also want to add logic to handle empty selection here in `submit_manifest_route` function in routes.py

mialy-defelice: I can add description of the default (`combo`).

The default is actually handed in `synapse.py` so its fine if a user does not fill out the `manifest_record_type`. This makes the default uniform across API and CLI.

linglp: I tested with not selecting anything for `manifest_record_type`, and the code would fail: 
```
  File "/Users/lpeng/Documents/schematic-git2/schematic/api/routes.py", line 397, in submit_manifest_route
    manifest_id = metadata_model.submit_metadata_manifest(
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/models/metadata.py", line 349, in submit_metadata_manifest
    manifest_id = syn_store.associateMetadataWithFiles(
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/store/synapse.py", line 1517, in associateMetadataWithFiles
    return manifest_synapse_file_id
UnboundLocalError: local variable 'manifest_synapse_file_id' referenced before assignment
```

By not selecting anything, I meant something like this: 
Screen Shot 2023-03-10 at 11 20 25 PM

mialy-defelice: Thanks for catching this, I didn't realize non-responses were handled so differently between the CLI and API. I'll address it!

mialy-defelice:
@GiaJordan @linglp 

The most recent pushes should address the comments above. I waited on changing the option names further till discussion with @milen-sage.

schematic/store/synapse.py
```diff
@@ -961,12 +959,15 @@ def format_row_annotations(self, se, sg, row, entityId, useSchemaLabel, hideBlan
...
-                    annos[anno_k] = anno_v
+                try:
+                    if isinstance(anno_v,float) and np.isnan(anno_v):
+                            annos[anno_k] = ""
+                    elif isinstance(anno_v,str) and re.fullmatch(csv_list_regex, anno_v) and rule_in_rule_list('list', sg.get_node_validation_rules(anno_k)):
+                        annos[anno_k] = anno_v.split(",")
+                    else:
+                        annos[anno_k] = anno_v
+                except:
+                    breakpoint()
```
GiaJordan, line 970: Sorry I didn't catch before, but should we raise a specific error here @mialy-defelice?

schematic/models/commands.py
milen-sage: We always assume file manifests are uploaded in addition to any other options. But I still like calling the manifest_record_type as 'file' or 'file_only' explicitly to indicate that only a manifest file would be uploaded. This can be our default setting. If we go with 'file_only' I think having something along the lines of 'table' or 'table_file' and 'table_entities' will do the trick. We will eventually deprecate 'table_entities'. I'll leave the wording of these parameters to you; we don't need to get stuck on that though as long as the naming is somewhat reasonable and the options are documented and work, CLI/API users will figure it out.

schematic/store/synapse.py
mialy-defelice: Oops I think I had hit a bug at one point and was just debugging, this was not an actual test. I'll remove it, thanks for catching it!!

schematic/models/commands.py
mialy-defelice: @milen-sage should we change the default in this PR or in a new one?

I will go with the following names:
[`file_only`, `file_and_entities`, `table_and_file`, `table_file_and_entities`] just so its more readable and since we aren't limited for space right now.

mialy-defelice:
@GiaJordan @linglp @milen-sage All comments should be addressed with the latest updates.

GiaJordan (approved):
Looks good!

PR #1177, opened 2023-03-07
Added a parameter to control if GE gets used when using `manifest/validate` endpoint 

linglp:
Added a parameter `restrict_rules` to control if GE should be used when using `manifest/validate` endpoint. This is needed for testing API on AWS and also bypassing reticulate error for Anthony


mialy-defelice (approved):
Looks good!

PR #1180, opened 2023-03-07
Propagate logger level entered in from command line to other schematic modules

GiaJordan:
Currently, passing in a parameter for `--verbosity` from the command line will only set the verbosity for the origin `schematic.xx.commands` module. For instance, if `--verbosity DEBUG` was set when running the `validate` command and `logging.debug` statements were added, only the statements in `models.commands.py` would show, and not in `metadata.py` or `df_utils.py` that are called subsequently.
Changes made in this PR takes the value passed in and uses it to set the verbosity of the logger for the `schematic` module, so that all submodules called inherit the appropriate level. The `schematic` module logger was set instead of the module under `schematic` so that if `utils` or modules outside the original `commands` module were called, loggers would still inherit the correct level.

If necessary, these changes can also be added to the testing modules files at a later time to allow for debugging/info messages as well

### DIFF
--- a/schematic/manifest/commands.py
+++ b/schematic/manifest/commands.py
@@ -15,7 +15,7 @@ from schematic.schemas.generator import SchemaGenerator
 from schematic.utils.google_api_utils import export_manifest_csv, export_manifest_excel, export_manifest_drive_service
 from schematic.store.synapse import SynapseStorage
 
-logger = logging.getLogger()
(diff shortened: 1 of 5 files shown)


mialy-defelice (approved):
This is great! Tested using the `submit` command and `DEBUG` verbosity level to actually see all the `debug` comments, compared to current `develop` branch.

PR #1190, opened 2023-03-12
Added function to calculate and clear cache 

linglp:
Added functionality to calculate and clear cache if exceeds estimated ephemeral storage limit. This is related to: [DCA-101](https://sagebionetworks.jira.com/browse/DCA-101)

[DCA-101]: https://sagebionetworks.jira.com/browse/DCA-101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -105,7 +105,6 @@ class SynapseStorage(BaseStorage):
         # try clearing the cache
         # scan a directory and check size of files
         if os.path.exists("/var/www/.synapseCache"):
-            #nbytes = sum(d.stat().st_size for d in os.scandir('/var/www/.synapseCache') if d.is_file())
             nbytes = get_dir_size("/var/www/.synapseCache")
             # if 19.5 GB has already been taken, purge cache prior to today
             if nbytes >= 20937965568:
@@ -126,7 +125,7 @@ class SynapseStorage(BaseStorage):
             if self.project_scope:
                 self.storageFileviewTable = self.syn.tableQuery(
                     f"SELECT * FROM {self.storageFileview} WHERE projectId IN {tuple(self.project_scope + [''])}"
-                , resultsAs="rowset").asDataFrame()
+                    ).asDataFrame()
             else:
                 # get data in administrative fileview for this pipeline
                 self.storageFileviewTable = self.syn.tableQuery(
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1,3 +1,4 @@
+import datetime
 from copy import deepcopy
 import os
 import uuid  # used to generate unique names for entities
@@ -16,7 +17,7 @@ import pandas as pd
 import re
 import synapseclient
 from time import sleep
-from ..utils.general import get_dir_size, convert_size
+from schematic.utils.general import get_dir_size, convert_size
 from synapseclient import (
     Synapse,
     File,
@@ -50,7 +51,6 @@ from schematic.store.base import BaseStorage
 from schematic.exceptions import MissingConfigValueError, AccessCredentialsError
 
 from schematic import CONFIG
-import datetime
 
 logger = logging.getLogger(__name__)
 
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -52,7 +52,7 @@ from schematic.exceptions import MissingConfigValueError, AccessCredentialsError
 
 from schematic import CONFIG
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("Synapse storage")
 
 class SynapseStorage(BaseStorage):
     """Implementation of Storage interface for datasets/files stored on Synapse.
@@ -115,12 +115,12 @@ class SynapseStorage(BaseStorage):
             if nbytes >= 20937965568:
                 an_hour_earlier = datetime.strftime(datetime.utcnow()- timedelta(hours = 1), '%s')
                 num_of_deleted_files = cache.purge(before_date = int(an_hour_earlier))
-                print(f'{num_of_deleted_files} number of files have been deleted from {root_dir}')
+                logger.info(f'{num_of_deleted_files} number of files have been deleted from {root_dir}')
             else:
                 # print remaining ephemeral storage on AWS 
                 remaining_space = 21474836480 - nbytes
                 converted_space = convert_size(remaining_space)
-                print(f'Estimated {remaining_space} bytes (which is approximately {converted_space}) remained in ephemeral storage after calculating size of .synapseCache.')
+                logger.info(f'Estimated {remaining_space} bytes (which is approximately {converted_space}) remained in ephemeral storage after calculating size of .synapseCache.')
 
     def _query_fileview(self):
         self._purge_synapse_cache()
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -102,14 +102,14 @@ class SynapseStorage(BaseStorage):
         self._query_fileview()
 
     def _purge_synapse_cache(self, root_dir: str = "/var/www/.synapseCache/"):
-        '''
+        """
         Purge synapse cache if it exceeds 7GB
         Args:
             root_dir: directory of the .synapseCache function
         Returns: 
             if size of cache reaches 7GB, return the number of files that get deleted
             otherwise, return the total remaining space (assuming total ephemeral storage is 20GB on AWS )
-        '''
+        """
         # try clearing the cache
         # scan a directory and check size of files
         cache = self.syn.cache
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -101,23 +101,25 @@ class SynapseStorage(BaseStorage):
 
         self._query_fileview()
 
-    def _purge_synapse_cache(self):
+    def _purge_synapse_cache(self, root_dir: str = "/var/www/.synapseCache/"):
         '''
-        Purge synapse cache if it exceeds 7 GB
+        Purge synapse cache if it exceeds 7GB
+        Args:
+            root_dir: directory of the .synapseCache function
+        Returns: 
+            if size of cache reaches 7GB, return the number of files that get deleted
+            otherwise, return the total remaining space (assuming total ephemeral storage is 20GB on AWS )
         '''
-        # define the root directory
-        # set maximum synapse cache allowed
-        # set maximum ephemeral stroage allowed on AWS
-        root_dir = "/var/www/.synapseCache/"
-        maximum_storage_allowed_cache_gb = 7
-        maximum_storage_allowed_cache_bytes = convert_gb_to_bytes(maximum_storage_allowed_cache_gb)
-        total_ephemeral_storag_gb = 20
-        total_ephemeral_storage_bytes = convert_gb_to_bytes(total_ephemeral_storag_gb)
-
         # try clearing the cache
         # scan a directory and check size of files
         cache = self.syn.cache
         if os.path.exists(root_dir):
+            # set maximum synapse cache allowed
+            # set maximum ephemeral stroage allowed on AWS
+            maximum_storage_allowed_cache_gb = 7
+            maximum_storage_allowed_cache_bytes = convert_gb_to_bytes(maximum_storage_allowed_cache_gb)
+            total_ephemeral_storag_gb = 20
+            total_ephemeral_storage_bytes = convert_gb_to_bytes(total_ephemeral_storag_gb)
             nbytes = get_dir_size(root_dir)
             # if 7 GB has already been taken, purge cache before 15 min
             if nbytes >= maximum_storage_allowed_cache_bytes:
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -104,8 +104,9 @@ class SynapseStorage(BaseStorage):
     def _purge_synapse_cache(self):
         # try clearing the cache
         # scan a directory and check size of files
-        if os.path.exists("/var/www/.synapseCache"):
-            nbytes = get_dir_size("/var/www/.synapseCache")
+        root_dir = self.syn.cache.cache_root_dir
+        if os.path.exists(root_dir):
+            nbytes = get_dir_size(root_dir)
             # if 19.5 GB has already been taken, purge cache prior to today
             if nbytes >= 20937965568:
                 c = self.syn.core.cache.Cache()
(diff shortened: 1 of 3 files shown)


schematic/store/synapse.py
```diff
@@ -100,14 +101,32 @@ def __init__(
...
+                today = datetime.strftime(datetime.utcnow(), '%s')
+                c.purge(before_date = int(today))
+            else:
+                # print remaining ephemeral storage on AWS 
+                remaining_space = 21474836480 - nbytes
+                converted_space = convert_size(remaining_space)
+                print(f'Estimated {remaining_space} bytes (which is approximately {converted_space}) remained in ephemeral storage after calculating size of .synapseCache.')
+
     def _query_fileview(self):
+        self._purge_synapse_cache()
```
thomasyu888, line 122: @linglp the function makes sense, but, could this be a problem if you're running this function in parallel?  What happens when a query happens and then a purge occurs immediately after?

I'm not sure if as.DataFrame would work. I think it could be worthwhile to remove the folder size condition and test this function in a docker container (so you don't mess with your own cache).   Run this command in parallel many times to see if you run into an error.

schematic/store/synapse.py
```diff
@@ -100,14 +101,32 @@ def __init__(
 
         self._query_fileview()
 
+    def _purge_synapse_cache(self):
+        # try clearing the cache
+        # scan a directory and check size of files
+        if os.path.exists("/var/www/.synapseCache"):
```
thomasyu888, line 107: There is a way of getting the synapseCache dir using the synapse connection object.  If that exists, maybe use that.

schematic/store/synapse.py
```diff
@@ -100,14 +101,32 @@ def __init__(
...
     def _query_fileview(self):
+        self._purge_synapse_cache()
         try:
             self.storageFileview = CONFIG["synapse"]["master_fileview"]
             self.manifest = CONFIG["synapse"]["manifest_basename"]
             if self.project_scope:
                 self.storageFileviewTable = self.syn.tableQuery(
                     f"SELECT * FROM {self.storageFileview} WHERE projectId IN {tuple(self.project_scope + [''])}"
-                ).asDataFrame()
+                , resultsAs="rowset").asDataFrame()
```
thomasyu888, line 129: Could you explain the behavior when you do `resultsAs=rowset` here when you do an asDataFrame at the end?

schematic/utils/general.py
```diff
@@ -26,3 +27,30 @@ def unlist(_list):
         return _list[0]
     else:
         return _list
+
+def get_dir_size(path='.'):
```
thomasyu888, line 31: Nit (nitpicky): I don't think this needs to be an key argument, it can probably be positional and required when you call this function.

get_dir_size(path)

schematic/utils/general.py
```diff
@@ -26,3 +27,30 @@ def unlist(_list):
...
+    '''
+    calculate total size of a directory
+    args: 
+    path: path to a folder or directory
+    '''
+    total = 0
+    with os.scandir(path) as it:
```
thomasyu888, line 38: The most common usecase of context managers are for patching/mocking and opening files in Python.  Due to this, I set out to find out why a context manager should be used here and this is what I found: https://github.com/python/cpython/issues/70182

Nice!

schematic/utils/general.py
```diff
@@ -26,3 +27,30 @@ def unlist(_list):
...
+    '''
+    convert bytes to a human readable format
+    '''
+    if size_bytes == 0:
+       return "0B"
+    size_name = ("B", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB")
+    i = int(math.floor(math.log(size_bytes, 1024)))
+    p = math.pow(1024, i)
+    s = round(size_bytes / p, 2)
+    return "%s %s" % (s, size_name[i])
```
thomasyu888, line 56: Nit: use f-string interpolation.

f"{s} {size_name[I]}"

schematic/utils/general.py
```diff
@@ -26,3 +27,30 @@ def unlist(_list):
...
+    return total
+
+def convert_size(size_bytes):
+    '''
+    convert bytes to a human readable format
+    '''
+    if size_bytes == 0:
+       return "0B"
+    size_name = ("B", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB")
+    i = int(math.floor(math.log(size_bytes, 1024)))
```
thomasyu888, line 53: Nit: `i` is not a descriptive variable.  I would move away from single letter variables for code readability.

schematic/store/synapse.py
```diff
@@ -100,14 +101,32 @@ def __init__(
 
         self._query_fileview()
 
+    def _purge_synapse_cache(self):
```
thomasyu888, line 104: Nit: Add doc string

schematic/utils/general.py
```diff
@@ -26,3 +27,30 @@ def unlist(_list):
...
+    total = 0
+    with os.scandir(path) as it:
+        for entry in it:
+            if entry.is_file():
+                total += entry.stat().st_size
+            elif entry.is_dir():
+                total += get_dir_size(entry.path)
+    return total
+
+def convert_size(size_bytes):
```
thomasyu888, line 46: Nit: Add in Arguments and type hinting

```
convert_size(size_bytes: int)
```

schematic/store/synapse.py
```diff
@@ -50,6 +50,7 @@
 from schematic.exceptions import MissingConfigValueError, AccessCredentialsError
 
 from schematic import CONFIG
+import datetime
```
thomasyu888, line 53: Nit: `datetime` is a python base packge, so I would add it to the top. ([pep8 standards for package imports](https://peps.python.org/pep-0008/#imports))

schematic/store/synapse.py
```diff
@@ -16,7 +16,7 @@
 import re
 import synapseclient
 from time import sleep
-
+from ..utils.general import get_dir_size, convert_size
```
thomasyu888, line 19: Nit: I would follow the same import structure through this package:

```
from schematic.utils.general import get_dir_size, convert_size
```

schematic/store/synapse.py
linglp: Thanks Tom. I think the function is: `self.syn.cache.cache_root_dir`

schematic/store/synapse.py
linglp: Thanks for pointing this out. I was trying to experiment if using "rowset" would work and forgot to remove it.

schematic/utils/general.py
linglp: I double checked by looking at the documentation related to `os.scandir` [here](https://docs.python.org/3/library/os.html#os.scandir). It seems like that even though `scandir.close()` will be called automatically, it is still advisable to use the with statement. I have left a comment about this in the code.

schematic/store/synapse.py
linglp: Thanks for pointing this out Tom. I tried using future concurrency to test out sending 20 concurrent requests. But maybe because clearing the cache took too long that I ended up getting: 
```
generated an exception:('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
generated an exception:('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
generated an exception:('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
generated an exception:('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
generated an exception:('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
generated an exception:('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
generated an exception:('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))

```

linglp: Note: problem about purging synapse cache before querying file view: two users could interrupt each other. For example, if user A download the file view, and user B purges the fileview before user A can read file view, then this would cause a problem for user A.

schematic/store/synapse.py
linglp: @thomasyu888 hmm if we change this line to `self.syn.cache.cache_root_dir`, I think this will impact local cache when running the endpoint locally? Should we avoid that by using `/var/www/.synapseCache` instead?

schematic/store/synapse.py
```diff
@@ -100,14 +101,35 @@ def __init__(
...
+        Purge synapse cache if it exceeds 19.5 GB
+        '''
+        # try clearing the cache
+        # scan a directory and check size of files
+        root_dir = self.syn.cache.cache_root_dir
+        if os.path.exists(root_dir):
+            nbytes = get_dir_size(root_dir)
+            # if 19.5 GB has already been taken, purge cache prior to today
+            if nbytes >= 20937965568:
+                c = self.syn.core.cache.Cache()
```
linglp, line 115: @thomasyu888 when I tested the purge functionality today, I noticed that this line would trigger an attribute error. I think `cache = self.syn.cache` and `cache.purge()` would work

schematic/store/synapse.py
```diff
@@ -100,23 +101,46 @@ def __init__(
...
+        # scan a directory and check size of files
+        root_dir = self.syn.cache.cache_root_dir
+        cache = self.syn.cache
+        if os.path.exists(root_dir):
+            nbytes = get_dir_size(root_dir)
+            # if 19.5 GB has already been taken, purge cache before an hour
+            if nbytes >= 20937965568:
+                an_hour_earlier = datetime.strftime(datetime.utcnow()- timedelta(hours = 1), '%s')
+                num_of_deleted_files = cache.purge(before_date = int(an_hour_earlier))
+                print(f'{num_of_deleted_files} number of files have been deleted from {root_dir}')
```
linglp, line 118: Here should I use a simple print statement or logger? I think print statement might be better since I could see it on AWS

schematic/store/synapse.py
thomasyu888: ^ That is up to you.  It might make most sense to hard code it if you only expect the cache dir to be in `/var/www/.synapseCache`

schematic/store/synapse.py
thomasyu888: hmm - a google search revealed this: https://stackoverflow.com/questions/37703609/using-python-logging-with-aws-lambda.  

If the rest of the package uses `logging`, I would use `logging` as well.

linglp: @thomasyu888 thanks for the link. I have updated the code to use logger/logging

schematic/store/synapse.py
```diff
@@ -100,23 +101,46 @@ def __init__(
...
+        '''
+        # try clearing the cache
+        # scan a directory and check size of files
+        root_dir = self.syn.cache.cache_root_dir
+        cache = self.syn.cache
+        if os.path.exists(root_dir):
+            nbytes = get_dir_size(root_dir)
+            # if 19.5 GB has already been taken, purge cache before an hour
+            if nbytes >= 20937965568:
+                an_hour_earlier = datetime.strftime(datetime.utcnow()- timedelta(hours = 1), '%s')
```
thomasyu888, line 116: I recommended an hour, but is this reasonable?  How fast did you encounter the storage space overfill again?

linglp: Note: I have already updated it to 15 min and also clear cache when it reaches 7GB

schematic/store/synapse.py
```diff
@@ -100,23 +101,46 @@ def __init__(
 
         self._query_fileview()
 
+    def _purge_synapse_cache(self):
+        '''
+        Purge synapse cache if it exceeds 7 GB
+        '''
+        # try clearing the cache
+        # scan a directory and check size of files
+        root_dir = self.syn.cache.cache_root_dir
```
linglp, line 110: change this to point to the AWS cache

schematic/store/synapse.py
```diff
@@ -100,23 +101,46 @@ def __init__(
...
+        cache = self.syn.cache
+        if os.path.exists(root_dir):
+            nbytes = get_dir_size(root_dir)
+            # if 7 GB has already been taken, purge cache before 15 min
+            if nbytes >= 7516192768:
```
linglp, line 115: Save the big number as a global variable and use that instead

linglp: Use GB instead of bytes and create a function to convert GB back to bytes

schematic/utils/general.py
thomasyu888, line 1: Nit: I would try to be consistent with all your functions.

1. Add type hinting as a best practice
2. Use a standard docstring format (e.g. [Google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html))
```
def my_function(argument: str) -> str:
    """
    my function here
    
    Args:
       argument: description here
    
    Returns:
       str:  A string
    """

File: schematic/store/synapse.py [hunk 17]
```diff
@@ -100,23 +101,53 @@ def __init__(
...
         self._query_fileview()
 
+    def _purge_synapse_cache(self):
+        '''
+        Purge synapse cache if it exceeds 7 GB
+        '''
+        # define the root directory
+        # set maximum synapse cache allowed
+        # set maximum ephemeral stroage allowed on AWS
+        root_dir = "/var/www/.synapseCache/"
```
thomasyu888, line 111: You could set this as an optional key argument of this function which makes it more flexible.

```
def _purge_synapse_cache(self, root_dir="/var..."):
```

schematic/store/synapse.py
```diff
@@ -100,23 +101,53 @@ def __init__(
...
+        maximum_storage_allowed_cache_gb = 7
+        maximum_storage_allowed_cache_bytes = convert_gb_to_bytes(maximum_storage_allowed_cache_gb)
+        total_ephemeral_storag_gb = 20
+        total_ephemeral_storage_bytes = convert_gb_to_bytes(total_ephemeral_storag_gb)
```
thomasyu888, line 115: Nit: You technically have no use for these if the `root_dir` doesn't exist, so it may make sense to move them into the if statement.

schematic/store/synapse.py
thomasyu888, line 1: Will there be tests written for this?

thomasyu888 (commented):
🚀 Great work @linglp !

schematic/utils/general.py
linglp: Thank you so much for all the great feedbacks @thomasyu888 . I have updated the doc string based on your suggestion. I learned a lot from having you review this PR. Thank you again!

schematic/store/synapse.py
linglp: Thanks @thomasyu888 . Let me open a separate PR for testing.

PR #1192, opened 2023-03-16
Added functionality to download a manifest by using the manifest id

linglp:
Related to the issue here: https://sagebionetworks.jira.com/browse/FDS-40

The goal of this PR is to create a new endpoint to allow users to download a manifest by using the manifest id. Previously, schematic only has `getDatasetManifest` function that downloads an existing manifest from a dataset (and would require querying the file view)

Highlight of the changes: 
- Added ManifestDownload class to handle downloading a manifest. I initially tried to liberate the download method under synapse storage class, but I felt synapse storage class is pretty huge and the code might be more organized if we move the manifest download out
- Before downloading a manifest,  I also added the functionality to check the type of an entity. Previously, there's a method under synapse storage does the same thing, so I move that function to utils. The API endpoint that uses the functionality of checking entity type also gets updated. 
- The old endpoint of downloading a manifest through asset view and dataset id was renamed to: `dataset/manifest/download`, and the new endpoint was named `manifest/download` that takes the manifest id as a parameter. 
- Tests


run_api.py
```diff
@@ -15,4 +15,4 @@
     # Launch app
     app = create_app()
     #CORS(app, resources={r"*": {"origins": "*"}})
-    app.run(host=host, port=port, debug=False)
\ No newline at end of file
+    app.run(host=host, port=port, debug=True)
```
linglp, line 18: will eventually switch to "False" after final testing

schematic_api/api/__main__.py
```diff
@@ -13,7 +13,7 @@ def main():
 
     # Launch app
     # CORS(app, resources={r"*": {"origins": "*"}})
-    app.run(host=host, port=port, debug=False)
+    app.run(host=host, port=port, debug=True)
```
linglp, line 16: this will also be reverted back to "false" before merging

schematic/store/synapse.py
```diff
@@ -395,85 +491,44 @@ def getDatasetManifest(
...
-                        manifest_data['name'] = new_manifest_filename
-                        manifest_data['filename'] = new_manifest_filename
-                        manifest_data['path'] = new_manifest_path_name
-
+            manifest_syn_id = self._get_manifest_id(manifest)
+            if downloadFile: 
+                md = ManifestDownload()
+                manifest_data = ManifestDownload.download_manifest(md, self.syn, manifest_id=manifest_syn_id, newManifestName=newManifestName, manifest_df=manifest)
+                if manifest_data == "":
+                    logger.debug(f"No manifest data returned. Please check if you have successfully downloaded manifest: {manifest_syn_id}")
```
linglp, line 530: should this be "logger.debug" or should this be a warning? I think most of the time when "manifest_data" returns an empty string, it usually indicates there's something wrong with the download_manifest function. Downstream function usually tries to extract information from manifest_data, and if it returns an empty string, usually the downstream function will break.

andrewelamb (approved):

andrewelamb:
@linglp Could you let me know when the Production API will have this?

[187](https://github.com/Sage-Bionetworks/schematic_db/pull/187) Should be merged at the same time.

schematic/store/synapse.py
```diff
@@ -51,9 +51,98 @@
 from schematic.exceptions import MissingConfigValueError, AccessCredentialsError
 
 from schematic import CONFIG
+from schematic.utils.general import entity_type_checking
 
 logger = logging.getLogger("Synapse storage")
 
+class ManifestDownload(object):
```
thomasyu888, line 58: @linglp Two things before reviewing more later - hope this helps.

1. My 2nd comment will feed into this, but when you use classes, you always want to start with "why do I need a class"?  I'm not saying this isn't a good use case of classes, but you're currently not utilizing the class structure.
1.  One example for a class is if you have a bunch of functions that take the same arguments. For example, you have  `syn` and `manifest_id` in each of the functions (meaning you are doing an action with `syn` and `manifest_id`.  You could do something like the below. 

    ```
     @dataclass
     class ManifestDownload:
            syn: synapseclient.Synapse
            manifest_id: str
     ```
     this will allow you to do something like this...
     ```
     md = ManifestDownload(syn=syn, manifest_id=manifest_id)
     md.download_manifest()
     ```

schematic/utils/general.py
```diff
@@ -69,4 +69,31 @@ def convert_gb_to_bytes(gb: int):
...
+
+    if type_entity == "<class 'synapseclient.table.EntityViewSchema'>":
+        return "asset view"
+    elif type_entity == "<class 'synapseclient.entity.Folder'>":
+        return "folder"
+    elif type_entity == "<class 'synapseclient.entity.File'>":
+        return "file"
+    elif type_entity == "<class 'synapseclient.entity.Project'>":
+        return "project"
+    else:
```
thomasyu888, line 98: So a number of things here.

1. Instead of using `str(type(...))`, you can do something like this: `isinstance(entity_name, synapseclient.table.EntityViewSchema)`. (Read up on `isinstance`)
2. You can return `entity_name.concreteType` if theres no matching entity type.

schematic/utils/general.py
```diff
@@ -69,4 +69,31 @@ def convert_gb_to_bytes(gb: int):
...
+    """
+    check the entity type of the id that needs to be downloaded
+    Args:
+        manifest_id: id of a manifest
+    Return: 
+        type_entity: type of the manifest being returned
+    """
+    # check the type of entity
+    try: 
+        entity_name = syn.get(manifest_id, downloadFile=False)
```
thomasyu888, line 85: Nit: Is this an `entity_name` or is it an `entity`?  You want to try to name your variables what they actually are.

schematic_api/api/openapi/api.yaml
```diff
@@ -101,8 +101,57 @@ paths:
         - Manifest Operations
   /manifest/download:
     get:
-      summary: Endpoint to download an existing manifest
-      description: Endpoint to download an existing manifest
+      summary: Endpoint to download an existing manifest by using a manifest id
+      description: Endpoint to download an existing manifest by using a manifest id
+      parameters:
+        - in: query
```
thomasyu888, line 107: This isn't great security practice, and I would strongly advise not adding any other features that leverages this until it is fixed.

schematic_api/api/openapi/api.yaml
```diff
@@ -101,8 +101,57 @@ paths:
         - Manifest Operations
   /manifest/download:
```
thomasyu888, line 102: If this is in your control, my advise is to redesign the API as such, and to not include `manifestId` in the query parameter.

```
GET /manifest/{manifestId}
```

schematic_api/api/openapi/api.yaml
```diff
@@ -101,8 +101,57 @@ paths:
...
+            text/csv:
+              schema:
+                type: string
+        "403":
+          description: You don't have access to the requested resources. 
+        "500":
+          description: Please check schematic log
+      tags:
+        - Manifest Operations            
+  /dataset/manifest/download:
```
thomasyu888, line 151: Same thought here about API design

```
GET /dataset/{datasetId}/manifest
```

schematic_api/api/routes.py
thomasyu888, line 1: Nit: Add docstrings, type hints, follow pep8 standards, and import statements

Edit: Personally, I would reorder all of the import statements using a tool called [isort](https://pycqa.github.io/isort/) in another PR.

schematic_api/api/routes.py
```diff
@@ -234,7 +239,15 @@ def parse_bool(str_bool):
         raise ValueError(
             "String boolean does not appear to be true or false. Please verify input."
         )
-        
+
+def return_as_json(manifest_local_file_path):
+    manifest_csv = pd.read_csv(manifest_local_file_path)
+    try:
+        manifest_json = json.loads(manifest_csv.to_json(orient="records"))
```
thomasyu888, line 246: Is there a reason you are doing this instead of doing?

```
manifest_csv.to_dict(orient="records")
```

schematic_api/api/routes.py
```diff
@@ -234,7 +239,15 @@ def parse_bool(str_bool):
...
+        return manifest_json
+    except Exception:
```
thomasyu888, line 248: It's not good practice to `except Exception` and `raise Exception`.  You want to catch the specific error.

schematic_api/api/routes.py
```diff
@@ -608,8 +620,48 @@ def get_viz_tangled_tree_layers(schema_url, figure_type):
...
+
+    # default as_json to True
+    if as_json is None:
+        as_json=True
+
+    # try logging in to asset store
+    try:
+        syn = store.login(input_token=input_token)
+    except Exception: 
+        raise SynapseAuthenticationError
```
thomasyu888, line 648: What is your rationale for this `try... except`? 

I think a `SynapseAuthenticationError` is very specific to not being able to log in (maybe due to wrong password).  There is also `SynapseNoCredentialsError` or `SynapseTimeoutError`, or `SynapseHTTPError`

schematic_api/api/routes.py
```diff
@@ -608,8 +620,48 @@ def get_viz_tangled_tree_layers(schema_url, figure_type):
...
+    try: 
+        md = ManifestDownload()
+        manifest_data = ManifestDownload.download_manifest(md, syn, manifest_id, new_manifest_name)
+        #return local file path
+        manifest_local_file_path = manifest_data['path']
+    except Exception: 
+        logger.error(f"Failed to download manifest: {manifest_id}")
```
thomasyu888, line 655: What is your reason for logging the error and raising it? (Is it for cloudwatch logs?)

schematic_api/api/routes.py
```diff
@@ -620,12 +672,14 @@ def download_manifest(input_token, dataset_id, asset_view, as_json, new_manifest
     manifest_data = store.getDatasetManifest(datasetId=dataset_id, downloadFile=True, newManifestName=new_manifest_name)
 
     #return local file path
-    manifest_local_file_path = manifest_data['path']
+    try:
```
thomasyu888, line 675: Can you explain what is going on here?  So, I'm assuming by reading the code that if the key `path` doesn't exist, the manifest wasn't downloaded. 

It's best practice to throw the `KeyError` then:

`except KeyError:`

tests/test_api.py
```diff
@@ -576,9 +575,61 @@ def test_get_datatype_manifest(self, client, syn_token):
                 "Year of Birth": "Int64",
                 "entityId": "string"}
 
+    # small manifest: syn51078535; big manifest: syn51156998
+    @pytest.mark.parametrize("manifest_id",["syn51156998", "syn51156998"])
```
thomasyu888, line 579: TIL (today I learned): I didn't realize you could chain these - which makes it more readable, but see my comment about `parametrize` below

tests/test_api.py
```diff
@@ -576,9 +575,61 @@ def test_get_datatype_manifest(self, client, syn_token):
...
+    def test_manifest_download(self, config, client, syn_token, manifest_id, new_manifest_name, as_json):
+        params = {
+            "input_token": syn_token,
+            "manifest_id": manifest_id,
+            "new_manifest_name": new_manifest_name, 
+            "as_json": as_json
+
+        }
+
+        response = client.get('http://localhost:3001/v1/manifest/download', query_string = params)
```
thomasyu888, line 591: Nit: it might make your life easier if you had a global config for (maybe a Jira ticket to refactor)

```
SERVER_URL = "http://localhost:3001/v1"
```

Then you can do

```
f"{SERVER_URL}/manifest/download"
```

tests/test_utils.py
```diff
@@ -50,6 +62,19 @@ def test_dict2list_with_list(self):
...
+        entity_type = entity_type_checking(syn, entity_id)
+        if entity_id == "syn27600053":
+            assert entity_type == "folder"
+        elif entity_id == "syn29862078":
+            assert entity_type == "file"
+        elif entity_id == "syn23643253":
+            assert entity_type == "asset view"
+        elif entity_id == "syn30988314":
+            assert entity_type == "folder"
+
```
thomasyu888, line 77:
```suggestion
    @pytest.mark.parametrize("entity_id, expected", [("syn27600053", "folder"), ("syn29862078", "file"), ("syn23643253", "asset view"), ("syn30988314", "folder")])
    def test_check_entity_type(self, synapse_store, entity_id, expected_entity_type):
        syn = synapse_store.syn
        entity_type = entity_type_checking(syn, entity_id)
        assert entity_type == expected_entity_type
```

tests/test_store.py
thomasyu888, line 1: Great initial start!  I think it will help to learn more about parametrize: https://docs.pytest.org/en/6.2.x/parametrize.html You use this a lot, but I'm not sure it is being properly used right now.

The important thing to remember is that if you are testing different functions of the code, it should probably be different testing functions.  For example: let's say your function raises a Value Error under one circumstance, but returns strings another.  Your unit tests are:

```
def test_my_function_returns_string()
# and
def test_my_function_raises_error()
```

schematic/utils/general.py
```diff
@@ -69,4 +69,31 @@ def convert_gb_to_bytes(gb: int):
     return: total number of bytes
     """
     bytes_to_return = gb * 1024 * 1024 * 1024
-    return bytes_to_return
\ No newline at end of file
+    return bytes_to_return
+
+
+def entity_type_checking(syn, manifest_id):
```
thomasyu888, line 75: Nit: What are you doing in this function? 

Are you checking the entity type?  Or are you mapping the entity type to a string?

schematic/store/synapse.py
```diff
@@ -606,7 +661,10 @@ def getProjectManifests(self, projectId: str) -> List[str]:
                         )
 
                     manifest_info = self.getDatasetManifest(datasetId,downloadFile=True)
-                    manifest_name = manifest_info["properties"]["name"]
+                    try:
+                        manifest_name = manifest_info["properties"]["name"]
```
thomasyu888, line 665: Are there ever scenarios where `manifest_info['properties']` doesn't exist?  if not, how about doing:

```
manifest_name = manifest_info["properties"].get("name", "")
```

You can then remove this later

```
manifest_name = ""
```

schematic/store/synapse.py
thomasyu888: If this breaks the downstream code, maybe it should be an error.  What is your rationale about not raising the error here?

schematic/store/synapse.py
```diff
@@ -51,9 +51,98 @@
...
+        manifest_data = ""
+
+        # check entity type
+        self._entity_type_checking(syn, manifest_id)
+
+        # download a manifest
+        try:
+            manifest_data = self._download_manifest_to_folder(syn, manifest_id)
+        except(SynapseUnmetAccessRestrictions, SynapseAuthenticationError):
+            if not manifest_df.empty:
```
thomasyu888, line 119: Can you explain this?  Are you downloading the manifest that is censored?

schematic/store/synapse.py
```diff
@@ -51,9 +51,98 @@
...
+                    manifest_data = self._download_manifest_to_folder(syn, new_manifest_id)
+                except (SynapseUnmetAccessRestrictions, SynapseAuthenticationError) as e:
+                    logger.error(f"You don't have access to the requested resource: {manifest_id}")
+                    raise (f"You don't have access to the requested resource: {manifest_id}")
+
+            else:
+                logger.error(f"You don't have access to the requested resource: {manifest_id}")
+
+        if newManifestName:
+            if os.path.exists(manifest_data['path']):
```
thomasyu888, line 133: I think this can be condensed into one line:

```
if newManifestName and os.path....
```

thomasyu888 (changes requested):
👍🏼 Great start @linglp!  I really appreciate you writing out the goal of this PR along with the highlight of the changes.  That really helped me orient me around the PR.

Additionally - I appreciate the tests that you wrote to ensure the stability of the features you added!  I made my comments - please feel free to drop some time on my calendar to discuss any of my comments.

schematic/store/synapse.py
linglp: Thank you so much Tom. I really like this idea.

schematic/store/synapse.py
linglp: hmm Thanks Tom. I think about this issue a bit more, and I think it would make more sense to turn it to logger.error...Previously, I was thinking that since we initialized `manifest_data` as an empty string in the beginning (see [here](https://github.com/Sage-Bionetworks/schematic/blob/05d2dede44e31121067cecdc1b2a9313860de7e1/schematic/store/synapse.py#:~:text=%22%22%22-,manifest_data%20%3D%20%22%22,-%23%20get%20a%20list)), then it might make sense to "allow" returning an empty string back? And to be honest, i am not sure if we should always: 
* raise an error when the request failed
* Or return an empty value so that the request would go through without raising an error. 
I see a mix in our current code base. But I think if the downstream code does not accept an empty value, maybe we could just catch the error here and make it simpler.

schematic/store/synapse.py
```diff
@@ -51,9 +51,98 @@
...
+        """
+        try downloading a manifest to local cache or a given folder
+        manifest
+        Args:
+            syn: an object of type synapseclient.
+            manifest_id: id of a manifest  
+        Return: 
+            manifest_data: A new Synapse Entity object of the appropriate type
+        """
+        if CONFIG["synapse"]["manifest_folder"]:
```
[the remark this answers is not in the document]
linglp: hmm so technically users should download the config file from our repo, and our default config.yml should have "manifest_folder" as a key there (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/config.yml#:~:text=manifest_folder%3A%20%27manifests%27)) I am thinking maybe we could deprecate the else statement

schematic/store/synapse.py
```diff
@@ -51,9 +51,98 @@
...
+            manifest_data = syn.get(
+                    manifest_id,
+                    downloadLocation=CONFIG["synapse"]["manifest_folder"],
+                    ifcollision="overwrite.local",
+                )
+        else:
+            manifest_data = syn.get(
+                        manifest_id,
+                    )
```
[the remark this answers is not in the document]
thomasyu888: Makes sense, I'll relay what I said in slack, which is to slowly lint each script with `black` and look through the code changes.  Personally, I've never encountered any code breaking changes with `black`

schematic/store/synapse.py
linglp: Thanks for the question Tom. This is I guess an edge case that I noticed that the old code is dealing with and I tried to preserve the old behavior. For example, for the dataset here: https://www.synapse.org/#!Synapse:syn27600053, we have two manifests: one is censored and one is uncensored. The uncensored one (see [here](https://www.synapse.org/#!Synapse:syn29862078)) has access restriction, so the old code would try to download the censored manifest.

schematic/utils/general.py
linglp: I am trying to return the entity type, and I forgot to change the name (after copying it from an old function). I will update the name of it. Thanks for pointing it out.

schematic/store/synapse.py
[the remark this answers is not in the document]
thomasyu888: You can always add a 
```
# TODO potentially deprecate if else statement because...
```

This way you make note of the issues in the code and remember to go back to it.

schematic/store/synapse.py
thomasyu888: Thanks, makes sense. Do you expect only to have 1 or 2 manifest in a folder?

schematic/store/synapse.py
thomasyu888: I think that depends on how the downstream code uses it.  If you don't want to break your downstream code, then it makes sense to catch the error here.  

You can also add a `# TODO` to potentially revisit.

schematic_api/api/__main__.py
thomasyu888: Nit: Add TODO, I feel like there might be a way to start the app with `--debug` flag that does `debug=debug`.  The reason I say TODO is because you don't want to change too many things in a PR, just try to stick with the original purpose of the PR.

schematic/store/synapse.py
linglp: Yes. For a given dataset, there should only be 0-2 manifest.

schematic/store/synapse.py
linglp: Thanks Tom. This is a good advice. I will make a comment and remind myself to revisit.

schematic_api/api/__main__.py
linglp: @thomasyu888 Thanks for the advice. I will add a comment to the code.

schematic/store/synapse.py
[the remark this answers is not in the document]
linglp: Thanks again. I will add that comment in the code.

schematic/store/synapse.py
linglp: I like what you suggested. Thank you Tom.

schematic/utils/general.py
```diff
@@ -69,4 +69,31 @@ def convert_gb_to_bytes(gb: int):
...
+    check the entity type of the id that needs to be downloaded
+    Args:
+        manifest_id: id of a manifest
+    Return: 
+        type_entity: type of the manifest being returned
+    """
+    # check the type of entity
+    try: 
+        entity_name = syn.get(manifest_id, downloadFile=False)
+    except:
```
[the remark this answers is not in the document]
linglp: Yes. Thank you Tom.

schematic/utils/general.py
linglp: @thomasyu888 Thank you. Using `isinstance` indeed makes my code looks cleaner.

schematic/store/synapse.py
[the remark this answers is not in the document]
linglp: hmm @thomasyu888 I tried linting `synapse.py` but since that file is huge, it looks like there are a lot of formatting changes. Should I create a separate PR to lint the module after this one gets merged? Since otherwise it would be difficult to see the changes that I made to the module. `general.py` is much shorter so I linted that in this PR.

schematic/store/synapse.py
thomasyu888: So what is the impact if this is an empty string.  By default if it is an empty string os.path.exists is going to be False.

Now if what you mean by doesnt exist is that there will be a KeyError, then you can do manifest_data.get('path')

schematic/store/synapse.py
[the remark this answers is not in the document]
thomasyu888: So, I would try to do linting of entire modules completely separate (general.py is probably fine).  I know there is actually a vscode module that automatically lints only the code you touch. I'd look into that.

There are also auto docstrings creators in vscode, If you have type hints in your function, it will create a nice docstrings template for you to fill out.

schematic_api/api/openapi/api.yaml
thomasyu888: This can be a Jira ticket

schematic_api/api/openapi/api.yaml
thomasyu888: This can also be a Jira ticket

schematic/store/synapse.py
linglp: Thanks for the suggestion Tom. If I use: `manifest_name = manifest_info["properties"].get("name", "")`, I don't need the try except?

thomasyu888: Well it seems like you still want to log the fact that you failed to download the manifest which is why this exists.  What happens when you fail to download the manifest, does `load_df` work?

linglp: No, I don't think load_df would work since that requires "path" as a key. Maybe I could try something like: 
```
                    manifest_name = manifest_info["properties"].get("name", "")

                    if not manifest_name:
                        logger.error(f'Failed to download manifests from {datasetId}') 
```

schematic/store/synapse.py
linglp: I think we could just do `manifest_data.get('path')` like you suggested to avoid key error. Thank you!

schematic/store/synapse.py
```diff
@@ -395,85 +494,46 @@ def getDatasetManifest(
...
-                        new_manifest_filename = newManifestName + '.csv'
-                        new_manifest_path_name = manifest_data['path'].replace(manifest['name'][0], new_manifest_filename)
-                        os.rename(manifest_data['path'], new_manifest_path_name)
-
-                        # Update file names/paths in manifest_data
-                        manifest_data['name'] = new_manifest_filename
-                        manifest_data['filename'] = new_manifest_filename
-                        manifest_data['path'] = new_manifest_path_name
-
+            manifest_syn_id = self._get_manifest_id(manifest)
```
thomasyu888, line 528: Excellent decision to split this off.  What you'll see in your development is that it really is all about how to simplify the code so your tests are easier to write.

schematic/store/synapse.py
```diff
@@ -606,7 +666,11 @@ def getProjectManifests(self, projectId: str) -> List[str]:
                         )
 
                     manifest_info = self.getDatasetManifest(datasetId,downloadFile=True)
-                    manifest_name = manifest_info["properties"]["name"]
+                    manifest_name = manifest_info["properties"].get("name", "")
+
+                    if not manifest_name:
```
thomasyu888, line 671: Great use of Python implicit booleanness!

schematic_api/api/routes.py
```diff
@@ -608,8 +620,48 @@ def get_viz_tangled_tree_layers(schema_url, figure_type):
...
+        file path of the downloaded manifest
+    """
+    # call config_handler()
+    config_handler()
+
+    # use Synapse Storage
+    store = SynapseStorage(input_token=input_token)
+
+    # default as_json to True
+    if as_json is None:
```
thomasyu888, line 641: I just realized this, is there a reason why you're not seeing `as_json=True` in the parameter section?

```
def download_manifest(input_token, manifest_id, new_manifest_name='', as_json=True)
```

schematic/store/synapse.py
linglp: Thanks Tom!

schematic/store/synapse.py
linglp: Thanks to your previous feedback.

schematic_api/api/openapi/api.yaml
linglp: Yeah thanks Tom. There are a bunch of endpoints that need redesign like this. I open a placeholder ticket: https://sagebionetworks.jira.com/browse/DCA-138

schematic_api/api/openapi/api.yaml
linglp: Thanks Tom. Thomas has mentioned similar things. Please see the jira ticket here: https://sagebionetworks.jira.com/browse/DCA-138

schematic_api/api/openapi/api.yaml
linglp: That makes sense. I will add a TODO here.

schematic_api/api/routes.py
linglp: Thanks Tom. I created a task for myself here: https://sagebionetworks.jira.com/browse/DCA-139

schematic_api/api/routes.py
linglp: Yes. Raising a key error here makes sense. Alternatively, I could use the .get function: 
```
manifest_data.get("path", "")
``` 
But since not getting the path key from manifest_data would be problematic, so maybe raising an error is better?

schematic_api/api/routes.py
linglp: That's a good catch. Thanks Tom. For some reasons, I was under the impression that if users select nothing in swagger UI, as_json will be default back to None. But I just double checked, and it seems like if I set as_json = True in parameter, this is sufficient to default as_json = True.

schematic_api/api/routes.py
linglp: Thanks for the suggestion Tom. I think maybe `SynapseNoCredentialsError` is not needed for now since the token is a required parameter. But I could add other errors. If I have SynapseHTTPError, can I raise it as: 
```
    except SynapseHTTPError as e:
        raise(f"Please see Synapse HTTP error status:{e.response.status_code}")
```

schematic_api/api/routes.py
linglp: Thanks Tom. To be honest, I am not sure when I should raise an error and when I should just log it (so I did both). I think cloudwatch log could also show the error that you raise in the code, so this is not a problem. Let me remove the part that "log" error and just keep the "raise"

schematic_api/api/routes.py
linglp: I just tried what you suggested and it worked too. Thanks Tom!

schematic_api/api/routes.py
linglp: Thanks Tom. I changed this line to raise JSONDecodeError

schematic_api/api/routes.py
thomasyu888: Try something like this: https://stackoverflow.com/questions/9157210/how-do-i-raise-the-same-exception-with-a-custom-message-in-python.  Right now, i think your code will fail.

```
except KeyError as e:
     raise KeyError(f'Failed to download manifest from dataset: {dataset_id}') from e
```
OR

```
except KeyError as e:
   logging.error(f'Failed to download manifest from dataset: {dataset_id}')
   raise e
```

In the future, you can imagine having custom exceptions for `Schematic` that make error catching and raising more robust.

schematic_api/api/routes.py
```diff
@@ -234,7 +240,15 @@ def parse_bool(str_bool):
...
         )
-        
+
+def return_as_json(manifest_local_file_path):
+    manifest_csv = pd.read_csv(manifest_local_file_path)
+    try:
+        manifest_json = manifest_csv.to_dict(orient="records")
+        return manifest_json
+    except JSONDecodeError as e: 
+        raise (f"Fail to return the manifest as a json") from e
```
thomasyu888, line 250: @linglp You'll want to be careful here:

This raises a different error than:
```
In [2]: raise ("foo")
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 raise ("foo")

TypeError: exceptions must derive from BaseException
```

This:
```
In [1]: raise ValueError("foo")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [1], in <cell line: 1>()
----> 1 raise ValueError("foo")

ValueError: foo
```

Notice how the first one is an actual error with how you've used the `raise` statement.

schematic_api/api/routes.py
```diff
@@ -608,8 +625,49 @@ def get_viz_tangled_tree_layers(schema_url, figure_type):
...
+    try:
+        syn = store.login(input_token=input_token)
+    except SynapseAuthenticationError as e:
+        logger.error(f'Your credentials are not valid. Please provide a valid synapse token')
+        raise e
+    except SynapseTimeoutError as e:
+        logger.error("Time out waiting for synapse to respond")
+        raise e
+    except SynapseHTTPError as e:
+        logger.error(f"A Synapse HTTP error occurred. Please see the error status:{e.response.status_code}")
```
linglp, line 654: Here I updated the try and except syntax. But when I intentionally put in a fake token, it still only showed me the error raised by synapse python client. I guess that's fine since synapse python client is also showing informative error message. But in case I want to change it, is there a way to update the syntax?

schematic_api/api/routes.py
linglp: Thanks Tom. We do have exception.py (see the document [here](https://github.com/Sage-Bionetworks/schematic/blob/develop-light-refactor-download-manifest/schematic/exceptions.py)). Do you think if i should create another class there for handling login error? I also noticed that we have `AccessCredentialsError` class (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop-light-refactor-download-manifest/schematic/exceptions.py#:~:text=class%20AccessCredentialsError(Exception))), but that one seems to only for handling not having access to a certain entity in synapse, and not really login error.

tests/test_api.py
linglp: Thanks Tom. I will add the TO DO like you suggested.

schematic_api/api/routes.py
linglp: Thanks Tom. I have updated the code to make sure that the right error message got raised.

schematic_api/api/routes.py
thomasyu888: I wouldn't worry about it in this particular PR, but just make note of it as a possible update in the future. What you're doing a lot here is catching and re-raising the exception for "better error messages". I don't think you need to try to catch every single error, sometimes its ok for functions to just fail, but in this particular `KeyError` scenario, I was referring to now this particular syntax:

```
raise()
```

Will fail without an Exception type.

schematic/store/synapse.py
```diff
@@ -51,9 +52,102 @@
...
+        """
+        # check the type of entity
+        entity_type = entity_type_mapping(self.syn, self.manifest_id)
+        if entity_type  != "file":
+            logger.error(f'You are using entity type: {entity_type}. Please try using a file')
+
+    @staticmethod
+    def download_manifest(self, newManifestName="", manifest_df=None):
+        """
+        Donwload a manifest based on a given manifest id.
```
thomasyu888, line 102: Nit: Typo in Download

schematic/store/synapse.py
```diff
@@ -51,9 +52,102 @@
...
+        Return: 
+             if the entity type is wrong, raise an error
+        """
+        # check the type of entity
+        entity_type = entity_type_mapping(self.syn, self.manifest_id)
+        if entity_type  != "file":
+            logger.error(f'You are using entity type: {entity_type}. Please try using a file')
+
+    @staticmethod
+    def download_manifest(self, newManifestName="", manifest_df=None):
```
thomasyu888, line 100: nit: Try your best in getting in the habit of adding type hints. It's extremely painful to code to best practices especially when you're busy, but it will become second nature to you the more you do it.

thomasyu888 (commented):
🔥 🔥 🔥 Great work!  The last step is probably to have Andrew test this code because he is going to be using it in `schematic_db` right?

I'll defer to how you guys want to test this, but I just had some minor comments.

schematic/store/synapse.py
andrewelamb: @thomasyu888 We are in the process of getting Schematic up to speed on the most up to date coding practices and this includes being fully typed and enforced with mypy. Its a pretty big codebase so it might take a while, but I agree it's very important. I know having schematic_db being fully typed has saved me a lot of possible bugs.

thomasyu888: Totally agree - it should be iterative. The intent here was to slowly guide new contributions or parts of new contributions to follow best practices so we don't add to the backlog of code that needs to be linted / refactored.

schematic/store/synapse.py
linglp: Thanks! Let me fix that

schematic_api/api/routes.py
linglp: Thanks Tom. I have updated the syntax there. I think instead of raising a KeyError, it makes more sense to raise a type error because manifest_data can be an empty string when it returns (and `manifest_data["path"]` would therefore not work)

schematic/store/synapse.py
linglp: Thanks Tom for catching all these. It has been a great learning experience for me.

linglp:
@mialy-defelice since Andrew is out for a week, do you have time to give this PR another test? This PR is for downloading a manifest, so maybe you could try downloading a manifest by using the new `/manifest/download` endpoint that only requires a manifest id and also the `/dataset/manifest/download` endpoint that would work the same as the old `/manifest/download`. 

@lakikowolfe Just a head up: after this PR gets merged and deployed to AWS, you would need to change the name of the url being called to `/dataset/manifest/download` to download a manifest using the dataset id.

schematic/store/synapse.py
```diff
@@ -51,9 +52,101 @@
...
+    def _entity_type_checking(self):
+        """
+        check the entity type of the id that needs to be downloaded
+        Return: 
+             if the entity type is wrong, raise an error
+        """
+        # check the type of entity
+        entity_type = entity_type_mapping(self.syn, self.manifest_id)
+        if entity_type  != "file":
+            logger.error(f'You are using entity type: {entity_type}. Please try using a file')
```
mialy-defelice, line 97: @linglp This is minor but can you be more explicit in this error. Instead of using 'try', be more declarative, `Please use a file.' or 'Please provide a file id.'

schematic/store/synapse.py
```diff
@@ -387,96 +462,79 @@ def getFilesInStorageDataset(
...
+        censored_regex=re.compile('.*censored.*')
+        censored = manifest['name'].str.contains(censored_regex)
+        if any(censored):
+            # Try to use uncensored manifest first
+            not_censored=~censored
+            if any(not_censored):
+                manifest_syn_id=manifest[not_censored]["id"][0]
+            # if only censored manifests are available, just use the first censored manifest
+            else: 
+                manifest_syn_id = manifest["id"][0]
```
mialy-defelice, line 482: @linglp this is actually a big assumption here. If someone isn't super familiar with the file structure expected and is just trying things out, this might return some confusing results. It would be nice to check if there are multiple manifests that _could_ be downloaded to count them, and raise a warning telling the user that its expected to only have a single manifest (two if censored and uncensored) and that you are just going to pull a single manifest (providing that manifest id).

mialy-defelice: This applies to lines 482 and 486.

schematic/store/synapse.py
linglp: @mialy-defelice sounds good. I could make that change.

schematic/store/synapse.py
linglp: Thanks for bringing this up. I think here i tried to preserve the old behavior that we have in our code base (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#:~:text=manifest_syn_id%3Dmanifest%5B%22id%22%5D%5B0%5D)) . If I understand it correctly, if there are uncensored manifest and censored manifest, the code would first try using the uncensored manifest. If there are multiple uncensored manifest, we would use the first one? And if there are multiple censored manifests available, we would download the first one as well. I agree with you that issuing a warning to the users and explaining the situation is better. Should I address the problem here or open a separate PR to address in develop?

mialy-defelice: Why don't you create an issue in Jira and use is situation as a user story.

tests/test_store.py
```diff
@@ -444,3 +466,65 @@ def test_upsertTable(self, helpers, synapse_store, config, projectId, datasetId)
...
         synapse_store.syn.delete(tableId)
+
+
+class TestDownloadManifest:
+    @pytest.mark.parametrize("datasetFileView", [{"id": ["syn51203973", "syn51203943"], "name": ["synapse_storage_manifest.csv", "synapse_storage_manifest_censored.csv"]}, {"id": ["syn51203973"], "name": ["synapse_storage_manifest.csv"]}, {"id": ["syn51203943"], "name": ["synapse_storage_manifest_censored.csv"]}])
+    def test_get_manifest_id(self, synapse_store, datasetFileView):
+        # rows that contain the censored manifest
+        datasetFileViewDataFrame = pd.DataFrame(datasetFileView)
+        row_censored = datasetFileViewDataFrame.loc[datasetFileViewDataFrame['name'] == "synapse_storage_manifest_censored.csv"]
+        if len(row_censored) > 0:
```
mialy-defelice, line 477: Im assuming if row_censored returns no censored manifests you will return columns but no rows, which is why you're looking for the length of the df? Can you do ` if not row_censored.empty` for a more compact solution?

schematic/store/synapse.py
linglp: Thanks Mialy. I wrote a jira issue here: https://sagebionetworks.jira.com/browse/FDS-345

tests/test_store.py
linglp: Yeah that works. Let me make that change. Thanks.

mialy-defelice (approved):
Thank for addressing my other comments.

My user testing produced the expected results for `dataset/manifest/download` and `manifest/download`. It looks great!

PR #1193, opened 2023-03-20
Create example data model for single rule benchmarking

GiaJordan:
Create example model with `MockComponent` attributes set to not required so that JSON schema validation only runs for the attribute being benchmarked. To be used for single rule benchmarking tests. Rules are explicitly set to their default behavior.


mialy-defelice (approved):
Looks good!

PR #1197, opened 2023-03-24
Added duration flag to show run time of tests

linglp:
Added duration flag to show run time of tests


linglp:
Close this PR since we no longer need this.

PR #1200, opened 2023-04-04
Add `IsNA` rule to validation suite | FDS-81 FDS-232

GiaJordan:
Add `IsNA` rule to check if manifest entries are marked as `Not Applicable`, to be used only in conjunction with other rules ~~using the `OR` operator to be implemented in #/1201~~.

This PR adds support for the combination of the new `IsNA` rule with `type` validation type rules `int`, `float`, and `num`, but not `str`. Implementation of rule combination functionality is intended to be temporary until after the refactor of the validation suite when general support for rule disjunction is implemented.

Tests are also included for the new rule. The data model attribute `Check NA` will test the behavior of the `int::IsNA` combo during the valid and invalid manifest tests. All combinations will be tested when `test_rule_combinations` is run.

The rule in its current form is only intended to be used by @andrewelamb for iAtlas, as such the addition of this rule **will not be documented** until rule disjunctions are officially supported and this rule is refactored.

### DIFF
--- a/schematic/models/GE_Helpers.py
+++ b/schematic/models/GE_Helpers.py
@@ -134,7 +134,8 @@ class GreatExpectationsHelpers(object):
             "protectAges": "expect_column_values_to_be_between",
             "unique": "expect_column_values_to_be_unique",
             "inRange": "expect_column_values_to_be_between",
-            
+            "IsNA": "expect_column_values_to_match_regex_list",
+
             # To be implemented rules with possible expectations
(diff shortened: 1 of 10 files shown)


GiaJordan:
@andrewelamb will perform tests with iAtlas manfiests

andrewelamb:
@GiaJordan This is working with an iAtlas manifest!

andrewelamb (approved):

GiaJordan:
@andrewelamb happy to hear; thank you for verifying!

PR #1202, opened 2023-04-06
feat: added nginx to encrypt the communication between ALB and API

linglp:
Why this is needed: https://sagebionetworks.jira.com/browse/DCA-98

## Changelog
* Changed base image from python3.10 to https://github.com/tiangolo/uwsgi-nginx-flask-docker
* added nginx
* configured nginx to use ssl by using a self-signed certificate
* added /health path to check status

## How to test the PR
* Please follow the instructions here: https://github.com/Sage-Bionetworks/schematic/pull/1202#discussion_r1166983663
* To test locally, please follow the instructions in schematic_api/api/ReadMe

## Note
* `tiangolo/uwsgi-nginx-flask` image was developed by the person who created FastAPI . Based on the ReadMe of `tiangolo/uwsgi-nginx-flask`, using FastAPI could potentially provide 8 times better performance, and usinf `tiangolo/meinheld-gunicorn-flask` could provide 4 times better performance. This PR is to make based on the assumption that we don't change our current framework
* To follow the best practices, the docker should be built using user `www-data`: https://sagebionetworks.jira.com/browse/FDS-411
* Note: as of May 23rd, this PR has been deployed and tested in schematic dev instance using schematic_profiler 

## Technical details 
More details related to the image: https://hub.docker.com/r/tiangolo/uwsgi-nginx-flask/

More details about making encryption work: 
Since Bruce mentioned in slack (see details [here](https://sagebionetworks.slack.com/archives/CEFQD0KU1/p1681247669102339?thread_ts=1681242034.812899&cid=CEFQD0KU1)) that we could potentially use a self-assigned certificate, I ended up giving up on certbox mentioned in the [documentation](https://www.digitalocean.com/community/tutorials/how-to-serve-flask-applications-with-uwsgi-and-nginx-on-ubuntu-22-04) provided by @tschaffter in [DCA-68](https://sagebionetworks.jira.com/browse/DCA-68). 

Here are the main resources that I ended up following: 
* https://www.digitalocean.com/community/tutorials/how-to-create-a-self-signed-ssl-certificate-for-nginx-in-ubuntu-16-04#create-a-configuration-snippet-pointing-to-the-ssl-key-and-certificate
* and here: https://www.humankode.com/ssl/create-a-selfsigned-certificate-for-nginx-in-5-minutes/
* https://faun.pub/setting-up-ssl-certificates-for-nginx-in-docker-environ-e7eec5ebb418

The hard part is to make use of the content in the resources mentioned above and apply them to the [Docker image](https://github.com/tiangolo/uwsgi-nginx-flask-docker/blob/master/docker-images/python3.10.dockerfile) that we are using. I digged into the base image and found out that the [entrypoint.sh](https://github.com/tiangolo/uwsgi-nginx-flask-docker/blob/master/docker-images/entrypoint.sh) generates a bunch of additional nginx config files in `/etc/nginx/conf.d` folder and there's also `/etc/nginx/nginx.conf` file on top that loads those additional nginx.conf by using the line below: 
```
    include /etc/nginx/conf.d/*.conf;
```

After many trials and errors, I learned that I have to make the following changes: 
* enable port 443 in docker-compose.yml
* modify `/etc/nginx/nginx.conf` to redirect traffic to HTTPS
* Add certificate.conf:  for nginx to handle HTTPS requests (note that we have to add location block there otherwise we will get error like "“schematic-api-aws | 2023/04/13 18:15:36 [error] 197#197: *70 open() "/etc/nginx/html/favicon.ico" failed (2: No such file or directory), client: 172.20.0.1, server: localhost, request: "GET /favicon.ico HTTP/2.0", host: "localhost", referrer: "https://localhost/v1/ui/"”) 
* Add self-signed.conf: for including two certificate and key generated by SSL
* Add self-params.conf (maybe optional if not using dhparam.pem?): for adding path to dhparam: `ssl_dhparam /etc/ssl/dhparam.pem`

I also learned through experiments that I can't just modify `/etc/nginx/nginx.conf` by copying and pasting a new one because the child docker image inherits the entrypoint from the parent. If I just copied and pasted the new `nginx.conf`, it would get overwritten once the bash script "entrypoint.sh" starts running.

The only solution seems to be to modify the "entrypoint.sh"... But I tried that and it seemed like that the new `entrypoint.sh` didn't get executed.. In the end, I decided to create a new bash script and use that as an entrypoint. Hence you see the following in Dockerfile: 
```
WORKDIR ${ROOT}
COPY ./uwsgi-nginx-entrypoint.sh ./entrypoint2.sh
COPY ./uwsgi-nginx-entrypoint.sh ./uwsgi-nginx-entrypoint2.sh
RUN chmod +x uwsgi-nginx-entrypoint2.sh
RUN chmod +x entrypoint2.sh
RUN chown -R nginx /uwsgi-nginx-entrypoint2.sh
RUN chown -R nginx /entrypoint2.sh
```
and finally overwriting the entrypoint in the parent image: 
```
# specify entrypoint again to generate config
ENTRYPOINT ["/entrypoint2.sh"]
```

And interestingly, CMD has to be reset once ENTRYPOINT gets reset. 

In the end, everything seems to start working. You should be able to see swagger UI by in firefox and chrome by visiting: https://localhost/v1/ui/

When I did: `curl https://127.0.0.1/v1/ui --insecure`, it returned: 
```
<!doctype html>
<html lang=en>
<title>Redirecting...</title>
<h1>Redirecting...</h1>
<p>You should be redirected automatically to the target URL: <a href="https://127.0.0.1/v1/ui/">https://127.0.0.1/v1/ui/</a>. If not, click the link.
```

[DCA-68]: https://sagebionetworks.jira.com/browse/DCA-68?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


schematic_api/Dockerfile
thomasyu888, line 1: Hey @linglp I'm trying to build this to test your solution.  Do you have instructions on building this container?

thomasyu888: Fantastic, I follow these instructions and it worked!

```
curl https://thomasyu888-psychic-acorn-7ggr9vx9p62xpg9-443.preview.app.github.dev/v1/ui/ --insecure
<html>
<head><title>302 Found</title></head>
<body>
<center><h1>302 Found</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

linglp: Thanks for trying it out Tom!

linglp:
I also spent some time reading about nginx metrics report. Based on the article [here](https://www.datadoghq.com/blog/how-to-collect-nginx-metrics/), for an open source nginx, we could see: the number of accepted connection, the number of connection handled, the number of connection dropped, the number of active connection, and total number of requests. And nginx plus could show number of 4XX codes and number of 5XX codes. Even though cloud watch could provide similar metrics, I want to be able to view the metrics more easily. 

If I added the following in certificate.conf: 
```
    location /nginx_status {
        stub_status; 
        allow 172.28.0.0/16; 
        deny all; 
    }
```

and also modify the docker-compose.yml: 
```
networks:
  schematic:
    name: schematic
    ipam:
      driver: schematic
      config:
        - subnet: "172.28.0.0/16"
```

When I visited https://127.0.0.1/nginx_status, I could see metrics like: 
```
Active connections: 1 
server accepts handled requests
 28 28 74 
Reading: 0 Writing: 1 Waiting: 0 
```
The interpretation: 
* number of active connection: 1
* number of accepted connection: 28
* number of handled connection: 28
* total number of handled requests: 74
* calculate request per connection: total number of handled requests (74)/number of handled connections(28) = 2.64

schematic_api/Dockerfile
```diff
@@ -51,7 +77,15 @@ COPY  schematic_api ./schematic_api
...
-RUN chown -R www-data:www-data /opt/app/great_expectations/
+
+# copy tests folder because some endpoints by default download to the tests folder
+COPY tests ./tests
+
+# change permission
+RUN chown -R www-data:www-data ${APP_DIR}
+
+# allow downloading to synapse cache 
+RUN chown -R www-data:www-data /root
```
jesusaurus, line 88: Why is the synapse cache under `/root`? Can we configure it to be in another location, maybe `/var/tmp/synapse`?

jesusaurus: @linglp It's probably not a big deal in our case, but `/root` is the home directory of the root user, and is not really an appropriate place for services to store data. See https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html for more info on the standard linux filesystem hierarchy.

And if its the default directory, I guess that means we're running the service as the root user (which can be a potential security issue). If that is the case, we should eventually change the container to run as a service user instead of the root user, but that can be a challenging migration that I don't want to block this PR on. Here's a short article I found on that: https://www.howtogeek.com/devops/why-processes-in-docker-containers-shouldnt-run-as-root/

tschaffter: I agree with the proposal of @jesusaurus. Installing and running the app as a non-root user is more secure. This work should be tracked in a separate ticket.

certificate.conf
```diff
@@ -0,0 +1,22 @@
...
+        try_files $uri @app;
+    }
+    location @app {
+        include uwsgi_params;
+        uwsgi_pass unix:///tmp/uwsgi.sock;
+    }
+    location /static {
+        alias /app/static;
+    }
+    location /health {
```
tschaffter, line 18: @linglp This is an elegant solution. Currently this endpoint does not seem to be documented in the OpenAPI description of Schematic. That may be fine if you consider the health endpoint is "private", i.e. should not be exposed to the external users.

tschaffter (commented):
Just posted a few comments. I'll review this PR more in depth this week.

schematic_api/Dockerfile
linglp: Thanks both. I created an issue here: https://sagebionetworks.jira.com/browse/DCA-173 to keep track of this issue.

env.example
```diff
@@ -1,5 +1,4 @@
 # API server
 SERVER_PROTOCOL=http://
 SERVER_DOMAIN=localhost
-SERVER_PORT=7080
```
linglp, line 4: @tschaffter Based on the line of code [here](https://github.com/Sage-Bionetworks/schematic/pull/1202/files#diff-a0f6c3ca194fef3726d45adc2ac51fb28aa30a82b8d7b7ab0f80a50496dc31e9:~:text=default%20to%2080-,USE_LISTEN_PORT,-%3D%24%7BLISTEN_PORT%3A%2D), the bash script is looking for env variable `USE_LISTEN_PORT`. Do you think if we should let users define `USE_LISTEN_PORT` in their own environment file? And we could also change `docker-compose.yml` to something like: 
```
    ports: 
      - "${LISTEN_PORT}:80"
      - "443:443"
  ```

**Review by tschaffter — APPROVED** [review]
There is a small modification I suggested to @linglp during a call about how to set the port of the Docker container. Otherwise this PR looks good.

File: main.py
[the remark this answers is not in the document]
Comment (mialy-defelice) [reply main.py:1]: opp i understand how its being used now.

**Review by mialy-defelice — COMMENTED** [review]
Looked through the code. I am not an expert in this domain so I'm relying a bit on Thomas S. And Tom Y. for their expertise here. I will test now.

**Comment by mialy-defelice** [comment]
Tested and got the API running using GitHub code spaces.

File: schematic_api/api/README.md [hunk 4]
```diff
@@ -18,13 +18,17 @@ If you define `APP_PORT` as `3001` in `docker-compose.yml`, you should be able t
 After installing uWSGI, flask and connexion, to start a docker container that runs flask application with uWSGI:
 1) Comment out the first part of `docker-compose.yml` and focus only on building container `schematic-aws`.
 2) Get `service_account_creds.json` by doing `schematic init --config /path/to/config.yml`. 
-3) Make a copy of `env.example` and rename it as `.env` and keep it in the same directory as `env.example` By default, schematic uses port 7080. If port 7080 is not available, please update `SERVER_PORT` in .env file. 
+3) Make a copy of `env.example` and rename it as `.env` and keep it in the same directory as `env.example` By default, schematic uses port 80. If port 80 is not available, please update `USE_LISTEN_PORT` in .env file.
```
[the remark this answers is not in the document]
linglp: Thank you for pointing this out. I changed the default port to 81.

schematic_api/api/README.md
```diff
@@ -18,13 +18,17 @@ If you define `APP_PORT` as `3001` in `docker-compose.yml`, you should be able t
 After installing uWSGI, flask and connexion, to start a docker container that runs flask application with uWSGI:
```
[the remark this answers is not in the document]
linglp: Thanks Mialy. I just reorganized the instructions.

schematic_api/api/README.md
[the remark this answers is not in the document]
linglp: Thanks @mialy-defelice I updated the instructions based on your suggestion

linglp:
@mialy-defelice thanks for your review. I updated the ReadMe.md based on your advice. Let me know what you think

mialy-defelice (approved):
Thanks for changing the Readme its much clearer now!

PR #1203, opened 2023-04-11
Remove restrictions on rule number and allowed combinations

GiaJordan:
Remove checks for amount of rules specified for a given attribute, and checks ensuring only compatible and verified rule combinations are provided


mialy-defelice (approved):
I checked that it worked on my end.
Looks good!

PR #1205, opened 2023-04-18
Added configuration to view nginx status

linglp:
Added configuration to view nginx status

Depends on #1202


linglp:
close this for now since we no longer need it.

PR #1207, opened 2023-04-20
Remove requirement for `.synapseConfig` file to use `upsert` feature

GiaJordan:
Update credential gathering process for support of `schematic_db` use so that other log in methods besides the `.synapseConfig` file are supported. 
The API endpoints and relevant functions were also simplified to remove a parameter that was a duplicate of another.
Table upsert tests were also updated to ensure "update" functionality works as expected

### DIFF
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -256,7 +256,7 @@ paths:
           schema:
             type: string
             nullable: true
-          description: Data Model Component
+          description: Data Model Component. Providing a data type will also validate the manfiest against the rules specified for the data type's attributes.
           example: Patient
           required: true
         - in: query
--- a/api/openapi/api.yaml
+++ b/api/openapi/api.yaml
@@ -24,7 +24,7 @@ paths:
             https://raw.githubusercontent.com/Sage-Bionetworks/schematic/develop/tests/data/example.model.jsonld
           required: true
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
           description: Token is needed if getting an existing manifest on AWS
@@ -105,7 +105,7 @@ paths:
       description: Endpoint to download an existing manifest
       parameters:
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
             nullable: false
@@ -281,7 +281,7 @@ paths:
           description: If True, validation suite will only run with in-house validation rule. If False, the Great Expectations suite will be utilized and all rules will be available.
           required: true
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
             nullable: false
@@ -438,7 +438,7 @@ paths:
       operationId: api.routes.get_manifest_datatype
       parameters:
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
             nullable: false
@@ -475,7 +475,7 @@ paths:
       operationId: api.routes.get_storage_projects
       parameters:
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
             nullable: false
@@ -504,7 +504,7 @@ paths:
       operationId: api.routes.get_storage_projects_datasets
       parameters:
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
             nullable: false
@@ -541,7 +541,7 @@ paths:
       operationId: api.routes.get_files_storage_dataset
       parameters:
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
             nullable: false
@@ -594,7 +594,7 @@ paths:
       operationId: api.routes.get_asset_view_table
       parameters:
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
             nullable: false
@@ -631,7 +631,7 @@ paths:
       operationId: api.routes.get_project_manifests
       parameters:
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
             nullable: false
@@ -679,7 +679,7 @@ paths:
       operationId: api.routes.check_entity_type
       parameters:
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
             nullable: false
@@ -717,7 +717,7 @@ paths:
       operationId: api.routes.check_if_files_in_assetview
       parameters:
         - in: query
-          name: input_token
+          name: access_token
           schema:
             type: string
             nullable: false
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1726,6 +1726,15 @@ class SynapseStorage(BaseStorage):
         dataset_index = self.storageFileviewTable["id"] == datasetId
         dataset_row = self.storageFileviewTable[dataset_index]
 
+        # re-query if nothing found
+        if len(dataset_row) == 0:
+            sleep(5)
+            self._query_fileview()
+            # Subset main file view
+            dataset_index = self.storageFileviewTable["id"] == datasetId
+            dataset_row = self.storageFileviewTable[dataset_index]
+
+
         # Return `projectId` for given row if only one found
         if len(dataset_row) == 1:
             dataset_project = dataset_row["projectId"].values[0]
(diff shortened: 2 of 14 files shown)


GiaJordan:
@afwillia just a heads up that the input parameters for the api endpoints will be changing when this PR is merged, we can coordinate when this gets merged depending on how long updating the front end will take

afwillia (commented):
This looks good to me - I just need to know when the dev AWS instance gets updated so I can change the function parameters in DCA

api/openapi/api.yaml
```diff
@@ -631,7 +631,7 @@ paths:
       operationId: api.routes.get_project_manifests
       parameters:
         - in: query
-          name: input_token
```
linglp, line 634: I am a little bit concerned of changing all the `input_token` to `access_token`. That means all the downstream applications using schematic APIs need to update their syntax.

schematic/store/synapse.py
```diff
@@ -1724,12 +1720,19 @@ def getDatasetProject(self, datasetId: str) -> str:
...
-        self._query_fileview()
-
         # Subset main file view
         dataset_index = self.storageFileviewTable["id"] == datasetId
         dataset_row = self.storageFileviewTable[dataset_index]
 
+        # re-query if no datasets found
+        if len(dataset_row) == 0:
+            sleep(5)
+            self._query_fileview()
```
linglp, line 1730: why do we need to re-query if no datasets found?

schematic/store/synapse.py
```diff
@@ -1946,6 +1949,50 @@ def replaceTable(synStore, tableToLoad: pd.DataFrame = None, tableName: str = No
...
+            # check which credentials are provided in file
+            if config.has_option('authentication', 'username'):
+                username = config.get('authentication', 'username')
+            if config.has_option('authentication', 'authtoken'):
+                authtoken = config.get('authentication', 'authtoken')
+        
+        # raise error if required credentials are not found
+        if not username and not authtoken:
+            raise NameError(
+                "Username or authtoken credentials could not be found in the environment, synapse object, or the .synapseConfig file"
```
linglp, line 1987: if both "username" and "authtoken" are not available, should the error message say something like "Username **and** authtoken credentials XXX"

linglp:
@GiaJordan In addition to the comments above, I am also a bit confused why you needed to change getDatasetProject function. Maybe we could discuss more about this off line because after PR here: https://github.com/Sage-Bionetworks/schematic/pull/1192, there are two ways to download manifests. One way is to use manifest id (which will be faster), and the other way is to use dataset id which involves downloading the asset view. The change will be in develop once  https://github.com/Sage-Bionetworks/schematic/pull/1142 gets merged.

schematic/store/synapse.py
GiaJordan: > if both "username" and "authtoken" are not available, should the error message say something like "Username **and** authtoken credentials XXX"

yes, it should say "and", thanks for catching!


> Can we combine the two raise statements and do something like:
> 
> ```
>         if not username or not authtoken:
>             raise NameError(
>                 "Username or authtoken credentials could not be found in the environment, synapse object, or the .synapseConfig file"
> ```

I wanted to separate out the error messages so that users would know exactly what was missing with less chance for confusion from saying "this or this is missing"

schematic/store/synapse.py
```diff
@@ -1946,6 +1949,50 @@ def replaceTable(synStore, tableToLoad: pd.DataFrame = None, tableName: str = No
...
+            authtoken = env_access_token
+
+        # Get token from authorization header
+        if 'Authorization' in synStore.syn.default_headers:
+            authtoken = synStore.syn.default_headers['Authorization'].split('Bearer ')[-1]
+
+        # retrive credentials from synapse object
+        synapse_object_creds = synStore.syn.credentials
+        if hasattr(synapse_object_creds, 'username'):
+            username = synapse_object_creds.username
```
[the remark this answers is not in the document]
GiaJordan: Yeah, that is what I had intended, they should be the same but I wanted to include both to be safe

schematic/store/synapse.py
GiaJordan: Before, it was re-querying every time the method was called. Now it will only re-query if no datasets were found as a last attempt to not error out.

api/openapi/api.yaml
GiaJordan: Yeah, we'd need to coordinate with those users like we are with @afwillia to update the endpoints, but I think this would result in a descriptive parameter name while also eliminating duplicate parameters.

GiaJordan:
> This looks good to me - I just need to know when the dev AWS instance gets updated so I can change the function parameters in DCA

@afwillia cool, will do. Thanks!
Also gonna tag @lakikowolfe to give you a heads up as well

schematic/store/synapse.py
GiaJordan: @milen-sage lingling mentioned that you might've wanted this to be removed? Is that the case?

schematic/store/synapse.py
```diff
@@ -1946,6 +1949,50 @@ def replaceTable(synStore, tableToLoad: pd.DataFrame = None, tableName: str = No
...
+        if os.path.exists(CONFIG.SYNAPSE_CONFIG_PATH):
+            config = synStore.syn.getConfigFile(CONFIG.SYNAPSE_CONFIG_PATH)
+
+            # check which credentials are provided in file
+            if config.has_option('authentication', 'username'):
+                username = config.get('authentication', 'username')
+            if config.has_option('authentication', 'authtoken'):
+                authtoken = config.get('authentication', 'authtoken')
+        
+        # raise error if required credentials are not found
```
linglp, line 1984: Could you add a comment here and clarify that without having a username, the code would still work?

lakikowolfe:
> > This looks good to me - I just need to know when the dev AWS instance gets updated so I can change the function parameters in DCA
> 
> @afwillia cool, will do. Thanks! Also gonna tag @lakikowolfe to give you a heads up as well

Thanks for the tag @GiaJordan this will effect DFA. A few questions:

- What's the time frame on this merge? 
- my understanding is that the main breaking change is `input_token` -> `access_token`. Is that correct?

GiaJordan:
@lakikowolfe 

- hopefully sometime this week, I think atp we're just waiting for @milen-sage to chime in on the fileview querying
- that's correct, the endpoints will need to be updated with the new param name

schematic/store/synapse.py
milen-sage: We can set expectations that if datasets were just added they may not immediately appear in schematic and DCA. That said, as @GiaJordan noted, this code now only retries if there are no datasets found (e.g. datasets may have been added for a first time in a project and may take a few seconds to be indexed). In the latter case the extra 5s is not too bad. As long as we are not adding the 5s overhead for every query, we can leave it in.

GiaJordan: @milen-sage cool, thank you for clarifying

schematic/store/synapse.py
```diff
@@ -1724,12 +1720,19 @@ def getDatasetProject(self, datasetId: str) -> str:
...
         """
 
-        self._query_fileview()
-
         # Subset main file view
         dataset_index = self.storageFileviewTable["id"] == datasetId
         dataset_row = self.storageFileviewTable[dataset_index]
 
+        # re-query if no datasets found
+        if len(dataset_row) == 0:
```
mialy-defelice, line 1728: Is `dataset_row` a df?
If so could you do something like `if dataset_row.empty` and `if not dataset_row.empty`?

mialy-defelice (changes requested):
I added a comment during my review. Other than that it logically looks good to me. It looks like you're also coordinating the merge with others, so it looks fine to me to merge.

schematic/store/synapse.py
GiaJordan: @mialy-defelice yes; I could use `dataset_row.empty` in place of this line. As far as using `if not dataset_row.empty`; it seems like there's a reason why it's checking for exactly one, as opposed to just any, but I'm not completely sure why so I'm hesitant to change it right now

mialy-defelice (approved):
Looks good!

GiaJordan:
I made one more change to improve security; when performing upserts through the api, the token will be passed as a parameter through to the upsert function instead of taking it from the default headers of the synapse object

PR #1210, opened 2023-04-27
Downloaded manifests to temporary folders on AWS

linglp:
Related to the bug reported by @andrewelamb here: https://sagebionetworks.jira.com/browse/FDS-299

Highlights: 
* Added function in general.py that could create temporary folders in specified directory. 
* when downloading a manifest on AWS, downloading the manifests to a temporary location.
* Updated permission in Dockerfile to make sure that the user that we defined in uWSGI.ini could create folders on AWS. 

Will merge to develop after #1142

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -45,6 +45,8 @@ from schematic_db.synapse.synapse import SynapseConfig
 from schematic_db.rdb.synapse_database import SynapseDatabase
 from schematic_db.schema.schema import get_key_attribute
 
+from synapseclient.entity import File
+
 from schematic.utils.df_utils import update_df, load_df
 from schematic.utils.general import create_temp_folder
 from schematic.utils.validate_utils import comma_separated_list_regex, rule_in_rule_list
@@ -67,15 +69,14 @@ class ManifestDownload(object):
     syn: synapseclient.Synapse
     manifest_id: str
 
-    def _download_manifest_to_folder(self):
+    def _download_manifest_to_folder(self) -> File:
         """
         try downloading a manifest to local cache or a given folder
         manifest
         Return: 
-            manifest_data: A new Synapse Entity object of the appropriate type
+            manifest_data: synapse file entity
         """
-         # TO DO: potentially deprecate the if else statement because "manifest_folder" key always exist in config
-
+        # TO DO: potentially deprecate the if else statement because "manifest_folder" key always exist in config
         # on AWS, to avoid overriding manifest, we download the manifest to a temporary folder
         if "SECRETS_MANAGER_SECRETS" in os.environ:
             temporary_manifest_storage = "/var/tmp/temp_manifest_download"
@@ -101,7 +102,7 @@ class ManifestDownload(object):
                     )
         return manifest_data 
 
-    def _entity_type_checking(self):
+    def _entity_type_checking(self) -> str:
         """
         check the entity type of the id that needs to be downloaded
         Return: 
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -7,6 +7,7 @@ import atexit
 import logging
 import secrets
 from dataclasses import dataclass
+import tempfile
 
 # allows specifying explicit variable types
 from typing import Dict, List, Tuple, Sequence, Union
@@ -45,6 +46,7 @@ from schematic_db.rdb.synapse_database import SynapseDatabase
 from schematic_db.schema.schema import get_key_attribute
 
 from schematic.utils.df_utils import update_df, load_df
+from schematic.utils.general import create_temp_folder
 from schematic.utils.validate_utils import comma_separated_list_regex, rule_in_rule_list
 from schematic.schemas.explorer import SchemaExplorer
 from schematic.schemas.generator import SchemaGenerator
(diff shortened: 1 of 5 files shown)


schematic/store/synapse.py
```diff
@@ -65,18 +69,31 @@ class ManifestDownload(object):
...
     manifest_id: str
 
-    def _download_manifest_to_folder(self):
+    def _download_manifest_to_folder(self) -> File:
         """
         try downloading a manifest to local cache or a given folder
         manifest
         Return: 
-            manifest_data: A new Synapse Entity object of the appropriate type
+            manifest_data: synapse file entity
```
andrewelamb, line 77: The type should be in parenthesis:

manifest_data(File): synapse file entity

Perhaps the description could be more descriptive as well like "A Synapse file entity of the downloaded manifest"

schematic/store/synapse.py
```diff
@@ -65,18 +69,31 @@ class ManifestDownload(object):
...
         """
-         # TO DO: potentially deprecate the if else statement because "manifest_folder" key always exist in config
-        if CONFIG["synapse"]["manifest_folder"]:
+        # TO DO: potentially deprecate the if else statement because "manifest_folder" key always exist in config
+        # on AWS, to avoid overriding manifest, we download the manifest to a temporary folder
+        if "SECRETS_MANAGER_SECRETS" in os.environ:
```
andrewelamb, line 81: This is a really complex series of if/elif/else statements, and really hard to parse. Is there any way to clean this up and make it easier to understand? If not I would recommend adding more documentation.

schematic/utils/general.py
```diff
@@ -114,7 +116,21 @@ def entity_type_mapping(syn, entity_id):
...
+    """This function creates a temporary directory in the specified directory 
+    Args:
+        path(str): a directory path where all the temporary files will live
+    Returns: returns the absolute pathname of the new directory.
+    """
+    # Create a temporary directory in the specified directory
+    try:
+        path = tempfile.mkdtemp(dir=path)
+        return path
+    except FileNotFoundError:
```
andrewelamb, line 130: Why is the same exception being raised after being caught?
If that is intended, I would [chain](https://docs.python.org/3/tutorial/errors.html#exception-chaining) this exception.

schematic/store/synapse.py
```diff
@@ -65,18 +69,31 @@ class ManifestDownload(object):
...
         """
         try downloading a manifest to local cache or a given folder
         manifest
         Return: 
-            manifest_data: A new Synapse Entity object of the appropriate type
+            manifest_data: synapse file entity
         """
-         # TO DO: potentially deprecate the if else statement because "manifest_folder" key always exist in config
-        if CONFIG["synapse"]["manifest_folder"]:
+        # TO DO: potentially deprecate the if else statement because "manifest_folder" key always exist in config
```
andrewelamb, line 79: This todo should be removed and a jira ticket started instead.

andrewelamb:
@linglp Should I check to see if I can recreate the bug via swagger with this branch?

schematic/store/synapse.py
thomasyu888: +1. This chunk of code seems to be logic for `_determine_download_location`.  The main question I have is this: 

* Why does "SECRETS_MANAGER_SECRETS" existing in `os.environ` take precedence over `CONFIG['synapse']['manifest']`?

thomasyu888 (commented):
Nice review by Andrew!  Just some minor comments on my end since you added me as reviewer.  Thanks for including me!

schematic/store/synapse.py
linglp: @thomasyu888 That's because the config on AWS also has "manifest_folder" key. If I have `if CONFIG["synapse"]["manifest_folder"]` takes precedence, then the logic of downloading to a temporary folder can't be triggered. 

I think an alternative solution is to use a different config for deploying to AWS. Since in the future, we would use the develop branch to deploy to AWS, we can't add a different key in config.yml to indicate deployment. But maybe we could have a separate config.yml just for deployment? Will that bring in more complexity?

schematic/utils/general.py
linglp: Thanks for the feedbacks @andrewelamb and @thomasyu888 . I will update the code to remove try except.

schematic/store/synapse.py
linglp: Thanks Andrew. Let me update the comment like you suggested.

schematic/store/synapse.py
thomasyu888: Thanks - this is helpful. Similar to how AWS does credentials and profiles, maybe something similar can be done here? Can open a jira ticket to discuss.

I don't know enough of the codebase, but potentially you can see if there are duplicated keys in your CONFIG and look into simplifying functions to take in the the 'synapse' or 'aws' parameter instead of if-else statements.  You could have different "profile" configurations like you said when deploying differently.   I'll defer this decision to the FAIR team.

linglp:
> @linglp Should I check to see if I can recreate the bug via swagger with this branch?

Hmm you could check if it works locally. But I don't think if you could check if that resolves the error on AWS by testing it locally.

andrewelamb:
@linglp Yeah I guess we'll have to wait till its in AWS.

tests/test_utils.py
```diff
@@ -77,6 +78,20 @@ def test_entity_type_mapping_invalid_entity_id(self, synapse_store):
         with pytest.raises(SynapseHTTPError) as exception_info:
             entity_type_mapping(syn, "syn123456")
 
+    def test_download_manifest_to_temp_folder(self):
+        # define temporary folder path
+        current_dir = os.getcwd()
```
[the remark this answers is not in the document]
linglp: Thanks Tom. This is a great suggestion.

schematic/store/synapse.py
linglp: @andrewelamb I discussed this issue with Milen in our 1:1. He suggested that let's leave this code here for now and see if the solution works for fixing your bug. If it does, we could circle back to this issue and discuss with the team what would be a good way to handle downloading manifests on AWS v.s. downloading manifests locally

schematic/store/synapse.py
```diff
@@ -65,18 +69,31 @@ class ManifestDownload(object):
...
+                os.mkdir("/var/tmp/temp_manifest_download")
+            download_location = create_temp_folder(temporary_manifest_storage)
+
+        elif CONFIG["synapse"]["manifest_folder"]:
+            download_location=CONFIG["synapse"]["manifest_folder"]
+
+        else:
+            download_location=None
+
+        if download_location:
```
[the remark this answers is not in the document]
linglp: Thanks Tom. I will take notes and update the code.

andrewelamb (approved):

MiekoHash:
Hi @linglp @andrewelamb Is this something we can plan for sprint 15 to finish up verification on AWS?

PR #1213, opened 2023-05-03
FDS-293 Fix column mismatch for Excel File Based Manifest Generation 

mialy-defelice:
This PR addresses the issue brought up in [FDS-293](https://sagebionetworks.jira.com/browse/FDS-293).

**For Context:**
The crux of the issue that was happening relates to how manifests are generated when there is an issue between the columns in the manifest being downloaded from Synapse and the columns that are expected based on the Schema. For the CSV and excel downloads, the handling was conducted such that if there were columns in the schema that weren't in the downloaded manifest, these would be handled and these columns would be added to the manifest. But if there were columns in the downloaded manifest that weren't in the schema, the columns would be simply added, but the ordering of the data was not adjusted accordingly.

**Additionally**
The handling of the different outputs was being handled separately allowing this divergence to occur, where you could get the expected output when exporting a CSV but not if you were exporting to excel. A user would presumably expect all the data to be the same but simply the output type to be different.

**What this PR Does**
I moved the handling of the columns and data upstream of the data output in a new function called `ManifestGenerate._update_dataframe_with_existing_df`. This new function handles all the logic for updating manifest to have all the expected columns using Pandas. The pandas DF is then handed of to `_handle_output_format_logic` which figures out the proper format to export to and calls the necessary functions.

There are also differences in the handling of `output_format` and `sheet_url`. There is a small change that handles this difference in output type a bit more clearly, in the future though we would want to consider depreciating `sheet_url`.

All this works functionally right now, but is a less than ideal solution. Since we are going to be doing a larger refactor in the future I don't think this PR needs to perform an intermediate refactor, but I do think there is a lot of room for improvement in the structure of the Manifest Generator now that we are expecting outputs other than Google Sheet.

[FDS-293]: https://sagebionetworks.jira.com/browse/FDS-293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -1256,13 +1256,16 @@ class ManifestGenerator(object):
         )
         return manifest_url
 
-    def _get_mismatched_columns(self, headers_1, headers_2):
-        """
+    def _get_mismatched_columns(self, headers_1:list , headers_2:list) -> list:
+        """Compare two colunm sets and get cols that are in headers_1, but not headers_2
+        Args:
+            headers_1 (list): list of column headers
+            headers_2 (list): list of column headers
+        Returns: 
+            list: column headers in headers_1 but not headers_2
 
         """
-        # Compare two colunm sets and get cols that are in headers_1, but not headers_2
-        mismatched_columns = set(headers_1) - set(headers_2)
-        return mismatched_columns
+        return set(headers_1) - set(headers_2)
 
 
     def set_dataframe_by_url(
@@ -1374,8 +1377,6 @@ class ManifestGenerator(object):
 
         # Generate empty manifest using `additional_metadata`
         manifest_url = self.get_empty_manifest()
-        
-        #TODO: MIALY Make sure this functionality is preserved, should be bc there isnt another manifest we are merging with.
         manifest_df = self.get_dataframe_by_url(manifest_url)
 
         # Annotations clashing with manifest attributes are skipped
@@ -1437,22 +1438,6 @@ class ManifestGenerator(object):
         """
         ## TO DO: deprecate sheet_url parameter and simplify the logic here
 
-        # check if output_format parameter gets set. If not, check the sheet_url parameter
-        
-        '''
-        if not output_format: 
-            # if the sheet_url parameter gets set to True, return a google sheet url 
-            if sheet_url: 
-                # populate google sheet with dataframe
-                manifest_sh = self.set_dataframe_by_url(empty_manifest_url, dataframe)
-                return manifest_sh.url
-                # else, return a pandas dataframe
-            else: 
-                return dataframe
-        else:
-        '''
-        # if not saving excel delete after header is made.
-
         # if the output type gets set to "dataframe", return a data frame 
         if output_format == "dataframe":
             """
@@ -1476,13 +1461,12 @@ class ManifestGenerator(object):
 
             return output_file_path
         
-        # for all other cases, return a google sheet url after populating the dataframe
-        # TODO: add "google_sheet" as an output choice
+        # Return google sheet if sheet_url flag is raised.
+        # TODO: Depreciate sheet URL and add google_sheet as an output_format choice.
         elif sheet_url: 
             manifest_sh = self.set_dataframe_by_url(manifest_url=empty_manifest_url, manifest_df=dataframe, out_of_schema_columns=out_of_schema_columns)
             return manifest_sh.url
 
-        # new default is dataframe?
         else:
             return dataframe
 
@@ -1541,7 +1525,6 @@ class ManifestGenerator(object):
             updated_df, out_of_schema_columns = self._update_dataframe_with_existing_df(empty_manifest_url=empty_manifest_url, existing_df=manifest_record[1])
 
             # determine the format of manifest
-            #result = self._handle_output_format_logic(output_format = output_format, output_path = output_path, sheet_url = sheet_url, empty_manifest_url=empty_manifest_url, dataframe = manifest_record[1])
             result = self._handle_output_format_logic(output_format = output_format,
                                                       output_path = output_path,
                                                       sheet_url = sheet_url,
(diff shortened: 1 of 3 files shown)


mialy-defelice:
[Link to sheet detailing testing](https://docs.google.com/spreadsheets/d/1Qgt4CHtnvHVeve2I_7rGdZ5fusFLrhO-K3kv9Kbwwik/edit#gid=1219624880)

schematic/manifest/generator.py
```diff
@@ -1254,18 +1256,19 @@ def get_empty_manifest(self, json_schema_filepath=None):
...
-        out_of_schema_columns = set(manifest_df_header) - set(wb_header)
-
-        return out_of_schema_columns
+    def _get_missing_columns(self, headers_1:list , headers_2:list) -> list:
+        """Compare two colunm sets and get cols that are in headers_1, but not headers_2
+        Args:
+            headers_1 (list): list of column headers
+            headers_2 (list): list of column headers
+        Returns: 
+            list: column headers in headers_1 but not headers_2
```
linglp, line 1265: will this truly return a list? I did a simple experiment: 
```
x = set(["apple", "banana", "cherry"])
y = set(["google", "microsoft", "apple"])

z = x-y

print(z)
```
Here, i am seeing: `{'cherry', 'banana'}` which is a set

schematic/manifest/generator.py
```diff
@@ -1254,18 +1256,19 @@ def get_empty_manifest(self, json_schema_filepath=None):
...
 
+        """
+        return set(headers_1) - set(headers_2)
 
     def set_dataframe_by_url(
-        self, manifest_url: str, manifest_df: pd.DataFrame
+        self, manifest_url: str, manifest_df: pd.DataFrame, out_of_schema_columns=None,
     ) -> ps.Spreadsheet:
```
linglp, line 1272: add type hinting of `out_of_schema_columns` and add a brief description of the parameter in the docstring.

schematic/manifest/generator.py
```diff
@@ -1458,56 +1419,59 @@ def export_sheet_to_excel(self, title: str = None, manifest_url : str = None, ou
...
         else:
             output_excel_file_path = os.path.abspath(os.path.join(os.getcwd(), file_name))
+        
         # export the manifest to excel
         export_manifest_drive_service(manifest_url, file_path=output_excel_file_path, mimeType = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet')
         
         return output_excel_file_path
 
-    def _handle_output_format_logic(self, output_format: str = None, output_path: str = None, sheet_url: bool = None, empty_manifest_url: str = None, dataframe: pd.DataFrame = None):
+    def _handle_output_format_logic(self, output_format: str = None, output_path: str = None, sheet_url: bool = None, empty_manifest_url: str = None, dataframe: pd.DataFrame = None, out_of_schema_columns=None):
```
linglp, line 1428: add typing of `out_of_schema_columns` and description in doc string

schematic/manifest/generator.py
```diff
@@ -1458,56 +1419,59 @@ def export_sheet_to_excel(self, title: str = None, manifest_url : str = None, ou
...
+        # if the output type gets set to "excel", return an excel spreadsheet
+        elif output_format == "excel":              
+            # export manifest url that only contains column headers to Excel
+            output_file_path = self.export_sheet_to_excel(title = self.title,
+                                                          manifest_url = empty_manifest_url,
+                                                          output_location = output_path,
+                                                          )
+            
+            if not os.path.exists(output_file_path): 
+                logger.error(f'Export to Excel spreadsheet fail. Please make sure that file path {output_file_path} is valid')
```
linglp, line 1460: Maybe raise an error here is better than logging an error?

schematic/manifest/generator.py
mialy-defelice: Yep you're right I can change the docs.

schematic/manifest/generator.py
```diff
@@ -1458,56 +1419,59 @@ def export_sheet_to_excel(self, title: str = None, manifest_url : str = None, ou
...
 
-                return output_file_path
-            # for all other cases, return a google sheet url after populating the dataframe
-            else: 
-                manifest_sh = self.set_dataframe_by_url(empty_manifest_url, dataframe)
-                return manifest_sh.url
+        # if the output type gets set to "dataframe", return a data frame 
+        if output_format == "dataframe":
+            """
+            TODO: for backwards compatibility may need to return downloaded CSV only with no updates wrt to the schema.
```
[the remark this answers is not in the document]
mialy-defelice: In the older version, dataframe was returned before any other manipulations were done to check for columns and add new columns to the df. This is now done upstream of this function, so all the outputs are the same regardless of output type. But if someone was expecting to just download the manifest from Synapse and not check the manifest against the model then we might need to consider that option. Though they could just use the synapse client to do that.

schematic/manifest/generator.py
```diff
@@ -1584,45 +1558,86 @@ def get_manifest(
...
+        Handle scenario when existing manifest does not match new manifest template due to changes in the data model:
+            the sheet column header reflect the latest schema the existing manifest column-set may be outdated
+            ensure that, if missing, attributes from the latest schema are added to the column-set of the existing manifest so that the user can modify their data if needed
+            to comply with the latest schema.
+        Args:
+            empty_manifest_url (str): Google Sheet URL with an empty manifest with headers.
+            existing_df (Pd.DataFrame): df of an existing manifest
+
+        Returns:
+            updated_df (Pd.DataFrame): existing_df with new_columns added.
```
linglp, line 1584: Based on the code, it seems like we are returning both updated_df and out_of_schema_columns?

schematic/manifest/generator.py
mialy-defelice: This part was not actually added by me, but I can change to raise an error instead.

linglp:
@mialy-defelice I think this looks very good overall. The code looks a lot more cleaner and more organized. Thanks for the update.

schematic/manifest/generator.py
mialy-defelice: Oop yep, ill update the returns.

schematic/manifest/generator.py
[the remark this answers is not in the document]
linglp: hmm I think if they just want to download a manifest, they could use the manifest/download endpoint instead of manifest generate?

schematic/manifest/generator.py
mialy-defelice: I am actually going to move the path check upstream of this to the CLI. Check for it in an update.

GiaJordan (approved):
Looks good @mialy-defelice!

PR #1221, opened 2023-05-15
Hot fix - clear storage when reaching 1gb. Log file storage distribution 

linglp:


linglp:
No longer need this. Close it for now

PR #1228, opened 2023-05-25
Restrict `typing-extensions` package to versions before `v4.6.0`

GiaJordan:
This PR temporarily fixes the issue described in FDS-478 where tests would fail because of using `isinstance()` in the process of executing the code as normal.

From the `typing-extensions` [changelog](https://github.com/python/typing_extensions/blob/main/CHANGELOG.md), in versions `v4.6.0` and up (released May 22, 2023 so 3 days ago):
> Backport https://github.com/python/cpython/pull/26067 (originally by Yurii Karabas), ensuring that isinstance() calls on protocols raise TypeError when the protocol is not decorated with @runtime_checkable. Patch by Alex Waygood.

This matches the [errors raised](https://github.com/Sage-Bionetworks/schematic/actions/runs/5071241815/jobs/9107380281?pr=1227) during tests when we updated our `poetry.toml` file and the dependencies in the `.lock` file were allowed to update as necessary.

Until our relevant classes or possibly in the python client are updated this will cause errors when we run our tests. Pinning the version gives us time to continue development on more pressing issues until we have the bandwidth to make updates to comply with the new requirements. `typing-extensions` was not a direct dependency for schematic beforehand, but was required by our dependencies so the version was relatively unlocked before this PR.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -555,7 +555,7 @@ grpcio-gcp = ["grpcio-gcp (>=0.2.2,<1.0dev)"]
 
 [[package]]
 name = "google-api-python-client"
-version = "2.86.0"
+version = "2.87.0"
 description = "Google API Client Library for Python"
 category = "main"
 optional = false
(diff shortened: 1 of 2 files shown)


milen-sage (approved):
@GiaJordan I see docutils and mistune packages are removed from poetry.lock. It's likely the removal wouldn't have impact. Do you know why thy are no longer needed though?

GiaJordan:
@milen-sage for sure!

PR #1229, opened 2023-05-26
Optimize performance in table upsert backwards compatibility scenarios

GiaJordan:
Add functionality to the `synapseStorage` object to make calls to the synapse REST API by way of the functionality exposed in the `synapsePythonClient`.

The method for making the calls implemented in the `synapseStorage` object is intended for general use, but due to current deadlines and time constraints is not fully generalized and is mostly relevant to the two calls that are necessary for renaming table columns when necessary during upsert operations. Also, while async calls are made, they are not necessarily used asynchronously on the schematic side, again due to time constraints, so future development should be done on these fronts to improve performance and usability.

Also changed the structure of the `TableOperations` class since it now has more helper methods and allowing instances to reference themselves simplifies things.


GiaJordan:
@linglp the code for the API calls will be in this PR

GiaJordan:
Testing results with new method:

<table>
<thead>
<tr>
<th>Performance Time (seconds)</th>
<th>CLI using python client ops</th>
<th>CLI using synapse API calls</th>
<th>API using python client ops</th>
<th>API using synapse API calls</th>
</thead>
<tbody>
<tr>
<td>Adding 200 rows to a table with ~1 existing row</td>
<td>47</td>
<td>43</td>
<td>116</td>
<td>42</td>
</tr>
<tr>
<td>Adding 200 rows to a table with ~200 existing rows</td>
<td>41</td>
<td>43</td>
<td>148</td>
<td>48</td>
</tr>
</tbody>


@milen-sage @AmyHeiser

schematic/store/synapse.py
```diff
@@ -2130,50 +2214,45 @@ def upsertTable(synStore: SynapseStorage, sg: SchemaGenerator, tableToLoad: pd.D
...
+                self._update_table_uuid_column(sg)
+                synapseDB.upsert_table_rows(table_name=self.tableName, data=self.tableToLoad)
             # Raise if other error
             else:
                 raise ex
 
-        return existingTableId
+        return self.existingTableId
 
-    def _update_table_uuid_column(synStore: SynapseStorage, table_id: str, sg: SchemaGenerator,) -> None:
```
GiaJordan, line 2163: @thomasyu888 here's the relevant chunk of code for renaming a column in an existing table

schematic/store/synapse.py
```diff
@@ -296,6 +296,62 @@ def wrapper(*args, **kwargs):
                     raise ex
         return wrapper
 
+    def send_api_request(self, request_type: str, uri: str, body: Union[Dict, str], headers: Dict = None, endpoint: str = None, **kwargs):
```
thomasyu888, line 299: You might be aware of this, but the synapse client does have these:

```
import synapseclient
syn = synapseclient.login()
syn.restPOST
syn.restGET
syn.restPUT
syn.restDELETE
```

Those are heavily tested, so it might be nice to use those if it captures everything you need.

schematic/store/synapse.py
```diff
@@ -2188,70 +2267,73 @@ def _update_table_uuid_column(synStore: SynapseStorage, table_id: str, sg: Schem
...
+                        "maximumSize": 64,
+                        "maximumListLength": 1,
+                    }
+
+                    # Send POST /column request to define new column and get new column ID
+                    newColResponse = self.synStore.send_api_request(
+                        request_type = "restPOST",
+                        uri = "https://repo-prod.prod.sagebase.org/repo/v1/column",
+                        body = columnModelDict,
+                    )
```
thomasyu888, line 2289: Nit: This can be replaced with

```
self.synStore.syn.restPOST("column", body=columnModelDict)
```

schematic/store/synapse.py
```diff
@@ -2188,70 +2267,73 @@ def _update_table_uuid_column(synStore: SynapseStorage, table_id: str, sg: Schem
...
+                    }
+
+                    # Build body for POST request
+                    schemaChangeBody = {
+                        "entityId": self.existingTableId,
+                        "changes": [columnChangeDict],
+                    }
+
+                    # Send POST request to change column name
+                    schemaChangeResponse = self.synStore.send_api_request(
```
thomasyu888, line 2310: Same thing here about using existing client `syn.restPOST`.  I do like the `send_api_request` function though!

tests/test_store.py
```diff
@@ -218,6 +218,32 @@ def test_getDatasetManifest(self, synapse_store, downloadFile):
             # return manifest id
             assert manifest_data == "syn51204513"
 
+    def test_synapse_rest_request(self, synapse_store):
```
thomasyu888, line 221: Using existing client functionality will allow you to remove this test.  That said, i think what's more important is to test `_update_table_uuid_column` this function.

schematic/store/synapse.py
thomasyu888, line 2289: This can be done with the client:

```
new_col = synapseclient.Column(name='Id', columnType='STRING', maximumSize=64, maximumListLength=1)
self.synStore.syn.store(new_col)
```

thomasyu888 (commented):
nice work @GiaJordan ! Just some minor comments that might help with maintenance of the code by offloading some of that to the Python client.

Overall, I think FAIR should decide on a standard: snake case or camelCase and stick with it for consistent code readability.

schematic/store/synapse.py
```diff
@@ -1975,62 +2065,56 @@ def createTable(synStore: SynapseStorage, tableToLoad: pd.DataFrame = None, tabl
...
-        current_table = synStore.syn.get(existingTableId)
-        current_columns = synStore.syn.getTableColumns(current_table)
+        current_table = self.synStore.syn.get(self.existingTableId)
+        current_columns = self.synStore.syn.getTableColumns(current_table)
         for col in current_columns:
             current_table.removeColumn(col)
 
-        if not tableName:
-            tableName = datasetName + 'table'
+        if not self.tableName:
```
thomasyu888, line 2105: Nit: `self.tableName` is default value is None and its type is `str`. I recommend to be specific

```
if self.tableName is not None
```

Not that this will happen, but if you set the value as `0`. you will get something like this:

```
tableName = 0
datasetName = "foo"
if not tableName:
    tableName = datasetName + 'table'
print(tableName)
# footable
```

schematic/store/synapse.py
GiaJordan: @thomasyu888 for sure, hope you find it useful
For this chuck of code, right now I don't plan for this to be reused. When speaking with @andrewelamb he mentioned that this isn't something we want to be doing often, in the context of treating Synapse tables like a database. The preferred method would be for users to rebuild their schema and reupload if they need to make a change to a column. This chunk enables backwards compatibility because of a change on our end so that users wouldn't have to manually do it themselves for this special case.

schematic/store/synapse.py
GiaJordan: Thanks for mentioning; those methods are what we're interfacing with in this method. Throughout the codebase we've been trying to use general functions where possible to simplify maintenance in the future, and I was trying to continue that here. My thinking was also that when we do the asset store refactor, part of the common store interface will be a method to send an API request to the respective stores. Does this approach make sense or am I overcomplicating things here?

tests/test_store.py
GiaJordan: @thomasyu888 thanks for pointing that out! before this PR we weren't able to easily test `_update_table_uuid_column` due to the difficulty in triggering specific case where it's actually needed and the operations it was doing. It might be easier now that we've switched to using the rest calls

schematic/store/synapse.py
GiaJordan: Would the value returned by `self.synStore.syn.store(new_col)` be the same as what we're currently getting now / does the `syn.store` operation actually store anything in this case or just create a column type object? Also, do you know if there's a performance difference between using the API and the client for this operation?

schematic/store/synapse.py
thomasyu888: Ah - my apologies!  I should have looked at the code closer, but I had mistaken `request` for the `requests` package and I thought you were doing raw `requests.GET/POST/...` calls over the client.

There are multiple ways to think about this.  Since the asset store for 'tables' specifically are things like postgres, synapse tables, Bigquery, naturally it'll be different API calls which might be harder to encapsulate.  In terms of file stores, that would be using something like fsspec (which we do have on our roadmap to build out fs-synapse), which would be pretty different from storing files into tables.

heres some pseudo code for OOP structure.  Some of this could be abstracted in `schematic-db`, so this might be unnecessary.  I encourage everyone to learn about [SOLID principles](https://www.freecodecamp.org/news/solid-principles-explained-in-plain-english/)

```
from abc import ABC

@dataclass
class BaseAssetStore(ABC):
     connection_env: str
     ....
     @abstractmethod
     def store_table(self):
           ....

class SynapseStore(BaseAssetStore):
     def store_table():
           # your implementation here

@dataclass
class BaseTableUpdate(ABC):
    asset_store: BaseAssetStore,   # This is where you put in your store logic for different stores. Those could have different base classes and configurations
    tableToLoad: pd.DataFrame = None, 
    tableName: str = None, 
    datasetId: str = None, 
    existingTableId: str = None,
    restrict: bool = False
    
    @abstract_method
    def createTable():
         ...

@dataclass
class SynapseTableUpdate(BaseTableUpdate)
    asset_store: SynapseStore
    
    def createTable():
         ...
....

```

schematic/store/synapse.py
thomasyu888: Yes, the value would be the same.  `syn.store` just creates a new column type object, basically the only difference here is you don't need to test all of this since it's tested in the client.

Behind the scenes, the client just does the API call.  If you're interested, `syn.store` will look for this function in any class: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/f3698af6c7d361c5a22d102d2625152933809ceb/synapseclient/table.py#L1517.  And it will call the `restPOST` command.  

You are essentially duplicating code and also formatting your own json when there's a Python object that you can use.

schematic/store/synapse.py
thomasyu888: I definitely find this helpful!  That makes sense to me - Synapse Tables is definitely not a database.  In terms of backwards compatibility - I leave it to you all to decide whats best.  What I look for code is

1. Reusability
2. Readabilty
3. Testability

I challenge everyone to think broader.  Yes, this particular code may not be re-used here, but table column renames is probably something that is helpful to the Sage community.  By pulling that core functionality out, it would be easier to test, and you could take that directly and create a PR into the python client.

schematic/store/synapse.py
andrewelamb: I saw my name mentioned!

I think the eventual asset store refactor will do exactly this @thomasyu888, like schematic db does with [database backends](https://github.com/Sage-Bionetworks/schematic_db/blob/develop/schematic_db/rdb/rdb.py#L39)

GiaJordan:
@thomasyu888 thank you so much for your comments, we're going to create tickets for these so that we can further optimize these portions of code in the future. From talking with @milen-sage during our 1 on 1 today we're going to implement your comments in coming PRs so that we can still deliver this functionality before I'm out next week.

schematic/store/synapse.py
GiaJordan: Ahh I see, that makes sense. I think in the future we'll switch to using the client for this operation and the api call for the renaming

schematic/store/synapse.py
GiaJordan: That's a good point, the calls for other stores will likely be different than what we're using for synapse currently.
@linglp here's a heads up that this will likely change in the future, including a switch to just using the client functions directly at some point

schematic/store/synapse.py
```diff
@@ -2188,70 +2267,73 @@ def _update_table_uuid_column(synStore: SynapseStorage, table_id: str, sg: Schem
...
+                        "changes": [columnChangeDict],
+                    }
+
+                    # Send POST request to change column name
+                    schemaChangeResponse = self.synStore.send_api_request(
+                        request_type = "restPOST",
+                        uri = f"https://repo-prod.prod.sagebase.org/repo/v1/entity/{self.existingTableId}/table/transaction/async/start",
+                        body = schemaChangeBody,
+                    )
                     # Exit iteration; only concerned with `Uuid` column
```
thomasyu888, line 2315: I didn't realize this, but I dug deeper and I actually already have a function that does this: `syn._async_table_update` .  You can specify `wait=False` to speed up your code if you don't want to wait for the update to happen.


```suggestion
                   syn._async_table_update("syn51659783", changes=[columnChangeDict])
```

schematic/store/synapse.py
linglp: Thanks for the example @thomasyu888. I really like the idea of separating asset store base class and table operation. I think it makes sure that each class focuses on one thing. Thanks @GiaJordan for the reminder. I will keep in mind that the structure of the code might be changed in the future.

schematic/store/synapse.py
linglp: @thomasyu888 for the example you gave, should the code be something like: 
```
from synapseclient.table import Column
new_col = Column(name='Id', columnType='STRING', maximumSize=64, maximumListLength=1) 
newColResponse = syn.store(new_col)
```

@GiaJordan if using the client directly is simplier, could you make the change in this PR and switch back to use the client? That will be easier to maintain in the future.

linglp: I also dig into the client code more, and just like @thomasyu888 described, I think here is a duplicate of what the client is doing here: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/f3698af6c7d361c5a22d102d2625152933809ceb/synapseclient/client.py#L1035

When storing a column, `restPOST` is taking: `obj.postURI()` which is `/column`, and `obj.json` is the json object: 
```
{ "columnType": "STRING", "concreteType": "org.sagebionetworks.repo.model.table.ColumnModel", "id": "192889", "maximumListLength": 1, "maximumSize": 64, "name": "Id" }
```
which is like what you are doing with the `send_api_request` function

schematic/store/synapse.py
linglp: @thomasyu888 I also noticed that there's a function called `_async_table_update` in client. Is that function also doing something similar? I saw: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/f3698af6c7d361c5a22d102d2625152933809ceb/synapseclient/client.py#L3439

thomasyu888: See a comment I made below with this suggestion.

schematic/store/synapse.py
thomasyu888: Yup, that's right!

tests/test_store.py
linglp, line 221: @GiaJordan since I have merged my PR into your PR, this test will need to be updated?

schematic/store/synapse.py
```diff
@@ -296,6 +298,62 @@ def wrapper(*args, **kwargs):
                     raise ex
         return wrapper
 
+    def send_api_request(self, request_type: str, uri: str, body: Union[Dict, str], headers: Dict = None, endpoint: str = None, **kwargs):
```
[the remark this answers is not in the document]
GiaJordan: @linglp we can remove

tests/test_store.py
GiaJordan: We'll remove it along with `send_api_request`

GiaJordan: that change has been pushed now

schematic/store/synapse.py
```diff
@@ -2130,50 +2159,46 @@ def upsertTable(synStore: SynapseStorage, sg: SchemaGenerator, tableToLoad: pd.D
...
             # Raise if other error
             else:
                 raise ex
 
-        return existingTableId
-
-    def _update_table_uuid_column(synStore: SynapseStorage, table_id: str, sg: SchemaGenerator,) -> None:
+        return self.existingTableId
+    
+    @profile(sort_by='cumulative', strip_dirs=True)
```
linglp, line 2187: @linglp @GiaJordan let me remove this before merge

linglp: @GiaJordan I just removed this line.

linglp (approved):

schematic/store/synapse.py
GiaJordan: Thank you!

PR #1219, opened 2023-05-12
Feature fds 273 coniguration

andrewelamb:
fixes [fds-273](https://sagebionetworks.jira.com/browse/FDS-273)

For a plan of what was to be done please see the [refactor document.](https://docs.google.com/document/d/1o5jTeQy0OV5h2F0Nz52bAbaLhdBxOv9LGl2HKdkAsK4/edit)

PR #1233, opened 2023-06-06
fix: Fixed unit test in test_manifest

linglp:
As @andrewelamb mentioned in the comment of this PR: https://github.com/Sage-Bionetworks/schematic/pull/1219#issuecomment-1572933084, the test wasn't set up correctly. This is the fix to make sure that both cases get run.


andrewelamb (approved):

andrewelamb:
@linglp I'm approving since this works for now,  but I'll have to change this test in the configuration.py refactor.

PR #1235, opened 2023-06-08
Replaced synapse API calls with synapse Python client call 

linglp:
## Changelog
* Replaced synapse API call with synapse python client call to reduce complexity

## Context
Based on previous discussion with @GiaJordan and @milen-sage, it seems like making direct synapse API calls improve performance of the `table upsert` functionality. Initially, I was hoping to see if I could borrow similar approach to improve performance of `manifest/download`. 

But based on @thomasyu888 's comment in PR here: https://github.com/Sage-Bionetworks/schematic/pull/1229, the synapse python client already includes some functionalities that Gianna implemented. It seems like we just need to find the right synapse python client functions to call. I then investigated by replacing direct synapase API calls with functions in synapse python client and tested out the performance. It turns out that we could maintain the same (or even slightly better performance) by calling synapse python client. 

## Steps that I took: 
* I started by testing the performance of the `_update_table_uuid_column` function in the `develop` branch by using `profile` function in `utils.general.py`. It seems like the function took 11.695 seconds to finish See the report here: [_update_table_uuid_column_develop.txt](https://github.com/Sage-Bionetworks/schematic/files/11691407/_update_table_uuid_column_develop.txt)
* I then used Gianna's branch `develop-Uuid-API-calls-FDS-493` and tested out the function of `_update_table_uuid_column`, and it seems like the function finish much faster. See the report here: 
[_update_table_uuid_column_gianna_pr.txt](https://github.com/Sage-Bionetworks/schematic/files/11691447/_update_table_uuid_column_gianna_pr.txt)
* Based on Gianna's PR, I then replaced the direct API calls with synapse python client functions that Tom mentioned and tested again, this time function finished slightly faster than above[_update_table_uuid_column.txt](https://github.com/Sage-Bionetworks/schematic/files/11691458/_update_table_uuid_column.txt)

## To reproduce test results above: 
* Launch schematic API by using `poetry run python3 run_api.py`
* Manually update the table schema here: https://www.synapse.org/#!Synapse:syn51512388/tables/ and change the `Id` to `Uuid`. (Since the goal is to test out the performance of renaming column from `Uuid` to `Id`, we have to ensure the table is using `Uuid` before upsert) 
* In swagger UI interface, try submitting manifest by using the following parameter: 
  * schema_url: https://raw.githubusercontent.com/Sage-Bionetworks/schematic/develop/tests/data/example.model.jsonld
  * data_type: MockRDB
  * dataset_id: syn51680230
  * manifest_record_type: table_and_file
  * access_token: your synapse token
  * asset_view: syn51680311
  * table_manipulation: `upsert`
  * use_schema_label: false
  * file_name: `schematic/tests/data/mock_manifests/rdb_table_manifest_upsert.csv`

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -2285,12 +2285,9 @@ class TableOperations:
                         "maximumListLength": 1,
                     }
 
-                    # Send POST /column request to define new column and get new column ID
-                    newColResponse = self.synStore.send_api_request(
-                        request_type = "restPOST",
-                        uri = "https://repo-prod.prod.sagebase.org/repo/v1/column",
-                        body = columnModelDict,
-                    )
+                    new_col = Column(name='Id', columnType='STRING', maximumSize=64, maximumListLength=1)
+                    newColResponse = self.synStore.syn.store(new_col)
+
 
                     # Define columnChange body
                     columnChangeDict = {
@@ -2304,19 +2301,7 @@ class TableOperations:
                         ]
                     }
 
-                    # Build body for POST request
-                    schemaChangeBody = {
-                        "entityId": self.existingTableId,
-                        "changes": [columnChangeDict],
-                    }
-
-                    # Send POST request to change column name
-                    schemaChangeResponse = self.synStore.send_api_request(
-                        request_type = "restPOST",
-                        uri = f"https://repo-prod.prod.sagebase.org/repo/v1/entity/{self.existingTableId}/table/transaction/async/start",
-                        body = schemaChangeBody,
-                    )
-                    # Exit iteration; only concerned with `Uuid` column
+                    self.synStore.syn._async_table_update(table=self.existingTableId, changes=[columnChangeDict], wait=False)
                 break
 
         return
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -2233,7 +2233,6 @@ class TableOperations:
         except(SynapseHTTPError) as ex:
             # If error is raised because Table has old `Uuid` column and not new `Id` column, then handle and re-attempt upload
             if 'Id is not a valid column name or id' in str(ex):
-                print('triggering this function call that I want to test')
                 self._update_table_uuid_column(sg)
                 synapseDB.upsert_table_rows(table_name=self.tableName, data=self.tableToLoad)
             # Raise if other error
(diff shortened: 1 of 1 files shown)


schematic/store/synapse.py
```diff
@@ -2231,14 +2233,16 @@ def upsertTable(self, sg: SchemaGenerator,):
...
                 self._update_table_uuid_column(sg)
                 synapseDB.upsert_table_rows(table_name=self.tableName, data=self.tableToLoad)
             # Raise if other error
             else:
                 raise ex
 
         return self.existingTableId
-
+    
+    @profile(sort_by='cumulative', strip_dirs=True)
```
linglp, line 2244: will be removed after other people confirm that performance is the same as PR: https://github.com/Sage-Bionetworks/schematic/pull/1229

schematic/store/synapse.py
```diff
@@ -2281,12 +2284,9 @@ def _update_table_uuid_column(self, sg: SchemaGenerator,) -> None:
...
                     }
 
-                    # Send POST /column request to define new column and get new column ID
-                    newColResponse = self.synStore.send_api_request(
-                        request_type = "restPOST",
-                        uri = "https://repo-prod.prod.sagebase.org/repo/v1/column",
-                        body = columnModelDict,
-                    )
+                    new_col = Column(name='Id', columnType='STRING', maximumSize=64, maximumListLength=1)
+                    newColResponse = self.synStore.syn.store(new_col)
```
thomasyu888, line 2288: Nit: Why camel case here?

thomasyu888 (approved):
Looks great! Going to pre-approve.  This is a great PR - it's scoped to one specific change which makes it easy to review.

schematic/store/synapse.py
linglp: hmm I copied Gianna's variable name in her original PR. But I think our code base is mostly using snake case. I should just switch that back to snake case.

GiaJordan:
I believe I was testing from the beginning to end of [this function](https://github.com/Sage-Bionetworks/schematic/blob/develop-Uuid-API-calls-FDS-493-2/schematic_api/api/routes.py#L373)

GiaJordan:
@linglp thank you for gathering this information!

PR #1237, opened 2023-06-14
Allow storage/project/datasets to take in parent_id and benefactor_id

mialy-defelice:
Addresses [FDS-542](https://sagebionetworks.jira.com/browse/FDS-542).

**Background:**

This PR addresses an issue in the AD knowledge portal where they were unable to view datasets within their 'project'. 

What was being attempted is as follows:

**Figure 1:** Benefactor Folder to the dataset folder.
Screen Shot 2023-06-13 at 2 01 33 PM
**Figure 2:** Parent Folder to the dataset folder.
Screen Shot 2023-06-13 at 2 01 51 PM
**Figure 3:** View of the asset view showing projectId, parentId, and benefactorId for the dataset.
Screen Shot 2023-06-13 at 2 02 25 PM
**Figure 4:** View of the REST API endpoint being used.
Screen Shot 2023-06-13 at 2 03 06 PM

The AD Knowledge Portal is the overarching “Project” with a ProjectId SynId = `syn2580853`. This is the immutable projectID no matter what is submitted through the schematic REST API (see figure 3). 

For the DCA they want the study that is surfaced to the user to be “ABC-DS”. Within that folder wanted to find the dataset Genomic Variants (SNP array). To do this they added the annotation `contentType=dataset` to the Genomic Varients folder (`syn38195237`). 

In the API the following was provided:

`asset_view :  syn51324810`

`project_id : syn38190930`

Within schematic when looking at the fileview it can see that the folder Genomic Varients folder (`syn38195237`) is annotated as a dataset, however when trying to check that its in the correct project_id folder, it cant validate that. This is because the `project_id` provided in the API is not its true `project_id`, rather it is the `benefactorId` (again see the third figure).

 After discussing with Anthony, I will expose some new arguments to the current function and endpoint. I can allow users to submit a parentId or a benefactorId. If one of these options is provided I can check that there is a folder with `contentType=dataset` in the given benefactor or parentfolder.

This change will allow users to search the project, parent, or benefactor folders for datasets.

**This PR**
Implements the change discussed above. Users can optionally pass a `parentId` or a `benefactorId`. Cannot provide both parent and benefactor Ids.  If one of these folder IDs is provided a further check will be done to see if the dataset resides in the given folder, within the overall project.

**NOTE**
Designed to throw an error if both a benefactor ID and parent ID are supplied.

**To test**
In the API _for AMPAD_, provide the following to the REST API (must have sufficient permissions):
`V1/storage/project/datasets`
- `Asset View: syn51324810`
- `ProjectId: syn2580853`
- `BenefactorId: syn38190930`

Should return:
```
[
  [
    "syn38195237",
    "Genomic Variants (SNP array)"
  ]
]
```

In the API with standard Schematic Testing Permissions:
`V1/storage/project/datasets`

- `Asset View: syn51707141 `
- `ProjectId: syn23643250`
- `ParentId: syn51707122 `

Should return: 
```
[
  [
    "syn51707123",
    "TestDatasetFolder"
  ]
]
```

Note: BenefactorID seems to be often the same as the ProjectId and seems to do largely with inheritance of permissions...

[FDS-542]: https://sagebionetworks.jira.com/browse/FDS-542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -381,7 +381,7 @@ class SynapseStorage(BaseStorage):
 
         return sorted_projects_list
 
-    def getStorageDatasetsInProject(self, projectId: str) -> List[str]:
+    def getStorageDatasetsInProject(self, projectId: str, parentId: str = '', benefactorId: str = '') -> List[str]:
         """Gets all datasets in folder under a given storage project that the current user has access to.
 
         Args:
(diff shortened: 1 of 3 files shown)


mialy-defelice:
@linglp @afwillia I have thought about it a bit more... we might not need to limit the user to only providing a benefactor **or** parent id. I think it would also make sense for users to want to provide a ProjectId, BenefactorID and Parent ID. Let me know if you have any thoughts.

mialy-defelice:
Closing this PR bc we are not going to do it.

PR #1240, opened 2023-06-15
Fix df_utils/load_df so it more accurately captures integers to prevent Regex Errors.

mialy-defelice:
**Background**
In issue [FDS-547](https://sagebionetworks.jira.com/browse/FDS-547) HTAN reported an interesting error where their regex validation rule was not evaluating properly `re.match("\d+$|Not\sApplicable$|unknown$")`. The regex  is looking for either a string of digits of arbitrary length or "Not Applicable" or "unknown".

The problem was the 0's in their validation column were being converted to "0.0" even through they were entered as "0". This conversion was causing them to fail the validation .

The problem was specific to when "0"s were submitted in the same column as longer digits. For manifests with only "0"s in the column a quirk of pandas was allowing them to not be converted to float and thus to pass validation. 

**Solution**
The essence of this issue boils down to how "0"s can be treated either like numbers or bool (`false`). The func `load_df` is doing a lot of processing of imported manifests to make sure the import of values into schematic is accurate to the users intent (rather than just converting everything to str in bulk). 

In short:

1. The manifest is read as a table where all the values are strings.
2. `ints`: `load_df` attempts to find all the strings that should be integers and stores them in a data frame (`ints`) where each cell records the integer (if it can be converted to an int) or the boolean value `false` if it cannot.
3. `floats`: `load_df` then attempts to convert all the cells it can to float. If it cant it keeps those values as string. This df is a mix of strings and floats.
4. `processed_df`: after some string processing, `load_df` then attempts to layer the ints back into the `floats` df (that already contains the floats and strs). This should ensure that all the values we need as `ints` are preserved as integers. The problem arises bc the way the masking was applied, it look at the `ints` data frame, and looked for if `ints!=False`. In this case all of the 0's that were recorded then evaluated as a boolean false. So none of the 0 integers were layered back over the floats and strings they were expected to replace. Thus the 0's were converted to 0.0. This caused them to fail validation when the regex was applied.
5.  An interesting case: when the entire column to be converted to Float is composed of 0's they are not converted to 0.0...rather they were retained as integers. In this case, it didn't matter that they were not overlaid later as ints bc they were never converted to float in the first place. This is how this error got through.

To get ensure this error does not happen in the future I replaced the check for if an integer was present with a more explicit type check rather than using the boolean check: `ints!=False`. 

The type check is very simply: `ints.applymap(pd.api.types.is_integer)`.

I am additionally adding to the validation test suite to make sure this is checked for in the future.

**To Test**
Via CLI:
Validate against the test manifest `tests/data/example.model.csv`
```
schematic model --config config.yml validate -dt MockComponent -mp tests/data/mock_manifests/Valid_Test_Manifest.csv 
```
[FDS-547]: https://sagebionetworks.jira.com/browse/FDS-547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schematic/utils/df_utils.py
+++ b/schematic/utils/df_utils.py
@@ -60,6 +60,9 @@ def load_df(file_path, preserve_raw_input=True, data_model=False, **load_args):
             pandarallel.initialize(verbose = 1)
             ints = org_df.parallel_applymap(lambda x: np.int64(x) if str.isdigit(x) else False, na_action='ignore').fillna(False)
 
+        # Identify cells converted to intergers
+        ints_tf_df = ints.applymap(pd.api.types.is_integer)
+
         # convert strings to numerical dtype (float) if possible, preserve non-numerical strings
         for col in org_df.columns:
             float_df[col]=pd.to_numeric(float_df[col], errors='coerce')
@@ -68,9 +71,9 @@ def load_df(file_path, preserve_raw_input=True, data_model=False, **load_args):
         
         # Trim nans and empty rows and columns
         processed_df = trim_commas_df(float_df)
-        
+
         # Store values that were converted to type int in the final dataframe
-        processed_df=processed_df.mask(ints != False, other = ints)  
+        processed_df=processed_df.mask(ints_tf_df, other = ints)
         
         # log manifest load and processing time
         logger.debug(f"Load Elapsed time {perf_counter()-t_load_df}")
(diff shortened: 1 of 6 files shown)


linglp:
Thanks @mialy-defelice for documenting the problem. I played with the original code, and I think the problem described above can be summarized in the code below: 
```
test_int = pd.DataFrame({"test_column": [0, 2, "string", "string", 123456897, 0, 25689]})
float_df_test = pd.DataFrame({"test_column": [0.0, 2.0, False, False, 123456897.0, 0.0, 25689.0]})
float_df_masked = float_df_test.mask(test_int !=False, other=test_int)
```

When printed out the float_df_mask: 
```
  test_column
0         0.0
1           2
2      string
3      string
4   123456897
5         0.0
6       25689
```

It shows that all big integers were converted back but 0.0 is still kept as float. 
When I looked at the code, it doesn't seem like we have a unit test for load_df which I think we probably should. Should I open an issue for this?

linglp (approved):

PR #1252, opened 2023-06-29
Only add validation rules to google sheets

mialy-defelice:
This PR addresses the issue where validation rules applied to Excel did not work as expected. In this PR I changed the manifest generation processing to only allow validation rules to be applied to Google Sheets. I tested excel and google sheets manifests with and without annotations. I was able to see VRs with Google Sheets but not excel, as expected. See [here](https://docs.google.com/spreadsheets/d/1x9dOCDlc5GFiDhRZzxg4-nuse-QSm41tLgn6ODAPXc8/edit?usp=sharing), for some CLI testing commands (for HTAN BU the master fileview is `syn22125563`).


schematic/manifest/generator.py
```diff
@@ -1236,7 +1242,7 @@ def _gather_all_fields(self, fields, json_schema):
         )
         return required_metadata_fields
 
-    def get_empty_manifest(self, json_schema_filepath=None):
+    def get_empty_manifest(self, json_schema_filepath=None, sheet_url=None):
```
linglp, line 1245: since that we will deprecate `sheet_url` in the future, is there a way to get away from adding `sheet_url` parameters to more places? What about using `output_format` parameter?

linglp:
What is the Jira issue related to this PR? I don't remember that we talked about removing validation rules from Excel.. is that a request that we want to support?

schematic/manifest/generator.py
mialy-defelice: For now this is the only direct way of specifying that we are using google sheets. I imagine `sheet_url` will be depreciated when you refactor the generator, and in that case I think the rewrite will be significant enough that specifying `sheet_url` here shouldn't be too much of a burden.

mialy-defelice:
> What is the Jira issue related to this PR? I don't remember that we talked about removing validation rules from Excel.. is that a request that we want to support?

The only validation rules that are actually applied to GoogleSheets or Excel for _real-time_ validation is `regex match`.  Removing _all_ validation rules from excel in this context functionally means just removing `regex match` since Excel cannot apply `regex`. 

To be more explicit in what is getting removed, I updated the code to more specifically only apply `regex match` for google sheets, but this means we will be extracting validation rules even when we know we will not be using them. If we determine this is undesirable I can move the check back up to only extract VRs if creating google sheets.

linglp:
@mialy-defelice The code looks good, even though I am wondering if there's a way to remove all validation rules after the manifest URL gets created. Also, I ran `test_api.py` to make sure that the code change doesn't affect anything. There's only one error from the tests, and it seems like that it is related to not using the latest version of develop. I believe if you pull the latest from develop, that error should be resolved.

mialy-defelice:
> @mialy-defelice The code looks good, even though I am wondering if there's a way to remove all validation rules after the manifest URL gets created. Also, I ran `test_api.py` to make sure that the code change doesn't affect anything. There's only one error from the tests, and it seems like that it is related to not using the latest version of develop. I believe if you pull the latest from develop, that error should be resolved.

Okay, I pulled from develop and resolved conflicts yesterday are you using the latest commit I pushed?
I would rather not add anything to the google sheet in the first place rather than try to remove them later since its so tricky working with the sheets and is a bit unnecessary. I think with the refactor we can be a lot more deliberate about how the sheets and excel are created so they don't rely so heavily on one another.

schematic/manifest/generator.py
```diff
@@ -1254,7 +1257,7 @@ def get_empty_manifest(self, json_schema_filepath=None, strict=None):
         )
 
         manifest_url = self._create_empty_gs(
```
linglp, line 1259: I noticed that `strict` in not in the doc string of function `_create_empty_gs`. Could you add that?

schematic/manifest/generator.py
```diff
@@ -1102,10 +1103,11 @@ def _create_requests_body(
         requests_body = {}
         requests_body["requests"] = []
         for i, req in enumerate(ordered_metadata_fields[0]):
-            # Gather validation rules and valid values for attribute
-            validation_rules = self.sg.get_node_validation_rules(req)
+            # Gather validation rules and valid values for attribute, if using google sheets.
 
-            if validation_rules:
+            validation_rules = self.sg.get_node_validation_rules(req)
+
```
linglp, line 1109: I think it will also be worth adding some comments here to explain that we are only adding regex match to google sheet

schematic/manifest/generator.py
```diff
@@ -1303,7 +1306,7 @@ def set_dataframe_by_url(
 
         return sh
 
-    def get_dataframe_by_url(self, manifest_url: str) -> pd.DataFrame:
+    def get_dataframe_by_url(self, manifest_url: str, strict=None) -> pd.DataFrame:
```
linglp, line 1309: I saw that `strict` is not being used. Could you remove it?

schematic/manifest/generator.py
```diff
@@ -1353,7 +1356,7 @@ def map_annotation_names_to_display_names(
         return annotations.rename(columns=label_map)
 
     def get_manifest_with_annotations(
-        self, annotations: pd.DataFrame, strict: bool=None,
+        self, annotations: pd.DataFrame, strict: bool=None, sheet_url:bool=None,
     ) -> Tuple[ps.Spreadsheet, pd.DataFrame]:
         """Generate manifest, optionally with annotations (if requested).
```
linglp, line 1362: Could you also add comments here so that the doc string matches the parameters/

schematic/manifest/generator.py
mialy-defelice: Yep!

schematic/manifest/generator.py
mialy-defelice: Done!

schematic/manifest/generator.py
mialy-defelice: Done!

linglp (approved):
It looks good overall. Thanks @mialy-defelice

PR #1254, opened 2023-07-05
BugFix: Fixed an issue where files on synapse were not being annotated correctly

GiaJordan:
Fixed an issue where annotations for files were not being added correctly. 
For file-based metadata, the table ID was being added for the `entityId` column, and this synId was then used as the target for which entity the annotations should be applied to. This PR now stores the entityId of the files themselves so that they can be annotated properly.

## To Test

- Submit [manifest](https://docs.google.com/spreadsheets/d/1q4VaK4ta2WEx4O2u9IztGrn2MJ01Rirc9Ca2OI_FoF4/edit#gid=0) with `-mrt table_and_file` and `-tm replace`
- Check annotations for dummy dataset files in dataset syn39241199


schematic/store/synapse.py
```diff
@@ -600,6 +594,33 @@ def updateDatasetManifestFiles(self, sg: SchemaGenerator, datasetId: str, store:
         manifest = manifest.fillna("") 
         
         return manifest_id, manifest
+    
+    def _get_file_entityIDs(self, manifest: pd.DataFrame, dataset_files: List, only_new_files: bool = False):
```
linglp, line 598: Can you also create a test for `_get_file_entityIDds` function?

schematic/store/synapse.py
```diff
@@ -1343,15 +1364,32 @@ def add_entities(
...
         for idx, row in manifest.iterrows():
-            if not row["entityId"] and (manifest_record_type == 'file_and_entities' or 
-                manifest_record_type == 'table_file_and_entities'):
-                manifest, entityId = self._create_entity_id(idx, row, manifest, datasetId)
-            elif not row["entityId"] and manifest_record_type == 'table_and_file':
-                # If not using entityIds, fill with manifest_table_id so 
-                row["entityId"] = manifest_synapse_table_id
-                manifest.loc[idx, "entityId"] = manifest_synapse_table_id
-                entityId = ''
+            if 'filename' not in [col.lower() for col in manifest.columns]:
```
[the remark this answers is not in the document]
linglp: @GiaJordan I tried commented out this if statement and the else statement, and it seems like the annotation could still be added correctly. The code that you added from 1368-1376 is sufficient to fix the issue

linglp:
Thanks @GiaJordan I think the functionality works in general. I have tested by using the submit endpoint in restAPI: 
* schema_url: https://raw.githubusercontent.com/Sage-Bionetworks/schematic/develop/tests/data/example.model.jsonld
* data_type: empty
* dataset_id: syn39241199
* manifest_record_type: "table_and_file"
* asset_view: syn51707141
* table_manipulation: replace

I have tested after deleting the `id` and `entity_id` column and made sure that annotation gets added correctly to dataset: 
Screen Shot 2023-07-10 at 2 53 35 PM
file:

schematic/store/synapse.py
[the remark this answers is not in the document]
GiaJordan: The bug was caused by assigning the `manifest_synapse_table_id` to the `entityId` column for every row. Adding in this line (in addition to line 1369: `if 'filename' in [col.lower() for col in manifest.columns]:`) completes the behavior split for file based and non-file based metadata. We still want the last `else` clause to trigger so that the annotations get assigned appropriately for each row/file.

schematic/store/synapse.py
```diff
@@ -1343,6 +1369,18 @@ def add_entities(
             manifest (pd.DataFrame): modified to add entitiyId as appropriate.
 
         '''
+
+        # Expected behavior is to annotate files if `Filename` is present regardless of `-mrt` setting
+        if 'filename' in [col.lower() for col in manifest.columns]:
+            # get current list of files and store as dataframe
+            dataset_files = self.getFilesInStorageDataset(datasetId)
```
mialy-defelice, line 1376: can you update the variable naming here to make the differences more clear, to me its not clear upon reading how `files` differs exactly from `dataset_files`, since it takes `dataset_files` as a parameter, furthermore `files` still contains information for the entire dataset.

Maybe something along the lines of: `filenames_list` and `file_info_dict`

mialy-defelice (changes requested):

mialy-defelice:
Tested via the API and it works for me as well.

schematic/store/synapse.py
GiaJordan: Ahh I see how that could be confusing, I'll update

GiaJordan:
> @GiaJordan Do we have tests to make sure that annotations get added correctly for both file-based manifest and non file-based manifests? Could you also link the Jira issue?

We do now, and [sure](https://sagebionetworks.jira.com/browse/FDS-654)

schematic/store/synapse.py
```diff
@@ -600,6 +594,38 @@ def updateDatasetManifestFiles(self, sg: SchemaGenerator, datasetId: str, store:
...
+        """
+        files = {"Filename": [], "entityId": []}
+
+        if only_new_files:
+            if manifest is None:
+                raise UnboundLocalError(
+                    "No manifest was passed in, a manifest is required when `only_new_files` is True."
+                )
+            
+            # find new files if any
```
linglp, line 617: Note: New files are files that are not in the manifest. @GiaJordan Could you add a comment here to clarify?

tests/test_store.py
```diff
@@ -113,9 +114,37 @@ def test_getFileAnnotations(self, synapse_store):
 
         assert expected_dict == actual_dict
 
-    def test_annotation_submission(self, synapse_store, helpers, config: Configuration):
-        manifest_path = "mock_manifests/annotations_test_manifest.csv"
-
+    @pytest.mark.parametrize('only_new_files',[True, False])
+    def test_file_entityIds(self, helpers, synapse_store, only_new_files):
```
linglp, line 118: could you change the function name to `test_get_file_entityIds`

GiaJordan: sure

mialy-defelice (commented):
Running `pytest tests/test_store.py` returned the following warning though the tests passed:

```tests/test_store.py::TestDownloadManifest::test_entity_type_checking[syn29862078]
  /Users/mialydefelice/Library/Caches/pypoetry/virtualenvs/schematicpy-rs9bwhSl-py3.9/lib/python3.9/site-packages/synapseclient/client.py:744: UserWarning: 
  This entity has access restrictions. Please visit the web page for this entity (syn.onweb("syn29862078")). Look for the "Access" label and the lock icon underneath the file name. Click "Request Access", and then review and fulfill the file download requirement(s).
  ```
  Is this expected?

GiaJordan:
> Is this expected?

@mialy-defelice yes, there's a censored and uncensored manifest in that dataset, and schematic will try to download the one with sensitive info first and if the user doesn't have access it will use the censored one

mialy-defelice (approved):
Okay looks good!

PR #1259, opened 2023-07-11
add Component as a required attribute to the example data model

mialy-defelice:
This PR addresses the issue where the desired functionality is to have Component show up as a required column in Example manifests. The data model previously did not contain `Component` as an attribute. This change adds `Component` as a Required attribute. 

Test via CLI with:
```
schematic manifest -c config.yml get -dt Patient -p tests/data/example.model.jsonld -s
```

### DIFF
--- a/tests/data/example.model.csv
+++ b/tests/data/example.model.csv
@@ -1,4 +1,5 @@
 Attribute,Description,Valid Values,DependsOn,Properties,Required,Parent,DependsOn Component,Source,Validation Rules
+Component,,,,,TRUE,,,,
 Patient,,,"Patient ID, Sex, Year of Birth, Diagnosis, Component",,FALSE,DataType,,,
 Patient ID,,,,,TRUE,DataProperty,,,
 Sex,,"Female, Male, Other",,,TRUE,DataProperty,,,
--- a/tests/data/example.model.jsonld
+++ b/tests/data/example.model.jsonld
@@ -1973,6 +1973,23 @@
             }
         },
         {
+            "@id": "bts:Component",
+            "@type": "rdfs:Class",
+            "rdfs:comment": "TBD",
+            "rdfs:label": "Component",
+            "rdfs:subClassOf": [
+                {
+                    "@id": "schema:Thing"
+                }
+            ],
+            "schema:isPartOf": {
+                "@id": "http://schema.biothings.io"
+            },
+            "sms:displayName": "Component",
+            "sms:required": "sms:true",
+            "sms:validationRules": []
+        },
+        {
             "@id": "bts:Patient",
             "@type": "rdfs:Class",
             "rdfs:comment": "TBD",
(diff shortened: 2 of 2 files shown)


GiaJordan (changes requested):
I was able to generate a manifest with the `Component` column highlighted as required!
Before we merge, could you also regenerate the `example.single_rule.model.jsonld` file from the `example.model.csv` file? Its the same MockComponent but each attribute has `Required` set to `FALSE` and the validation rules have the default `error` or `warning` specified explicitly. I could also open up another PR targeting this branch with those changes

mialy-defelice:
> Thank you for catching this!
> I was able to generate a manifest with the `Component` column highlighted as required.
> Before we merge, could you also regenerate the `example.single_rule.model.jsonld` file from the `example.model.csv` file? Its the same MockComponent but each attribute has `Required` set to `FALSE` and the validation rules have the default `error` or `warning` specified explicitly. I could also open up another PR targeting this branch with those changes.


Do you have that model csv?

GiaJordan:
> Do you have that model csv?

@mialy-defelice no, I had been using the `example.model.csv` file so that each change in that model wouldn't have to be replicated in another csv

mialy-defelice:
> > Do you have that model csv?
> 
> @mialy-defelice no, I had been using the `example.model.csv` file so that each change in that model wouldn't have to be replicated in another csv

I think in this case its worthwhile to maintain a separate csv or to create a script so that for each change to the data model one wouldn't have to remake the `.model.csv`

GiaJordan:
> > > Do you have that model csv?
> > 
> > 
> > @mialy-defelice no, I had been using the `example.model.csv` file so that each change in that model wouldn't have to be replicated in another csv
> 
> I think in this case its worthwhile to maintain a separate csv or to create a script so that for each change to the data model one wouldn't have to remake the `.model.csv`

Okay, that makes sense

GiaJordan:
> > > > > Do you have that model csv?
> > > > 
> > > > 
> > > > @mialy-defelice no, I had been using the `example.model.csv` file so that each change in that model wouldn't have to be replicated in another csv
> > > 
> > > 
> > > I think in this case its worthwhile to maintain a separate csv or to create a script so that for each change to the data model one wouldn't have to remake the `.model.csv`
> > 
> > 
> > Okay, that makes sense
> 
> I was going to update the change here, but I think now we should open a separate PR since its a bit outside the scope of this one to add a new csv as well.

sounds good!

AmyHeiser:
@mialy-defelice  Is this related to an FDS ticket?

mialy-defelice:
> @mialy-defelice Is this related to an FDS ticket?

@AmyHeiser No, I think bc it was so small the decision was to not make a ticket. But I think this issue originally came out of a user testing situation experienced by you and Milen,... so idk if there's a ticket for that?

GiaJordan:
> I was going to update the change here, but I think now we should open a separate PR since its a bit outside the scope of this one to add a new csv as well.

https://sagebionetworks.jira.com/browse/FDS-1239

GiaJordan (approved):

PR #1260, opened 2023-07-11
[bug fix] Fixed errors when calling `get_empty_manifest`

linglp:
## Context
`get_empty_manifest` is supposed to take in two parameters, json_schema_filepath, strict.. but in some parts of the code base, we are calling it only with `strict` parameter

Related to: https://sagebionetworks.jira.com/browse/FDS-712
## Changelog
Updated how `get_empty_manifest` is called in the code base.

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -9,7 +9,7 @@ import pandas as pd
 from pathlib import Path
 import pygsheets as ps
 from tempfile import NamedTemporaryFile
-from typing import Dict, List, Tuple, Union
+from typing import Dict, List, Tuple, Union, Optional
 
 from schematic.schemas.generator import SchemaGenerator
 from schematic.utils.google_api_utils import (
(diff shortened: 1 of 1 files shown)


schematic/manifest/generator.py
```diff
@@ -1236,18 +1236,19 @@ def _gather_all_fields(self, fields, json_schema):
         )
         return required_metadata_fields
 
-    def get_empty_manifest(self, json_schema_filepath=None, strict=None):
+    def get_empty_manifest(self, strict: bool=None, json_schema_filepath: str=None,):
```
linglp, line 1239: Use optional boolean

schematic/manifest/generator.py
```diff
@@ -1236,18 +1236,19 @@ def _gather_all_fields(self, fields, json_schema):
         )
         return required_metadata_fields
 
-    def get_empty_manifest(self, json_schema_filepath=None, strict=None):
+    def get_empty_manifest(self, strict: Optional[bool], json_schema_filepath: str=None):
         """Create an empty manifest using specifications from the
         json schema.
         Args:
+            strict (bool): strictness with which to apply validation rules to google sheets.
```
mialy-defelice, line 1243: Can you just add here: True, blocks incorrect entries, False, raises a warning

linglp: Yes. Thank Mialy! I have addressed that.

mialy-defelice (approved):
Looks good! Just to note, we are introducing a lot of changes to the same area so need to be conscious of possible merge conflicts.

PR #1268, opened 2023-07-19
[bug fix] Updated .synapseCache, functions to calculate cache, and cleared manifests before each download

linglp:
Related to: https://sagebionetworks.jira.com/browse/FDS-768

Should merge after: [1269](https://github.com/Sage-Bionetworks/schematic/pull/1269)

## Context
Schematic ran into "disk space error" again on AWS: 
```
7/18/2023, 9:25:13 AM PDT	OSError: [Errno 28] No space left on device: '/root/.synapseCache/216/127095216'
```
After some investigation, I discovered that the code is not purging cache correctly and also after nginx deployment, cache is now stored in a different folder. 

## Changelog
* Updated paths of .synapseCache on AWS
* Updated to use `du -sh` to check size of cache
* Deleted cache that was 15 min old when cache reaches 1GB instead of 7GB
* Since we also download manifests when triggering `manifest/download` endpoint, clear those manifests before each download 

## Test
* Confirm deleting cache is working now in AWS
Kept calling `storage/project/dataset` with HTAN asset view and project id
* Checked "ephemeral storage" chart on AWS. The chart below shows a little dent which means that the cache actually gets deleted 
Screen Shot 2023-07-27 at 4 32 44 PM

* Also checked the log and saw that log files printed out the number of files that got deleted. 
Screen Shot 2023-07-27 at 3 53 01 PM
Screen Shot 2023-07-27 at 4 31 29 PM


schematic/store/synapse.py
```diff
@@ -43,17 +43,17 @@
 from schematic_db.rdb.synapse_database import SynapseDatabase
 
 
-from schematic.utils.df_utils import update_df, load_df, col_in_dataframe, populate_df_col_with_another_col
```
linglp, line 46: `populate_df_col_with_another_col` function only shows up in df_utils and test_utils, and is not used anywhere else.

schematic/store/synapse.py
```diff
@@ -7,7 +7,7 @@
 import logging
 import secrets
 from dataclasses import dataclass
-import tempfile
```
linglp, line 10: made sure that tempfile is not used here

schematic/store/synapse.py
```diff
@@ -75,12 +75,16 @@ def _download_manifest_to_folder(self) -> File:
         """
         if "SECRETS_MANAGER_SECRETS" in os.environ:
             temporary_manifest_storage = "/var/tmp/temp_manifest_download"
+            # clear out all the existing manifests
```
linglp, line 78: clear all the existing manifests before downloading new manifests.

schematic/store/synapse.py
```diff
@@ -182,36 +186,30 @@ def __init__(
...
         self.storageFileview = CONFIG.synapse_master_fileview_id
         self.manifest = CONFIG.synapse_manifest_basename
+        self.root_synapse_cache = "/root/.synapseCache"
         self._query_fileview()
 
-    def _purge_synapse_cache(self, root_dir: str = "/var/www/.synapseCache/", maximum_storage_allowed_cache_gb=7):
+    def _purge_synapse_cache(self, maximum_storage_allowed_cache_gb=1):
         """
-        Purge synapse cache if it exceeds 7GB
+        Purge synapse cache if it exceeds 1GB
```
linglp, line 194: purge synapse cache if it exceeds 1GB.

schematic/store/synapse.py
```diff
@@ -182,36 +186,30 @@ def __init__(
         self.project_scope = project_scope
         self.storageFileview = CONFIG.synapse_master_fileview_id
         self.manifest = CONFIG.synapse_manifest_basename
+        self.root_synapse_cache = "/root/.synapseCache"
```
linglp, line 189: default the root of synapse cache to `/root/.synapseCache`. Note: this should not affect users when they use schematic locally because `self.root_synapse_cache` does not get used else where other than clearing .synapseCache on AWS.

linglp: In the future, `root_synapse_cache` should probably be an argument when instantiating `SynapseStorage` class

tests/test_utils.py
```diff
@@ -39,8 +41,59 @@ def synapse_store():
...
+    def test_check_synapse_cache_size(self,tmp_path):
+        mock_synapse_cache_dir = tmp_path / ".synapseCache"
+        mock_synapse_cache_dir.mkdir()
+
+        mock_synapse_table_query_csv = mock_synapse_cache_dir/ "mock_synapse_table_query.csv"
+        mock_synapse_table_query_csv.write_text("example file for calculating cache")
+
+        file_size = check_synapse_cache_size(mock_synapse_cache_dir)
+
+        if IN_GITHUB_ACTIONS:
```
linglp, line 92: For some reasons, when running in github action, the file size changes

tests/test_utils.py
```diff
@@ -39,8 +41,59 @@ def synapse_store():
         synapse_store = SynapseStorage()
     yield synapse_store
 
-
 class TestGeneral:
+    def test_clear_synapse_cache(self, tmp_path):
```
linglp, line 45: Create a mock file under mock folder .synapseCache and make sure that cache can be cleared correctly using `clear_synapse_cache` function.

tests/test_utils.py
```diff
@@ -1,34 +1,36 @@
-import logging
 import json
```
linglp, line 1: reordering import statement using `isort`

schematic_api/Dockerfile
```diff
@@ -73,7 +73,9 @@ WORKDIR ${APP_DIR}
 
 # copy other files to app/app
 # Note: run_api.py is not needed
-COPY ./pyproject.toml ./poetry.lock ./config.yml ./main.py ./
+
```
linglp, line 76: fixed the issue related to docker file after configuration class refactor. This part was borrowed from PR: https://github.com/Sage-Bionetworks/schematic/pull/1269

schematic/utils/general.py
```diff
@@ -57,24 +60,66 @@ def get_dir_size(path: str):
...
+        datetime:  return result of date time calculation
+    """
+    if before_or_after=="before": 
+        date_time_result = input_date - timedelta(minutes=minutes)
+    else:
+        date_time_result = input_date + timedelta(minutes=minutes)
+    return date_time_result
+
+
+def check_synapse_cache_size(directory='/root/.synapseCache')-> Union[float, int]:
```
linglp, line 81: I found the previous way of calculating size of .synapseCache a bit confusing. When I debugged the space limit error, I used `du -sh <path to .synapseCache>` command, so I think I should make it consistent and use this command when I calculate cache

schematic/store/synapse.py
```diff
@@ -182,36 +186,30 @@ def __init__(
...
+            dir_size_bytes = check_synapse_cache_size(directory=self.root_synapse_cache)
+            # if 1 GB has already been taken, purge cache before 15 min
+            if dir_size_bytes >= maximum_storage_allowed_cache_bytes:
+                num_of_deleted_files = clear_synapse_cache(self.syn.cache, minutes=15)
+                logger.info(f'{num_of_deleted_files}  files have been deleted from {self.root_synapse_cache}')
             else:
-                remaining_space = total_ephemeral_storage_bytes - nbytes
-                converted_space = convert_size(remaining_space)
-                logger.info(f'Estimated {remaining_space} bytes (which is approximately {converted_space}) remained in ephemeral storage after calculating size of .synapseCache excluding OS')
+                logger.info(f'the total size of .synapseCache is: {nbytes} bytes')
```
linglp, line 212: OS actually can take around 14-17% of our ephemeral storage (20GiB), so I think the previous calculation is a bit confusing. Instead of "guessing" how much space we have, I think printing out the size of .synapseCache is more useful

schematic/store/synapse.py
thomasyu888: Nit: update the doc.  this function is purging synapse cache when it exceeds a certain hard drive cap. The default is 1gb but it can be changed.

schematic/utils/general.py
```diff
@@ -57,24 +60,66 @@ def get_dir_size(path: str):
                 total += get_dir_size(entry.path)
     return total
 
+def calculate_datetime(minutes: int, input_date: datetime, before_or_after: str = "before") -> datetime:
```
thomasyu888, line 63: One thought to simplify this function is to allow for negative or positive minutes. Not sure if that's allowed in the timedelta though.

tests/test_utils.py
thomasyu888: Weird. This should be a comment.

schematic/store/synapse.py
thomasyu888: This can be a comment

tests/test_utils.py
```diff
@@ -39,8 +41,59 @@ def synapse_store():
...
+        time.sleep(1)
+
+        # clear cache
+        my_cache = cache.Cache(cache_root_dir=mock_synapse_cache_dir)
+        clear_synapse_cache(my_cache, minutes=0.0001)
+        # make sure that cache files are now gone
+        assert os.path.exists(mock_synapse_table_query_csv) == False
+        assert os.path.exists(mock_cache_map) == False
+
+    def test_calculate_datetime(self):
```
thomasyu888, line 73: Nit: sometimes simple is best so no need to use parametrize but these are technically two separate test cases.

thomasyu888 (commented):
This looks great!! Just some minor comments

tests/test_utils.py
linglp: Thanks Tom. I will add this as a comment.

schematic/store/synapse.py
```diff
@@ -182,36 +186,30 @@ def __init__(
...
+        Purge synapse cache if it exceeds 1GB
         Args:
-            root_dir: directory of the .synapseCache function
-            maximum_storage_allowed_cache_gb: the maximum storage allowed before purging cache. Default is 7 GB. 
-
+            maximum_storage_allowed_cache_gb: the maximum storage allowed before purging cache. Default is 1 GB. 
         Returns: 
-            if size of cache reaches a certain threshold (default is 7GB), return the number of files that get deleted
-            otherwise, return the total remaining space (assuming total ephemeral storage is 20GB on AWS )
+            if size of cache reaches a certain threshold (default is 1GB), return the number of files that get deleted
```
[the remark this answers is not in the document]
linglp: Good point. Thank you! I will update the comment

schematic/store/synapse.py
linglp: Thanks. I will add a comment here.

tests/test_utils.py
linglp: Good point. Thanks Tom. Let me separate them

schematic/utils/general.py
```diff
@@ -57,24 +60,66 @@ def get_dir_size(path: str):
...
+
+    Args:
+        input_date (datetime): date time object provided by users
+        minutes (int): number of minutes
+        before_or_after (str): default to "before". if "before", calculate x minutes before current date time. if "after", calculate x minutes after current date time. 
+
+    Returns:
+        datetime:  return result of date time calculation
+    """
+    if before_or_after=="before":
```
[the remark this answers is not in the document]
linglp: Thanks Tom. I will add some code to catch exception.

schematic/utils/general.py
```diff
@@ -57,24 +60,66 @@ def get_dir_size(path: str):
...
+    elif "M" in size:
+        size_in_mb = float(size.rstrip('M'))
+        byte_size = size_in_mb * 1000000
+    elif "G" in size: 
+        size_in_gb = float(size.rstrip('G'))
+        byte_size = convert_gb_to_bytes(size_in_gb)
+    elif "B" in size:
+        byte_size = float(size.rstrip('B'))
+    else:
+        logger.error('Cannot recongize')
```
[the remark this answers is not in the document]
linglp: Thanks Tom. Let me improve this error message.

schematic/utils/general.py
```diff
@@ -57,24 +60,66 @@ def get_dir_size(path: str):
...
-        return "0B"
-    size_name = ("B", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB")
-    # calculate the log of size (in bytes) to base 1024 and run it down to the nearest integer
-    index_int = int(math.floor(math.log(size_bytes, 1024)))
-    # return the value of 1024 raised to the power of index
-    power_cal = math.pow(1024, index_int)
-    # convert bytes to a different unit if applicable
-    size_bytes_converted = round(size_bytes / power_cal, 2)
-    return f"{size_bytes_converted} {size_name[index_int]})"
+    command = ['du', '-sh', directory]
```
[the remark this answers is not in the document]
linglp: Thanks for the note Tom. I added an additional comment here

schematic/utils/general.py
```diff
@@ -57,24 +60,66 @@ def get_dir_size(path: str):
...
+    output = subprocess.check_output(command).decode('utf-8')
```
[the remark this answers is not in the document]
linglp: Thanks for pointing this out. I updated the code to use `subprocess.run`

schematic/store/synapse.py
linglp: Good point. Thanks Tom

tests/test_utils.py
```diff
@@ -39,8 +41,68 @@ def synapse_store():
...
+        expected_result_date_after = datetime.strptime("07/20/23 17:46:34", '%m/%d/%y %H:%M:%S')
+        assert minutes_after == expected_result_date_after
+
+    def test_calculate_datetime_raise_error(self):
+        with pytest.raises(ValueError):
+            input_date = datetime.strptime("07/20/23 17:36:34", '%m/%d/%y %H:%M:%S')
+            minutes = calculate_datetime(input_date=input_date, minutes=10, before_or_after="error")
+    
+        
+    def test_check_synapse_cache_size(self,tmp_path):
```
linglp, line 91: Just add a comment so that users using windows won't be surprised that the test fail. Or add pytest mark to mark this test and skip this test on windows.

GiaJordan (approved):

tests/test_utils.py
linglp: Yes sure. I could add that

PR #1270, opened 2023-07-23
[bug fix] Updated manifest download endpoint to avoid querying file view

linglp:
Related to the Jira issue here: https://sagebionetworks.jira.com/browse/FDS-776

## Changelog
Avoid instantiating file view before calling the login method. Since we could not cache file view, this change will allow manifest/download to be faster and avoid downloading the same file repeatedly.


mialy-defelice (approved):
Confirmed that I could still return a manifest with changes incorporated.

PR #1272, opened 2023-07-27
update channel id to `fair-data-team` channel

GiaJordan:
the `fair-data` slack channel has been archived. This PR updates the channel id for release notifications to notify the `fair-data-team` channel instead, along with the existing behavior of notifying the `fair-data-tools` channel


linglp (approved):
Thanks. This makes sense!

GiaJordan:
> Thanks. This makes sense!

no prob, thanks for reviewing!

PR #1277, opened 2023-07-31
Add API endpoint that returns the current version of schematic

GiaJordan:
This PR adds an endpoint `/version` to the API to be used when schematic is deployed to AWS or running in a docker container. The endpoint will return the version of schematic, as identified by the tag that triggered the creation of the docker image. The tag is extracted from GitHub and is assigned as an environment variable `VERSION` in the `DockerFile` this is then available when a container is run.

Currently not usable when running the API outside of a docker container, it will return a `500 SERVER ERROR` response.

The schematic version tag was also added as an environment variable to the non-api image as well.

## To Test

- Pull the latest test image based on this branch 
  - `docker pull ghcr.io/sage-bionetworks/schematic:0.1.58-beta`
- Run the API with a docker container
  - `docker run --rm -p 3001:3001 -v $(pwd):/schematic -w /schematic --name schematic -e GE_HOME=/usr/src/app/great_expectations/ ghcr.io/sage-bionetworks/schematic:0.1.58-beta python run_api.py`
- Use SwaggerUI to use the api and use the `/version` [endpoint](http://localhost:3001/v1/ui/#/Version/schematic_api.api.routes.get_schematic_version)
- verify that the endpoint returns the version os schematic used to create the image, in this case `v0.1.58-beta`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,17 @@
 # FROM python:3.10.6
 FROM python:3.10.8-slim-bullseye
 
+# add version as a build argument
+ARG tag
+
 ENV PYTHONFAULTHANDLER=1 \
   PYTHONUNBUFFERED=1 \
   PYTHONHASHSEED=random \
   PIP_NO_CACHE_DIR=off \
   PIP_DISABLE_PIP_VERSION_CHECK=on \
   PIP_DEFAULT_TIMEOUT=200 \
-  POETRY_VERSION=1.2.0
+  POETRY_VERSION=1.2.0 \
+  VERSION=$tag
 
 WORKDIR /usr/src/app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,13 @@
 # FROM python:3.10.6
 FROM python:3.10.8-slim-bullseye
 
-# add version as a build argument
-ARG TAG
-
 ENV PYTHONFAULTHANDLER=1 \
   PYTHONUNBUFFERED=1 \
   PYTHONHASHSEED=random \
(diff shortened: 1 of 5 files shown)


schematic_api/api/routes.py
```diff
@@ -815,3 +815,10 @@ def get_nodes_display_names(schema_url: str, node_list: list[str]) -> list:
     node_display_names = gen.get_nodes_display_names(node_list, mm_graph)
     return node_display_names
 
+def get_schematic_version() -> str:
+    """
+    Return the current version of schematic
+    """
+    if "VERSION" in os.environ:
+        version = os.environ["VERSION"]
+    return version
```
linglp, line 824: If users use this locally, I think it would make sense to provide a more specific error message so that users know that this endpoint can only be used within a docker container. The current message: "UnboundLocalError: local variable 'version' referenced before assignment" is confusing for users

linglp (changes requested):

schematic_api/api/routes.py
GiaJordan: That makes sense

Dockerfile
```diff
@@ -1,13 +1,17 @@
 # FROM python:3.10.6
 FROM python:3.10.8-slim-bullseye
 
+# add version as a build argument
```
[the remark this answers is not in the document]
GiaJordan: Ahh I see, yeah I can

linglp (approved):
@GiaJordan looks good

lakikowolfe (commented):
output looks good!

PR #1278, opened 2023-08-04
[bug fix] Do not pull annotations when use_annotations set to False and there is no existing manifests

linglp:
## Changelog
* When there is no existing manifests, only pull annotations when `use_annotations = True`.
* Currently, the logic of filling in Filename and `entityId` is in function `updateDatasetManifestFiles`, and it only pulls in `Filename` and `entityId` when there is an existing manifest. I changed the behavior so that when there is no existing manifests and `use_annotations = False`, filename and entityId in manifest still get filled. 

## Discussion 
In the old code, if there is no existing manifest, but a component is file-base, and there is annotation but users are not interested in annotations, there is this statement here that only give users `Filename`, `eTag`, and `entityId`
```
            else:
                #Subset columns if no interested in user-defined annotations and there are files present
                if self.is_file_based:
                    annotations = annotations[["Filename", "eTag", "entityId"]]
```
Can we just delete this part? Since if there is no existing manifest and `use_annotations = False`, we are no longer pulling annotations

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -462,7 +462,6 @@ class ManifestGenerator(object):
                 )
             else:
                 self.additional_metadata["Component"] = [self.root]
-
         return
 
     def _get_additional_metadata(self, required_metadata_fields: dict) -> dict:
@@ -1527,7 +1526,7 @@ class ManifestGenerator(object):
         manifest_record = store.updateDatasetManifestFiles(self.sg, datasetId = dataset_id, store = False)
 
         # get URL of an empty manifest file created based on schema component
-        empty_manifest_url = self.get_empty_manifest(strict=strict, sheet_url=sheet_url)
+        empty_manifest_url = self.get_empty_manifest(strict=strict, sheet_url=True)
 
         # Populate empty template with existing manifest
         if manifest_record:
@@ -1547,35 +1546,45 @@ class ManifestGenerator(object):
             return result
 
         # Generate empty template and optionally fill in with annotations
+        # if there is no existing manifest and use annotations is set to True, 
+        # pull annotations (in reality, annotations should be empty when there is no existing manifest)
         else:
             # Using getDatasetAnnotations() to retrieve file names and subset
             # entities to files and folders (ignoring tables/views)
-
             annotations = pd.DataFrame()
-            if self.is_file_based:
-                annotations = store.getDatasetAnnotations(dataset_id)
+            if self.use_annotations:
+                if self.is_file_based:
+                    annotations = store.getDatasetAnnotations(dataset_id)
+                    # in the old code, we take a subset of columns if no interested in user-defined annotations and there are files present
 
-            # if there are no files with annotations just generate an empty manifest
-            if annotations.empty:
-                manifest_url = self.get_empty_manifest(strict=strict)
-                manifest_df = self.get_dataframe_by_url(manifest_url)
-            else:
+                # if there are no files with annotations just generate an empty manifest
+                if annotations.empty:
+                    #manifest_url = self.get_empty_manifest(strict=strict)
+                    manifest_df = self.get_dataframe_by_url(empty_manifest_url)
+
+            # else:
                 # Subset columns if no interested in user-defined annotations and there are files present
-                if self.is_file_based and not self.use_annotations:
-                    annotations = annotations[["Filename", "eTag", "entityId"]]
-
-                # Update `additional_metadata` and generate manifest
-                manifest_url, manifest_df = self.get_manifest_with_annotations(annotations, sheet_url=sheet_url, strict=strict)
+                # if self.is_file_based:
+                #     annotations = annotations[["Filename", "eTag", "entityId"]]
+        
+            # Update `additional_metadata` and generate manifest
+            manifest_url, manifest_df = self.get_manifest_with_annotations(annotations, sheet_url=sheet_url, strict=strict)
 
             # Update df with existing manifest. Agnostic to output format
             updated_df, out_of_schema_columns = self._update_dataframe_with_existing_df(empty_manifest_url=empty_manifest_url, existing_df=manifest_df)
+
+            # if dataset id is provided, try getting a list of filenames and fill in to manifest
+            if dataset_id:
+                dataset_files, updated_manifest = store.fill_in_entity_id_filename(dataset_id, updated_df)
+            else:
+                updated_manifest = updated_df
 
             # determine the format of manifest that gets return 
             result = self._handle_output_format_logic(output_format = output_format,
                                                       output_path = output_path,
                                                       sheet_url = sheet_url,
                                                       empty_manifest_url=empty_manifest_url,
-                                                      dataframe = updated_df,
+                                                      dataframe = updated_manifest,
                                                       out_of_schema_columns = out_of_schema_columns,
                                                       )
             return result
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -541,6 +541,28 @@ class SynapseStorage(BaseStorage):
 
         return result_dict
 
+    def fill_in_entity_id_filename(self, datasetId: str, manifest: pd.DataFrame):
+        # get current list of files
+        dataset_files = self.getFilesInStorageDataset(datasetId)
+
+        # update manifest with additional filenames, if any
+        # note that if there is an existing manifest and there are files in the dataset
+        # the columns Filename and entityId are assumed to be present in manifest schema
+        # TODO: use idiomatic panda syntax
+        if dataset_files:
+            new_files = self._get_file_entityIds(dataset_files=dataset_files, only_new_files=True, manifest=manifest)
+
+            # update manifest so that it contain new files
+            new_files = pd.DataFrame(new_files)
+            manifest = (
+                pd.concat([manifest, new_files], sort=False)
+                .reset_index()
+                .drop("index", axis=1)
+            )
+
+        manifest = manifest.fillna("") 
+        return dataset_files, manifest
+        
     def updateDatasetManifestFiles(self, sg: SchemaGenerator, datasetId: str, store:bool = True) -> Union[Tuple[str, pd.DataFrame], None]:
         """Fetch the names and entity IDs of all current files in dataset in store, if any; update dataset's manifest with new files, if any.
 
@@ -572,16 +594,10 @@ class SynapseStorage(BaseStorage):
         # note that if there is an existing manifest and there are files in the dataset
         # the columns Filename and entityId are assumed to be present in manifest schema
         # TODO: use idiomatic panda syntax
-        if dataset_files:
-            new_files = self._get_file_entityIds(dataset_files=dataset_files, only_new_files=True, manifest=manifest)
 
-            # update manifest so that it contain new files
-            new_files = pd.DataFrame(new_files)
-            manifest = (
-                pd.concat([manifest, new_files], sort=False)
-                .reset_index()
-                .drop("index", axis=1)
-            )
+        dataset_files, manifest = self.fill_in_entity_id_filename(datasetId, manifest)
+        if dataset_files:
+
 
             # update the manifest file, so that it contains the relevant entity IDs
             if store:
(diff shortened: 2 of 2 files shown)


schematic/manifest/generator.py
```diff
@@ -1547,35 +1546,45 @@ def get_manifest(
...
+                    annotations = store.getDatasetAnnotations(dataset_id)
+                    # in the old code, we take a subset of columns if no interested in user-defined annotations and there are files present
 
-            # if there are no files with annotations just generate an empty manifest
-            if annotations.empty:
-                manifest_url = self.get_empty_manifest(strict=strict)
-                manifest_df = self.get_dataframe_by_url(manifest_url)
-            else:
-                # Subset columns if no interested in user-defined annotations and there are files present
-                if self.is_file_based and not self.use_annotations:
```
linglp, line 1564: not sure if I could just delete this part. See description in the PR

schematic/store/synapse.py
```diff
@@ -615,7 +630,14 @@ def _get_file_entityIds(self, dataset_files: List,  only_new_files: bool = False
             
             # find new files (that are not in the current manifest) if any
             for file_id, file_name in dataset_files:
-                if not file_id in manifest["entityId"].values:
+                if "entityId" in manifest.columns:
```
linglp, line 633: This logic here needs to be clean up. But i think it should be a separate PR to refactor this part

schematic/store/synapse.py
```diff
@@ -541,6 +541,28 @@ def getDataTypeFromManifest(self, manifestId:str):
 
         return result_dict
 
+    def fill_in_entity_id_filename(self, datasetId: str, manifest: pd.DataFrame):
```
linglp, line 544: This function is for filling "entity id" column and "file name" column

schematic/store/synapse.py
```diff
@@ -572,16 +594,10 @@ def updateDatasetManifestFiles(self, sg: SchemaGenerator, datasetId: str, store:
...
-            # update manifest so that it contain new files
-            new_files = pd.DataFrame(new_files)
-            manifest = (
-                pd.concat([manifest, new_files], sort=False)
-                .reset_index()
-                .drop("index", axis=1)
-            )
 
             # update the manifest file, so that it contains the relevant entity IDs
             if store:
```
linglp, line 603: not sure why we would need `if store` logic here, but that should be a separate discussion

schematic/store/synapse.py
linglp, line 633: The original code here assumes that when filling in "Filename" column and "entityId" column, those two columns are already present in the existing manifest. But when `use_annotations` is set to True and there is no existing manifest, we still want File name and "entityId" column to be filled. See comments from line 637-640

schematic/manifest/generator.py
```diff
@@ -1547,35 +1546,42 @@ def get_manifest(
...
-                    annotations = annotations[["Filename", "eTag", "entityId"]]
-
-                # Update `additional_metadata` and generate manifest
-                manifest_url, manifest_df = self.get_manifest_with_annotations(annotations, sheet_url=sheet_url, strict=strict)
+            if self.use_annotations:
```
linglp, line 1555: change the behavior so that we only call `getDatasetAnnotations` when use_annotations is set to True

schematic/store/synapse.py
```diff
@@ -538,6 +538,88 @@ def getDataTypeFromManifest(self, manifestId:str):
...
+        # remove the first row 
+        manifest_new = manifest_new.iloc[1:, :]
+        manifest_new = manifest_new.fillna("")
+
+        # reset index again
+        manifest_new = manifest_new.reset_index(drop=True)
+
+        return manifest_new
+
+    def fill_in_entity_id_filename(self, datasetId: str, manifest: pd.DataFrame) -> Tuple[List, pd.DataFrame]:
```
linglp, line 592: The logic for this function previously exists in  `updateDatasetManifestFiles` function. I wrapped that part there and packaged the logic in this function

schematic/store/synapse.py
```diff
@@ -538,6 +538,88 @@ def getDataTypeFromManifest(self, manifestId:str):
 
         return result_dict
 
+    def _get_files_under_datasets(self, datasetId: str) -> dict:
```
linglp, line 541: change it to `_get_files_metadata_from_dataset` since we are not actually getting the files

schematic/store/synapse.py
```diff
@@ -538,6 +538,88 @@ def getDataTypeFromManifest(self, manifestId:str):
...
+
+        Returns:
+            pd.DataFrame: returns a pandas dataframe 
+        """
+        # get file names and entity ids of a given dataset 
+        dataset_files_dict = self._get_files_under_datasets(datasetId)
+        new_files = pd.DataFrame(dataset_files_dict)
+        # concatenate the two dataframes together 
+        manifest = (
+                pd.concat([manifest, new_files], sort=False)
```
linglp, line 574: Make sure that we already have `Filename` column we should account for it. The sorting should be done according to the sorting function in manifest generator.

schematic/store/synapse.py
```diff
@@ -538,6 +538,88 @@ def getDataTypeFromManifest(self, manifestId:str):
...
+                .reset_index()
+                .drop("index", axis=1)
+        )
+        # fill the rest of component column by component name
```
linglp, line 578: Check if "Component" even exists

linglp: If there is no component, then obviously don't need to fill the "Component" column

linglp: take a look at _add_root_to_component function. It fill "Component" column when `self.is_file_based and 'Filename' in self.additional_metadata:`

schematic/store/synapse.py
```diff
@@ -569,25 +655,16 @@ def updateDatasetManifestFiles(self, sg: SchemaGenerator, datasetId: str, store:
...
+        dataset_files, manifest = self.fill_in_entity_id_filename(datasetId, manifest)
+        if dataset_files:
             # update the manifest file, so that it contains the relevant entity IDs
             if store:
                 manifest.to_csv(manifest_filepath, index=False)
 
                 # store manifest and update associated metadata with manifest on Synapse
                 manifest_id = self.associateMetadataWithFiles(sg, manifest_filepath, datasetId)
 
-        manifest = manifest.fillna("")
```
linglp, line 590: this line was removed because `fill_in_entity_id_filename` already fills na with empty string

GiaJordan:
Verifying that the expected behavior RE manifest generation with and without annotations is observed:

- from the API as specified
- from the command line as specified
- from the command line as .CSVs
- from the command line as .XLSXs

schematic/manifest/generator.py
```diff
@@ -1547,35 +1546,42 @@ def get_manifest(
...
+                if self.is_file_based:
+                    annotations = store.getDatasetAnnotations(dataset_id)
+                    # in the old code, we take a subset of columns if no interested in user-defined annotations and there are files present
+
+                # if there are no files with annotations just generate an empty manifest
+                if annotations.empty:
+                    manifest_df = self.get_dataframe_by_url(empty_manifest_url)
+
+            # else:
```
GiaJordan, line 1564: Yeah I don't think we need this part. Is the `eTag` is just a hash for the annotations?

schematic/store/synapse.py
```diff
@@ -538,6 +538,94 @@ def getDataTypeFromManifest(self, manifestId:str):
...
+            a dictionary that contains filename and entityid under a given datasetId or None if there is nothing under a given dataset id are not available
+        """
+        dataset_files = self.getFilesInStorageDataset(datasetId)
+        if dataset_files:
+            dataset_file_names_id_dict = self._get_file_entityIds(dataset_files, only_new_files=only_new_files, manifest=manifest)
+            return dataset_file_names_id_dict
+        else:
+            return None
+
+    def add_entity_id_and_filename(self,  datasetId: str, manifest: pd.DataFrame) -> pd.DataFrame:
```
GiaJordan, line 559: Not that we have to implement now, but do you think in the future it would be useful to add one method as a wrapper that calls either `add_entity_id_and_filename` or `fill_in_entity_id_filename` depening on the content of the manifest so that users (other devs) don't have to worry about if the columns exist in the manifest before hand?

GiaJordan (changes requested):

schematic/store/synapse.py
```diff
@@ -569,25 +657,16 @@ def updateDatasetManifestFiles(self, sg: SchemaGenerator, datasetId: str, store:
...
-
-            # update manifest so that it contain new files
-            new_files = pd.DataFrame(new_files)
-            manifest = (
-                pd.concat([manifest, new_files], sort=False)
-                .reset_index()
-                .drop("index", axis=1)
-            )
 
+        dataset_files, manifest = self.fill_in_entity_id_filename(datasetId, manifest)
```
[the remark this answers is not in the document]
linglp: Very good catch. Thank you so much. I think we don't need those parts because `fill_in_entity_id_filename` function already calls getFilesInStorageDataset

schematic/store/synapse.py
```diff
@@ -538,6 +538,94 @@ def getDataTypeFromManifest(self, manifestId:str):
...
+            # update Filename column
+            # add entityId column to the end
+            manifest_dict.update(dataset_files_dict)
+            
+            # if the component column exists in existing manifest, fill up that column 
+            if "Component" in manifest_dict.keys():
+                manifest_dict["Component"] = manifest_dict["Component"] * max(1, len(manifest_dict["Filename"]))
+            
+            # turn dictionary back to a dataframe
+            manifest_dict_index = pd.DataFrame.from_dict(manifest_dict, orient='index')
```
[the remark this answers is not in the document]
linglp: Yeah thanks for pointing this out. Let me update the name of variable.

schematic/manifest/generator.py
```diff
@@ -1547,35 +1546,42 @@ def get_manifest(
...
+            #     Subset columns if no interested in user-defined annotations and there are files present
+            #     if self.is_file_based:
+            #         annotations = annotations[["Filename", "eTag", "entityId"]]
+            # Update `additional_metadata` and generate manifest
+            manifest_url, manifest_df = self.get_manifest_with_annotations(annotations, sheet_url=sheet_url, strict=strict)
 
             # Update df with existing manifest. Agnostic to output format
             updated_df, out_of_schema_columns = self._update_dataframe_with_existing_df(empty_manifest_url=empty_manifest_url, existing_df=manifest_df)
```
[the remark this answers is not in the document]
linglp: I agree`updated_manifest` is not ideal but this is the current behavior: https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/manifest/generator.py#L1571

schematic/store/synapse.py
```diff
@@ -538,6 +538,94 @@ def getDataTypeFromManifest(self, manifestId:str):
...
+
+            # drop index
+            manifest_dict_updated = manifest_dict_updated.reset_index(drop=True)
+
+            return manifest_dict_updated
+        else:
+            return manifest
+
+    def fill_in_entity_id_filename(self, datasetId: str, manifest: pd.DataFrame) -> Tuple[List, pd.DataFrame]:
+        """fill in Filename column and EntityId column. This function assumes that entityId column and Filename column already exist
```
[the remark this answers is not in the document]
linglp: Thanks Gianna! I double checked and I think if filename column and entityId column are not present, they will be created

schematic/store/synapse.py
linglp: Yes definitely. I am trying not to make this PR changes too many things, so I don't want to refactor this part. But I completely agree that in the future, how we fill up `filename` and `entityId` column should be the same across code base.

linglp: See https://sagebionetworks.jira.com/browse/FDS-958

schematic/manifest/generator.py
linglp: Yeah I think so.. I agree we could remove that part. I removed it in the PR here: https://github.com/Sage-Bionetworks/schematic/pull/1279

schematic/store/synapse.py
GiaJordan: @linglp keeping the changed in this PR more focused makes sense; thanks for creating the issue!

schematic/manifest/generator.py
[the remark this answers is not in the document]
GiaJordan: Ahh I see, perhaps we could just change it in the future if the need arises

GiaJordan (approved):

PR #1279, opened 2023-08-09
Refactored logic when pulling annotations

linglp:


schematic/manifest/generator.py
```diff
@@ -1526,7 +1523,7 @@ def get_manifest(
         manifest_record = store.updateDatasetManifestFiles(self.sg, datasetId = dataset_id, store = False)
 
         # get URL of an empty manifest file created based on schema component
-        empty_manifest_url = self.get_empty_manifest(strict=strict, sheet_url=sheet_url)
```
linglp, line 1529: I think the codes following this line assumes that here we are generating a manifest url, so I hard-coded sheet_url=True

schematic/manifest/generator.py
```diff
@@ -1390,19 +1387,19 @@ def get_manifest_with_annotations(
...
-        manifest_url = self.get_empty_manifest(sheet_url=sheet_url, strict=strict)
+        # With annotations added, regenerate empty manifest
+        manifest_url = self.get_empty_manifest(sheet_url=True, strict=strict)
         manifest_df = self.get_dataframe_by_url(manifest_url=manifest_url)
 
         # Annotations clashing with manifest attributes are skipped
         # during empty manifest generation. For more info, search
         # for `additional_metadata` in `self.get_empty_manifest`.
         # Hence, the shared columns need to be updated separately.
-        if self.is_file_based and self.use_annotations:
```
linglp, line 1400: since get_manifest_with_annotations function now only gets called when "is_file_based" = true and use_annotations=true, I remove the if statement here

schematic/manifest/generator.py
```diff
@@ -1364,19 +1364,16 @@ def map_annotation_names_to_display_names(
         return annotations.rename(columns=label_map)
 
     def get_manifest_with_annotations(
-        self, annotations: pd.DataFrame, sheet_url:bool=None, strict: Optional[bool]=None,
```
linglp, line 1367: I remove sheet_url parameter from the function here because it is no longer needed.

schematic/manifest/generator.py
```diff
@@ -1390,19 +1387,19 @@ def get_manifest_with_annotations(
         self.additional_metadata = annotations_dict
 
         # Generate empty manifest using `additional_metadata`
-        manifest_url = self.get_empty_manifest(sheet_url=sheet_url, strict=strict)
+        # With annotations added, regenerate empty manifest
+        manifest_url = self.get_empty_manifest(sheet_url=True, strict=strict)
```
linglp, line 1391: Later, `set_dataframe_by_url` function uses manifest_url as an input parameter, so I think here manifest_url has to be an actual google sheet url. I then hard-coded sheet_url=True

schematic/manifest/generator.py
```diff
@@ -1555,33 +1552,25 @@ def get_manifest(
             if self.use_annotations:
                 if self.is_file_based:
                     annotations = store.getDatasetAnnotations(dataset_id)
-                    # in the old code, we take a subset of columns if no interested in user-defined annotations and there are files present
-
-                # if there are no files with annotations just generate an empty manifest
-                if annotations.empty:
```
linglp, line 1561: I changed it so that `get_dataframe_by_url ` do not get called when use_annotations is set to True and is_file_base is also True

mialy-defelice (approved):

PR #1280, opened 2023-08-09
[bug fix] Skipped submission when running API tests remotely

linglp:
Related to Jira issue: https://sagebionetworks.jira.com/browse/FDS-887
## Changelog
skip submission test when running test_api.py remotely

Note: The github action could still time-out after 6 hours: https://github.com/Sage-Bionetworks/schematic/actions/runs/5812972451. But at least all the tests have passed there.

### DIFF
--- a/.github/workflows/api_test.yml
+++ b/.github/workflows/api_test.yml
@@ -76,7 +76,7 @@ jobs:
         if: ${{ inputs.perform_benchmarking }}
         run: >
           source .venv/bin/activate;
-          pytest -m "schematic_api"
+          pytest -m "schematic_api and not submission"
 
       - name: Run API tests + Exclude Benchmarks
         env:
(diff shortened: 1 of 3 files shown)


GiaJordan (changes requested):
Looks good, could you also add the `submission` mark to the `pyproject.toml` markers as well?

Also are these tests passing for you? I'm getting `500` responses for `test_submit_manifest_table_and_file_replace`, `test_submit_manifest_file_only_replace`, and `test_submit_manifest_json_str_replace`. But it could just be that the tests aren't getting my credentials correctly or I dont have access to something on synapse. Either way just wanted to check.

linglp:
@GiaJordan I just double checked and made sure that all the tests related to submission work. I wonder why they didn't work for you... could you show me the error?

GiaJordan:
@linglp okay so after double checking it seems like it's not getting any credentials so I'm not logged in. Where are the tests supposed to be getting the creds from?

linglp:
> @linglp okay so after double checking it seems like it's not getting any credentials so I'm not logged in. Where are the tests supposed to be getting the creds from?

Could you try again? i recently updated the sharing settings of some of the test dataset folders. You should be able to run them now.

GiaJordan (approved):
The submission tests pass for me now and can be controlled with the mark!

PR #1281, opened 2023-08-10
[bug fix] Added includesType parameter when using walk function 

linglp:
Related to Jira: https://sagebionetworks.jira.com/browse/FDS-883

## Changelog
* added IncludeTypes parameter when calling `getFilesInStorageDataset`
* added unit test for `getFilesInStorageDataset` function 
* re-order import statement 

Note: when writing unit test, I noticed that we are not catching cases where `fileNames` is not None in `getFilesInStorageDataset` function. See jira issue FDS 889 for details

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -35,7 +35,7 @@ from synapseclient.entity import File
 from synapseclient.table import CsvFileTable, build_table, Schema
 from synapseclient.annotations import from_synapse_annotations
 from synapseclient.core.exceptions import SynapseHTTPError, SynapseAuthenticationError, SynapseUnmetAccessRestrictions
-from synapseutils import walk
+import synapseutils
 from synapseutils.copy_functions import changeFileMetaData
 
 import uuid
@@ -413,7 +413,7 @@ class SynapseStorage(BaseStorage):
         """
 
         # select all files within a given storage dataset folder (top level folder in a Synapse storage project or folder marked with contentType = 'dataset')
-        walked_path = walk(self.syn, datasetId, includeTypes=["folder", "file"])
+        walked_path = synapseutils.walk(self.syn, datasetId, includeTypes=["folder", "file"])
 
         file_list = []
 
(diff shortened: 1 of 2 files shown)


schematic/store/synapse.py
```diff
@@ -413,7 +413,7 @@ def getFilesInStorageDataset(
         """
 
         # select all files within a given storage dataset folder (top level folder in a Synapse storage project or folder marked with contentType = 'dataset')
-        walked_path = walk(self.syn, datasetId)
+        walked_path = walk(self.syn, datasetId, includeTypes=["folder", "file"])
```
thomasyu888, line 416: Nit: I like to be explicit and do synapseutils.walk because there is an os.walk.

thomasyu888 (approved):
Going to pre-approve, but this looks good.

Might want to confirm with everyone that file and folder is all you want.

mialy-defelice (approved):
Checked current output on develop and on this branch and see it working as expected. Great catch @linglp !

schematic/store/synapse.py
linglp: Sounds good. Thanks for the suggestion Tom.

tests/test_store.py
```diff
@@ -232,6 +236,29 @@ def test_getDatasetProject(self, dataset_id, synapse_store):
...
+            [("test_folder", "syn124")],
+            [("test_file", "syn126")],
+        ),
+        (
+            (os.path.join("parent_folder", "test_folder"), "syn124"),
+            [],
+            [("test_file_2", "syn125")],
+        ),
+        ]
+        expected_return_full_path = [('syn126', 'parent_folder/test_file'), ('syn125', 'parent_folder/test_folder/test_file_2')]
```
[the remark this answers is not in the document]
linglp: Thanks Tom. I updated the code there!

linglp:
Per discussion in code review, we should avoid calling `walk` function since we already have file view.

PR #1282, opened 2023-08-16
Add unit test for `add_root_to_component` function 

linglp:
Related to https://sagebionetworks.jira.com/browse/FDS-910

## Changelog
* added unit test for `add_root_to_component` function

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -439,7 +439,7 @@ class ManifestGenerator(object):
 
         return required_metadata_fields
 
-    def _add_root_to_component(self, required_metadata_fields:dict):
+    def _add_root_to_component(self, required_metadata_fields: Dict[str, List]):
         """If 'Component' is in the column set, add root node as a
         metadata component entry in the first row of that column.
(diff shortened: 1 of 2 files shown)


schematic/manifest/generator.py
```diff
@@ -439,7 +439,7 @@ def _gather_dependency_requirements(self, json_schema, required_metadata_fields)
 
         return required_metadata_fields
 
-    def _add_root_to_component(self, required_metadata_fields):
+    def _add_root_to_component(self, required_metadata_fields:dict):
```
andrewelamb, line 442: One way to be more specific than `dict` is to type the keys and/or values of the dict. For example if you know the keys are always strings you can do `dict[str, Any]`, or if you knows the values are always integers you can do `dict[str, int]`

andrewelamb (approved):

schematic/manifest/generator.py
linglp: @andrewelamb Thanks for the tip. I could add `dict[str, Any]`

PR #1283, opened 2023-08-24
23.8.2

andrewelamb:


linglp:
@GiaJordan @mialy-defelice @andrewelamb 

Sorry! I should have reviewed this release PR sooner. When I ran `test_api.py` locally, I noticed that there are two categories of tests that are failing: 
* Tests that are related to generating manifests as an excel spreadsheet
* Tests that are related to downloading manifests

For generating manifests an an excel spreadsheet, please see more details and the fix in this PR: https://github.com/Sage-Bionetworks/schematic/pull/1284. Apparently when doing testing for the PRs related to fixing the use_annotations parameter issues, I didn't test thoroughly enough. Luckily the tests caught the error. I don't think we should release until #1284 gets merged. I will update the description of this release PR to "depends on" #1284 

For the manifest download endpoint, I think the tests are outdated. The issue is we switched to download manifest to a temporary folder but the tests are still checking if we are downloading manifests to the old location. I am a bit surprised that we didn't catch this issue earlier since we have been using Github workflows to run `test_api.py` the tests seemed to be passing there. Anyway please refer to this Jira issue FDS 793 to update the test related to manifest/download. (Note: I don't want to link the Jira issue here to avoid associating this PR with the Jira key for FDS 793)

GiaJordan (approved):
Approving based on @andrewelamb's documentation of test results and per his conversation with @milen-sage about moving ahead with the release due to the low severity of the found bugs

PR #1284, opened 2023-08-29
[bug fix] Fix issues when there is no dataset ids available

linglp:


GiaJordan (approved):
Error was observed on `develop`, indicated tests pass on this branch and the error was not observed.

Good catch!

PR #1285, opened 2023-08-31
Develop fds 978 entity based annotations

GiaJordan:


andrewelamb:
@GiaJordan I'm not seeing any tests for the new functionality.

GiaJordan:
@andrewelamb No, this PR is restoring logic was [already present](https://github.com/Sage-Bionetworks/schematic/blame/757abd1a0a7b4e79b1f70d998cf676c3e137e7cb/schematic/manifest/generator.py#L1559) in the codebase and is not novel. Adding an additional test did not seem necessary

andrewelamb:
@GiaJordan If case this was already covered by existing tests, wouldn't those have been failing up till now?

GiaJordan:
@andrewelamb it seems covering this case [increases runtime](https://github.com/Sage-Bionetworks/schematic/blob/da1d59958b5f6f73d06c2423bf4137ece64d99a9/tests/test_manifest.py#L157), so that's why nothing was failing

andrewelamb (approved):

GiaJordan:
@andrewelamb , @mialy-defelice After our discussion I added a test for this case. Parameterizing the other fixtures would've likely caused issues for the existing tests that used them, so to simplify things the one for this case is separate. It will run twice, once for a dataset where there is a manifest to pull from and once where the dataset is empty.

I've also gone through the CLI and API tests that @mialy-defelice performs when changing this part of the manifest generator. All were fine except for using the `-a` flag with the HTAN manifests (tests in rows 4,5, and 6). This was a known issue though as they've added an annotation key to their files that we don't necessarily need to support at this time. Those rows also experienced the same error on `develop` so it's not a branch specific problem.

Please let me know if either of you have questions!

andrewelamb (approved):

PR #1287, opened 2023-09-06
BugFix: `is_class_in_schema` no longer errors out when a class is not in schema

GiaJordan:
This PR fixes the behavior of `is_class_in_schema` so that it no longer raises an error when a class is not in a schema.

### DIFF
--- a/schematic/schemas/explorer.py
+++ b/schematic/schemas/explorer.py
@@ -184,17 +184,7 @@ class SchemaExplorer:
         return list(nodes)
 
     def is_class_in_schema(self, class_label):
-        
-        # accessing the node in the dictionary will error out if the class is not in the schema
-        try:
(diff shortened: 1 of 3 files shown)


schematic/schemas/explorer.py
```diff
@@ -184,7 +184,17 @@ def get_adjacent_nodes_by_relationship(
         return list(nodes)
 
     def is_class_in_schema(self, class_label):
-        if self.schema_nx.nodes[class_label]:
+        
+        # accessing the node in the dictionary will error out if the class is not in the schema
```
mialy-defelice: so sorry I forgot you can do `if class_label in self.schema_nx.nodes()` and that will give True if the class label is in the schema and false if not, the you can raise the helpful error.

mialy-defelice (approved):

PR #1288, opened 2023-09-08
Move API `access_token` parameter to request header

GiaJordan:
This PR pulls from some of the work done previously by @linglp in #1218 and [#1599](https://github.com/Sage-Bionetworks/sage-monorepo/pull/1599) and comments from @thomasyu888, but with modifications. It is applied to `schematic` to enhance security of the API and apps deployed on AWS that use schematic as the backend.

Synapse Access Tokens are now implemented as Bearer Authentication tokens in the request headers instead of as query parameters sent in the request body. Providing this token is required for all endpoints where it is an option, with the exception of `/manifest/generate`, where it is acceptable to not provide a token if not required for accessing resources on synapse, to preserve the existing functionality.

Tests that previously had `access_token` as a parameter were marked in #1289 for test comparison purposes. They've also been updated to pass in the token in the request header as part of this PR. All of these tests are passing bar 7, which are also failing on #1289, an analogue of `develop` with only the markers changed. I believe some API tests had been identified as having problems before @linglp's leave, these are likely among them.

Tests of the `/manifest/generate` endpoint that do not require an access token to be provided are also marked and are all passing.

Access tokens are no longer visible in the logs, confirmed on the dev instance
image

### To Test

- Try the endpoints that use the access token from SwaggerUI
- Try interacting with the API now that `access_token` is no longer a query parameter and instead should be included in the header
- (note) you should be able to just pass in the auth token without the `Bearer: ` prefix when working through Swagger

Depends on #1289

### DIFF
--- a/schematic_api/api/openapi/api.yaml
+++ b/schematic_api/api/openapi/api.yaml
@@ -25,6 +25,7 @@ paths:
       description: Endpoint to create dynamically create metadata manifest files
       security:
         - bearerAuth: []
+        - {}
       parameters:
         - in: query
           name: schema_url
--- a/schematic_api/api/openapi/api.yaml
+++ b/schematic_api/api/openapi/api.yaml
@@ -115,15 +115,9 @@ paths:
     get:
       summary: Endpoint to download an existing manifest by using a manifest id
       description: Endpoint to download an existing manifest by using a manifest id
+      security:
+        - bearerAuth: []
       parameters:
-        - in: query
-          name: access_token
-          schema:
-            type: string
-            nullable: false
-          description: Token
-          example: Token
-          required: true
         - in: query
           name: manifest_id
           schema:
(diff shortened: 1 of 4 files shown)


GiaJordan:
@lakikowolfe @afwillia @andrewelamb 
You all can test the changes to authorization with this branch, the `0.1.60-beta` schematic image on the GHCR, or on the dev instance of the api. Please let me know how it goes.

cc @MiekoHash

afwillia:
Right on, thanks @GiaJordan. I was able to get `manifest/generate` working by removing the `access_token` argument and adding the `Authoriation` header like so `httr::add_headers(Authorization = sprintf("Bearer %s", access_token))` (@lakikowolfe)

lakikowolfe:
Working through updating all my functions now but so far Auth in the header is working as expected. Thanks @GiaJordan for the heads up and @afwillia for the tip!

GiaJordan:
Perf, thanks you guys @lakikowolfe @afwillia

andrewelamb (approved):

GiaJordan:
@AmyHeiser we decided that we're going to not include this in the upcoming release, and include it in the following one right?

AmyHeiser:
@GiaJordan correct, we have it scheduled for 23.9.3

PR #1290, opened 2023-09-13
Bug Fix: Adding annotations to files with -mrt file_only parameter

mialy-defelice:
Relates to the following issues reported from AMPAD and VEOIBD:
https://sagebionetworks.jira.com/browse/FDS-1028
https://sagebionetworks.jira.com/browse/FDS-1032

Annotations were not being added when using `-mrt file_only` parameter.

I did some testing and discovered the fix developed in FDS-654 worked but it relied on using the `file_and_table` parameter.
The difference in these parameters, is that the function previously called `add_entities` was under a flag (`with_entities`) that was set to `false` for `file_only`, `true` for` file_and_entites`. There was not flag controlling adding entities when  `file_and_table` is selected.

I removed the flag for `file_only` and `file_and_entities`. The function to add annotations to entities actually already controls for MRT selection and was named in a misleading way. It only adds entities when `file_and_entities` is selected, otherwise if file_only is selected, it gets the file entitiyIds and adds annotations to those files. I renamed the function to `add_annotations_to_entities_files`.

PR #1289, opened 2023-09-08
Develop markers for api tests according to credentials required FDS-1026

GiaJordan:
PR adds markers for API tests the send synapse credentials in the request, and those that send a `None` value for the credentials in the request. Aside from the new markers, the code is the same as what is on `develop`. 
7 of these tests are failing. These were probably among the ones identified before by @linglp 
```
FAILED tests/test_api.py::TestSynapseStorage::test_get_dataset_files[None-True] - AssertionError: assert ['syn23643255', 'schematic - main/DataTypeX/Sample_A.txt'] in [['syn23643259', 'schematic - ...
FAILED tests/test_api.py::TestSynapseStorage::test_get_dataset_files[None-False] - AssertionError: assert (['syn25705259', 'Boolean Test'] and ['syn23667202', 'DataTypeX_table'] in [['syn23643259', ...
FAILED tests/test_api.py::TestSynapseStorage::test_get_dataset_files[Sample_A.txt-True] - AssertionError: assert (['syn23643255', 'schematic - main/DataTypeX/Sample_A.txt'] and ['syn24226530', 'schematic -...
FAILED tests/test_api.py::TestManifestOperation::test_download_access_restricted_manifest - assert 200 == 500
FAILED tests/test_api.py::TestManifestOperation::test_dataset_manifest_download[Test-None] - assert 500 == 200
FAILED tests/test_api.py::TestManifestOperation::test_dataset_manifest_download[Test-True] - assert 500 == 200
FAILED tests/test_api.py::TestManifestOperation::test_dataset_manifest_download[Test-False] - assert 500 == 200
```

PR #1292, opened 2023-09-20
v23.9.2

andrewelamb:
### Features:
- #1289 

### Bugs fixed:
- #1285
- #1287
- #1290

### DIFF
--- a/.github/workflows/api_test.yml
+++ b/.github/workflows/api_test.yml
@@ -76,7 +76,7 @@ jobs:
         if: ${{ inputs.perform_benchmarking }}
         run: >
           source .venv/bin/activate;
-          pytest -m "schematic_api"
+          pytest -m "schematic_api and not submission"
 
       - name: Run API tests + Exclude Benchmarks
         env:
@@ -85,4 +85,4 @@ jobs:
         if: ${{ false == inputs.perform_benchmarking }}
         run: >
           source .venv/bin/activate;
-          pytest -m "schematic_api and not rule_benchmark"
+          pytest -m "schematic_api and not submission and not rule_benchmark"
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -25,6 +25,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
           
+      - name: Set env variable for version tag
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+
       - name: Log in to the Container registry
         uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
         with:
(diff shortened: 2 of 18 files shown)


andrewelamb:
@AmyHeiser It looks like everything descrive [here](https://sagebionetworks.jira.com/projects/FDS/versions/34283/tab/release-report-all-issues) has been merged into develop.

@GiaJordan @mialy-defelice Is there anything else that we would want to add to this release? If not it seems like it's ready to be tested.

GiaJordan (approved):
## Testing Report

### Pytest
The tests skipped on GH actions were run, excluding the api benchmarking tests. 3 were failing:
```
FAILED tests/test_manifest.py::TestManifestGenerator::test_get_manifest_first_time[data_frame-skip_annotations-BulkRNAseqAssay]
FAILED tests/test_api.py::TestSynapseStorage::test_get_dataset_files[None-False]
FAILED tests/test_api.py::TestManifestOperation::test_download_access_restricted_manifest
```
I think we can proceed with the release in spite of these failures.

The first test should have been updated as part of a previous PR where a manifest column was removed, but the failure wasn't raised because the test wasn't run as part of our automated testing. @Mialy DeFelice has opened an [issue](https://sagebionetworks.jira.com/browse/FDS-1047) to remedy.

For the 2nd test I'm not sure if what is failing should be the intended behavior; it seems like the failure is because tables are not being returned in the list of files. Inspecting it shows that nothing besides files are returned in the response which is what I would expect the desired behavior to be. [Issue](https://sagebionetworks.jira.com/browse/FDS-1143).

For the 3rd, the entity on synapse no longer has access restrictions in place that the test expects so it is failing. [Issue](https://sagebionetworks.jira.com/browse/FDS-1144).

These are minor failures that aren't related to the changes in this release so proceeding seems safe.

### CLI
Manifest submission from the CLI worked as expected. `-mrt table_file_and_entities` was not tested due to previously identified issues. Manifest submission worked fine.

### API
All API endpoints are working as expected. `/model/submit` was tested with the different values for `-mrt` excluding `table_file_and_entities` because of the aforementioned issue. One thing to note is that if you're uploading a manifest with -mrt set to file_only but the manifest has the entityId of the table in the entityId column then the table will still be updated. We should probably document that somewhere @AmyHeiser or change the behavior if that's not intended.

**Overall we're good to proceed with the release of schematic `v23.9.2`**

MiekoHash:
@GiaJordan Thank you for the nice summary. Is there a reason why https://github.com/Sage-Bionetworks/schematic/pull/1290 is not listed under Bugs fixed? Is it bundled with something else?

andrewelamb:
@MiekoHash My, fault there was a typo in the notes, and I added the wrong issue.

MiekoHash:
Ah, great, thanks, @andrewelamb . @AmyHeiser This release is ready for your review. I tagged you on [this small fix](https://sagebionetworks.jira.com/browse/FDS-1026?focusedCommentId=184347). Thanks everyone.

andrewelamb:
@AmyHeiser @MiekoHash @milen-sage @mialy-defelice 
This was renamed to v23.9.3 due to the workflow issues.

PR #1293, opened 2023-09-20
Tests for DataModelRelationships class

andrewelamb:
fixes [fds-1058](https://sagebionetworks.jira.com/browse/FDS-1058)
fixes [fds-1075](https://sagebionetworks.jira.com/browse/FDS-1075)
fixes [fds-1076](https://sagebionetworks.jira.com/browse/FDS-1076)
fixes [fds-1077](https://sagebionetworks.jira.com/browse/FDS-1077)
fixes [fds-1078](https://sagebionetworks.jira.com/browse/FDS-1078)

### DIFF
--- a/tests/test_schemas.py
+++ b/tests/test_schemas.py
@@ -46,8 +46,8 @@ def DME(helpers, data_model_name='example.model.csv'):
     DME = DataModelGraphExplorer(graph_data_model)
     yield DME
 
-@pytest.fixture(name="dmr_object")
-def dmr():
+@pytest.fixture(name="dmr")
+def fixture_dmr():
     """Yields a data model relationships object for testing"""
     yield DataModelRelationships()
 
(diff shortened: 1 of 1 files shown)


andrewelamb:
@mialy-defelice Some comments:

```
for k, v in self.relationships_dictionary.items():
            try:
                if v['required_header']:
                    required_headers.append(v['csv_header'])
            except KeyError:
                print(f"Did not provide a 'required_header' key, value pair for the nested dictionary {k} : {key}")
```
I'm not entire sure what the intention of these code blocks are.  If I'm not mistaken, you will never get a `KeyError` because `v['required_header']` is in an `if` statement. If `'required_header'` isn't in `v` if it simply resolve as `False` and nothing will happen. 

In the f-string `f"Did not provide a 'required_header' key, value pair for the nested dictionary {k} : {key}"` There doesn't appear to be  a variable named `key`, should that be `v`?

I would change these blocks to:

```
for k, v in self.relationships_dictionary.items():
            if v['required_header']:
                required_headers.append(v['csv_header'])
            else
                print(f"Did not provide a 'required_header' key, value pair for the nested dictionary {k} : {v}")
```

andrewelamb:
@mialy-defelice Since this class and its methods don't take any variables I wasn't sure what to test. I wrote some to basically make sure the methods run and the return non-empty outputs. As the comment above describes I think the `KeyValue` errors are impossible to trigger, I didn't write anything that tests them. Let me know if there is anything else you would like me to test for!

andrewelamb:
@mialy-defelice I would also suggest to make this mypy friendly:

```
from typing import Dict # this can be removed
from typing import Any
def define_data_model_relationships(self) -> dict[str, Any]:
def define_required_csv_headers(self) ->  list[str]:
def define_edge_relationships(self) -> dict[str, Optional[str]]:
def define_value_relationships(self)  -> dict[str, Optional[str]]:
```

mialy-defelice:
> @mialy-defelice I would also suggest to make this mypy friendly:
> 
> ```
> from typing import Dict # this can be removed
> from typing import Any
> def define_data_model_relationships(self) -> dict[str, Any]:
> def define_required_csv_headers(self) ->  list[str]:
> def define_edge_relationships(self) -> dict[str, Optional[str]]:
> def define_value_relationships(self)  -> dict[str, Optional[str]]:
> ```

Assuming this was a comment for the relationships class, I just added reviewers to that PR, so please feel free to leave tons of comments there!

mialy-defelice:
> @mialy-defelice Since this class and its methods don't take any variables I wasn't sure what to test. I wrote some to basically make sure the methods run and the return non-empty outputs. As the comment above describes I think the `KeyValue` errors are impossible to trigger, I didn't write anything that tests them. Let me know if there is anything else you would like me to test for!

I think the only thing I would add is that for the helper classes like test_define_required_csv_headers that we test the output matches the exact current output. This would help in the instance that someone has accidentally changed something, or goes in and intentionally changes something... which shouldn't be happening. If we intend to add things to the relationships class, we can update the tests.

mialy-defelice:
But other than that it looks good.

andrewelamb:
@mialy-defelice Can I go ahead and merge this?

mialy-defelice:
@andrewelamb so sorry, I didn't see that you added to the tests last week. yep go ahead and merge!

PR #1302, opened 2023-09-22
Raise a helpful error when a datatype cannot be found in a schema 

GiaJordan:
For the linked issue, the error arose because the data type was not spelled according to the conventions of the class label in the schema. The error raised by networkx: return self._nodes[n] \ KeyError: 'AssayRNASeqMetadataTemplate' was not very good at communicating what was going wrong. I've wrapped another error around this message for the case of generating a manifest that should be more explanatory and offers suggestions for remedying.

Re-merge of the reverted PR #1286 (was reverted not due to issues but to allow for re-releasing a failed release without any other changes)

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -54,7 +54,11 @@ class ManifestGenerator(object):
         self.creds = services_creds["creds"]
 
         # schema root
-        self.root = root
+        if root:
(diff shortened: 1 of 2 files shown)


mialy-defelice (approved):
Approving since its just adding back a reverted change.

PR #1303, opened 2023-09-22
BugFix: Add functions to return a display name or requirement if not available.

mialy-defelice:
Resolves [FDS-1030](https://sagebionetworks.jira.com/browse/FDS-1030)

When using attributes from Biothings, they do not have fields for 'required' or 'displayName'. Added functions so this would not lead to a Key Error, so instead will return the ID with context stripped for the Display Name, and return the default of False for required.

[FDS-1030]: https://sagebionetworks.jira.com/browse/FDS-1030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schematic/schemas/generator.py
+++ b/schematic/schemas/generator.py
@@ -340,6 +340,19 @@ class SchemaGenerator(object):
 
         return node_required
 
+    def get_node_display_name(self, node_label: str, mm_graph: nx.MultiDiGraph) -> list:
+        """Get display name associated with a given node label, return id if no display name.
+        Args:
+            node_label, str: Node to retrieve display name for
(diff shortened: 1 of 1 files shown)


mialy-defelice:
@GiaJordan when you get a chance can you review this PR?

GiaJordan:
@mialy-defelice yes for sure!

GiaJordan (approved):
Testing with the parameters provided in the JIRA ticket was successful, a manifest was able to be downloaded

PR #1304, opened 2023-09-22
Schema Refactor: Unit Tests for DataModelParser

mialy-defelice:
Addresses the Story:
 [FDS-1056 - Tests for DataModelParser](https://sagebionetworks.jira.com/browse/FDS-1056)
and related tasks:
[FDS-1066 - test_get_base_schema_path](https://sagebionetworks.jira.com/browse/FDS-1066)
[FDS-1067 - test_get_model_type](https://sagebionetworks.jira.com/browse/FDS-1067)
[FDS-1068 - test_parse_model](https://sagebionetworks.jira.com/browse/FDS-1068)


tests/test_schemas.py
```diff
@@ -40,21 +40,55 @@ def generate_graph_data_model(helpers, path_to_data_model):
...
+
+        assert os.path.basename(biothings_path) == "biothings.model.jsonld"
+
+    @pytest.mark.parametrize("data_model", ['example.model.csv', 'example.model.jsonld'], ids=["csv", "jsonld"])
+    def test_get_model_type(self, helpers, data_model):
+        # Instantiate Data model parser.
+        data_model_parser = helpers.get_data_model_parser(data_model_name=data_model)
+
+        # Check the data model type
+        assert (data_model == 'example.model.csv') == (data_model_parser.model_type == 'CSV')
```
mialy-defelice, line 72: Dictionary to look up that the value is correct instead of having two assert statements.

tests/test_schemas.py
```diff
@@ -40,21 +40,55 @@ def generate_graph_data_model(helpers, path_to_data_model):
...
+        # Instantiate Data model parser.
+        data_model_parser = helpers.get_data_model_parser(data_model_name=data_model)
+        # Parse Model
+        model_dict = data_model_parser.parse_model()
+
+        # Get a key in the model
+        attribute_key = list(model_dict.keys())[0]
+
+        # Check that the structure of the model dictionary conforms to expectations.
+        assert True == (type(model_dict) == dict)
```
mialy-defelice, line 88: Get rid of True.

linglp (commented):
Sorry wrong comment. Will take a look again

tests/test_schemas.py
mialy-defelice: Done!

tests/test_schemas.py
mialy-defelice: Done

tests/test_schemas.py
GiaJordan: Yeah I agree, having just one assertion that checks the `DATA_MODEL_DICT` against the model parser model type would be easier to maintain; we wouldn't need to add new assertions each time we added a new supported format.

tests/test_schemas.py
```diff
@@ -40,21 +46,56 @@ def generate_graph_data_model(helpers, path_to_data_model):
...
+
+    @pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
+    def test_parse_model(self, helpers, data_model):
+        '''Test that the correct parser is called and that a dictionary is returned in the expected structure.
+        '''
+        # Instantiate Data model parser.
+        data_model_parser = helpers.get_data_model_parser(data_model_name=data_model)
+        
+        # Parse Model
+        model_dict = data_model_parser.parse_model()
```
GiaJordan, line 89: could the name of this be changed to not be as similar to the `DATA_MODEL_DICT`?

GiaJordan (changes requested):

tests/test_schemas.py
```diff
@@ -40,21 +46,56 @@ def generate_graph_data_model(helpers, path_to_data_model):
...
+        '''
+        # Instantiate Data model parser.
+        data_model_parser = helpers.get_data_model_parser(data_model_name='example.model.csv')
+
+        # Get path to default biothings model.
+        biothings_path = data_model_parser._get_base_schema_path(base_schema=None)
+
+        assert os.path.basename(biothings_path) == "biothings.model.jsonld"
+
+    @pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
```
linglp, line 72: Just like I mentioned in other PR, I think this part could be moved to the class level. Something like:
```
@pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
class TestDataModelParser:
```

So that we don't have to repeat this line for every single function

tests/test_schemas.py
```diff
@@ -40,21 +46,56 @@ def generate_graph_data_model(helpers, path_to_data_model):
...
+    def test_get_model_type(self, helpers, data_model):
+        # Instantiate Data model parser.
+        data_model_parser = helpers.get_data_model_parser(data_model_name=data_model)
+
+        # Check the data model type
+        assert (data_model == 'example.model.csv') == (data_model_parser.model_type == 'CSV')
+        assert (data_model == 'example.model.jsonld') == (data_model_parser.model_type == 'JSONLD')
+
+    @pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
+    def test_parse_model(self, helpers, data_model):
```
linglp, line 82: In general, this test looks good to me. But I do notice that `parse_model` function handles cases where the models provided are neither CSV nor JSON-LD, I think that part could be added to the test too to make sure that the error message gets raised.

linglp (approved):
Overall it looks good. I left some comments but I could pre-approve for now.

PR #1310, opened 2023-10-10
Refactor Schemas: unit tests FDS-1061 TestDataModelNodes

mialy-defelice:


tests/test_schemas.py
```diff
@@ -172,22 +216,274 @@ def test_sub_schema_graph(self):
...
+            #Looking up this way, in case we add empty defaults back to JSONLD it wont fail, but will only be absent in JSONLD not CSV.
+            if not node_dict['required'] == False:
+                assert DATA_MODEL_DICT[data_model] == 'JSONLD'
+
+    @pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
+    def test_generate_node(self, helpers, data_model):
+        # Test adding a dummy node
+        node_dict = {'label': 'test_label'}
+
+        path_to_data_model = helpers.get_data_path(data_model)
```
mialy-defelice, line 471: actually probably don't need path_to_data_model here.

tests/test_schemas.py
```diff
@@ -172,22 +216,274 @@ def test_sub_schema_graph(self):
...
+
+        # Get Graph
+        graph_data_model = generate_graph_data_model(helpers, data_model_name=path_to_data_model)
```
mialy-defelice, line 474: just pass data_model here.

tests/test_schemas.py
```diff
@@ -172,22 +216,274 @@ def test_sub_schema_graph(self):
...
+        # Make sure the nodes returned conform to expectations (values and order)
+        ## The parsing records display names for relationships for CSV and labels for JSONLD, so the expectations are different between the two.
+        if DATA_MODEL_DICT[data_model]=='CSV':
+            expected_nodes = ['Patient', 'Patient ID', 'Sex', 'Year of Birth', 'Diagnosis', 'Component', 'DataType']
+        elif DATA_MODEL_DICT[data_model] == 'JSONLD':
+            expected_nodes = ['Patient', 'PatientID', 'Sex', 'YearofBirth', 'Diagnosis', 'Component', 'DataType']
+
+        assert nodes == expected_nodes
+
+        # Ensure order is tested.
```
linglp, line 245: I am a bit confused here. Here it seems like we are ensuring "Patient" does not become the last item in a list?

tests/test_schemas.py
```diff
@@ -172,22 +216,274 @@ def test_sub_schema_graph(self):
...
+        reordered_nodes = nodes.copy()
+        reordered_nodes.remove('Patient')
+        reordered_nodes.append('Patient')
+        assert reordered_nodes != expected_nodes
+
+    @pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
```
linglp, line 251: Could be re-written like: 
```
@pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values())) class TestDataModelNodes:
```

tests/test_schemas.py
```diff
@@ -172,22 +216,274 @@ def test_sub_schema_graph(self):
...
+    def test_gather_all_nodes(self, helpers, data_model):
```
linglp, line 252: I understand what you are trying to do here. I am thinking since `gather_all_nodes` calls `gather_nodes` function, will it make sense to mock `gather_nodes` and then test if `gather_all_nodes` could return the correct output?

tests/test_schemas.py
```diff
@@ -172,22 +216,274 @@ def test_sub_schema_graph(self):
...
+        attr_rel_dictionary = data_model_parser.parse_model()
+
+        # Instantiate DataModelNodes
+        data_model_nodes = generate_data_model_nodes(helpers, data_model_name=data_model)
+
+        # Get properties in the data model
+        data_model_properties = data_model_nodes.get_data_model_properties(attr_rel_dictionary)
+        
+        # In the current example model, there are no properties, would need to update this section if properties are added.
+        assert data_model_properties == []
```
linglp, line 305: Yeah, this is a good point. I created a ticket here to track: https://sagebionetworks.jira.com/jira/software/c/projects/FDS/issues/FDS-1226

tests/test_schemas.py
```diff
@@ -172,22 +216,274 @@ def test_sub_schema_graph(self):
...
+            assert entry_type == 'class'
+
+        # Check that the added property is properly loaded as a property
+        assert data_model_nodes.get_entry_type('TestProperty') == 'property'
+
+    @pytest.mark.parametrize("data_model", list(DATA_MODEL_DICT.keys()), ids=list(DATA_MODEL_DICT.values()))
+    @pytest.mark.parametrize("rel_func", list(REL_FUNC_DICT.values()), ids=list(REL_FUNC_DICT.keys()))
+    @pytest.mark.parametrize("test_dn", list(TEST_DN_DICT.keys()), ids=list(TEST_DN_DICT.keys()))
+    @pytest.mark.parametrize("test_bool", ['True', 'False', True, False, 'kldjk'], ids=['True_str', 'False_str', 'True_bool', 'False_bool', 'Random_str'])
+    def test_run_rel_functions(self, helpers, data_model, rel_func, test_dn, test_bool):
```
linglp, line 355: Do we have unit tests for `get_attribute_display_name_from_label`, `parse_validation_rules`, `get_label_from_display_name`, `convert_bool_to_str`? If so, I think here we could simply mock responses from these four functions and test `run_rel_functions`.. Currently what you are doing here makes sense. But at the same time, I feel like this test becomes difficult to read because it gathers essentially four tests of four different functions

andrewelamb (approved):

linglp:
@mialy-defelice I left some comments for your reference. I think it general it looks good. Feel free to merge

tests/test_schemas.py
mialy-defelice: we do have tests for those, but the trick here is actually making sure the parameters are being passed properly through this function.. which has at times felt a bit chaotic, so its nice to call the functions this way.

tests/test_schemas.py
mialy-defelice: This is an additional maybe unnecessary test to ensure that the order and not just the contents is being tested in the previous test. this is sort of a future proof thing...

PR #1311, opened 2023-10-11
Add `access_token` to `/model/validate` header

GiaJordan:
Add `access_token` to the `/model/validate` endpoint, as an optional part of the header


GiaJordan:
@afwillia Since the access token needs to be optional for this endpoint, I added an exception that gets raised if users try to validate a manifest that uses a cross manifest validation rule without providing an access token in cased it would be useful to catch and display a message on the DCA side. I don't know if that's the case but just wanted to let you know

andrewelamb (approved):

PR #1312, opened 2023-10-11
Fix an issue where the asset view was not being set correctly for operations involving manifest validation

GiaJordan:
Add `asset_view` to `/model/validate` and `project_scope` to `/model/validate` and `model/submit` so that cross manifest validation operations locate the correct resources.

Depends on #1311 #1314

**If you run any of the changed unit tests please delete the manifests that were uploaded to datasets `syn52656106` or `syn52656104` immediately to avoid causing other branches tests to fail.** (Required until #1314 is merged)


linglp:
Did some tests manually below through API: 
1. Validate an invalid manifest using the /model/validate endpoint with `project_scope` and `asset_view` parameter. Ensure that errors could be returned as expected.
2. Validate a valid manifest using the /model/validate using asset view: syn23643253 and project scope: syn23643250. Everything seems to be working too. 
3. I also created a branch new project with a brand new dataset folder and asset view to test it out. (Asset view: syn52737292 and project scope: syn52737126) and by uploading a valid test manifest, the output that I got is: 
```
{
  "errors": [],
  "warnings": [
    [
      [],
      "Check Match at Least",
      "Value(s) [] from row(s) [] of the attribute Check Match at Least in the source manifest are missing.",
      []
    ]
  ]
}
```

I don't think this is expected because it doesn't show the values and rows that don't match. Also, this warning shouldn't be shown because in my test folder: https://www.synapse.org/#!Synapse:syn52737592, I have the same manifest. 

Another point to note: 
Currently, if there are no datasest ids or manifest ids found within a given project, the code would continue without giving any warning. For debugging, I think it will bee helpful to note that if there are no dataset ids found or no manifest ids found so that users know that cross manifest validation didn't really get triggered. 

Another observation is that if we have three columns that have validation rules that could trigger cross manifest, it seems like we are calling `cross_validation` three times? And each time that we call `cross_validation` function , we are downloading manifests within a given project. If this is correct, maybe this part could be optimized in the future..

GiaJordan:
> 3. I also created a brand new project with a brand new dataset folder and asset view to test it out. (Asset view: syn52737292 and project scope: syn52737126) and by uploading a valid test manifest, the output that I got is:
> 
> ```
> {
>   "errors": [],
>   "warnings": [
>     [
>       [],
>       "Check Match at Least",
>       "Value(s) [] from row(s) [] of the attribute Check Match at Least in the source manifest are missing.",
>       []
>     ]
>   ]
> }
> ```
> 
> I don't think this is expected because it doesn't show the values and rows that don't match. Also, this warning shouldn't be shown because in my test folder: https://www.synapse.org/#!Synapse:syn52737592, I have the same manifest.
> Another point to note: Currently, if there are no datasest ids or manifest ids found within a given project, the code would continue without giving any warning. For debugging, I think it will bee helpful to note that if there are no dataset ids found or no manifest ids found so that users know that cross manifest validation didn't really get triggered.

Discussed in CR, to be addressed in [FDS-1236](https://sagebionetworks.jira.com/browse/FDS-1236)

> Another observation is that if we have three columns that have validation rules that could trigger cross manifest, it seems like we are calling `cross_validation` three times? And each time that we call `cross_validation` function , we are downloading manifests within a given project. If this is correct, maybe this part could be optimized in the future..

Discussed in CR, to be addressed in [FDS-1237](https://sagebionetworks.jira.com/browse/FDS-1237)

GiaJordan:
@linglp 
> 3. I also created a brand new project with a brand new dataset folder and asset view to test it out. (Asset view: syn52737292 and project scope: syn52737126) and by uploading a valid test manifest, the output that I got is:
> 
> ```
> {
>   "errors": [],
>   "warnings": [
>     [
>       [],
>       "Check Match at Least",
>       "Value(s) [] from row(s) [] of the attribute Check Match at Least in the source manifest are missing.",
>       []
>     ]
>   ]
> }
> ```
> 
> I don't think this is expected because it doesn't show the values and rows that don't match. Also, this warning shouldn't be shown because in my test folder: https://www.synapse.org/#!Synapse:syn52737592, I have the same manifest.
> Another point to note: Currently, if there are no datasest ids or manifest ids found within a given project, the code would continue without giving any warning. For debugging, I think it will bee helpful to note that if there are no dataset ids found or no manifest ids found so that users know that cross manifest validation didn't really get triggered.

Discussed in CR, to be addressed in [FDS-1236](https://sagebionetworks.jira.com/browse/FDS-1236)

> Another observation is that if we have three columns that have validation rules that could trigger cross manifest, it seems like we are calling `cross_validation` three times? And each time that we call `cross_validation` function , we are downloading manifests within a given project. If this is correct, maybe this part could be optimized in the future..

Discussed in CR, to be addressed in [FDS-1237](https://sagebionetworks.jira.com/browse/FDS-1237)


[FDS-1236]: https://sagebionetworks.jira.com/browse/FDS-1236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

**cc:** @AmyHeiser

linglp (approved):
Thanks @GiaJordan ! i also confirmed that the submission is working fine.

GiaJordan:
@linglp thank you for reviewing!

PR #1313, opened 2023-10-11
Refactor Schemas: Unit tests for DataModelJSONLD FDS-1064

mialy-defelice:
Adds tests for the DataModelJSONLD Class.

### DIFF
--- a/tests/test_schemas.py
+++ b/tests/test_schemas.py
@@ -487,9 +487,38 @@ class TestDataModelJsonLd:
             assert 'sms' in object_template['sms:required']
         if '@id' in object_template:
             assert 'bts' in object_template['@id']
-        return
-    def test_clean_template(self):
-        return
+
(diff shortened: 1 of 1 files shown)


andrewelamb (approved):

tests/test_schemas.py
```diff
@@ -400,26 +401,242 @@ def test_get_json_validation_schema(self, helpers, data_model, source_node, sche
...
+        graph_data_model = generate_graph_data_model(helpers, data_model_name=data_model)
+
+        # Instantiate DataModelJsonLD
+        data_model_jsonld = DataModelJsonLD(Graph=graph_data_model)
+
+        # Get empty template
+        template = data_model_jsonld.class_template()
+
+        # Fill out template with some 'Sex' information
+        template['@id'] = 'Sex'
```
linglp, line 552: I think this is a good test too. But at the same time, I also wonder what if the 'id' that you enter here is not in the schema, then what would be the expected response. Should we consider adding that in the test too?

linglp (changes requested):

PR #1319, opened 2023-10-23
Add default value for `access_token` to another validation method

GiaJordan:


schematic/models/validate_manifest.py
```diff
@@ -62,7 +62,7 @@ def get_multiple_types_error(
         return ["NA", error_col, error_message, error_val]
 
     def validate_manifest_rules(
-        self, manifest: pd.core.frame.DataFrame, sg: SchemaGenerator, restrict_rules: bool, project_scope: List, access_token: str,
+        self, manifest: pd.core.frame.DataFrame, sg: SchemaGenerator, restrict_rules: bool, project_scope: List, access_token: str = None,
```
andrewelamb, line 65: I don't know if you want to change it now or when we type annotate this, but `access_token: str = None` should be 
 `access_token: Optional[str] = None`

andrewelamb (approved):

schematic/models/validate_manifest.py
GiaJordan: thanks for catching, I'll fix this one now

PR #1320, opened 2023-10-24
Reduce execution time for `Test Schematic` workflow

GiaJordan:
# Overview
This PR includes an array of improvements to the unit and integration tests for schematic, as well as minor changes to the code where relevant to correct errors and improve testing performance.

Previously the runtime for the  `Test Schematic` workflow was regularly 60-90 minutes, at some times even taking up to almost 3 hours to complete [x](https://github.com/Sage-Bionetworks/schematic/actions/runs/4168434979). As a result of the improvements in this PR **the workflow now routinely finishes in 25-35 minutes**, and often under 20 minutes [x](https://github.com/Sage-Bionetworks/schematic/actions/runs/6961430780) [x](https://github.com/Sage-Bionetworks/schematic/actions/runs/6961784718) [x](https://github.com/Sage-Bionetworks/schematic/actions/runs/6962664566) [x](https://github.com/Sage-Bionetworks/schematic/actions/runs/6962664566). 

**Note: this PR includes changes that will affect our github billing.**

Depends on #1331 

# Changelog
## Switch from `tenacity` to `pytest-rerunfailures` for tests
`tenacity` is no longer used to retry a certain test that may fail depending on the state of resources on synapse. Instead, `pytest-rerunfailures` has been implemented, that allows for a failing test to be re-run in another attempt at success before being considered a failure. The previous tenacity implementation was probably too generous in how much time was allowed between retrys in hindsight, but the initial implementation was overly cautious to reduce failures for reasons other than what was being tested. This change reduces the theoretical maximum time for the test from over 30 minutes plus execution time to only the execution time for the test to run twice.

## Tenacity correction in `synapse.py`
It appears that in a previous PR, a method was added to the `SynapseStorage` class but was accidentally added between an existing method and the decorator that was decorating it. That removed the retry functionality from `getDatasetProject` and then applied it to `checkIfinAssetView` which was likely not intended. The decorator allowed for a theoretical maximum wait time 70 seconds. I'm unsure if this wait sequence was being triggered by the method it was accidentally decorating, but it has been resolved now and is applied to the correct method.

## `synapse_store` fixture: `session` scope and moved to `conftest.py`
The setup time for 6-9 different tests clocked up to 900-1000 seconds (15-17 minutes) for each test ([x](https://github.com/Sage-Bionetworks/schematic/actions/runs/6933424771/job/18861390840) [x](https://github.com/Sage-Bionetworks/schematic/actions/runs/6933424771/job/18861390840) [x](https://github.com/Sage-Bionetworks/schematic/actions/runs/6933424771/job/18861390578) [x](https://github.com/Sage-Bionetworks/schematic/actions/runs/6933424771/job/18861390578) [x](https://github.com/Sage-Bionetworks/schematic/actions/runs/6933264279/job/18864822266) [x](https://github.com/Sage-Bionetworks/schematic/actions/runs/6933264279/job/18864822657)), **not including the time to actually execute the tests**. Upon investigation, all of the tests used the `synapse_store` fixture to generate a store object. This fixture was implemented twice separately, in `test_store.py` and `test_utils.py`, and a scope was not specified so the default `function` scope was in effect. Locally, the time to execute the fixture code and yield a `synapse_store` object was 4.5-5 seconds. Because the scope of this fixture was `function`, this fixture was being called 40 separate times over the test session: 34 times in `test_store.py` and 6 times in `test_utils.py`. This would account for 3:20 of execution time locally and was most likely the cause of the extended setup times of upwards of 1000 seconds experienced with these tests in the remote environment.
The fixture has been moved to `conftest.py` and its scope has been updated to `session` so that the code only needs to be executed once per run and can be shared across all the tests that require it. In my testing this did not lead to test failures among the tests that utilize this fixture.

## Switch to use larger runners
The runner used for the `Test Schematic` workflow  was switched from `ubuntu-latest`, which has 2 CPUs, 7GB of RAM, and 14GB of storage to `ubuntu-22.04-4core-16GBRAM-150GBSSD`, which has 4CPUs, 16GB of RAM, and 150GB of storage. This was [previously implemented](https://sagebionetworks.slack.com/archives/C0591MQA80Z/p1697566784426289) by @/tschaffter in the monorepo to solve a storage space issue. Note that time utilizing the larger runners is not free and is billed at [$0.016/minute](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates).

## Implement `pytest-xdist`
This package is used to distribute our tests across multiple CPUs. In conjunction with the larger runner now being used, we can now spread the test execution across 4 CPUs to run them concurrently instead of sequentially. This has a great effect on the execution time because tests that take longer to complete no longer block other tests from being run while they perform their operations.

# Future Improvements
Some tests still consistently take significantly longer than others to run (~900s): `tests/test_manifest.py::TestManifestGenerator::test_get_manifest_no_annos` and `tests/test_store.py::TestSynapseStorage::test_annotation_submission`. `tests/test_validation.py::TestManifestValidation::test_invalid_manifest` occasionally takes a significantly long time as well. We could look into these specifically in the future. It seems like the reason the tests currently take ~20 minutes instead of less than 10 is because of these specific tests.

We should evaluate the tests that were excluded to not add to the run time too much and consider running them as part of the standard checks.

We could also look into some of these changes and see if they are possible to apply to the API test suite and `Test Schematic API` workflow, depending on how soon the migration to the monorepo is occuring.


mialy-defelice:
Tested by running workflow.

mialy-defelice (approved):

PR #1325, opened 2023-11-07
update dockerfile poetry versions FDS-1325

GiaJordan:
Update poetry version in dockerfiles


linglp:
Note: the action to publish docker image failed: https://github.com/Sage-Bionetworks/schematic/actions/runs/6747810246/job/18344783240 when triggering a deployment of v23.11.1... this is probably because the poetry version got updated in:  #1316 . This PR is to make sure that we also update the poetry version in docker file.

linglp (approved):

PR #1332, opened 2023-11-28
fixed typing that API uses

andrewelamb:
- Fixes [fds-1342](https://sagebionetworks.jira.com/browse/FDS-1342)
- Fixes incorrect type hinting (parameters and return types) for methods that the API uses.

### DIFF
--- a/schematic/models/metadata.py
+++ b/schematic/models/metadata.py
@@ -186,8 +186,14 @@ class MetadataModel(object):
 
     # TODO: abstract validation in its own module
     def validateModelManifest(
-        self, manifestPath: str, rootNode: str, restrict_rules: bool = False, jsonSchema: str = None, project_scope: List = None, access_token: str = None,
-    ) -> List[str]:
+        self,
+        manifestPath: str,
+        rootNode: str,
+        restrict_rules: bool = False,
+        jsonSchema: Optional[str] = None,
+        project_scope: Optional[List] = None,
+        access_token: Optional[str] = None,
+    ) -> tuple[list, list]:
         """Check if provided annotations manifest dataframe satisfies all model requirements.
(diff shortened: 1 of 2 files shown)


schematic/models/metadata.py
```diff
@@ -292,13 +298,13 @@ def submit_metadata_manifest(
         dataset_id: str,
         manifest_record_type: str,
         restrict_rules: bool,
-        validate_component: str = None,
+        access_token: str,
```
GiaJordan, line 301: should this one be `Optional[str] = None` because it used to be `str = None`?

schematic/models/metadata.py
```diff
@@ -292,13 +298,13 @@ def submit_metadata_manifest(
...
+        validate_component: str = Optional[None],
```
GiaJordan, line 302: is this backwards or intentional?

GiaJordan (changes requested):

schematic/models/metadata.py
andrewelamb: Oopps, yep this is a mistake I'll fix this.

schematic/models/metadata.py
GiaJordan: Ahh right, I think I had thought this was the validation method. thanks for clarifying

GiaJordan (approved):

schematic/models/metadata.py
mialy-defelice: @andrewelamb @GiaJordan We need to keep the None default, since this change actually broke the CLI submit if you require the access token. The CLI gets the access token another way from the API. I will submit another PR to update typing `Optional[str]=None`

PR #1336, opened 2023-12-06
Resolved an issue where a visualization endpoint intermittently returned the wrong information

GiaJordan:
Was able to repro erroneous behavior on prod, staging, and a local `develop` api, for `visualize/tangled_tree/layers` endpoint only. 
Behavior was caused because over a session, the value for `all_layers` in `TangledTree.get_tangled_tree_layers` would be preserved between requests sent to the endpoint, with the new layer information just being appended to it. The endpoint then takes the first element of the `all_layers` list after it is returned, assuming that it is a list of length 1. What was actually happening is that the layers for each request were all being returned in this list and so when the first one was returned it could differ from what was requested in the most recent request.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -881,6 +881,7 @@ class TestSchemaVisualization:
 
     @pytest.mark.parametrize("figure_type", ["component", "dependency"])
     def test_visualize_tangled_tree_layers(self, client, figure_type, data_model_jsonld):
+        # TODO: Determine a 2nd data model to use for this test, test both models sequentially, add checks for content of response
         params = {
             "schema_url": data_model_jsonld,
             "figure_type": figure_type
--- a/schematic/visualization/tangled_tree.py
+++ b/schematic/visualization/tangled_tree.py
@@ -749,7 +749,6 @@ class TangledTree(object):
         '''
         # Gather the data model's, topological generations, nodes and edges
         topological_gen, nodes, edges, subg = self.get_topological_generations()
-        all_layers = []
 
         if self.figure_type == 'component':
             # Gather all source nodes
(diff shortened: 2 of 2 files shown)


linglp:
@GiaJordan thank you for the PR. Could you deploy this to AWS before merging to `develop`?

GiaJordan:
@linglp it's been deployed to the dev env. behavior is the same as on the local api

tests/test_api.py
```diff
@@ -881,6 +881,7 @@ def test_visualize_attributes(self, client, data_model_jsonld):
 
     @pytest.mark.parametrize("figure_type", ["component", "dependency"])
     def test_visualize_tangled_tree_layers(self, client, figure_type, data_model_jsonld):
+        # TODO: Determine a 2nd data model to use for this test, test both models sequentially, add checks for content of response
```
linglp, line 884: I think maybe we should open a Jira issue for this. I don't think we would remember to look at comments

linglp:
@mialy-defelice and @GiaJordan 
I think having `all_layers` default as a list is likely to cause error in `save_outputs` (see the line [here](https://github.com/Sage-Bionetworks/schematic/blob/e8270007db931c91b4039bfb99ed5eb110562791/schematic/visualization/tangled_tree.py#L691)). The post [here](https://stackoverflow.com/questions/366422/how-can-i-avoid-issues-caused-by-pythons-early-bound-default-parameters-e-g-m) explains mutable default argument, and I think that's what caused the issue and confusion. Can we make some changes and make sure that all_layers don't get initialized as a list?

linglp (approved):
@GiaJordan @mialy-defelice I tested the PR and it looks good to me.

PR #1337, opened 2023-12-07
Resolved issues with cross-manifest validation that occured when a manifest was only a single row

GiaJordan:
Updates the util function `iterable_to_str_list` to A) handle the case where a single string is passed in and B) handle the case where a non-iterable object, such as a numerical object, is passed in.

Fixes FDS-1384
Fixes FDS-1425

### DIFF
--- a/schematic/utils/validate_utils.py
+++ b/schematic/utils/validate_utils.py
@@ -76,7 +76,10 @@ def np_array_to_str_list(np_array):
     return np.char.mod('%d', np_array).tolist()
 
 def iterable_to_str_list(iterable):
-    "Parse an iterable into a list of strings"
+    """
+    Parse an object into a list of strings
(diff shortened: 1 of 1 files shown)


schematic/utils/validate_utils.py
```diff
@@ -76,10 +77,18 @@ def np_array_to_str_list(np_array):
 
 def iterable_to_str_list(iterable):
     "Parse an iterable into a list of strings"
```
andrewelamb, line 79: You might want to update the docstring to describe the inputs that are allowed (str, Number, iterable)

andrewelamb (approved):

schematic/utils/validate_utils.py
GiaJordan: good point, thanks!

PR #1346, opened 2024-01-12
Fixed an issue where empty rows would not be removed from manifests, and add test for `load_df` util function

GiaJordan:
Add test for `load_df` util and resolve issue where empty rows would not be removed from a manifest when types were inferred

Completes [FDS-1574](https://sagebionetworks.jira.com/browse/FDS-1574) and resolves [FDS-1575](https://sagebionetworks.jira.com/browse/FDS-1575)

[FDS-1574]: https://sagebionetworks.jira.com/browse/FDS-1574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FDS-1575]: https://sagebionetworks.jira.com/browse/FDS-1575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Depends on #1345


schematic/utils/df_utils.py
```diff
@@ -31,14 +36,14 @@ def load_df(file_path, preserve_raw_input=True, data_model=False, **load_args):
     # Read CSV to df as type specified in kwargs
     org_df = pd.read_csv(file_path, keep_default_na=True, encoding="utf8", **load_args)
 
+    # only trim if not data model csv
+    if not data_model:
```
linglp, line 40: @GiaJordan the change here makes sense to me. But I am also wondering what's the rationale of moving this block of code out of the original if statement?

tests/test_utils.py
```diff
@@ -255,6 +255,25 @@ def test_load_schema_org(self):
 
 
 class TestDfUtils:
```
linglp, line 257: do we have any code that use `data_model=True` option? If so, I would recommend adding a test. If not, I would recommend removing `data_model` parameter. I saw that `_convert_csv_to_data_model` function is using `data_model=True` option, but it doesn't seem like `_convert_csv_to_data_model` is being used.

tests/data/mock_manifests/Invalid_Test_Manifest.csv
```diff
@@ -1,4 +1,5 @@
 Component,Check List,Check Regex List,Check Regex Single,Check Regex Format,Check Regex Integer,Check Num,Check Float,Check Int,Check String,Check URL,Check Match at Least,Check Match at Least values,Check Match Exactly,Check Match Exactly values,Check Recommended,Check Ages,Check Unique,Check Range,Check Date,Check NA
 MockComponent,"ab,cd","ab,cd,ef",a,a,5.4,6,99.65,7,valid,https://www.google.com/,1738,1738,8085,98085,,6549,str1,70,32-984,7
 MockComponent,invalid list values,ab cd ef,q,m,0,c,99,5.63,94,http://googlef.com/,7163,51100,9965,71738,,32851,str1,30,notADate,9.5
-MockComponent,"ab,cd","ab,cd,ef",b,b,683902,6.5,62.3,2,valid,https://github.com/Sage-Bionetworks/schematic,8085,8085,1738,210065,,6550,str1,90,84-43-094,Not Applicable
\ No newline at end of file
+MockComponent,"ab,cd","ab,cd,ef",b,b,683902,6.5,62.3,2,valid,https://github.com/Sage-Bionetworks/schematic,8085,8085,1738,210065,,6550,str1,90,84-43-094,Not Applicable
```
linglp, line 4: is there any changes in this part? Or just the formatting?

tests/test_utils.py
```diff
@@ -255,6 +255,25 @@ def test_load_schema_org(self):
...
+    @pytest.mark.parametrize("preserve_raw_input", [True, False], ids=["Do not infer datatypes", "Infer datatypes"])
+    def test_load_df(self, helpers, preserve_raw_input):
+        test_col = "Check NA"
+        file_path = helpers.get_data_path("mock_manifests", "Invalid_Test_Manifest.csv")
+
+        df = df_utils.load_df(file_path, preserve_raw_input=preserve_raw_input, data_model=False)
+
+        assert df["Component"].dtype == "object"
+        assert df.shape[0] == 3
```
linglp, line 266: is this part of the test trying to make sure that the empty rows have been removed? If so, could you make a comment in the code and make it clearer? I would suggest doing something like: 
```
original_df = pd.read_csv(file_path)
assert original_df.shape[0] == 4
df = xxx (call the load_df function) 
assert df.shape[0] == 3
```
`data_model=True` parameter doesn't seem to be covered by test. If this is also a parameter that we use in the code, could you add test for it too?

schematic/utils/df_utils.py
linglp: I see. Thanks for the explanation.

tests/test_utils.py
linglp: Oh never mind. I just noticed that `parse_csv_model` function uses load_df and data_model=True. Since you are adding test for `load_df`, do you think you could also write a test to cover situations where data_model = True? If not, could you open a new ticket?

linglp (approved):
This looks good to me other than additional tests that could be added to load_df. I am going to approve for now.

tests/test_utils.py
GiaJordan: Thanks for confirming @linglp! I've opened up a ticket to cover that parameter's use

PR #1350, opened 2024-01-17
Refactor Schemas - Latest with Fixes.

mialy-defelice:
There was an error when merging up to to Refactor Schemas, so I re-opened the branch and merged Refactor Schemas down. This branch is able to be used directly.


linglp (approved):
@mialy-defelice thanks! I tested both APIs and CLIs and ran pytest locally. Everything is passing except some issues that are already in the develop branch. See fix here: https://github.com/Sage-Bionetworks/schematic/pull/1352

mialy-defelice:
@linglp Thank you! And thanks for tracking down the source of those errors :)

PR #1351, opened 2024-01-17
Move remained logic in `routes.py` related to manifest generator to library

linglp:


schematic/manifest/generator.py
```diff
@@ -1552,6 +1544,25 @@ def create_manifests(jsonld:str, data_types:list, access_token:Optional[str]=Non
         Returns:
             Union[List[str], List[pd.DataFrame], BinaryIO]: a list of Googlesheet URLs, a list of pandas dataframes or an Excel file.
         """
```
andrewelamb, line 1546: You should remove the BinaryIo type from the union. As far as I can tell when you pick excel its' returning a dataframe.

linglp: No, when you pick Excel, it returns an excel file.

linglp: Thanks for your feedback Andrew! I think if you pick "excel", the code would return a list of Excel file paths. I will correct the typing.

schematic/manifest/generator.py
```diff
@@ -1589,7 +1588,7 @@ def create_single_manifest(
             use_annotations (bool, optional): whether to use annotations. Defaults to False.
```
andrewelamb, line 1589: The docstring doesn't match the method arguments. Could you re-write to match?

schematic/manifest/generator.py
```diff
@@ -1655,8 +1639,27 @@ def create_manifests(
...
```
andrewelamb, line 1640: Change data_type -> data_types and dataset_id -> dataset_ids here

schematic/manifest/generator.py
```diff
@@ -1641,7 +1625,7 @@ def create_manifests(
         title: Optional[str] = None,
         strict: Optional[bool] = True,
         use_annotations: Optional[bool] = False,
-    ) -> Union[List[str], List[pd.DataFrame], BinaryIO]:
+    ) -> Union[List[str], List[pd.DataFrame]]:
```
andrewelamb, line 1628: If you pick excel as output type, do you still get a list[str] or just a str?

schematic/manifest/generator.py
andrewelamb: Should this be `excel file path` or can you get multiple excel sheets from this?

schematic/manifest/generator.py
```diff
@@ -1655,8 +1639,27 @@ def create_manifests(
...
+                    f"There is a mismatch in the number of data_types and dataset_id's that "
+                    f"submitted. Please check your submission and try again."
+                )
+
+            # Raise an error if used in conjunction with datatype = 'all_manifests'
+            if data_types[0] == "all manifests":
+                raise ValueError(
+                    f"When submitting 'all manifests' as the data_type cannot also submit dataset_id. "
+                    f"Please check your submission and try again."
+                )
```
andrewelamb, line 1661: Whats the reason for using f-strings here?

schematic_api/api/routes.py
```diff
@@ -258,43 +259,16 @@ def get_manifest_route(schema_url: str, use_annotations: bool, dataset_ids=None,
...
+    all_results = ManifestGenerator.create_manifests(path_to_data_model=schema_url, output_format=output_format, data_types=data_type, title=title, access_token=access_token, dataset_ids=dataset_id, strict=strict_validation, use_annotations=use_annotations)
+   
+    # return an excel file if output_format is set to "excel"
+    if output_format == "excel":
+        # should only contain one excel spreadsheet path
+        result = all_results[0]
+        dir_name = os.path.dirname(result)
+        file_name = os.path.basename(result)
+        mimetype='application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
+        return send_from_directory(directory=dir_name, path=file_name, as_attachment=True, mimetype=mimetype, max_age=0)
```
andrewelamb, line 271: I'm a little confused here. It looks like you are still using create_mainfests implying you cna get more than one manifest from this endpoint, but the arugment names are dataset_id and data_type. Shouldn't those be plural?

schematic/manifest/generator.py
linglp: the current behavior is that you would still get a list of excel file paths but only the first one would get returned as a file to the user (which is now being handled on the API side.). Now that I think about it more, to have the endpoint works more efficiently, probably it would make more sense to only return one excel file path here..but this is something that we could address in a different PR if needed.

schematic/manifest/generator.py
linglp: I think this should be excel file "paths" (instead of "path") because the current behavior is that the code would still return multiple excel file paths. The part about "returning excel sheet to the user" has moved to the API side (see routes.py).. and even if multiple excel file paths get returned by this function, only one file would get returned to users in routes.py

schematic/manifest/generator.py
linglp: This is legacy code. Let me change the f strings here.

schematic/manifest/generator.py
linglp: That's a good point. I have re-generated the docstring!

schematic/manifest/generator.py
linglp: Thanks. I have updated it.

schematic_api/api/routes.py
linglp: Yeah, they should be plural ideally. But if we want to change them to plural, we would also need to change the query parameters defined in `api.yml`. If we change that, we would also need to inform DFA, DCA so that they could change how they are calling schematic. Let me open a new issue for it and have them in a different PR. For this issue, I think it would be best if we just keep it simply for moving relevant parts from api to schematic library. 

See issue: https://sagebionetworks.jira.com/browse/FDS-1666

andrewelamb (approved):

PR #1360, opened 2024-02-02
Handle non-standard JSONLD Formats (no display name, and single dictionary entries)

mialy-defelice:
Used to help parse some exports from core models, but could have broader applicability.


thomasyu888:
:( Closing this due to age, but the PR is linked in the jira ticket for future reference

PR #1368, opened 2024-02-09
Feature: Component Based Rule Settings

mialy-defelice:
## Component Based Rule Setting

Component-Based Rule Setting is a powerful feature in data modeling that enables users to create rules tailored to specific subsets of components or manifests. This functionality was developed to address scenarios where a data modeler needs to enforce uniqueness for certain attribute values within one manifest while allowing non-uniqueness in another.

**Here's how it works:**

- **Rule Definition at Attribute Level:** Rules are defined at the attribute level within the data model.
- **Manifest-Level Referencing:** These rules can then be applied (or not) to specific manifests within the data model. This means that rules can be selectively enforced based on the manifest they're associated with.

This feature offers flexibility and applicability beyond its original use case. The new Component-Based Rule Setting feature provides users with the following options:

**Apply a Rule to All Manifests Except Specified Ones:** Users can now define a rule that applies to all manifests within the data model except for those explicitly specified. In cases where exceptions are specified, users have the flexibility to define unique rules for these exceptions or opt not to apply any rule at all.

**Specify a Rule for a Single Manifest:** Alternatively, users can specify a rule that applies to a single manifest exclusively. This allows for fine-grained control over rule enforcement at the manifest level.

**Unique Rules for Each Manifest:** Users can also define unique rules for each individual manifest within the data model. This enables tailored rule enforcement based on the specific requirements and characteristics of each manifest.

By leveraging the enhanced Component-Based Rule Setting feature, data modelers can efficiently enforce rules across their data models with greater precision and flexibility, ensuring data integrity while accommodating diverse use cases and requirements.

**Note**: All restrictions to rule combos and implementation also apply to component based rules.

As always try the rule combos with mock data to ensure they are working as intended before using in production.

**Format:**

- `^^`Double carrots indicates that Component based rules are being set
- Use `^^` to separate component rule sets
- `# In the first position (prior to the rule) to define the component/manifest to apply the rule to
- `# character cannot be used with out the `^^` to indicate component rule sets

**Use case:**

- Apply rule to all manifests except specified set.
  - `validation_rule^^#ComponentA`
  - `validation_rule^^#ComponentA^^#ComponentB`
- Apply a unique rule to each manifest.
  - `#ComponentA validation_rule_1^^#ComponentB validation_rule_2^^#ComponentC validation_rule_3`
- For specified manifest apply given validation rule, but for all others run a different rule
  - `#ComponentA validation_rule_1^^validation_rule_2`
  - `validation_rule_2^^#ComponentA validation_rule_1`
- Apply validation rule to only one manifest
  - `#ComponentA validation_rule_1^^`

### Testing Resources
**Test Models:**
TBD

**Test Manifest:**
[Example Biospecimen Manifest](https://docs.google.com/spreadsheets/d/19_axG2Zj7URk4CT5qYjH0HfpMIOQ1dYEPvyaazSVNZE/edit#gid=0)

[Example Patient Manifest](https://docs.google.com/spreadsheets/d/1IO0TkzwBX-lsu3rJDjWfgWYR6VlepingN9zuhkrgVUE/edit#gid=0)

### Testing Steps:

TBD


schematic/manifest/generator.py
```diff
@@ -751,7 +752,7 @@ def _request_regex_match_vr_formatting(
                 - Upon correct format entry, text will turn black.
                 - If incorrect format is entered a validation error will pop up.
         Input:
-            validation_rules: List[str], defines the validation rules
+            validation_rules: Any(List[str], Dict), defines the validation rules
                 applied to a particular column.
```
andrewelamb, line 756: I know this is the docstring, but this isn't how Any isused for typing. It also doesn't thta type in the function definition. It looks like list[str] is correct.

schematic/utils/schema_utils.py
```diff
@@ -2,13 +2,20 @@
...
 
 logger = logging.getLogger(__name__)
 
 DisplayLabelType = Literal["class_label", "display_label"]
 BLACKLISTED_CHARS = ["(", ")", ".", " ", "-"]
+DELIMITERS = {
+    "component_name_delimiter": "#",
+    "component_rules_delimiter": "^^",
+    "rule_delimiter": "::",
+}
```
andrewelamb, line 18: As mentioned in code review, I think the delimeters would be better as their own globals since you don't use the dict for anything besides looking up a specific value.

schematic/utils/schema_utils.py
```diff
@@ -207,15 +214,156 @@ def convert_bool_to_str(provided_bool: bool) -> str:
     return str(provided_bool)
 
 
-def parse_validation_rules(validation_rules: List[str]) -> List[str]:
+def get_individual_rules(rule: str, validation_rules: list[str]) -> list:
```
andrewelamb, line 217: Needs a docstring

schematic/utils/schema_utils.py
```diff
@@ -207,15 +214,156 @@ def convert_bool_to_str(provided_bool: bool) -> str:
...
+    # Get single rule
+    else:
+        validation_rules.append(rule)
+    return validation_rules
+
+
+def get_component_name_rules(
+    component_names: list, component_rule: str
+) -> Tuple[list, str]:
+    # If a component name is not attached to the rule, have it apply to all other components
```
andrewelamb, line 230: Needs a docstring

schematic/utils/schema_utils.py
```diff
@@ -207,15 +214,156 @@ def convert_bool_to_str(provided_bool: bool) -> str:
...
+    else:
+        component_names.append(
+            component_rule.split(" ")[0].replace(
+                DELIMITERS["component_name_delimiter"], ""
+            )
+        )
+        try:
+            assert component_names[-1] != " "
+        except ValueError:
+            print(
```
andrewelamb, line 243: This try/except block is a little wierd.. If you want to print that message when
`component_names[-1] != " "` then why not an `if` statement?
```
if component_names[-1] != " ":
    print(msg)
```

schematic/utils/schema_utils.py
```diff
@@ -207,15 +214,156 @@ def convert_bool_to_str(provided_bool: bool) -> str:
...
+                f"in the following rule: {component_rule}, "
+                f"please ensure there is not extra whitespace or non-allowed characters."
+            )
+        component_rule = component_rule.replace(component_rule.split(" ")[0], "")
+        component_rule = component_rule.strip()
+    return component_names, component_rule
+
+
+def parse_component_validation_rules(validation_rule_string: str) -> Dict:
+    component_names = []
```
andrewelamb, line 254: Needs a docstring

schematic/utils/schema_utils.py
```diff
@@ -207,15 +214,156 @@ def convert_bool_to_str(provided_bool: bool) -> str:
...
+            validation_rules = get_individual_rules(
+                rule=component_rule, validation_rules=validation_rules
+            )
+
+    # Ensure we collected the component names and validation rules like expected
+    try:
+        assert len(component_names) == len(validation_rules)
+    except ValueError:
+        print(
+            f"The number of components names and validation rules does not match "
```
andrewelamb, line 279: See above about try/except block

schematic/utils/schema_utils.py
```diff
@@ -207,15 +214,156 @@ def convert_bool_to_str(provided_bool: bool) -> str:
...
+            f"for validation rule: {validation_rule_string}."
+        )
+
+    validation_rules_dict = dict(
+        map(lambda i, j: (i, j), component_names, validation_rules)
+    )
```
andrewelamb, line 285: I think zip is the most pythonic way to create a dict from two lists:

```
keys = ['a', 'b', 'c']
values = [1, 2, 3]
dictionary = dict(zip(keys, values))
```

schematic/utils/schema_utils.py
```diff
@@ -207,15 +214,156 @@ def convert_bool_to_str(provided_bool: bool) -> str:
...
+    return validation_rules_dict
+
+
+def parse_single_set_validation_rules(validation_rule_string: str) -> list:
+    # Try to catch an improperly formatted rule
```
andrewelamb, line 290: Needs a docstring

schematic/utils/schema_utils.py
```diff
@@ -207,15 +214,156 @@ def convert_bool_to_str(provided_bool: bool) -> str:
...
+            return parse_component_validation_rules(
+                validation_rule_string=validation_rules[0]
+            )
+        # Parse rules that are set across *all* components/manifests
+        else:
+            return parse_single_set_validation_rules(
+                validation_rule_string=validation_rules[0]
+            )
+    else:
+        raise ValueError(
```
andrewelamb, line 332: So this is interesting. Since `validation_rules: Union[list, dict]` can only be a `dict` or `list` technically you don't need this last `else` statement. If other functions are sending things besides dict/list then validation_rules shouldadd those types to the Union, or just have it be `Any`

schematic/manifest/generator.py
mialy-defelice: Oh yes, I changed what I was taking in while working on it...and yea I originally thought Any worked like Union... Ill remove this.

schematic/utils/schema_utils.py
mialy-defelice: Well this is raising an Error if the assert doesn't work...I was following the pattern I had seen elsewhere in our code base but I could do an `if/else` and use `logger.error` instead like:
```
if component_names and component_names[-1] == " ":
    logger.error(" ...")
 ```

Comment (andrewelamb) [reply schematic/utils/schema_utils.py:243]: Yeah I've seen it elsewhere in the codebase... I'm not sure where it came from :)

**Comment by mialy-defelice** [comment]
@andrewelamb @GiaJordan I added an additional check for duplicate components in a rule, and updated the testing.

File: schematic/utils/schema_utils.py (hunk 5 above)
Comment (mialy-defelice) [reply schematic/utils/schema_utils.py:243]: Looking up docs, try:excepts seem standard...what is the method you would prefer?
https://docs.python.org/3/tutorial/errors.html

Comment (andrewelamb) [reply schematic/utils/schema_utils.py:259]: They both likely do the same thing. To me it's more about writing clear and readable code  and this is a case of control flow: If <something is the case> : print (msg).

Using the try/except tells me that you are expecting unavoidable exceptions somewhere else and you need to catch them so you cna handle them. By adding the assert staement the code is forcing an unesseccary exception just so that a try/except statement block can be used.

I don't think it's a big deal, I when I saw it, it took me a moment to understand why it was written this way.
I think this is much clearer:

```
if component_names and component_names[-1] == " ":
    logger.error(" ...")
```

andrewelamb (approved):

PR #1371, opened 2024-02-16
FDS-1721 pylint schema utils

andrewelamb:
1. Jira: [FDS-1721]
2. pylints utils/schema_utils.py

[FDS-1721]: https://sagebionetworks.jira.com/browse/FDS-1721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schematic/schemas/data_model_nodes.py
+++ b/schematic/schemas/data_model_nodes.py
@@ -155,7 +155,10 @@ class DataModelNodes:
             )
 
         elif rel_func == parse_validation_rules:
-            return parse_validation_rules(attr_relationships[csv_header])
+            rules = attr_relationships[csv_header]
+            if not isinstance(rules, dict) or isinstance(rules, list):
+                raise ValueError("Rules are not formatted properly, should be a dict or list")
+            return parse_validation_rules(rules)
 
         elif rel_func == get_label_from_display_name:
             return get_label_from_display_name(
--- a/schematic/schemas/data_model_nodes.py
+++ b/schematic/schemas/data_model_nodes.py
@@ -157,7 +157,9 @@ class DataModelNodes:
         elif rel_func == parse_validation_rules:
             rules = attr_relationships[csv_header]
             if not isinstance(rules, dict) or isinstance(rules, list):
-                raise ValueError("Rules are not formatted properly, should be a dict or list")
+                raise ValueError(
+                    "Rules are not formatted properly, should be a dict or list"
+                )
             return parse_validation_rules(rules)
 
         elif rel_func == get_label_from_display_name:
--- a/schematic/schemas/data_model_nodes.py
+++ b/schematic/schemas/data_model_nodes.py
@@ -156,11 +156,8 @@ class DataModelNodes:
 
         elif rel_func == parse_validation_rules:
             rules = attr_relationships[csv_header]
-            if not isinstance(rules, dict) or isinstance(rules, list):
-                raise ValueError(
-                    "Rules are not formatted properly, should be a dict or list"
-                )
-            return parse_validation_rules(rules)
+            if isinstance(rules, dict) or isinstance(rules, list):
+                return parse_validation_rules(rules)
 
         elif rel_func == get_label_from_display_name:
             return get_label_from_display_name(
--- a/schematic/utils/schema_utils.py
+++ b/schematic/utils/schema_utils.py
@@ -1,22 +1,34 @@
-import inflection
+"""Schema utils"""
+
+# pylint: disable=logging-fstring-interpolation
+
 import json
 import logging
-import networkx as nx
-import re
 import string
-from typing import List, Literal, Dict, Tuple, Union
+from typing import Literal, Union, Optional, Any
+
+import inflection
 
 
 logger = logging.getLogger(__name__)
 
 DisplayLabelType = Literal["class_label", "display_label"]
+EntryType = Literal["class", "property"]
 BLACKLISTED_CHARS = ["(", ")", ".", " ", "-"]
 COMPONENT_NAME_DELIMITER = "#"
 COMPONENT_RULES_DELIMITER = "^^"
 RULE_DELIMITER = "::"
 
 
-def attr_dict_template(key_name: str) -> Dict[str, dict[str, dict]]:
+def attr_dict_template(key_name: str) -> dict[str, dict[str, dict]]:
+    """_summary_
+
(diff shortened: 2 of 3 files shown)


schematic/utils/schema_utils.py
```diff
@@ -268,32 +308,37 @@ def get_component_name_rules(
...
+    Check if component names are repeated in a validation rule
+    Error Logged if a component name is duplicated.
+
     Args:
-        component_names, list[str]: list of components identified in the validation rule
-        validation_rule_str, str: validation rule, used if error needs to be raised.
-    Returns:
-        None
-    Raises: Error Logged if a component name is duplicated.
+        component_names (list[str]): _description_
```
mialy-defelice, line 316: list of components identified in the validation rule

schematic/utils/schema_utils.py
```diff
@@ -268,32 +308,37 @@ def get_component_name_rules(
...
+        validation_rule_string (str): _description_
```
mialy-defelice, line 317: validation rule, used if error needs to be raised.

mialy-defelice (changes requested):
I added some comments for updating descriptions in the docstrings.

schematic/utils/schema_utils.py
```diff
@@ -327,72 +372,74 @@ def parse_component_validation_rules(validation_rule_string: str) -> Dict:
...
+            f"The provided validation rule {validation_rule_string}, looks to be formatted as a "
+            "component based rule, but is missing the necessary formatting, "
+            "please refer to the SchemaHub documentation for more details."
         )
 
     return validation_rule_string.split(RULE_DELIMITER)
 
 
-def parse_validation_rules(validation_rules: Union[list, dict]) -> Union[list, dict]:
+def parse_validation_rules(validation_rules: Any) -> Optional[Union[list, dict]]:
```
mialy-defelice, line 395: In the PR for this function, for the typing, you indicated if we used union to restrict to list and dict, then we wouldn't have to add and additional else to look for other types. With it like this, its actually not being properly flagged when a str is passed. I think we should either add back the Union or add an additional test that would raise an exception for strings. Its passing the test right (`pytest tests/test_utils.py::TestSchemaUtils::test_parse_validation_rules[str_rule]` now because the asserts are in the try...I think we should update the test (perhaps in another PR) to move the asserts out of the try. Like this:

```try:
            parsed_validation_rules = parse_validation_rules(
                validation_rules=validation_rules
            )
        except:
            assert test_rule_name in ["str_rule", "duplicated_component"]

        if parsed_validation_rules:
            assert expected_parsed_rules == parsed_validation_rules
            assert test_rule_name not in ["str_rule", "duplicated_component"]
        else:
            test_rule_name in ["str_rule", "duplicated_component"]

**Review by mialy-defelice — CHANGES_REQUESTED** [review]

File: schematic/utils/schema_utils.py (hunk 3 above)
Comment (andrewelamb) [reply schematic/utils/schema_utils.py:395]: Yeah this is interesting, and I don't think there's one rigth answer. I guess it comes down to do you think its this functions job to handle strings (even if its to raise an error), or does it just handle lists/dicts and its up to the calling functions to make sure they are providing a list or dict? I would go with the latter, but we would need to make sure the functiosn that call it are doign so correctly.

Comment (andrewelamb) [reply schematic/utils/schema_utils.py:395]: So this is the only places this function is called besides tests:

Comment (andrewelamb) [reply schematic/utils/schema_utils.py]: Ok I tried the above and got a lot of errores in the tests

Comment (mialy-defelice) [reply schematic/utils/schema_utils.py]: I think I would rather explicitly restrict the inputs since we are expecting them to be only those, but anyone developing in the future might not know that, I would rather them be stopped.

Comment (andrewelamb) [reply schematic/utils/schema_utils.py]: @mialy-defelice I changes the code where it's caled to:

```
elif rel_func == parse_validation_rules:
    rules = attr_relationships[csv_header]
    if  isinstance(rules, dict) or isinstance(rules, list):
        return parse_validation_rules(rules)
```
The function is now only called for lists and dicts

The funciton now looks like:

```
def parse_validation_rules(validation_rules: Union[list, dict]) -> Union[list, dict]:
    """Split multiple validation rules based on :: delimiter

    Args:
        validation_rules (Union[list, dict]): List or Dictionary of validation rules,
            if list:, contains a string validation rule
            if dict:, key is the component the rule (value) is applied to

    Returns:
        Union[list, dict]: Parsed validation rules, component rules are output
          as a dictionary, single sets are a list.
    """
    if isinstance(validation_rules, dict):
        # Rules pulled in as a dict can be used directly
        return validation_rules

    else:
        # If rules are already parsed from the JSONLD
        if len(validation_rules) > 1 and isinstance(validation_rules[-1], str):
            return validation_rules
        # Parse rules set for a subset of components/manifests
        if COMPONENT_RULES_DELIMITER in validation_rules[0]:
            return parse_component_validation_rules(
                validation_rule_string=validation_rules[0]
            )
        # Parse rules that are set across *all* components/manifests
        return parse_single_set_validation_rules(
            validation_rule_string=validation_rules[0]
        )
```
Now itexplicityl only takes lists or dicts.

File: schematic/schemas/data_model_nodes.py [hunk 4]
```diff
@@ -155,7 +155,9 @@ def run_rel_functions(
             )
 
         elif rel_func == parse_validation_rules:
-            return parse_validation_rules(attr_relationships[csv_header])
+            rules = attr_relationships[csv_header]
```
mialy-defelice, line 158: Im clearly worried about people submitting odd things, but since we are gating here we won't hit a typing error if a string is submitted somehow, Can we raise an exception in an else?

mialy-defelice (changes requested):

schematic/schemas/data_model_nodes.py
andrewelamb: So I tried this at one point:

```
elif rel_func == parse_validation_rules:
    rules = attr_relationships[csv_header]
    if  isinstance(rules, dict) or isinstance(rules, list):
        return parse_validation_rules(rules)
   else:
      raise ValueError
```
but a bunch of tests failed. The way I have it now is closest to the current behavior, as there were not any exceptions raised before.

mialy-defelice (approved):

PR #1374, opened 2024-02-20
Avoid repeatedly calling `generate_data_model_graph`

linglp:
## Context: 
Related to https://sagebionetworks.jira.com/browse/FDS-1750

## Test
Since this affected multiple parts of the code base, I did: 
1) ran `test_api.py` locally and make sure that everything works there as expected
2) ran `model/componet-requirements` endpoint and used HTAN schema url as an input parameter
3) use `manifest/generate` endpoint to generate a new manifest

### DIFF
--- a/schematic/manifest/commands.py
+++ b/schematic/manifest/commands.py
@@ -153,7 +153,7 @@ def get_manifest(
 
     # Generate graph
     logger.info("Generating data model graph.")
-    graph_data_model = data_model_grapher.generate_data_model_graph()
+    graph_data_model = data_model_grapher.graph
 
     def create_single_manifest(data_type, output_csv=None, output_xlsx=None):
         # create object of type ManifestGenerator
--- a/schematic/visualization/attributes_explorer.py
+++ b/schematic/visualization/attributes_explorer.py
@@ -18,7 +18,7 @@ logger = logging.getLogger(__name__)
 class AttributesExplorer:
     """AttributesExplorer class"""
 
-    # pylint: disable=R0913
+    # pylint: disable=too-many-arguments
     def __init__(
         self,
         path_to_jsonld: str,
--- a/schematic/visualization/attributes_explorer.py
+++ b/schematic/visualization/attributes_explorer.py
@@ -24,9 +24,9 @@ class AttributesExplorer:
         self,
         path_to_jsonld: str,
         data_model_labels: DisplayLabelType,
-        data_model_grapher: DataModelGraph = None, 
-        data_model_graph_explorer: DataModelGraphExplorer=None,
-        parsed_data_model: Dict[str, Dict[str, Any]]=None
+        data_model_grapher: DataModelGraph = None,
+        data_model_graph_explorer: DataModelGraphExplorer = None,
+        parsed_data_model: Dict[str, Dict[str, Any]] = None,
     ) -> None:
         self.path_to_jsonld = path_to_jsonld
 
@@ -35,8 +35,8 @@ class AttributesExplorer:
         # Parse Model
         if not parsed_data_model:
             data_model_parser = DataModelParser(
-            path_to_data_model=self.path_to_jsonld,
-        )            
+                path_to_data_model=self.path_to_jsonld,
+            )
             parsed_data_model = data_model_parser.parse_model()
 
         # Instantiate DataModelGraph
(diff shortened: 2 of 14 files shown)


schematic/visualization/tangled_tree.py
mialy-defelice, line 1: The tangled tree actually calls the attributes explorer, both init functions overlap quite a bit. Can you let the attributes explorer optionally take in the graph data model and data model graph explorer. then put the parser under an `if not statement` so it only runs if the `graph_data_model` and `dmge` have not been supplied.

mialy-defelice (approved):
Thanks Lingling, I requested one change :)

mialy-defelice (changes requested):
Oops accidentally approved

schematic/visualization/tangled_tree.py
linglp: Thanks @mialy-defelice . This is a good point. I added those optional parameters as you suggested.

mialy-defelice:
Another update along the same lines, the `schemas/commands.py` file has a similar call:
```
# Generate graph
logger.info("Generating data model graph.")
graph_data_model = data_model_grapher.generate_data_model_graph()
```

mialy-defelice:
One more in `tests/test_api.py` in `get_MockComponent_attribute`

tests/test_api.py
```diff
@@ -106,7 +106,7 @@ def get_MockComponent_attribute():
     data_model_grapher = DataModelGraph(parsed_data_model)
 
     # Generate graph
-    graph_data_model = data_model_grapher.generate_data_model_graph()
+    graph_data_model = data_model_grapher.graph
```
linglp, line 109: @mialy-defelice Here I already made a change for `get_MockComponent_attribute()` function. Does this what you meant in your comment?

schematic/manifest/commands.py
```diff
@@ -153,7 +153,7 @@ def get_manifest(
 
     # Generate graph
     logger.info("Generating data model graph.")
-    graph_data_model = data_model_grapher.generate_data_model_graph()
+    graph_data_model = data_model_grapher.graph
```
linglp, line 156: @mialy-defelice I already made this change. Is this what you were referring to in your comment?

mialy-defelice: Ah yes!

tests/test_api.py
mialy-defelice: Yes, thank you!

mialy-defelice:
Okay this looks good to me, I am going to preemptively approve, assuming you run the relevant tests again after resolving conflicts with the attributes_explorer.

mialy-defelice (approved):

linglp:
@mialy-defelice Thanks. I have re-run `test_api.py` and everything is working as expected. I tried calling visualization endpoints from the APIs and they are working as expected as well. The only error is a known error: 
```
FAILED tests/test_api.py::TestSchemaVisualization::test_visualize_tangled_tree_layers[dependency] - assert 500 == 200
```

schematic/visualization/attributes_explorer.py
```diff
@@ -19,31 +18,38 @@
 class AttributesExplorer:
     """AttributesExplorer class"""
 
+    # pylint: disable=R0913
```
linglp, line 21: @andrewelamb @mialy-defelice can I add this line to bypass the complaint of "too many arguments": 
```
schematic/visualization/attributes_explorer.py:21:4: R0913: Too many arguments (6/5) (too-many-arguments)
```

I tried changing the class to `dataclass` but still I got: 
```
schematic/visualization/attributes_explorer.py:19:0: R0902: Too many instance attributes (11/7) (too-many-instance-attributes)
```

If this needs to be addressed, can we address it in a separate PR?

andrewelamb: I think that's ok to do. But to make it more clear I would do  #pylint: disable=too-many-arguments

andrewelamb: Yes I don't think dataclass is appropriate here. I would add `#pylint: disable=too-many-instance-attributes` for now.

linglp: @andrewelamb Thanks for the feedback. I will update this part based on what you advised. But since I am not using dataclass, I assume that adding `#pylint: disable=too-many-arguments` will be sufficient?

andrewelamb: Yep!

PR #1380, opened 2024-02-23
[bug fix] Updated workflow to install dependencies and root project together

linglp:
## Context
When installing dependencies, we had: 
```
        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
```
which made that github action skipped dependencies installation after `poetry install --only-root`. See screen shot that made the action failed. 

Screen Shot 2024-02-23 at 1 48 22 PM

I think it makes sense to just install root and dependencies together. 
I tested it locally to make sure that the project itself could be installed: 
```
(schematicpy-py3.10) lpeng@ip-192-168-0-2 schematic % poetry install --no-interaction --all-extras
Configuration file exists at /Users/lpeng/Library/Preferences/pypoetry, reusing this directory.

Consider moving TOML configuration files to /Users/lpeng/Library/Application Support/pypoetry, as support for the legacy directory will be removed in an upcoming release.
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: schematicpy (23.11.1)

```

### DIFF
--- a/.github/workflows/api_test.yml
+++ b/.github/workflows/api_test.yml
@@ -54,17 +54,11 @@ jobs:
           key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
 
       #----------------------------------------------
-      # install dependencies if cache does not exist
+      # install dependencies and root project
       #----------------------------------------------
-      - name: Install dependencies
+      - name: Install dependencies and root project
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-        run: poetry install --no-interaction --no-root --all-extras
-
-      #----------------------------------------------
-      #    install your root project, if required
-      #----------------------------------------------
-      - name: Install library
-        run: poetry install --no-interaction
+        run: poetry install --no-interaction --all-extras
 
       #----------------------------------------------
       #              run API test suite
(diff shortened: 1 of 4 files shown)


GiaJordan (changes requested):
Should the `pdoc.yml` workflow be updated as well?

linglp:
@GiaJordan thanks. I just updated the pdoc.yml. I was initially thinking of not updating it because pdoc.yml doesn't need API dependencies, for example uWSGI. But since this is more like an urgent fix, like me just update pdoc.yml to have it align with the rest of the workflow files. I will then open a separate issues for us to clean up the workflow files, the logic between them, and make it easier to see requirements of different files.

GiaJordan:
> @GiaJordan thanks. I just updated the pdoc.yml. I was initially thinking of not updating it because pdoc.yml doesn't need API dependencies, for example uWSGI. But since this is more like an urgent fix, like me just update pdoc.yml to have it align with the rest of the workflow files. I will then open a separate issues for us to clean up the workflow files, the logic between them, and make it easier to see requirements of different files.

Thank you, and that will be useful. Especially with the api being separated out from this repo and moving to the monorepo we could probably simplify a lot within the workflows
@linglp

GiaJordan:
@linglp is it possible to run the `pdoc` workflow once before merge with the latest change in?

GiaJordan (approved):
@linglp sorry for the confusion; given that the build passed we can leave it like you have now. Thanks again!

[🚀](https://emojipedia.org/rocket)

PR #1387, opened 2024-03-19
Match None Manifest Validation Rule FDS-1766

mialy-defelice:
### About this PR

This PR addresses [FDS-1766](https://sagebionetworks.jira.com/browse/FDS-1766).

The new rule works as follows:

- Format: 
   - `matchNone <targetComponent>.<targetAttribute> <scope> <raised message level>`
   - Will check that values of the specified attribute are not present in other manifests
   - Attribute will be invalid if the values are found in any other manifest
   - Default behavior: raises `warning`
- Example: 
   - matchNone Patient.PatientID set

      - Will check that the complete set of values from the specified attribute (that this rule is applied to), in the manifest that is currently being submitted for validation, is not present in any existing (target) Patient manifests that contains the PatientID attribute. This could be used to confirm that any set of  PatientIDs are not repeated in other manifest, for example.

   - matchNone Patient.PatientID value

      - Will check that each individual value from the specified attribute (that this rule is applied to), in the manifest that is currently being submitted for validation, is not present in any existing (target) Patient manifests that contains the PatientID attribute. This could be used to confirm that any PatientIDs are not repeated in other manifest, for example.

New CheckNone attributes and rules have been added to the `example.model.csv` for use in unit tests to confirm the rules are working as expected.

The new attributes have been added to the Synapse test files, and all cross manifest validation files have been relocated to the synapse project: [Schematic Cross Manifest Validation Testing,](https://www.synapse.org/#!Synapse:syn54126707/wiki/626665) to reduce testing time.

This PR also includes a minor refactor of the cross manifest validation portion of `validate_attribute` to avoid code duplication with addition of this new rule, and for code clarity and maintenance..

[FDS-1766]: https://sagebionetworks.jira.com/browse/FDS-1766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

--------------

### Testing this PR

This PR can be tested by making a copy of `tests/data/mock_manifests/Valid_Test_Manifest.csv`.
In `config.yml` set `master_fileview_id: "syn54126711"`

Run validation as follows:
```
schematic model -c config.yml validate -mp tests/data/mock_manifests/Valid_Test_Manifest_Copy.csv -dt MockComponent -ps syn54126707
```

Validation should pass.

Next, to make validation fail, change one of the values (column `checkMatchNone`)in your test manifest to `123`.

```
error: Value(s) ['123'] from row(s) ['3'] for the attribute Check Match None in the source manifest are not unique. Manifest(s) ['syn54126950'] contain duplicate values.
```

Additional  Potential Testing:

- Can do variations of this on your own, the goal to failure would be to make values in the original `Valid_Test_Manifest.csv`  columns `checkMatchNone` and `checkMatchNonevalues` overlap values in the Synapse test manifests: `[syn54126950, syn54127702, syn54127008]`
- Can also go into example data model `example.model.csv` and add new versions of the checkNone rule (change scope or error level).
   - Then, either convert to JSONLD or update the path to the data model in the config to the CSV.
- Can also run with parameters like `-rr` which goes through a separate initial process when forcing all the rules to be applied with `in house validation rules`.
- Test via API and CLI

### DIFF
--- a/schematic/models/validate_attribute.py
+++ b/schematic/models/validate_attribute.py
@@ -4,7 +4,7 @@ import re
 from time import perf_counter
 
 # allows specifying explicit variable types
-from typing import Any, Dict, List, Optional, Text, Literal, Union, Tuple
+from typing import Any, Optional, Text, Literal, Union
 from urllib import error
 from urllib.parse import urlparse
 from urllib.request import Request, urlopen
@@ -39,7 +39,7 @@ class GenerateError:
         error_message: str,
         invalid_entry: str,
         dmge: DataModelGraphExplorer,
-    ) -> Tuple[list[list[str]], list[list[str]]]:
+    ) -> tuple[list[list[str]], list[list[str]]]:
         """
         Purpose: Process error messages generated from schema
         Input:
@@ -69,7 +69,7 @@ class GenerateError:
         invalid_entry: str,
         dmge: DataModelGraphExplorer,
         val_rule: str,
-    ) -> Tuple[list[list[str]], list[list[str]]]:
+    ) -> tuple[list[list[str]], list[list[str]]]:
         """
         Purpose:
             If an error is found in the string formatting, detect and record
@@ -113,7 +113,7 @@ class GenerateError:
         attribute_name: str,
         invalid_entry: str,
         dmge: DataModelGraphExplorer,
-    ) -> Tuple[list[list[str]], list[list[str]]]:
+    ) -> tuple[list[list[str]], list[list[str]]]:
         """
         Purpose:
             Generate an logging error as well as a stored error message, when
@@ -153,7 +153,7 @@ class GenerateError:
         attribute_name: str,
         invalid_entry: str,
         dmge: DataModelGraphExplorer,
-    ) -> Tuple[list[list[str]], list[list[str]]]:
+    ) -> tuple[list[list[str]], list[list[str]]]:
         """
         Purpose:
             Generate an logging error as well as a stored error message, when
@@ -194,7 +194,7 @@ class GenerateError:
         invalid_entry: str,
         dmge: DataModelGraphExplorer,
         val_rule: str,
-    ) -> Tuple[list[list[str]], list[list[str]]]:
+    ) -> tuple[list[list[str]], list[list[str]]]:
         """
         Purpose:
             Generate an logging error as well as a stored error message, when
@@ -258,11 +258,11 @@ class GenerateError:
         val_rule: str,
         attribute_name: str,
         dmge: DataModelGraphExplorer,
-        matching_manifests: list[str] = [],
+        matching_manifests: list[str] = None,
         manifest_id: Optional[list[str]] = None,
         invalid_entry: Optional[list[str]] = None,
         row_num: Optional[list[str]] = None,
-    ) -> List[str]:
+    ) -> list[str]:
         """
         Purpose:
             Generate an logging error as well as a stored error message, when
@@ -289,7 +289,7 @@ class GenerateError:
             )
 
         elif "matchExactly" in val_rule:
-            if matching_manifests != []:
+            if matching_manifests and matching_manifests != []:
                 error_message = f"All values from attribute {attribute_name} in the source manifest are present in {len(matching_manifests)} manifests instead of only 1."
                 error_message += (
                     f" Manifests {matching_manifests} match the values in the source attribute."
@@ -330,7 +330,7 @@ class GenerateError:
         dmge: DataModelGraphExplorer,
         row_num=None,
         invalid_entry=None,
-    ) -> Tuple[list[str], list[str]]:
+    ) -> tuple[list[str], list[str]]:
         """
         Purpose:
             Generate an logging error or warning as well as a stored error/warning message when validating the content of a manifest attribute.
@@ -468,7 +468,7 @@ class GenerateError:
         error_col: str,
         error_message: str,
         error_val: Union[str, list[str]],
-    ) -> Tuple[list[str], list[str]]:
+    ) -> tuple[list[str], list[str]]:
         """
         Purpose:
             Log and store error messages in a list for further storage.
@@ -524,9 +524,11 @@ class ValidateAttribute(object):
         - Add year validator
         - Add string length validator
     """
+    def __init__(self, dmge: DataModelGraphExplorer):
+        self.dmge=dmge
 
     def get_target_manifests(
-        target_component, project_scope: list, access_token: str = None
+        self, target_component, project_scope: list, access_token: str = None
     ):
         t_manifest_search = perf_counter()
         target_manifest_ids = []
@@ -566,8 +568,7 @@ class ValidateAttribute(object):
         self,
         val_rule: str,
         manifest_col: pd.core.series.Series,
-        dmge: DataModelGraphExplorer,
-    ) -> Tuple[list[list[str]], list[list[str]], pd.core.series.Series]:
+    ) -> tuple[list[list[str]], list[list[str]], pd.core.series.Series]:
         """
         Purpose:
             Determine if values for a particular attribute are comma separated.
@@ -606,7 +607,7 @@ class ValidateAttribute(object):
                         attribute_name=manifest_col.name,
                         list_error=list_error,
                         invalid_entry=manifest_col[i],
-                        dmge=dmge,
+                        dmge=self.dmge,
                         val_rule=val_rule,
                     )
                     if vr_errors:
@@ -623,8 +624,7 @@ class ValidateAttribute(object):
         self,
         val_rule: str,
         manifest_col: pd.core.series.Series,
-        dmge: DataModelGraphExplorer,
-    ) -> Tuple[list[list[str]], list[list[str]]]:
+    ) -> tuple[list[list[str]], list[list[str]]]:
         """
         Purpose:
             Check if values for a given manifest attribue conform to the reguar expression,
@@ -665,7 +665,7 @@ class ValidateAttribute(object):
         errors = []
         warnings = []
 
-        validation_rules = dmge.get_node_validation_rules(
+        validation_rules = self.dmge.get_node_validation_rules(
             node_display_name=manifest_col.name
         )
         if validation_rules and "::" in validation_rules[0]:
@@ -689,7 +689,7 @@ class ValidateAttribute(object):
                             module_to_call=reg_exp_rules[1],
                             attribute_name=manifest_col.name,
                             invalid_entry=manifest_col[i],
-                            dmge=dmge,
+                            dmge=self.dmge,
                         )
                         if vr_errors:
                             errors.append(vr_errors)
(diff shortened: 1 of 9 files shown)


schematic/models/validate_attribute.py
```diff
@@ -254,11 +258,11 @@ def generate_cross_warning(
         val_rule: str,
         attribute_name: str,
         dmge: DataModelGraphExplorer,
-        matching_manifests=[],
-        missing_manifest_ID=None,
-        invalid_entry=None,
-        row_num=None,
-    ) -> Tuple[list[list[str]], list[list[str]]]:
+        matching_manifests: list[str] = [],
```
andrewelamb, line 261: Could we change the default here to `None` as well?

schematic/models/validate_attribute.py
```diff
@@ -1,9 +1,11 @@
+import builtins
 import logging
 import re
 from time import perf_counter
 
 # allows specifying explicit variable types
-from typing import Optional, Union, Tuple
+from typing import Any, Dict, List, Optional, Text, Literal, Union, Tuple
```
andrewelamb, line 7: You don't need to import `Dict` `List` or `Tuple` anymore, you cna just use the lower case verisosn as types `list`, `dict` `tuple`

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
                                 warnings.append(vr_warnings)
         return errors, warnings
 
-    def cross_validation(
-        self,
+    def _parse_validation_log(
+        validation_log: Dict[str, pd.core.series.Series]
+    ) -> tuple([[], [], []]):
```
andrewelamb, line 888: The return type here should be `tuple[list[str], list[str], list[str]]`

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+
+        # Parse invalid_rows and invalid_values so they can be used to raise warnings/errors
+        invalid_rows = iterable_to_str_list(set(validation_rows))
+        invalid_entries = iterable_to_str_list(set(validation_values))
+
+        return invalid_rows, invalid_entries, manifest_ids
+
+    def _merge_format_invalid_rows_values(
+        series_1: pd.core.series.Series, series_2: pd.core.series.Series
+    ) -> tuple([[], []]):
```
andrewelamb, line 915: The return type here should be tuple[list[str], list[str]]

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+            + 2
+        )
+        # Parse invalid_rows and invalid_values so they can be used to raise warnings/errors
+        invalid_rows = np_array_to_str_list(invalid_rows)
+        invalid_entry = iterable_to_str_list(invalid_values.squeeze())
+        return invalid_rows, invalid_entry
+
+    def _format_invalid_row_values(
+        invalid_values: Dict[str, pd.core.series.Series]
+    ) -> tuple([[], []]):
```
andrewelamb, line 944: See above

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+        invalid_rows = np_array_to_str_list(invalid_rows)
+        invalid_entry = iterable_to_str_list(invalid_values)
+        return invalid_rows, invalid_entry
+
+    def _gather_set_warnings_errors(
         val_rule: str,
-        manifest_col: pd.core.series.Series,
-        project_scope: list,
+        source_attribute: str,
+        set_validation_store: list[{}, [], {}],
```
andrewelamb, line 964: `[{}, [], {}]` isn't a valid type. If its always three things as described belowe perhaps it should be a tuple instead of a list.

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
         dmge: DataModelGraphExplorer,
-        access_token: str,
-    ) -> list[list[str]]:
+    ) -> tuple([[], []]):
```
andrewelamb, line 966: See above

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+        """Based on the cross manifest validation rule, and in set rule scope, pass variables to
+        _get_cross_errors_warnings
+            to log appropriate error or warning.
+        Args:
+            val_rule, str: Validation Rule
+            source_attribute, str: Source manifest column name
+            set_validation_store, list[{},[],{}]: contains the missing_manifest_log, present_manifest_log,
```
andrewelamb, line 973: I'm curious why these three things are saved as one variable. The name isn't very helpful.

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+
+    def _get_cross_errors_warnings(
+        val_rule: str,
+        attribute_name: str,
+        dmge: DataModelGraphExplorer,
+        row_num: Optional[list[str]] = None,
+        matching_manifests: Optional[list[str]] = [],
+        manifest_id: Optional[list[str]] = None,
+        invalid_entry: Optional[list[str]] = None,
+    ) -> tuple([[], []]):
```
andrewelamb, line 1044: Same as above

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+        if vr_errors:
+            errors.append(vr_errors)
+        if vr_warnings:
+            warnings.append(vr_warnings)
+        return errors, warnings
+
+    def _gather_value_warnings_errors(
+        val_rule: str,
+        source_attribute: str,
+        value_validation_store: list[{}, {}, {}],
```
andrewelamb, line 1080: Same as above

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+        dmge: DataModelGraphExplorer,
+    ) -> tuple([[], []]):
```
andrewelamb, line 1082: same as above

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+        target_manifest_id: str,
+        missing_manifest_log: Dict[str, pd.core.series.Series],
+        present_manifest_log: Dict[str, pd.core.series.Series],
+        repeat_manifest_log: Dict[str, pd.core.series.Series],
+    ) -> tuple(
+        [
+            Dict[str, pd.core.series.Series],
+            Dict[str, pd.core.series.Series],
+            Dict[str, pd.core.series.Series],
+        ],
```
andrewelamb, line 1150: this should be
```
tuple[
            dict[str, pd.core.series.Series],
            dict[str, pd.core.series.Series],
            dict[str, pd.core.series.Series],
]

File: schematic/models/validate_attribute.py [hunk 13]
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+                        target_manifest[target_attribute],
+                    ],
+                    join="outer",
+                    ignore_index=True,
+                )
+            else:
+                concatenated_target_column = target_manifest[target_attribute]
+            concatenated_target_column = concatenated_target_column.astype("object")
+
+        return concatenated_target_column
```
andrewelamb, line 1241: I'm having a really hard time figuring out what this method is doing. I would suggest some mroe comments.

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+
+    def _run_validation_across_targets_value(
+        manifest_col: pd.core.series.Series,
+        concatenated_target_column: pd.core.series.Series,
+    ) -> tuple([pd.core.series.Series, pd.core.series.Series, pd.core.series.Series]):
```
andrewelamb, line 1246: When typing you don't use the parentheis (generally) just square brackets

andrewelamb:
@mialy-defelice I know this isn't you, but do you knwo why in model/validate_attribute.py. most of the classes don't have the `self` argument? Thes shoudl either have the `self` arguement, get turned into class methods, or get turned into functions at some point.

schematic/models/validate_attribute.py
```diff
@@ -868,186 +883,590 @@ def url_validation(
...
+            # From the concatenated target column, for value scope, run validation
+            (
+                missing_values,
+                duplicated_values,
+                repeat_values,
+            ) = ValidateAttribute._run_validation_across_targets_value(
+                manifest_col=manifest_col,
+                concatenated_target_column=target_column,
+            )
+            validation_store = [missing_values, duplicated_values, repeat_values]
```
andrewelamb, line 1408: I woudl recommend validation_store being a tuple for now inseated of a list. At some point it shoudl probably be it's own class.

schematic/models/validate_manifest.py
```diff
@@ -216,9 +218,12 @@ def validate_manifest_rules(
...
                     # Validate for each individual validation rule.
-                    validation_method = getattr(
-                        ValidateAttribute, validation_types[validation_type]["type"]
-                    )
+                    try:
+                        validation_method = getattr(
+                            ValidateAttribute, validation_types[validation_type]["type"]
+                        )
+                    except:
+                        breakpoint()
```
andrewelamb, line 226: I'm assuming this breakpoint got left in by accident?

mialy-defelice:
> @mialy-defelice I know this isn't you, but do you knwo why in model/validate_attribute.py. most of the classes don't have the `self` argument? Thes shoudl either have the `self` arguement, get turned into class methods, or get turned into functions at some point.

Okay the reason for this is a bit complicated, but `self` is actually in the ValidateManifest class...so when I was trying to use `self` with these it was erroring... but now as I type this, I just realized that validate_attribute doesnt have an `__init__`... I didnt realize that before, I thought it was bc of how the methods were being called... but let me add an init and add self...

schematic/models/validate_manifest.py
mialy-defelice: Yes, I forgot that was there! thank you!

schematic/models/validate_attribute.py
mialy-defelice: Okay, so im storing sets of validation outputs that correspond to the scope (since this is how they are bundled to collect the data and to raise errors later). I did this bc otherwise I would have to either return 3 different variables (in order to still bundle by scope, but not package in a list), or output all 6 then import them where relevant. I did this just to simplify the amount of imports/exports. Its called `set_validation_store` bc its the set scope validation store, there is a `value_validation_store` and they are both output as `validation_store`, but for the import I changed the name so it would be more clear the type of validation store that is expected...
Does that make sense?

mialy-defelice:
Note: Tests failing on dev fail here:
```
FAILED tests/test_api.py::TestSchemaVisualization::test_visualize_tangled_tree_layers[dependency] - assert 500 == 200```

mialy-defelice:
Note as discussed during code review, there are restrictions/assumptions in the code that mean for cross manifest valdiation I need to use the term `match` and and cannot use the term `Unique` for naming the rule. We can look at these assmptions at a later date.

mialy-defelice:
@andrewelamb all your comments have been addressed :)

andrewelamb (approved):

PR #1389, opened 2024-03-20
[bug fix] Fix check_synapse_cache_size function; allow file size to be float

linglp:
## Context
We encountered this error on AWS side (see Jira ticket [here](https://sagebionetworks.jira.com/browse/FDS-1870)) This error was first caught by running schematic profiler, but it is now affecting a lot of endpoints, including manifest/generate, model/submit, /storage/project/datasets and so on on dev instance. The same bug also appears in staging and prod. The only reason that we are not receiving reports about this error from staging/prod is most likely because the cache size has not yet reached "GB" level

## Problem
`convert_gb_to_bytes` has wrong typing, and this line: `size_in_gb = float(size.rstrip("G"))` was changed to `size_in_gb = int(size.rstrip("G"))` which caused the problem. Test also didn't catch this error because it was testing situations where cache size is XXKB, but not when cache size is XXGB. 

## Solution
1. I removed convert_gb_to_bytes function from the code base. The function was doing a simple calculation that I think we don't need a function to do it. This also resolve the "wrong typing" issue of the function since this function no longer exists. 
2. The previous `test_check_synapse_cache_size` function creates a small test file to test `check_synapse_cache_size`. This means that the test was only testing when the cache size is "KB". I updated the test so that test files with various sizes (KB, MB, GB) are all tested. 

## Test
1) Make sure all the current tests are working on GH. 
2) I ran schematic profiler while monitoring AWS logs. I could see purging cache is working: 
Screen Shot 2024-03-21 at 6 20 23 PM
3) Make sure all the tests are returning 200 on schematic profiler: https://www.synapse.org/#!Synapse:syn51385540/tables/query/eyJzcWwiOiJTRUxFQ1QgKiBGUk9NIHN5bjUxMzg1NTQwIiwgImluY2x1ZGVFbnRpdHlFdGFnIjp0cnVlLCAib2Zmc2V0Ijo1NzUsICJsaW1pdCI6MjV9


tests/test_utils.py
```diff
@@ -159,22 +159,37 @@
...
 
+test_disk_storage = [
+    (2, 4000, 16000), 
+    (1000, 4000, 16000), 
+    (2000000, 1900000, 2000000), 
+    (1073741825, 1073741824, 1181116006.4), 
+]
+
+@pytest.fixture()
+def create_temp_query_file(tmp_path, request):
```
linglp, line 170: This function was created to create temporary files of various size. This utilizes: 
* tmp_path: https://docs.pytest.org/en/6.2.x/tmpdir.html which doesn't require clean up 
* using indirect parameterization: https://docs.pytest.org/en/7.1.x/example/parametrize.html#indirect-parametrization

mialy-defelice:
@linglp can you put the ticket number in the branch name so its tracked?

mialy-defelice (approved):

tests/test_utils.py
```diff
@@ -159,22 +161,55 @@
...
+    (2, 4000, 16000),
+    (1000, 4000, 16000),
+    (2000000, 1900000, 2000000),
+    (1073741825, 1073741824, 1181116006.4),
+]
+
+
+# create temporary files with various size based on request
+@pytest.fixture()
+def create_temp_query_file(
```
linglp, line 174: @andrewelamb could you double check the type hinting here? Does this look correct? i have never added type hinting to fixture so I am not so sure.

tests/test_utils.py
```diff
@@ -159,22 +161,55 @@
...
+
+    Args:
+        tmp_path (Path): temporary file path
+        request (any): a request for a fixture from a test
+
+    Yields:
+        Generator[Tuple[Path, Path, Path]]: return path of mock synapse cache directory, mock table query folder and csv
+    """
+    # define location of mock synapse cache
+    mock_synapse_cache_dir = tmp_path / ".synapseCache/"
```
andrewelamb, line 187: What does the "/" operator do here?

tests/test_utils.py
```diff
@@ -7,6 +7,9 @@
 import time
 from datetime import datetime
 from unittest import mock
+from pathlib import Path
+from typing import Union, Tuple, Generator
```
andrewelamb, line 11: You don't need to import `Tuple` anymore, you cna just use `tuple`

tests/test_utils.py
```diff
@@ -159,22 +161,55 @@
...
+    (2000000, 1900000, 2000000),
+    (1073741825, 1073741824, 1181116006.4),
+]
+
+
+# create temporary files with various size based on request
+@pytest.fixture()
+def create_temp_query_file(
+    tmp_path: Path, request: FixtureRequest
+) -> Generator[Tuple[Path, Path, Path], None, None]:
```
andrewelamb, line 176: You cna just use `tuple` here as opposed to `Tuple`

tests/test_utils.py
linglp: Oh I see. Interesting. Thanks for letting me know. I will change to use tuple.

tests/test_utils.py
linglp: Okay I will make that change.

tests/test_utils.py
linglp: It constructs a new path by appending ".synapseCache/" to `tmp_path`. This creates a new pathlib object.

PR #1369, opened 2024-02-13
Enable single rule disjunction

GiaJordan:
## Changelog
This PR enables single rule disjunction for the existing validation rules through the `IsNA` rule modifier. It can be used with each rule except the `recommended` rule by including it in the specification for validation rules as another rule (ie with the `::` operator).

When in use, Validation rules will raise an error or warning or no message as appropriate for each metadata entry that is provided, and will not raise any message for metadata entries that are omitted. The attribute should be marked as `FALSE` for the `REQUIRED` column in the data model to use this modifier.

A data model specific to these requirements has been created to monitor the status of each rule combined with this modifier and tests have been created as well.

- - - 

## Note for Reviewers:
The line endings for some files were changed and so the diffs can show entire file changes as part of this PR. Before you begin reviewing, under `Files changed>Settings` toggle `Hide Whitespace` to ON to get an accurate record of what actual code was changed.
image

PR #1390, opened 2024-03-25
Allow non-entry if Required=False

mialy-defelice:
### Note

This is a variation on another [PR enable Single Rule Disjunction.](https://github.com/Sage-Bionetworks/schematic/pull/1369).
Talked with Amy and Milen and decided to proceed with this version.

### Background

Users were experiencing an issue where when they used validation rules, the `Required` column entries were no longer working as expected.

If `Required` was set to `False` then a warning would appear if a user did not submit a value. If the user tried to set the messaging to `error` level, it would error out if a user did not submit a value, even though values are not required.

This PR fixes this issue. In this version, the expected behavior is restored and for attributes with validation rules, if `Required=False` and data is not provided then no error or warning is raised. But errors or warnings are still raised if incorrect data _is_ entered.

The key difference withthis PR and 1369 is that one does not need to add the `isNA` valdation rule to every single rule where you want non-values to be allowed, rather this is now the expected behavior. If you want to make it mandatory to add values, then you would set `Required=True`. Note, the `IsNA` rule still exists for the occasions you want a user to be able to enter the string `Not Applicable`.

### Testing

To unit the unit tests, I added a new version of the `example.model.jsonld` where the MockComponent Template attributes are set to `Required=False`. I then made a new mock manifest where I do not provide values for at least one row for each attribute column. In this set of cases, the validation should all still pass. 

To test on your own you can inspect these files then run them with the `validate` command. Can also make your own manifests and augment the data model to work as expected.

Let me know if there are any questions!


mialy-defelice:
@andrewelamb @linglp 
This PR is an alternative to the Single Rule Disjunction PR. 
It should be a simpler solution. I can speak more to the differences in our next code review or in person during home week :)

mialy-defelice:
Putting back to draft while I fix up some edge cases and add some additional testing.

mialy-defelice:
Note: Tests failing on dev fail here:
```
FAILED tests/test_api.py::TestSchemaVisualization::test_visualize_tangled_tree_layers[dependency] - assert 500 == 200```

**Comment by mialy-defelice** [comment]
Note, need to check with conditional requirements, to test if required checks are working properly.

**Comment by mialy-defelice** [comment]
@linglp @andrewelamb you both can start review if you want, but I realized the checks for if a column are required are too simple, I need to take conditionality into account. 

I will need time to update this PR. Moving back to draft for now.

**Comment by mialy-defelice** [comment]
@andrewelamb and @linglp Okay...It was a simpler fix than I initially thought it would be. Moving back to ready to review :)

File: schematic/models/validate_attribute.py [hunk 1]
```diff
@@ -37,7 +38,7 @@ def generate_schema_error(
         row_num: str,
         attribute_name: str,
         error_message: str,
-        invalid_entry: str,
+        invalid_entry,
```
andrewelamb, line 41: You could put `typing.Any` for these

andrewelamb:
@mialy-defelice I know this was not part of this PR, but do you know why the  `models.validate_attribute.GenerateError` is missing the self arguemtn for its methods?

schematic/models/validate_attribute.py
```diff
@@ -378,11 +385,148 @@ def generate_content_error(
             error_row=error_row,
             error_col=attribute_name,
             error_message=error_message,
-            error_val=error_val,
+            error_val=invalid_entry,
         )
 
         return error_list, warning_list
 
+    def _get_rule_attributes(
```
andrewelamb, line 393: This is missing a self argument. (Or it should be a class emthod or a function)

schematic/models/validate_attribute.py
```diff
@@ -378,11 +385,148 @@ def generate_content_error(
...
+        val_rule: str, error_col_name: str, dmge: DataModelGraphExplorer
+    ) -> tuple[[str, str, str]]:
```
andrewelamb, line 395: The return statement is returnign 6 things, but this shows only 3. Also the double brackets are uneeded.

schematic/models/validate_attribute.py
```diff
@@ -378,11 +385,148 @@ def generate_content_error(
...
+        """Extract different attributes from the given rule
+        Args:
+            val_rule, str:
+            error_col_name, str
+            dmge, DataModelGraphExplorer Object
+        Returns:
+        """
```
andrewelamb, line 402: This docstring is incomplete

schematic/models/validate_attribute.py
```diff
@@ -378,11 +385,148 @@ def generate_content_error(
...
+        return (
+            rule_parts,
+            rule_name,
+            specified_level,
+            is_schema_error,
+            col_is_recommended,
+            col_is_required,
+        )
+
+    def _get_is_na_allowed(
```
andrewelamb, line 422: See above about missing self argument

schematic/models/validate_attribute.py
```diff
@@ -378,11 +385,148 @@ def generate_content_error(
...
+            node_display_name=node_display_name
+        )
+
+        # Determine if IsNA is one of the rules, if it is return True, else return False
+        if rule_in_rule_list("IsNA", validation_rule_list):
+            return True
+        else:
+            return False
+
+    def _get_error_value_is_na(
```
andrewelamb, line 443: Same as above

schematic/models/validate_attribute.py
```diff
@@ -378,11 +385,148 @@ def generate_content_error(
...
+        elif (
+            (error_val is None)
+            or pd.isnull(error_val)
+            or (error_val == "<NA>" and not col_is_required)
+        ):
+
+        """
+        not_applicable_strings = [
+            "not applicable",
+        ]
```
andrewelamb, line 461: Is this meant to be added to at a later time?

schematic/models/validate_attribute.py
```diff
@@ -524,6 +661,33 @@ class ValidateAttribute(object):
     def __init__(self, dmge: DataModelGraphExplorer) -> None:
         self.dmge = dmge
 
+    def get_no_entry(self, entry, node_display_name):
```
andrewelamb, line 664: Missing a docstring

schematic/models/validate_attribute.py
```diff
@@ -524,6 +661,33 @@ class ValidateAttribute(object):
...
+        # check if <NA> in list let pass if not required and no value is recorded.
+        no_entry = False
+        # if not col_is_required:
+        na_allowed = GenerateError._get_is_na_allowed(
+            node_display_name=node_display_name, dmge=self.dmge
+        )
+        value_is_na = GenerateError._get_error_value_is_na(
```
andrewelamb, line 671: If you are usign these methods from a different class, they shouldn't be priavte

schematic/models/validate_attribute.py
```diff
@@ -524,6 +661,33 @@ class ValidateAttribute(object):
...
+            entry,
+            na_allowed,
+        )
+        if value_is_na:
+            no_entry = True
+
+        return no_entry
+
+    def get_entry_has_value(
```
andrewelamb, line 680: missing docstring

schematic/models/validate_attribute.py
```diff
@@ -594,9 +758,19 @@ def list_validation(
...
             for i, list_string in enumerate(manifest_col):
-                if not re.fullmatch(csv_re, list_string):
+                list_error = None
+                entry_has_value = self.get_entry_has_value(
+                    entry=list_string,
+                    node_display_name=manifest_col.name,
+                )
+
+                if not isinstance(list_string, str) and entry_has_value:
+                    list_error = "not_a_string"
```
andrewelamb, line 772: Whats is the purpose of this if block? `list_error` only gets used in the next elif block

schematic/models/validate_attribute.py
```diff
@@ -594,9 +758,19 @@ def list_validation(
             list_robustness = "strict"
 
         if list_robustness == "strict":
+            manifest_col = manifest_col.astype(str)
```
andrewelamb, line 761: What if `list_robustness != "strict"` `manifest_col` has no value and line 790 will trigger an error

schematic/models/validate_attribute.py
```diff
@@ -857,26 +1039,57 @@ def url_validation(
...
+                    if not urlparse(url).scheme:
+                        url = "http://" + url
+                    try:
+                        # Check that the URL points to a working webpage
+                        # if not log the appropriate error.
+                        request = Request(url)
+                        response = urlopen(request)
+                        valid_url = True
+                        response_code = response.getcode()
+                    except:
```
andrewelamb, line 1053: Do you want to catch every exception here?

schematic/models/validate_attribute.py
```diff
@@ -857,26 +1039,57 @@ def url_validation(
...
+                    # add scheme to the URL if not currently added.
+                    if not urlparse(url).scheme:
+                        url = "http://" + url
+                    try:
+                        # Check that the URL points to a working webpage
+                        # if not log the appropriate error.
+                        request = Request(url)
+                        response = urlopen(request)
+                        valid_url = True
+                        response_code = response.getcode()
```
andrewelamb, line 1052: `response_code` doesn't appear to be used

schematic/models/validate_attribute.py
```diff
@@ -1027,6 +1240,47 @@ def _gather_set_warnings_errors(
 
         return errors, warnings
 
+    def _remove_non_entry_from_invalid_entry_list(
+        self,
+        invalid_entry: Optional[list[str]],
+        row_num: Optional[list[str]],
+        attribute_name: str,
+    ):
```
andrewelamb, line 1248: Could you add the return type here?

schematic/models/validate_attribute.py
```diff
@@ -1027,6 +1240,47 @@ def _gather_set_warnings_errors(
...
+                    idx_to_remove.append(idx)
+
+            # If indices are recorded for NA values to remove, remove them and the corresponding row
+            if idx_to_remove:
+                for idx in sorted(idx_to_remove, reverse=True):
+                    del invalid_entry[idx]
+                    del row_num[idx]
+                # Perform check to make sure length of invalid_entry and row_num is the same. If not that would suggest
+                # there was an issue recording or removing values.
+                if len(invalid_entry) != len(row_num):
```
andrewelamb, line 1277: It says `row_num` is optional but it appears the method is assumign its a list

schematic/models/validate_attribute.py
mialy-defelice: GenerateError does not have an init, so its not called for here. If you want though, I can refactor GenerateError to add an init.

schematic/models/validate_attribute.py
```diff
@@ -378,11 +385,148 @@ def generate_content_error(
...
+        """Determine if the erroring value is NA
+        Args:
+            error_val: erroneous value
+        Returns:
+            bool: Returns True, if the error value is evaluated to be NA, and False if not
+        elif (
+            (error_val is None)
+            or pd.isnull(error_val)
+            or (error_val == "<NA>" and not col_is_required)
+        ):
```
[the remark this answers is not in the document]
mialy-defelice: Ah yes, I stashed it there when making some changes, so could see it easily if my new solution didnt work.. Ill delete it :)

schematic/models/validate_manifest.py
```diff
@@ -65,6 +65,30 @@ def get_multiple_types_error(
             error_val = f"Multiple Rules: list not first"
         return ["NA", error_col, error_message, error_val]
 
+    def check_max_rule_num(
+        self, validation_rules: str, col: pd.core.series.Series, errors: list[list[str]]
+    ) -> list[list[str]]:
+        """
+        validation_rules, str:
+        errors, list:
+        """
```
andrewelamb, line 74: Incomplete docsring

schematic/models/validate_attribute.py
mialy-defelice: Im not sure, possibly? I just moved it around, but it was in another function previously. I did a tiny little refactor of some functions.

schematic/models/validate_manifest.py
```diff
@@ -65,6 +65,30 @@ def get_multiple_types_error(
...
+        # Check that attribute rules conform to limits:
+        # IsNa is operates differently than most rules, do not consider it as a rule for evaluating
+        # if the number of rule pairs has been exceeded.
+        combined_rules = validation_rules.copy()
```
andrewelamb, line 78: why make a copy here?

schematic/models/validate_manifest.py
```diff
@@ -65,6 +65,30 @@ def get_multiple_types_error(
             error_val = f"Multiple Rules: list not first"
         return ["NA", error_col, error_message, error_val]
 
+    def check_max_rule_num(
+        self, validation_rules: str, col: pd.core.series.Series, errors: list[list[str]]
+    ) -> list[list[str]]:
+        """
+        validation_rules, str:
+        errors, list:
```
andrewelamb, line 73: incomplete docstring

schematic/models/validate_manifest.py
```diff
@@ -65,6 +65,30 @@ def get_multiple_types_error(
             error_val = f"Multiple Rules: list not first"
         return ["NA", error_col, error_message, error_val]
 
+    def check_max_rule_num(
+        self, validation_rules: str, col: pd.core.series.Series, errors: list[list[str]]
```
andrewelamb, line 69: I think validation rules should be a `list[str]` given how its being copied and having items removed from it

schematic/models/validate_attribute.py
andrewelamb: Oh its another one of these. Don't worry about it for now then. It really doesn't make sense as a class if it has no state, and there's nothing to initialize. Thsi shoudl probably be refactor into util functions.

schematic/models/validate_attribute.py
andrewelamb: If not line 465 could be
`error_val_is_na = error_val.lower() == "not applicable"`

schematic/models/validate_attribute.py
```diff
@@ -803,46 +1008,23 @@ def url_validation(
...
+            if entry_has_value:
+                # Check if a random phrase, string or number was added and
+                # log the appropriate error.
+                if not isinstance(url, str) or not (
+                    urlparse(url).scheme
+                    + urlparse(url).netloc
+                    + urlparse(url).params
+                    + urlparse(url).query
+                    + urlparse(url).fragment
+                ):
```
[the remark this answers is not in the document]
mialy-defelice: Dont understand in terms of URL parse? Or checking if the entry has a value?

schematic/models/validate_attribute.py
mialy-defelice: Just to note, i didnt write this section, Its appearing as a change bc I tab indented it bc I added an if. My guess is that if getting the response code doesnt work the idea is to raise an error.

schematic/models/validate_attribute.py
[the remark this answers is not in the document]
andrewelamb: It's the

( urlparse(url).scheme
                    + urlparse(url).netloc
                    + urlparse(url).params
                    + urlparse(url).query
                    + urlparse(url).fragment
)

What are the `+` doing here?

schematic/models/validate_attribute.py
mialy-defelice: I could make a ticket for a light refactor. I think we could probably put the DMGE an init, or it could be util...

schematic/models/validate_attribute.py
mialy-defelice: Im going to leave it for now, bc there is actually another ticket involving valid values allowing the string 'None'. We might need to add it here.

schematic/models/validate_attribute.py
mialy-defelice: I think its not meant to be used, but rather used to catch errors. if there is an error generating the response code, then the except will raise an error.

schematic/models/validate_attribute.py
andrewelamb: Up to you!

schematic/models/validate_attribute.py
mialy-defelice: manifest_col is imported, if the list robustness is strict its just converted here to type str for processing within this if.

schematic/models/validate_attribute.py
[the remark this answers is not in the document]
mialy-defelice: It looks like its a string concatenation, that would effectively rebuild a URL. Its looking to see if the parts to a URL can be extracted without requiring specific parts. So if the URL is not a string, and its not able to be parsed into any part, then its assumed the entry by the user is "random"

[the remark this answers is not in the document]
mialy-defelice: But I'll add that as a docstring.

schematic/models/validate_attribute.py
andrewelamb: Oh wait yes, disregard this comment :)

schematic/models/validate_attribute.py
mialy-defelice: I added a check for row_num as I am checking for invalid_entry

schematic/models/validate_manifest.py
mialy-defelice: Youre right!

mialy-defelice:
@andrewelamb All your comments should be addressed :)

andrewelamb:
@mialy-defelice Sounds good! I'll go ahead and approve

andrewelamb (approved):

PR #1391, opened 2024-03-25
Fds 1796 schema convert export pickle

afwillia:
The intent of this PR is to allow the `convert` CLI command to export the graph. It creates a new option, `output_type` with the values `jsonld`, `graph`, or `all`. `graph` exports the graph as a pickle, `jsonld` exports schema as a jsonld, and `all` does both.

Updated the CLI tests to ensure the file written by `convert` is read correctly.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -3,7 +3,6 @@ import os
 import pytest
 import pickle
 import json
-import tempfile
 
 from click.testing import CliRunner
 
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -46,6 +48,8 @@ def assert_expected_file(self, result, output_path):
 
     def test_schema_convert_cli(self, runner, helpers):
         data_model_csv_path = helpers.get_data_path("example.model.csv")
+        data_model_pickle_path = helpers.get_data_path("example.model.pickle")
```
mialy-defelice, line 51: For these I would try to find a way to supply these parameters via parmeterize, you can send the models via one parameter and the output type parameter via another. 
Then you could call the appropriate runner, based on your current parameters, do some checks to see where you care and call the appropriate asserts. Does that make sense?

mialy-defelice (changes requested):
Looks good so far, I requested some changes in testing and  I will test for function tomorrow :)

mialy-defelice:
@afwillia Also note your tests are failing only bc of black needing to be run on a couple files, if you run black and re-commit, it should get further :)

tests/test_cli.py
afwillia: yeah, this makes sense. I'll play around with it!

PR #1394, opened 2024-04-01
Add synapse cache as a parameter to SynapseStorage

andrewelamb:
- Fixes [FDS-1446]
- Adds synapse cache path as parameter to SynapseStorage class
- caceh path is used to create Synapse Clinet object on login

[FDS-1446]: https://sagebionetworks.jira.com/browse/FDS-1446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/test_store.py
```diff
@@ -92,9 +93,14 @@ def test_init(self):
 
 
 class TestSynapseStorage:
-    def test_init(self, synapse_store):
+    def test_init(self, synapse_store:SynapseStorage) -> None:
         assert synapse_store.storageFileview == "syn23643253"
         assert isinstance(synapse_store.storageFileviewTable, pd.DataFrame)
+        assert synapse_store.root_synapse_cache == ".synapseCache"
+
+    def test__purge_synapse_cache(self, synapse_store:SynapseStorage) -> None:
```
linglp, line 101: I think this is testing if the `purge_synapse_cache` function can be called. But this is not testing if the `purge_synapse_cache`  function is working as expected. Could you add a test to test if `_purge_synapse_cache` is working as expected?

schematic/store/synapse.py
```diff
@@ -268,30 +270,53 @@ def _query_fileview(self):
             raise AccessCredentialsError(self.storageFileview)
 
     @staticmethod
-    def login(token=None, access_token=None):
+    def login(
+        synapse_cache_path: str = ".synapseCache",
```
linglp, line 274: same as above

schematic/store/synapse.py
```diff
@@ -202,12 +203,14 @@ def __init__(
         token: Optional[str] = None,  # optional parameter retrieved from browser cookie
         access_token: Optional[str] = None,
         project_scope: Optional[list] = None,
+        synapse_cache_path: str = ".synapseCache",
```
linglp, line 206: This will change the current default behavior. By default, synapse python client set the cache directory in root. For example, for me, the cache directory is: `/Users/lpeng/.synapseCache`. But here, if you set synapse_cache_path default as `.synapseCache`, then it will change the cache directory as the same folder as I am running schematic. The change of `.synapseCache` directory doesn't really matter for me, but I think it might impact external users. I don't really see a compelling reason to change the default cache directory (I think our goal is just to make the .synapseCache folder location configuration). Can we observe the current behavior?

linglp (changes requested):

.gitignore
```diff
@@ -177,6 +177,8 @@ tests/data/schema.gpickle
 # Created during testting
 Example*
 manifests/*
+test_cache_dir
```
linglp, line 180: when will test_cache_dir get created? If it is created by test, it should be cleaned up after the test is done?

andrewelamb: This now gets deleted in the `login` test

schematic/store/synapse.py
```diff
@@ -268,30 +270,53 @@ def _query_fileview(self):
             raise AccessCredentialsError(self.storageFileview)
 
     @staticmethod
-    def login(token=None, access_token=None):
+    def login(
```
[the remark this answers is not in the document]
andrewelamb: There's now a test for the login method

.gitignore
linglp: should this line be removed?

tests/test_store.py
```diff
@@ -92,9 +94,25 @@ def test_init(self):
...
         assert synapse_store.storageFileview == "syn23643253"
         assert isinstance(synapse_store.storageFileviewTable, pd.DataFrame)
+        assert synapse_store.root_synapse_cache.endswith(".synapseCache")
+
+    def test__purge_synapse_cache(self, synapse_store:SynapseStorage) -> None:
+        """Tests SynapseStorage._purge_synapse_cache"""
+        size_before_purge = check_synapse_cache_size(synapse_store.root_synapse_cache)
+        synapse_store._purge_synapse_cache(maximum_storage_allowed_cache_gb=0.000001, minute_buffer=0)
+        size_after_purge = check_synapse_cache_size(synapse_store.root_synapse_cache)
+        assert size_before_purge > size_after_purge
```
linglp, line 107: I just ran this and saw that over 2000 files in my cache folder got deleted.... this is a bit scary.. 😦 I think here you should create a mock .synapseCache folder (and clean up after test is done) so that the tests won't actually modify the real .synapseCache folder stored in our root.

linglp (changes requested):

linglp:
@andrewelamb could you also merge the latest develop into your PR and test again? I saw that the error in `check_synapse_cache_size` function has not yet gotten fixed in your PR, but it has been fixed with the latest develop

.gitignore
andrewelamb: removed

schematic/store/synapse.py
```diff
@@ -241,7 +240,10 @@ def _purge_synapse_cache(self, maximum_storage_allowed_cache_gb=1):
             dir_size_bytes = check_synapse_cache_size(directory=self.root_synapse_cache)
             # if 1 GB has already been taken, purge cache before 15 min
             if dir_size_bytes >= maximum_storage_allowed_cache_bytes:
-                num_of_deleted_files = clear_synapse_cache(self.syn.cache, minutes=15)
+                logging.warning("TEST")
```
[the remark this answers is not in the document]
andrewelamb: removed

andrewelamb:
@linglp I did the requsted fixes and merged in develop

.gitignore
```diff
@@ -177,6 +177,7 @@ tests/data/schema.gpickle
 # Created during testting
 Example*
 manifests/*
+.synapseCache
```
linglp, line 180: @andrewelamb why do we need this?

linglp (approved):
Thanks for your work Andrew. I approve for now. But if the line in git ignore is not needed, please remove it. Thank you

PR #1396, opened 2024-04-03
Fds 1796 schema convert export pickle

afwillia:
Identical PR to [1391](https://github.com/Sage-Bionetworks/schematic/pull/1391), except this is from a branch in the sage-bionetworks schematic repo, not my fork. Hoping this makes the tests that require credentials pass.


tests/test_cli.py
```diff
@@ -44,24 +46,29 @@ def assert_expected_file(self, result, output_path):
...
-
-        label_type = 'class_label'
+    @pytest.mark.parametrize(
+        "model, output, expected",
+        [
+            ("tests/data/example.model.csv", "tests/data/example.model.pickle", 0),
+            ("tests/data/example.model.csv", "tests/data/example.model.jsonld", 0),
+        ],
+    )
+    def test_schema_convert_cli(self, runner, helpers, model, output, expected):
```
mialy-defelice, line 56: can you also test the situation where you are explicitly passing the output_type parameter instead of using the path to infer the output type?
You could call parameterize several times:
Once pass just the example model
Next pass both `--output_jsonld` with different paths
Next pass` --output_type`, and account for all output types.

Doing it this way will test a bunch of combinations.

Then you could call the runner in different combos
1. with `output_jsonld`
2. with `output_type`
3. with both

What would you expect to happen if someone passed a `.pickle` `--output_jsonld` but then said the output type was graph? Vice versa?

I know this might be a lot, we could leave it for now and create an additional ticket.

mialy-defelice (changes requested):
Going to do functional testing now :)

tests/test_cli.py
mialy-defelice: Would also want to pass none for these variables to ensure the defaults work as expected

schematic/schemas/commands.py
```diff
@@ -105,22 +115,38 @@ def convert(schema, data_model_labels, output_jsonld):
                 for w in war:
                     logger.warning(w)
 
+    if output_jsonld is None:
```
mialy-defelice, line 118: These need to be updated to look for `.pickle` as well. For now if you put a `.pickle` file extension, for `output_jsonld` it will not be removed, so the resultant file will have two `.pickle` extensions.

schematic/schemas/commands.py
```diff
@@ -105,22 +115,38 @@ def convert(schema, data_model_labels, output_jsonld):
...
+        "By default, the JSON-LD output will be stored alongside the first "
+        f"input CSV or JSON-LD file. In this case, it will appear here: '{output_jsonld}'. "
+        "You can use the `--output_jsonld` argument to specify another file path."
+    )
+
+    if output_type in ["graph", "all"]:
+        logger.info("Export graph to pickle if requested")
+        output_graph = output_file_no_ext + ".pickle"
+        try:
+            with open(output_graph, "wb") as file:
```
mialy-defelice, line 133: Following convention can you move the pickle dump to `utils/schema_utils.py`, you can call it something like `export_graph` to mimic `export_schema` that is used to export the JSONLD.

schematic/schemas/commands.py
```diff
@@ -56,7 +57,16 @@ def schema():  # use as `schematic model ...`
     metavar="<OUTPUT_PATH>",
     help=query_dict(schema_commands, ("schema", "convert", "output_jsonld")),
 )
-def convert(schema, data_model_labels, output_jsonld):
+@click.option(
```
mialy-defelice, line 60: For backwards and forwards compatibility can we let --output_jsonld be named --output_path as well? can you look if thats possible? We could then depreciate --output_jsonld over time.

mialy-defelice (changes requested):

mialy-defelice:
I checked functionality and it looked good, except as noted above with the `.pickle` naming.

Saved the pickle and opened it and it looked good :)

tests/test_cli.py
afwillia: @mialy-defelice took me a minute to figure out parametrize, but how does this look? It's still a WIP since I haven't implemented all the scenarios you mentioned, but I want to make sure this is the structure you imagined.

mialy-defelice: So yes, it looks right in that youre calling parameterize twice, but I would drop some of the tuples, bc some of the data repeats you dont need it:
```
    @pytest.mark.parametrize(
        "output_path",
        [
            "tests/data/example.model.pickle",
            "tests/data/example.model.jsonld",
        ],
        ids=["output_path_pickle", "output_path_jsonld"],
    )
    @pytest.mark.parametrize(
        "output_type",
        [
            "jsonld",
            "graph",
            "all"
        ],
        ids=["output_type_jsonld", "output_type_graph", "output_type_all"],
    )
    def test_schema_convert_cli(self, runner, output_path, output_type):
        model = "tests/data/example.model.csv"
        expected = 0
        ```

**Review by mialy-defelice — APPROVED** [review]

File: schematic/schemas/commands.py [hunk 5]
```diff
@@ -114,35 +126,48 @@ def convert(
...
+        "You can use the `--output_jsonld` argument to specify another file path."
+    )
+
+    if output_type in ["graph", "all"]:
+        logger.info("Export graph to pickle.")
+        output_graph = output_file_no_ext + ".pickle"
+        try:
+            export_graph(graph_data_model, output_graph)
+        except SystemExit as error:
+            raise error
```
andrewelamb, line 149: This try/except block doesn't seem to do anything. Maybe I don't understand?

schematic/schemas/commands.py
```diff
@@ -114,35 +126,48 @@ def convert(
...
+
+    if output_type == "graph":
+        click.echo(f"Graph created {output_graph}")
+        return 0
+
```
andrewelamb, line 154: I don't think I understand the logic behind the `output_type` parameter

schematic/schemas/commands.py
```diff
@@ -59,9 +59,21 @@ def schema() -> None:  # use as `schematic model ...`
...
+    help=query_dict(schema_commands, ("schema", "convert", "output_type")),
+)
 def convert(
-    schema: Any, data_model_labels: DisplayLabelType, output_jsonld: Optional[str]
-) -> None:
+    schema: Any,
+    data_model_labels: DisplayLabelType,
+    output_jsonld: Optional[str],
+    output_type: Optional[str],
+    output_path: Optional[str],
```
GiaJordan, line 75: I'd suggest removing `output_jsonld` as a parameter and use only `output_path` in combination with `output_type` with the latter controlling the extension

tests/test_cli.py
```diff
@@ -46,32 +48,117 @@ def assert_expected_file(self, result, output_path):
...
+            schema,
+            [
+                "convert",
+                model,
+                "--output_type",
+                output_type
+            ]
+        )
+
+        assert resultThree.exit_code == expected
```
GiaJordan, line 105: check for the creation of the file as an asssertion in addition to getting the expected output. you can also parse the file and check its contents here. Delete after the check

tests/test_cli.py
```diff
@@ -46,32 +48,117 @@ def assert_expected_file(self, result, output_path):
         except:
             pass
 
-    def test_schema_convert_cli(self, runner, helpers):
-        data_model_csv_path = helpers.get_data_path("example.model.csv")
+    @pytest.mark.parametrize(
+        "output_path",
+        [
+            "tests/data/example.model.pickle",
```
GiaJordan, line 54: Specify a test specific path for the files used in this test...

schematic/utils/schema_utils.py
```diff
@@ -500,3 +501,20 @@ def get_json_schema_log_file_path(data_model_path: str, source_node: str) -> str
...
+    """Write object to a pickle file.
+    Args:
+        schema, dict: A data model graph to export
+        file_path, str: File to create
+    """
+    try:
+        with open(file_path, "wb") as file:
+            pickle.dump(schema, file)
+        logger.info(f"The graph was created and saved to '{file_path}'.")
+    except SystemExit as error:
```
thomasyu888, line 516: nit: Are there other errors we need to look out for? or is this the only one

tests/test_cli.py
```diff
@@ -46,32 +48,117 @@ def assert_expected_file(self, result, output_path):
...
+    @pytest.mark.parametrize(
+        "output_type",
+        [
+            "jsonld",
+            "graph",
+            "all"
+        ],
+        ids=["output_type_jsonld", "output_type_graph", "output_type_all"]
+    )
+    def test_schema_convert_cli(self, runner, output_path, output_type):
```
thomasyu888, line 68: Nit: These are different test cases, per set of parameters.  I'm not certain it's required to split each parameter setting into separate tests, but it would make it more readable.

schematic/utils/schema_utils.py
```diff
@@ -500,3 +501,20 @@ def get_json_schema_log_file_path(data_model_path: str, source_node: str) -> str
         prefix = prefix_root
     json_schema_log_file_path = f"{prefix}.{source_node}.schema.json"
     return json_schema_log_file_path
+
+
+def export_graph(schema: dict, file_path: str) -> None:
```
thomasyu888, line 506: Please add a test for this function

schematic/schemas/commands.py
```diff
@@ -114,35 +126,48 @@ def convert(
                 for warning in war:
                     logger.warning(warning)
 
+    if output_path:
+        output_jsonld = output_path
+
+    if output_jsonld is None:
+        output_file_no_ext = re.sub("[.](jsonld|csv|pickle)$", "", schema)
```
BryanFauble, line 133: nit: Try to include full words in the name of variables. I assume `ext` here means `extension`, but I wouldn't have a clue if I am not already looking at how it's created.

schematic/schemas/commands.py
```diff
@@ -114,35 +126,48 @@ def convert(
...
+        output_file_no_ext = re.sub("[.](jsonld|csv|pickle)$", "", output_jsonld)
+
+    logger.info(
+        "By default, the JSON-LD output will be stored alongside the first "
+        f"input CSV or JSON-LD file. In this case, it will appear here: '{output_jsonld}'. "
+        "You can use the `--output_jsonld` argument to specify another file path."
+    )
+
+    if output_type in ["graph", "all"]:
+        logger.info("Export graph to pickle.")
```
BryanFauble, line 144: Look to include more context within log messages. Such as: What graph was exported to pickle?

schematic/schemas/commands.py
```diff
@@ -114,35 +126,48 @@ def convert(
...
+        output_graph = output_file_no_ext + ".pickle"
+        try:
+            export_graph(graph_data_model, output_graph)
+        except SystemExit as error:
+            raise error
+
+    if output_type == "graph":
```
BryanFauble, line 151: nit: This if block can be within the scope of `if output_type in ["graph", "all"]:`

schematic/schemas/commands.py
```diff
@@ -59,9 +59,21 @@ def schema() -> None:  # use as `schematic model ...`
...
+    default="jsonld",
+    help=query_dict(schema_commands, ("schema", "convert", "output_type")),
+)
 def convert(
-    schema: Any, data_model_labels: DisplayLabelType, output_jsonld: Optional[str]
-) -> None:
+    schema: Any,
+    data_model_labels: DisplayLabelType,
+    output_jsonld: Optional[str],
+    output_type: Optional[str],
```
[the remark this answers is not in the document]
BryanFauble: Personally, I really like the use of ENUMs that distill down to strings. This post covers it a bit:

https://stackoverflow.com/questions/58608361/string-based-enum-in-python

By using an ENUM you don't need to use strings for your conditional checks in the code. Basically, you'd accept an ENUM here then use the Enum class in your compares.


Here is a small snippet that shows what this looks like:

```
from enum import Enum


class OutputType(str, Enum):
    jsonld = 'jsonld'
    graph = 'graph'
    all = 'all'

    def __str__(self) -> str:
        # See https://peps.python.org/pep-0663/
        return self.value


def convert(value: OutputType) -> None:
    if value == OutputType.jsonld:
        print(f'Found the enum: {value}')
        print(value)
    elif value == OutputType.graph:
        print(f'Found the enum: {value}')
        print(value)
    elif value == OutputType.all:
        print(f'Found the enum: {value}')
        print(value)
    else:
        print(f"Invalid enum value: {value}")


convert(OutputType.jsonld)
convert(OutputType.graph)
convert(OutputType.all)
convert('jsonld')
convert('graph')
convert('all')
```

schematic/utils/schema_utils.py
```diff
@@ -500,3 +501,20 @@ def get_json_schema_log_file_path(data_model_path: str, source_node: str) -> str
...
+    Args:
+        schema, dict: A data model graph to export
+        file_path, str: File to create
+    """
+    try:
+        with open(file_path, "wb") as file:
+            pickle.dump(schema, file)
+        logger.info(f"The graph was created and saved to '{file_path}'.")
+    except SystemExit as error:
+        logger.error(
```
BryanFauble, line 517: Use `logger.exception` here instead of `.error`. By using `.exception` it'll automatically add in the stack trace information to the log message.

schematic/schemas/commands.py
afwillia: I remember Mialy mentioning keeping this for backwards compatibility for now. Can we deprecate it in a future release?

thomasyu888:
Closing due to other PR

PR #1398, opened 2024-04-05
Cross Manifest Validation: Allow first manifest to submit, if no target manifest to validate against.

mialy-defelice:
[FDS-1603](https://sagebionetworks.jira.com/browse/FDS-1603)

This PR fixes the issue where users were unable to submit a manifest with Cross Manifest validation rules (and error logging) if a manifest was not already uploaded to synapse.

In this PR if a project does not contain manifests with target columns to validate against, then submission is allowed to proceed. A warning is raised to let users know that there is not a target manifest to validate against.

Tests were added to check that validation is passing as expected.

[FDS-1603]: https://sagebionetworks.jira.com/browse/FDS-1603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


schematic/models/validate_attribute.py
```diff
@@ -390,6 +390,73 @@ def generate_content_error(
 
         return error_list, warning_list
 
+    def generate_no_cross_warning(
+        dmge: DataModelGraphExplorer, attribute_name: str, val_rule: str
+    ) -> list[list[str]]:
```
andrewelamb, line 395: This should be tuple[list[str], list[str]]

schematic/models/validate_attribute.py
```diff
@@ -390,6 +390,73 @@ def generate_content_error(
...
+            error_row=None,
+            error_col=attribute_name,
+            error_message=error_message,
+            error_val="No Invalid Entry Recorded",
+        )
+        if nv_errors:
+            errors.append(nv_errors)
+        if nv_warnings:
+            warnings.append(nv_warnings)
+        return errors, warnings
```
andrewelamb, line 458: I don't understand why the return isn't just:
`return nv_errors, nv_warnings`
?

schematic/models/validate_attribute.py
```diff
@@ -1397,6 +1467,27 @@ def _gather_value_warnings_errors(
             )
         return errors, warnings
 
+    def _check_if_target_manifest_is_empty(
+        self,
+        target_manifest: pd.core.series.Series,
+        target_manifest_empty: list[bool],
+        column_names: dict[str, str],
+    ):
```
andrewelamb, line 1475: Missing docstring

schematic/models/validate_attribute.py
```diff
@@ -1397,6 +1467,27 @@ def _gather_value_warnings_errors(
...
+        )
+
+        # Check if the entire target_manifest is empty
+        if sum(target_manifest_dupe.isnull().values.all(axis=0)) == len(
+            target_manifest_dupe.columns
+        ):
+            target_manifest_empty.append(True)
+        else:
+            target_manifest_empty.append(False)
+        return target_manifest_empty
```
andrewelamb, line 1489: Why add the result to the list and not just return a bool?

schematic/models/validate_attribute.py
```diff
@@ -1408,10 +1499,15 @@ def _run_validation_across_targets_set(
...
     ) -> tuple[
-        dict[str, pd.core.series.Series],
-        dict[str, pd.core.series.Series],
-        dict[str, pd.core.series.Series],
+        tuple[
+            dict[str, pd.core.series.Series],
+            dict[str, pd.core.series.Series],
+            dict[str, pd.core.series.Series],
+            bool,
+        ]
```
andrewelamb, line 1510: This doesn't match what's being returned.

schematic/models/validate_attribute.py
```diff
@@ -1462,15 +1568,25 @@ def _run_validation_across_targets_set(
                     # If there are missing values in the target manifest, log the manifest
                     # and the missing values.
                     missing_manifest_log[target_manifest_id] = missing_values
-
-        return missing_manifest_log, present_manifest_log, repeat_manifest_log
+        target_attribute_in_manifest_list.append(target_attribute_in_manifest)
```
andrewelamb, line 1571: This isn't part of the pull request but line 1566:
`present_manifest_log.append(target_manifest_id)` 
is a dictionary which you can't append to.

schematic/models/validate_attribute.py
```diff
@@ -1509,7 +1628,15 @@ def _gather_target_columns_value(
...
-        return concatenated_target_column
+        target_manifest_empty = self._check_if_target_manifest_is_empty(
+            target_manifest, target_manifest_empty, column_names
+        )
+        target_attribute_in_manifest_list.append(target_attribute_in_manifest)
+        return (
+            concatenated_target_column,
+            target_attribute_in_manifest_list,
+            target_manifest_empty,
+        )
```
andrewelamb, line 1639: This doesn't match the return type. IT looks like its
```
tuple[
  pd.core.series.Series,
  list[bool],
  bool
]

File: schematic/models/validate_attribute.py [hunk 8]
```diff
@@ -1589,11 +1716,33 @@ def _run_validation_across_target_manifests(
...
+                    # From the concatenated target column, for value scope, run validation
+                    (
+                        missing_values,
+                        duplicated_values,
+                        repeat_values,
+                    ) = self._run_validation_across_targets_value(
+                        manifest_col=manifest_col,
+                        concatenated_target_column=target_column,
+                    )
+                    validation_store = (missing_values, duplicated_values, repeat_values)
```
andrewelamb, line 1745: Is this code that got commented accidentally?

schematic/models/validate_attribute.py
```diff
@@ -1617,6 +1770,10 @@ def _run_validation_across_target_manifests(
         # Start timer
         start_time = perf_counter()
 
+        # if not target_manifest_ids:
+        # target_manifet_present = False
+        # return (start_time, target_manifet_present)
+        # else:
```
andrewelamb, line 1776: Same as above

schematic/models/validate_attribute.py
```diff
@@ -1649,40 +1810,59 @@ def _run_validation_across_target_manifests(
...
-                manifest_col=manifest_col,
-                concatenated_target_column=target_column,
-            )
-            validation_store = (missing_values, duplicated_values, repeat_values)
+        # Check if a target manifest has been found, if so do not export validation_store
+        if not target_attribute_in_manifest:
+            return (start_time, target_attribute_in_manifest)
+        elif sum(target_manifest_empty) > 0:
+            return (start_time, "values not recorded in targets stored")
+        else:
```
andrewelamb, line 1845: These returns don't match to return type for this method

schematic/models/validate_attribute.py
mialy-defelice: No, I put it here while I was moving things around and then forgot, its gone now.

schematic/models/validate_attribute.py
mialy-defelice: Thanks, this was a previous approach that didnt work out and I forgot to remove it completely.

schematic/models/validate_attribute.py
mialy-defelice: I updated the return types

mialy-defelice:
@andrewelamb I addressed your comments :)

mialy-defelice:
@andrewelamb Looks like the tests are failing now bc of the google credentials. I'll kick them off again when things are good.

schematic/models/validate_attribute.py
andrewelamb: I'm not seeing the for loop...

schematic/models/validate_attribute.py
```diff
@@ -1406,12 +1504,18 @@ def _run_validation_across_targets_set(
...
         repeat_manifest_log: dict[str, pd.core.series.Series],
+        target_attribute_in_manifest_list: list[bool],
+        target_manifest_empty: list[bool],
     ) -> tuple[
-        dict[str, pd.core.series.Series],
-        dict[str, pd.core.series.Series],
-        dict[str, pd.core.series.Series],
+        tuple[
+            dict[str, pd.core.series.Series],
+            dict[str, pd.core.series.Series],
```
andrewelamb, line 1514: The second dict in this is tuple is the `present_manifest_log` which is a `list[str]`

schematic/models/validate_attribute.py
```diff
@@ -1577,7 +1712,25 @@ def _run_validation_across_target_manifests(
         val_rule: str,
         manifest_col: pd.core.series.Series,
         target_column: pd.core.series.Series,
-    ) -> tuple[[float, list]]:
+    ) -> tuple[
+        float,
+        Union[
```
andrewelamb, line 1717: The outer `Union` here is redundant as it only contains the inner `Union`

mialy-defelice: Okay its fixed, It should have two unions, but I accidentally put bool and str within the second union.

schematic/models/validate_attribute.py
mialy-defelice: The function is called by `_run_validation_across_targets_set` which is called within a for loop in the function `_run_validation_across_target_manifests` at line 1808:

```
for target_manifest_id, target_dataset_id in zip(
            target_manifest_ids, target_dataset_ids
        ):
            # Pull manifest from Synapse
            entity = synStore.getDatasetManifest(
                datasetId=target_dataset_id, downloadFile=True
            )
            # Load manifest
            target_manifest = pd.read_csv(entity.path)

            # Get manifest column names
            column_names = self._get_column_names(target_manifest=target_manifest)

            # Read each target manifest and run validation of current manifest column (set) against each
            # manifest individually, gather results
            if "set" in rule_scope:
                (
                    (
                        missing_manifest_log,
                        present_manifest_log,
                        repeat_manifest_log,
                    ),
                    target_attribute_in_manifest_list,
                    target_manifest_empty,
                ) = self._run_validation_across_targets_set(
                    val_rule=val_rule,
                    column_names=column_names,
                    manifest_col=manifest_col,
                    target_attribute=target_attribute,

...etc...

```
Its basically attempting (like target_attribute_in_manifest_list) to save data for all the target manifests that are being evaluated.

Im doing them in slightly different ways bc I could reuse `_check_if_target_manifest_is_empty` for both value and set operations but i had to do it differently for set and value to get `target_attribute_in_manifest_list` bc of how the targets are processed differently.

andrewelamb (approved):

PR #1401, opened 2024-04-10
add functionality to set requirements based per component

mialy-defelice:
Allow users to be able to set attribute requirements per component, using a combination of component based rule pairs in conjunction with the new "rule" `required`.

Like such:
`#Patient unique warning^^#Biospecimen unique required error`

This rule indicates that the attribute is required for the Biospecimen manifest and not the Patient Manifest. To work properly the Required field should also be set to FALSE.

Extra tests were added to cover a variety of test cases.

fixes [FDS-1626]

[FDS-1626]: https://sagebionetworks.jira.com/browse/FDS-1626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


schematic/schemas/data_model_graph.py
```diff
@@ -226,6 +229,64 @@ def get_adjacent_nodes_by_relationship(
...
+            validation_rules, list[str]: validation rules list for a given node and component.
+        """
+        # get any additional validation rules associated with this node (e.g. can this node be mapped to a list of other nodes)
+        node_validation_rules = self.get_node_validation_rules(
+            node_display_name=node_display_name
+        )
+
+        # Parse the validation rules per component if applicable
+        if node_validation_rules and isinstance(node_validation_rules, dict):
+            node_validation_rules = extract_component_validation_rules(
```
andrewelamb, line 249: The return type from `get_node_validation_rules` is a list, your'e checking if `node_validation_rules` is a dict

schematic/schemas/data_model_graph.py
```diff
@@ -226,6 +229,64 @@ def get_adjacent_nodes_by_relationship(
...
+                manifest_component=manifest_component,
+                validation_rules_dict=node_validation_rules,
+            )
+        return node_validation_rules
```
andrewelamb, line 253: The return type for this method is `list[str]` but the return type for `get_node_validation_rules` is list, which may get returned, and the return type for `extract_component_validation_rules` is list[Union[str, list]]`. Perpahs the returr type for this method should just be `list`

schematic/schemas/data_model_graph.py
```diff
@@ -226,6 +229,64 @@ def get_adjacent_nodes_by_relationship(
...
+            manifest_component=manifest_component, node_display_name=node_display_name
+        )
+
+        # Check if the valdation rule specifies that the node is required for this particular component.
+        if rule_in_rule_list("required", node_validation_rules):
+            node_required = True
+            # To prevent any unintended errors, ensure the Required field for this node is False
+            if (
+                self.get_node_required(node_display_name=node_display_name)
+                and node_required
```
andrewelamb, line 277: I don't think `and node_required` is needed here as it will always be true by it being se to true at the beginging of this if block

schematic/schemas/data_model_graph.py
mialy-defelice: Ah this is bc in between my original writing of this function, and the implementation of component based validation rules, we introduced validation rule being recorded as dictionaries as well as lists... so ill need to update that typing.

schematic/schemas/data_model_graph.py
mialy-defelice: Good call!

linglp:
@mialy-defelice 
Thanks for the PR. I tested it by doing the following: 
By using a modified data model: 

Note: 
1) I added: #Patient unique warning^^#Biospecimen unique required error for "Patient ID" 
2) set required to "FALSE" for "Patient ID"
[example.model-test.csv](https://github.com/Sage-Bionetworks/schematic/files/14984769/example.model-test.csv)


Did the following tests:
1) could convert CSV to jsonLD
> Note: when I made an error of not setting required column of "Patient ID" to FALSE, with the same validation rule, the jsonld conversion failed. I think the error message could be more helpful? 

2) I updated the config.yml to use the csv data model and run validation command: `schematic model --config config.yml validate --manifest_path /Users/lpeng/Downloads/test_required_biospecimen_2.csv --data_type Biospecimen`
- Please see the csv here: [test_required_biospecimen_2.csv](https://github.com/Sage-Bionetworks/schematic/files/14984585/test_required_biospecimen_2.csv)
- Since I didn't provide a patient ID in the manifest, I got an expected error: `error: '' should be non-empty
[['4', 'Patient ID', "'' should be non-empty", '']]`

3) To make sure that `required` does not affect `unique` rule, I also tested by creating a duplicated Patient ID in the manifest above. Got an expected error: `error: Column Patient ID has the duplicate value(s) ['123'] in rows: ['2', '4'].`

4) For validating Patient manifest, I tried: `schematic model --config config.yml validate --manifest_path /Users/lpeng/Downloads/test_required_patient_1.csv --data_type Patient`
* Please see csv here: 
[test_required_patient_1.csv](https://github.com/Sage-Bionetworks/schematic/files/14984634/test_required_patient_1.csv)
* Didn't get an error which is expected because PatientID isn't required for Patient manifest

5) However, when I tried changing the validation rule to: `#Patient unique required warning^^#Biospecimen unique required error`, I got an error for Patient manifest instead of a warning: `error: '' should be non-empty
[['542', 'Patient ID', "'' should be non-empty", '']]` --- <ins> Is that expected? </ins>

6) If I removed "unique" and changed the validation rule to: `#Patient required warning^^#Biospecimen unique required error`, I got a schematic error: 
```
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.10/bin/schematic", line 6, in <module>
    sys.exit(main())
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.10/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.10/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.10/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.10/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.10/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.10/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.10/lib/python3.10/site-packages/click/decorators.py", line 45, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/models/commands.py", line 274, in validate_manifest
    errors, warnings = metadata_model.validateModelManifest(
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/models/metadata.py", line 275, in validateModelManifest
    errors, warnings, manifest = validate_all(
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/models/validate_manifest.py", line 348, in validate_all
    manifest, vmr_errors, vmr_warnings = vm.validate_manifest_rules(
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/models/validate_manifest.py", line 180, in validate_manifest_rules
    ge_helpers.build_expectation_suite()
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/models/GE_Helpers.py", line 330, in build_expectation_suite
    self.add_expectation(
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/models/GE_Helpers.py", line 374, in add_expectation
    expectation_type=validation_expectation[rule.split(" ")[0]],
KeyError: 'required'
```
This seems like that it triggers an error related to great expectations. Is this expected?

tests/test_validation.py
```diff
@@ -515,6 +515,141 @@ def test_in_house_validation(self, helpers, dmge):
...
+            # since the valid value order isnt set in error reporting, check a portion of the expected output
+            # Check the error row is expected
+            assert expected_error[1] in errors[1]
+            # Check that one of the values for the expected valid values is present
+            # Extract a valid value
+            valid_value = expected_error[2].split(',')[-1].split(']')[0].strip(' ').strip("\'")
+            assert  valid_value in errors[1][2]
+            assert warnings==[]
+
+        return
```
linglp, line 651: This should be removed right?

mialy-defelice:
@linglp Thank you so much for the thorough review!!

See my responses below :)

> 1. could convert CSV to jsonLD
> 
> > Note: when I made an error of not setting required column of "Patient ID" to FALSE, with the same validation rule, the jsonld conversion failed. I think the error message could be more helpful?

I could not reproduce this error. We could open another ticket for a data model validator function, but that would only run after the point where you are getting an error. I need additional details to try to reproduce, including your CSV and JSONLD data model.


> 5. However, when I tried changing the validation rule to: `#Patient unique required warning^^#Biospecimen unique required error`, I got an error for Patient manifest instead of a warning: `error: '' should be non-empty [['542', 'Patient ID', "'' should be non-empty", '']]` ---  Is that expected?

This is expected because rule modifiers like "warning" do not work on `required` since `required` works from JSON Schema validation, which defaults to Error logging. Its not a true rule, and I'll expand on this more in SchemaHub docs.


> 6. If I removed "unique" and changed the validation rule to: `#Patient required warning^^#Biospecimen unique required error`, I got a schematic error:

> This seems like that it triggers an error related to great expectations. Is this expected?

This error is occuring because again required is not an actual rule, its a rule modifier itself. I had a check to skip traditional validation if required is provided as the only rule, but did not add checks to ensure it would work properly if it was combined with modifiers.

There is now a check included in `validate_utils.py` that is called by `GE_helpers`. I also added a test to ensure this function is working as intended in `test_utils.py`

linglp:
@mialy-defelice 
Thanks for your clarification. I did more testing: 

Rule 1: #Patient unique warning^^#Biospecimen unique required error
“Required” column is set to False

Expected behavior v.s. actual behavior : 
* When patient id is not unique in patient manifest, should raise a warning
    * Indeed saw a warning: Column Patient ID has the duplicate value(s) ['541'] in rows: ['541', '542'].
* When Patient id is not unique in biospecimen manifest, should raise an error
    * Indeed saw an error:  error: Column Patient ID has the duplicate value(s) ['456'] in rows: ['3', '4'].
* When Patient id is not provided in biospecimen manifest, should raise an error 
    * Indeed saw an error: [[‘4’, 'Patient ID', "'' should be non-empty", '']]
* When patient id is not provided in patient manifest, should not raise an error 
    * Indeed do not see an error or a warning 


Rule 2: #Patient^^#Biospecimen required error
“Required” column is set to False
Expected behavior v.s. actual behavior : 
* When patient id is not provided in patient manifest, should not raise an error
    * saw: Your manifest has been validated successfully. There are no errors in your manifest, and it can be submitted without any modifications.
* When patient id is not provided in biospecimen manifest, should raise an error
    * Indeed saw an error: '' should be non-empty [['4', 'Patient ID', "'' should be non-empty", '']]


Rule 3: ^^#Biospecimen required error
“Required” column is set to False
Expected behavior v.s. actual behavior : 
* When patient id is not provided in biospecimen manifest, should raise an error 
    * Indeed saw an error: error: '' should be non-empty [['4', 'Patient ID', "'' should be non-empty", '']]
* When patient id is not provided in patient manifest, should not raise an error
    * saw: Your manifest has been validated successfully. There are no errors in your manifest, and it can be submitted without any modifications.

schematic/schemas/data_model_graph.py
```diff
@@ -226,6 +229,67 @@ def get_adjacent_nodes_by_relationship(
...
+
+        # Parse the validation rules per component if applicable
+        if node_validation_rules and isinstance(node_validation_rules, dict):
+            node_validation_rules = extract_component_validation_rules(
+                manifest_component=manifest_component,
+                validation_rules_dict=node_validation_rules,
+            )
+        return node_validation_rules
+
+    def get_component_node_required(
```
linglp, line 256: I think it will be worthwhile writing a unit test for get_component_node_required function? If this can't be added to this PR, maybe we could create a separate ticket?

I also noticed that `get_json_validation_schema` function calls: `get_component_node_validation_rules` function before calling get_component_node_required function. But get_component_node_required function calls `get_component_node_validation_rules`  function again. Is this intended? I think these are duplicated? And technically, `get_component_node_required` could just get output from `get_component_node_validation_rules`?

linglp (changes requested):

schematic/schemas/data_model_graph.py
linglp: Based on discussion with Mialy over slack, this "duplication" is expected because we want get_component_node_required to be called independent of get_component_node_validation_rules in the future.

schematic/models/GE_Helpers.py
```diff
@@ -6,7 +6,7 @@
 import numpy as np
 
 # allows specifying explicit variable types
-from typing import Any, Dict, Optional, Text, List
+from typing import Any, Dict, Optional, Text, List, Literal
```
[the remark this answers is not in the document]
mialy-defelice: Good catch, i forgot to remove it after changing how i did some things.

schematic/schemas/data_model_graph.py
mialy-defelice: I actually added the ability to optionally provide the component validation rules, and also added them in the JSON Schema code.

tests/test_validation.py
linglp, line 651: This "return" can be deleted?

linglp (approved):
I think it looks good to me. Thanks for your hard work on this Mialy

PR #1409, opened 2024-04-16
fix: added project scope parameter in test_api.py

linglp:
## Cause: 
This is related to cross manifest validation, and this manifest: [syn27600110](https://www.synapse.org/#!Synapse:syn27600110)  does not have an “Id” column

More context in: https://sagebionetworks.jira.com/browse/FDS-1965

## Solution
Since after [Pull Request #1387](https://github.com/Sage-Bionetworks/schematic/pull/1387), we moved cross manifest validation to separate folder, submission test in `test_api.py` should also be given a proper project scope.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1086,7 +1086,7 @@ class TestManifestOperation:
             elif python_version == "3.9":
                 dataset_id = "syn52656104"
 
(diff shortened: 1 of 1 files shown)


mialy-defelice (approved):
Tests passed on my end.

PR #1412, opened 2024-04-16
Bugfix: FDS-1686 test manifest being changed

andrewelamb:
- Fixes [FDS-1686]
- `schematic/tests/data/mock_manifests/test_BulkRNAseq.csv` is no longer having columns added as part of test
- (tes now creates a copy to add columsn to, and then deletes copy)

[FDS-1686]: https://sagebionetworks.jira.com/browse/FDS-1686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


linglp:
Yes, I think your solution will work. But I am wondering if you could wrap this idea of "making a copy of the original file and then delete the copy after the test is done" into a fixture so that it could be used in other tests? Here's a function that I wrote that could achieve similar purpose: https://github.com/Sage-Bionetworks/schematic/blob/cd9d0da516c9d382b364ccbdc5375f592d01c460/tests/test_metadata.py#L28 But I am thinking we could make it more generic and save it in conftest.py so that it could be used across tests. What do you think?

andrewelamb:
@linglp Can you check again?

tests/conftest.py
```diff
@@ -128,3 +125,49 @@ def synapse_store(request):
...
 
     yield synapse_store
+
+
+# These fixtures make copies of existing test manifests.
+# These copies can the be altered by a given test, and the copy will eb destroyed at the 
+# end of the test
+
+@pytest.fixture(scope="function")
+def test_bulkrnaseq(helpers: Helpers) -> Generator[str, None, None]:
```
linglp, line 135: Hi @andrewelamb! I think this function can be re-written in a better way that not only works for "mock_manifests/test_BulkRNAseq.csv", but also works for other file path that we might want to use in the future: 
```
@pytest.fixture(scope="function")
def temporary_file_copy(request, helpers: Helpers) -> Generator[str, None, None]:
    file_name = request.param
    # original file copy
    original_test_path = helpers.get_data_path(f"mock_manifests/{file_name}")
    # Copy the original CSV file to a temporary directory
    temp_csv_path = helpers.get_data_path(f"mock_manifests/test_{file_name}_copy.csv")
    
    shutil.copyfile(original_test_path, temp_csv_path)
    yield temp_csv_path
    # Teardown
    if os.path.exists(temp_csv_path):
        os.remove(temp_csv_path)
```

I have tested that the above would work. To call the fixture in other function, you could do something like: 
```
    @pytest.mark.parametrize("temporary_file_copy", ["mock_manifests/test_BulkRNAseq.csv"])
    def test_submit_metadata_manifest(
        self,
        temporary_file_copy,
        helpers: Helpers,
        file_annotations_upload: bool,
        restrict_rules: bool,
        data_model_labels: str,
        hide_blanks: bool,
        validate_component: Optional[str],
    ) -> None:
```

linglp (changes requested):

tests/conftest.py
andrewelamb: What is the request here `def temporary_file_copy(request, helpers: Helpers) -> Generator[str, None, None]:` ?

andrewelamb: Also that sounds good, but I think we're beyond the scope of this issue. Since this PR is almost a month old,  think we should get this approved and save the above improvment for another issue. Thanks!

linglp: More about request: https://docs.pytest.org/en/6.2.x/example/simple.html#request-example

linglp (approved):
@andrewelamb you could merge now. I updated the code based on my own suggestion :) I re-ran both `test_metadata.py` and `test_store.py` locally to make sure that everything passed there.

PR #1413, opened 2024-04-16
feat: turn pyopenssl package optional 

linglp:
## Cause
Related to security issues discovered in this PR: https://github.com/Sage-Bionetworks/sage-monorepo/pull/2619. `cryptography` which is a dependency of pyopenssl has high security risk. We can't update pyopenssl in mono repo until that it becomes an optional dependency in schematic. 

## Solution 
Updated toml file to turn `pyopenssl ` optional

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -3075,7 +3075,7 @@ testutils = ["gitpython (>3)"]
 name = "pyopenssl"
 version = "23.3.0"
 description = "Python wrapper module around the OpenSSL library"
-optional = false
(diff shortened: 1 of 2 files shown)


andrewelamb (approved):

linglp:
Note: I also tested by building a beta docker image: `ghcr.io/sage-bionetworks/schematic:0.1.67-beta` and it seems like that schematic docker container works locally. This should be okay to merge.

PR #1414, opened 2024-04-17
Automation of github release and tag, and publish to pypi

andrewelamb:
- Fixes [FDS-1613]
- Previous publish workflow renamed to publish-tag
- Previous publish workflow now only trgiggered on non-standard tags being created (ie. 24.4.1.dev)
- New publish workflow (publish-main) triggered on push to main
- New publish workflow: The github tag and release are created
- New publish workflow: Pypi release is done on push to main
- New publish workflow: Automates the creation of the release tag using todays date and number of releases previously this month

The tag regex has been improved to:

`^v[1-9][0-9]\.[0-9]+\.[0-9]\.[a-z]+$`

- The periods have been escaped as those are wild cards without doing so
- The starting and ending special characters added
- The `\.[a-z]+` added means that 
- `v24.4.1` or `v24.4.1-dev` won't trigger the workflow
- `v24.4.1.dev` will trigger the workflow

[FDS-1613]: https://sagebionetworks.jira.com/browse/FDS-1613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/publish.yml
```diff
@@ -1,17 +1,21 @@
 name: Publish to PyPI
 on:
   push:
-    tags:
-      - 'v[1-9][0-9].[0-9]+.[0-9]+'
```
linglp, line 5: I saw that this is no longer included? Does this mean that we can't trigger a dev release manually?

andrewelamb: Good catch, I've added it back in

andrewelamb:
@linglp @mialy-defelice Coudl you look at the changes I just made? To solve the problems we dicussed on Wednesday, there are now two workflows.

.github/workflows/publish_tag.yml
```diff
@@ -2,7 +2,7 @@ name: Publish to PyPI
 on:
   push:
     tags:
-      - 'v[1-9][0-9].[0-9]+.[0-9]+'
+      - 'v[1-9][0-9].[0-9]+.[0-9].[a-z]+'
     branches: [main]
```
linglp, line 6: Even if we want to keep two workflows, can we clean up this workflow more? Here, this workflow will still get triggered because `branches: [main]`

.github/workflows/publish_tag.yml
```diff
@@ -114,4 +114,4 @@ jobs:
           # For posting a simple plain text message
           slack-message: "Schematic has just been released. Check out new version: ${{ github.ref_name }}"
         env:
-          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
```
linglp, line 117: Also, if this workflow is for triggering a dev release, I don't think we need to post a message on slack

linglp (changes requested):

.github/workflows/publish_tag.yml
andrewelamb: Oh maybe I misunderstand how this works, but I thought this would require both the branch to be main and the tag to match. Otherwise this would have been triggered before every time we pushed to main right?

.github/workflows/publish_tag.yml
andrewelamb: Removed that

.github/workflows/publish_tag.yml
andrewelamb: I went ahead and removed that line

linglp:
@andrewelamb could you also make sure tags like `v0.1.16-beta` won't trigger the workflow and make a PYPI release? Because that's a tag that we want to use to build a docker image for docker deployment, and not for PYPI release.

linglp (changes requested):
the old workflow is still referencing  override_version.py right? So if you delete  override_version.py, the old workflow will fail?

andrewelamb:
@linglp Good point I'll make that fix

linglp:
@andrewelamb additionally, I tried testing if dev tags will get tirggered by your regex, but it didn't seem to work. Github regex is different from regular regex... I found out when I was working on the docker publish workflow that even though an external tool might say that xx would work given a regex rule, it might still not work for github. 

To reproduce : 
1) check out branch: `develop-test-fds-1613`
2) create a tag, like v24.4.1.dev
3) push that dev to schematic repo.. 

When I checked the github action tab, the workflow didn't get triggered.

linglp:
@andrewelamb if we are not going to continue working on this, can this PR be closed?

andrewelamb:
@linglp Yep!

andrewelamb:
I'm going to leave the branch for now in case anyoen wants to take a crack at it in the near future.

PR #1415, opened 2024-04-17
Automatically save the JSON Validation schema

mialy-defelice:
See [FDS-871](https://sagebionetworks.jira.com/browse/FDS-871) for background.

This PR simply adds back functionality to save the JSON Validation schema.
It seems to have been inadvertently removed in a previous PR.

To test simply generate a manifest then go to where the model is stored and find the JSON Validation schema. 

Note: If desired in another PR we can add options to customize storage location and whether or not to make it, but for now just restoring a function that was present and users need.

[FDS-871]: https://sagebionetworks.jira.com/browse/FDS-871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andrewelamb:
Also as an aside that function is rediculously large. How did it get what way?

mialy-defelice:
> @mialy-defelice Would it make sense to add a test that confirms it has this behavior? INHO it should alreayd have that test, and then removing that would have been caught. (I suppose unless the person removing it also removed the test)

Yea I can write a test to see that its been created.

mialy-defelice:
> Also as an aside that function is rediculously large. How did it get what way?

I have no idea, it predates me. I looked into doing a rewrite of it a couple years ago, but it pretty efficient so we decided to just leave it for now, since for the most part people weren't ever going in it to edit anything. Before I started most of the functions were actually so big it was almost impossible to read them. I think it was the schema explorer actually had a whole section (~1000 lines) duplicated. The manifest generator is unwieldy now, but it used to be almost a single function...

andrewelamb (approved):

PR #1427, opened 2024-05-09
Allaway/testing truncation

allaway:


thomasyu888:
Confirmed with Robert that we can close! Thanks for the contribution!

PR #1430, opened 2024-05-10
feat: added tracing to schematic code base for manifest generate and manifest submission 

linglp:
## Context
Related to: https://sagebionetworks.jira.com/browse/FDS-2121

Added HTTP tracing in schematic
* added open telemetry and jaeger in schematic code base. Added them as optional dependency
* added tracing for `manifest/generate` endpoint

## Usage: 
* run jaeger server locally: 
```
docker run --name jaeger \
  -e COLLECTOR_OTLP_ENABLED=true \
  -p 16686:16686 \
  -p 4318:4318 \
  jaegertracing/all-in-one:latest
```
And then call a schematic API endpoint locally
You could then see HTTP request gets captured. 

### Tracing helps understand manifest submission latency 
See issue: https://sagebionetworks.jira.com/browse/FDS-1983

* To reproduce: 
* schema_url: https://raw.githubusercontent.com/ncihtan/data-models/v24.3.1/HTAN.model.jsonld
* dataset_id: syn60523752
* asset_view: syn60523751

Note: the whole request takes around 10min to complete! But you should be able to see that the majority of latency is coming from storing annotations. 

### Tracing helps understand the latency of manifest generation
Screen Shot 2024-06-16 at 7 39 38 PM
From the screen shot above, you could see 1) how schematic communicates with synapse python client; 2) a visual representation of some operation run time in schematic.


schematic_api/api/routes.py
```diff
@@ -40,10 +41,63 @@
...
+tracer = trace.get_tracer("Schematic")
+
+def trace_function_params():
+    """capture all the parameters of API requests
+    """
+    def decorator(func):
+        """create a decorator
+        """
+        @wraps(func)
+        def wrapper(**kwargs: Any):
```
BryanFauble, line 86: Be careful only including `**kwargs` here. It's been a little bit since I've written a wrapper, but IIRC you should also be accepting `*args` too in the event you are decorating a function where non-kw arguments were used.

https://python.plainenglish.io/python-decorators-args-kwargs-oversimplified-5902a201717c

BryanFauble (approved):
Just a few comments, otherwise this looks great to me!

This is an excellent starting point and later on as you get more insight into what's going on you'll want to start adjusting exactly what you are/are not tracing.

schematic_api/api/routes.py
linglp: Thanks for the comment! I added *args to allow any number of positional arguments to be used as well.

andrewelamb (approved):

PR #1432, opened 2024-05-13
FDS-511 removed typing extensions lock

andrewelamb:
- Fixes [FDS-511]
- Removes typing-extensions lock

[FDS-511]: https://sagebionetworks.jira.com/browse/FDS-511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -4862,4 +4862,4 @@ aws = ["uWSGI"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.9.0,<3.11"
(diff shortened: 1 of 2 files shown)


linglp (approved):
seems like just removing locking the library version but when I did `poetry show typing-extensions', it seems like that the package is still installed.

andrewelamb:
@linglp Yep! Its used by ipython, which is used by great expectations. This just allows for poetry to use more recent versions if our dependencies need to.

PR #1435, opened 2024-05-31
BugFix: Manifest version increment

GiaJordan:
Fixes FDS-1875

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1330,7 +1330,10 @@ class SynapseStorage(BaseStorage):
         ).id
 
         synapseutils.copy_functions.changeFileMetaData(
-            syn=self.syn, entity=manifest_synapse_file_id, downloadAs=file_name_new
+            syn=self.syn,
+            entity=manifest_synapse_file_id,
+            downloadAs=file_name_new,
+            forceVersion=False,
         )
 
         return manifest_synapse_file_id


linglp:
@GiaJordan I am curious if you have thought of adding a test to make sure that version is only incremented by 1. 

But I have also tested using `manifest/submit` endpoint and make sure that version increments by 1 per submission.

GiaJordan:
@linglp I did consider it in addition to the manual tests I did but in the end didn't feel it was necessary

andrewelamb:
@GiaJordan I might have missed it somewehre, but why does github think the whole file changed according to the diff?

andrewelamb:
@GiaJordan If some behavior changed(I can't tell what changed) then I would think you need to add/update the testing for that. We are moving away form all the manual testing we currently do.

GiaJordan:
> I might have missed it somewehre, but why does github think the whole file changed according to the diff?

The line endings changed, I believe as a result of the formatting running on my machine, but beyond that I'm not sure why

GiaJordan:
> @GiaJordan If some behavior changed(I can't tell what changed) then I would think you need to add/update the testing for that. We are moving away form all the manual testing we currently do.

It's for the version increment counter, I didn't mean it would require manual testing moving forward, but if you feel it's appropriate to add tests for that then I can add

andrewelamb:
@GiaJordan You should look into the line endings, it's making the coide review veyr difficult :)

andrewelamb:
@GiaJordan If it's worht manual testing, it's worth writign a test for so we can make sure going forward we don't accidentally lose this behavior.

afwillia:
@GiaJordan saw this in my email, maybe [this link is helpful for line endings](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings?platform=windows). Essentially it suggests running `git config --global core.autocrlf true` to automatically convert line endings to mac/linux on commit.

GiaJordan:
@linglp @andrewelamb I've looked through the existing manifest upload tests. Can either of you explain what you're envisioning for a test for this given the other tests for manifest upload are using patching and aren't actually uploading a manifest?

GiaJordan:
@andrewelamb if you change the "hide whitespace" setting you'll be able to see
image
My question is more on how that should be done considering the current way manifest upload tests are written would not allow for such a check

GiaJordan:
> @GiaJordan So it sounds like we are currently missing any integration tests for the `upload_manifest_file`?

@andrewelamb that's what it looks like, though I thought there used to be

andrewelamb (approved):

andrewelamb:
@GiaJordan I went ahead and approved then

linglp:
@GiaJordan and @andrewelamb Since `changeFileMetaData` function is provided by synapse python client, I don't think we need to add test for this function. To make sure version gets increment correctly, I agree that we should add an integration test. Part of the reason that this is confusing is because we don't have a separation of unit test and integration test. I would advocate of prioritizing:  https://sagebionetworks.jira.com/browse/FDS-811.

PR #1440, opened 2024-06-07
feat: set annotations in an async way when submitting a manifest

linglp:
## Context
related to: [FDS-2113 Store annotations in an async way when submitting manifests](https://sagebionetworks.jira.com/browse/FDS-2113?atlOrigin=eyJpIjoiZGM0Njc0ODM0YTRhNGRlMzhiMTAwODhiZTdiZDc0MTUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -68,6 +68,8 @@ from schematic.utils.schema_utils import get_class_label_from_display_name
 from schematic.store.base import BaseStorage
 from schematic.exceptions import AccessCredentialsError
 from schematic.configuration.configuration import CONFIG
+from synapseclient.annotations import Annotations
+import asyncio
 
 logger = logging.getLogger("Synapse storage")
 
@@ -1634,6 +1636,33 @@ class SynapseStorage(BaseStorage):
             self.syn.set_annotations(annos)
         return
 
+    async def _add_annotations_async(
+        self,
+        dmge: DataModelGraphExplorer,
+        row: pd.Series,
+        entityId: str,
+        hideBlanks: bool,
+        annotation_keys: str,
+    ) -> None:
+        """add annotations to entity ids in an asynchronous way 
+
+        Args:
+            dmge: DataModelGraphExplorer object,
+            row: current row of manifest being processed
+            entityId (str): synapseId of entity to add annotations to
+            hideBlanks: Boolean flag that does not upload annotation keys with blank values when true. Uploads Annotation keys with empty string values when false.
+            annotation_keys:  (str) display_label/class_label(default), Determines labeling syle for annotation keys. class_label will format the display
+                name as upper camelcase, and strip blacklisted characters, display_label will strip blacklisted characters including spaces, to retain
+                display label formatting while ensuring the label is formatted properly for Synapse annotations.
+        """
+        # Format annotations for Synapse
+        annos = self.format_row_annotations(
+            dmge, row, entityId, hideBlanks, annotation_keys
+        )
+
+        if annos:
+            await annos.store_async()
+
     def _create_entity_id(self, idx, row, manifest, datasetId):
         """Helper function to generate an entityId and add it to the appropriate row in the manifest.
         Args:
@@ -1653,7 +1682,7 @@ class SynapseStorage(BaseStorage):
         manifest.loc[idx, "entityId"] = entityId
         return manifest, entityId
 
-    def add_annotations_to_entities_files(
+    async def add_annotations_to_entities_files(
         self,
         dmge,
         manifest,
@@ -1694,6 +1723,7 @@ class SynapseStorage(BaseStorage):
             ).drop("entityId_x", axis=1)
 
         # Fill `entityId` for each row if missing and annotate entity as appropriate
+        set_annotations_requests=[]
         for idx, row in manifest.iterrows():
             if not row["entityId"] and (
                 manifest_record_type == "file_and_entities"
@@ -1713,8 +1743,11 @@ class SynapseStorage(BaseStorage):
 
             # Adding annotations to connected files.
             if entityId:
-                self._add_annotations(dmge, row, entityId, hideBlanks, annotation_keys)
+                #self._add_annotations(dmge, row, entityId, hideBlanks, annotation_keys)
+                set_annotations_requests.append(asyncio.create_task(self._add_annotations_async(dmge, row, entityId, hideBlanks, annotation_keys)))
                 logger.info(f"Added annotations to entity: {entityId}")
+        # execute all requests of setting annotations
+        responses = await asyncio.gather(*set_annotations_requests, return_exceptions=True)
         return manifest
 
     def upload_manifest_as_table(
@@ -1767,7 +1800,7 @@ class SynapseStorage(BaseStorage):
         )
 
         if file_annotations_upload:
-            manifest = self.add_annotations_to_entities_files(
+            manifest = asyncio.run(self.add_annotations_to_entities_files(
                 dmge,
                 manifest,
                 manifest_record_type,
@@ -1775,7 +1808,7 @@ class SynapseStorage(BaseStorage):
                 hideBlanks,
                 manifest_synapse_table_id,
                 annotation_keys,
-            )
+            ))
         # Load manifest to synapse as a CSV File
         manifest_synapse_file_id = self.upload_manifest_file(
             manifest,
@@ -1840,14 +1873,14 @@ class SynapseStorage(BaseStorage):
             manifest_synapse_file_id (str): SynID of manifest csv uploaded to synapse.
         """
         if file_annotations_upload:
-            manifest = self.add_annotations_to_entities_files(
+            manifest = asyncio.run(self.add_annotations_to_entities_files(
                 dmge,
                 manifest,
                 manifest_record_type,
                 datasetId,
                 hideBlanks,
                 annotation_keys=annotation_keys,
-            )
+            ))
 
         # Load manifest to synapse as a CSV File
         manifest_synapse_file_id = self.upload_manifest_file(
@@ -1917,7 +1950,7 @@ class SynapseStorage(BaseStorage):
         )
 
         if file_annotations_upload:
-            manifest = self.add_annotations_to_entities_files(
+            manifest = asyncio.run(self.add_annotations_to_entities_files(
                 dmge,
                 manifest,
                 manifest_record_type,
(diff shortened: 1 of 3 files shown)


schematic/store/synapse.py
```diff
@@ -1713,8 +1753,47 @@ def add_annotations_to_entities_files(
...
+                                requests.add(
+                                    asyncio.create_task(
+                                        self.store_async_annotation(
+                                            annotation_dict=annos
+                                        )
+                                    )
+                                )
+
+                    except Exception as e:
+                        raise RuntimeError(f"failed with { repr(e) }.")
```
BryanFauble, line 1796: I think the syntax you would want here would be something like
`raise RuntimeError("message") from e`

https://stackoverflow.com/questions/24752395/python-raise-from-usage

schematic/store/synapse.py
```diff
@@ -1713,8 +1753,47 @@ def add_annotations_to_entities_files(
...
-                logger.info(f"Added annotations to entity: {entityId}")
+                # Format annotations for Synapse
+                annos_task = asyncio.create_task(
+                    self.format_row_annotations(
+                        dmge, row, entityId, hideBlanks, annotation_keys
+                    )
+                )
+                requests.add(annos_task)
+
+                while requests:
```
BryanFauble, line 1764: This while loop is getting quite a bit nested, it might be a bit cleaner to extract this out to a function call. I am doing something similar here:
https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1107/files#diff-2dda10b4479134475ffeee7529eec430e40f828246ff532daa91b867d8d411b5R314
`async def _execute_download_tasks(self, download_tasks: Set[asyncio.Task]) -> None:`

schematic/store/synapse.py
```diff
@@ -11,6 +11,7 @@
 import secrets
```
BryanFauble, line 11: nit: I have swapped to using isort to handle ordering imports, that way I never need to think about where to add the import as this will auto sort them for me:

image

schematic/store/synapse.py
```diff
@@ -1335,8 +1338,43 @@ def upload_manifest_file(
...
+        """
+        annotation_data = Annotations.from_dict(
+            synapse_annotations=annotation_dict["annotations"]["annotations"]
+        )
+        annotation_class = Annotations(
+            annotations=annotation_data,
+            etag=annotation_dict["annotations"]["etag"],
+            id=annotation_dict["annotations"]["id"],
+        )
+        return await annotation_class.store_async(self.syn)
```
BryanFauble, line 1374:
```suggestion
        return await annotation_class.store_async(synapse_client=self.syn)
```

Personally, I really like using kw arguments. I am also considering a requirement that `synapse_client` is always a kwargument for all functions. What if your opinion on that change?

My thought: If I make a change to any of these functions my preference would be to order it before `synapse_client`. However, if I do not enforce it be referenced by kwarg, then it would be a breaking change to add an argument earlier in the list.

Further reading: https://realpython.com/python-asterisk-and-slash-special-parameters/#in-short-the-python-asterisk-and-slash-control-how-to-pass-values-to-functions

schematic/store/synapse.py
```diff
@@ -1653,7 +1692,41 @@ def _create_entity_id(self, idx, row, manifest, datasetId):
         manifest.loc[idx, "entityId"] = entityId
         return manifest, entityId
 
-    def add_annotations_to_entities_files(
+    async def _store_annos(self, requests):
```
BryanFauble, line 1695: nit: Typing for arguments and return.

I think it's the `pylance` extension that I have installed that will help point these out automatically for me. This is an example of a random file I opened:

image

tests/test_store.py
```diff
@@ -483,6 +480,65 @@ def test_get_files_metadata_from_dataset(self, synapse_store):
...
+            etag="mock etag",
+            id="mock syn_id",
+        )
+
+        with patch(
+            "schematic.store.synapse.Annotations.store_async",
+            new_callable=AsyncMock,
+            return_value=expected_dict,
+        ) as mock_store_async:
+            result = await synapse_store.store_async_annotation(annos_dict)
```
BryanFauble, line 537: Should we also try to assert that result.annotations is equal to expected_dict.annotations here?

schematic/store/synapse.py
linglp: Thanks for the suggestion Bryan! i used isort in the past too. See a separate PR here: https://github.com/Sage-Bionetworks/schematic/pull/1441

schematic/store/synapse.py
linglp: Thanks Bryan! I already addressed that :)

schematic/store/synapse.py
linglp: Thanks for your help! I already addressed that :)

tests/test_store.py
linglp: can I just add: `assert result==expected_dict` instead?

BryanFauble: idk - can you? Does it pass if you use that assert?

linglp: Yes. That works!

schematic/store/synapse.py
linglp: Thanks for the comment Bryan! I personally really like keyword arguments as well. I think when reviewing the code, it is clearer what is being passed in what parameters for functions that require keyword arguments. 

hmm how many functions are you thinking of changing? I think probably it is not a good idea to change functions in `synapseclient.client` since that will introduce breaking changes that we would need to cope with immediately. But since schematic just started to utilize functions from `synapseclient.models`,  we have time to adapt :)

BryanFauble: > hmm how many functions are you thinking of changing? I think probably it is not a good idea to change functions in synapseclient.client since that will introduce breaking changes that we would need to cope with immediately. But since schematic just started to utilize functions from synapseclient.models, we have time to adapt :)

I would only be making changes in the new code. Everything else would remain as is to allow for backwards compatibility.

schematic/store/synapse.py
```diff
@@ -1917,14 +1974,16 @@ def upload_manifest_combo(
...
         if file_annotations_upload:
-            manifest = self.add_annotations_to_entities_files(
-                dmge,
-                manifest,
-                manifest_record_type,
-                datasetId,
-                hideBlanks,
-                manifest_synapse_table_id,
-                annotation_keys=annotation_keys,
+            manifest = asyncio.run(
```
BryanFauble, line 1977: Using asyncio.run will work (for most cases). The cases where this will not work is if you are running this from within a jupyter notebook. If this is a concern read below:

https://blog.jupyter.org/ipython-7-0-async-repl-a35ce050f7f7

The problem is that the notebook will already have an Asyncio event loop running, and you cannot using `asyncio.run` by default from within an event loop. To get around this in the python client I wrote this wrapper:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v4.3.0/synapseclient/core/async_utils.py#L80-L99

Here is an example of where I used that code:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v4.3.0/synapseutils/sync.py#L1225-L1233

tests/test_store.py
```diff
@@ -956,27 +1097,40 @@ def test_add_annotations_to_entities_files(
...
+        
+        async def mock_process_store_annos(requests):
+            await asyncio.sleep(0.1)
+
         with patch(
             "schematic.store.synapse.SynapseStorage.getFilesInStorageDataset",
             return_value=files_in_dataset,
         ):
-            manifest_df = pd.DataFrame(original_manifest)
+            with patch('schematic.store.synapse.SynapseStorage.format_row_annotations', return_value=mock_format_row_annos, new_callable=AsyncMock) as mock_format_row:
```
BryanFauble, line 1110: I recommend to update your precommit black config to include these files as well:

https://github.com/Sage-Bionetworks/schematic/blob/develop/.pre-commit-config.yaml#L13

BryanFauble (approved):
Just a few comments around areas to give some more thought. Otherwise this is great!

tests/test_store.py
linglp: Thanks. Will be resolved in PR: https://github.com/Sage-Bionetworks/schematic/pull/1442

schematic/store/synapse.py
linglp: I think we don't need to support async runs in jupyter notebook. But what you wrote is interesting! I saw that you used: `nest_asyncio.apply(loop=loop)`. Is the idea here that you are trying to add another async loop? And you will run both loops until they both complete by doing: `return loop.run_until_complete(coroutine)`?

tests/test_store.py
```diff
@@ -541,6 +542,89 @@ async def test_store_async_annotation(self, synapse_store: SynapseStorage) -> No
             assert isinstance(result, Annotations)
 
 
+    async def test_process_store_annos_failure(self, synapse_store: SynapseStorage) -> None:
+        """test _process_store_annos function when there's an error either getting or storing annotations
+        """
+        async def mock_failure_coro():
+            await asyncio.sleep(0.1)
```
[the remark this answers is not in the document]
linglp: I see. Can I then change it to something like: 
```
        async def mock_failure_coro():
            await print('mock failure coro')
            raise ValueError("sample error")
```

schematic/store/synapse.py
BryanFauble: > nest_asyncio.apply(loop=loop). Is the idea here that you are trying to add another async loop?

Not exactly. This is attaching the new async function to the existing asyncio loop that is already running. Even if the code that you called this from isn't an `async def` function. It's a bit hacky, but the use-case is that we need to be able to support running in a notebook seamlessly.

> And you will run both loops until they both complete by doing: return loop.run_until_complete(coroutine)?

This will run until the coroutine I specifically passed to it is completed. This doesn't end the entrant asyncio loop.

tests/test_store.py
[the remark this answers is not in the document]
BryanFauble: Just:
```
        async def mock_failure_coro():
            raise ValueError("sample error")
```

Should be sufficient.

schematic/store/synapse.py
linglp: I see. Thanks for the explanation!

tests/test_store.py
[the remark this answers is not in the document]
linglp: Thanks for your suggestion Bryan! I have removed the asyncio.sleep in all the test :)

andrewelamb (approved):

PR #1444, opened 2024-06-19
technical debt: added unit test for changeFileMetaData function

linglp:
## Changes
* Added test for changeFileMetaData function
* related to PR: https://github.com/Sage-Bionetworks/schematic/pull/1435
* Also related to: FDS-1875

Note: @andrewelamb @GiaJordan even though we don't have integration test for: upload_manifest_file function , I realized that unit test could still be added to ensure that forceVersion gets called with False. I think that will ensure that version gets incremented correctly.

### DIFF
--- a/tests/test_store.py
+++ b/tests/test_store.py
@@ -1014,17 +1014,32 @@ class TestManifestUpload:
             }
         )
         with patch("synapseclient.Synapse.store") as syn_store_mock, patch(
-            "synapseutils.copy_functions.changeFileMetaData"
-        ):
+            "schematic.store.synapse.synapseutils.copy_functions.changeFileMetaData"
+        ) as mock_change_file_metadata:
             syn_store_mock.return_value.id = "mock manifest id"
+            mock_component_name = "BulkRNA-seqAssay"
             mock_file_path = helpers.get_data_path(mock_manifest_file_path)
             mock_manifest_synapse_file_id = synapse_store.upload_manifest_file(
                 manifest=test_df,
                 metadataManifestPath=mock_file_path,
                 datasetId="mock dataset id",
                 restrict_manifest=True,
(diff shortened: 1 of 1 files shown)


linglp:
@BryanFauble I realized that very interestingly, I could do: 
``` 
from unittest.mock import ANY
            mock_change_file_metadata.assert_called_once_with(
                forceVersion=False,
                syn=ANY,
                entity=ANY,
                downloadAs=ANY,
            )
```

If I just want to ensure that forceVersion=False gets called, can I do something like above in the future? Because the essence of this PR: https://github.com/Sage-Bionetworks/schematic/pull/1435 here is to ensure that forceVersion=False.

andrewelamb (approved):

BryanFauble:
> @BryanFauble I realized that very interestingly, I could do:
> 
> ```
> from unittest.mock import ANY
>             mock_change_file_metadata.assert_called_once_with(
>                 forceVersion=False,
>                 syn=ANY,
>                 entity=ANY,
>                 downloadAs=ANY,
>             )
> ```
> 
> If I just want to ensure that forceVersion=False gets called, can I do something like above in the future? Because the essence of this PR: #1435 here is to ensure that forceVersion=False.

Yup, what your doing there works to verify that a function is being called with your expected parameters. A small reccomendation thought is to limit or remove using ANY. ANY can be used, however, a better practice is to assert that the function is called with exactly what you are expecting. If can help catch small bugs that might be introduced as your making changes.

linglp:
@BryanFauble thank you so much for your comment. I think ANY can maybe also used when a parameter accepts unpredictable value, such as time stamp. In the code, I wrote the test to ensure that exact parameters with expected values are called, but good to hear about your perspective of using ANY.

PR #1450, opened 2024-06-27
fix: created decorator that works for async function

linglp:
## Context
Related to this issue: https://sagebionetworks.jira.com/browse/FDS-2181

## Changes: 
* created the decorator that works for async format_row_annotation to accommodate the current expected behavior: when trying to add annotations to an entity id, and that entity id is in the trash can, return a warning instead of an error. 
* added unit tests for the following: 
1) added a test for `format_row_annotation` to make sure that when entity id is in trash can, the function would return None and a warning will be logged. This is the current behavior. If not followed, integration test: test_upsertTable would fail. 
2) added a test for `process_store_annos` to make sure that storing annotations is not triggered when annotations are empty

Note:  format_row_annotations is still under tested because of technical debt in previous development of schematic (independent of async). Please see FDS 2138 for further testing and refactoring this function

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -348,6 +348,7 @@ class SynapseStorage(BaseStorage):
                     return None
                 else:
                     raise ex
+
         return wrapper
 
     def getStorageFileviewTable(self):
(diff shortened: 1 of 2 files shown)


andrewelamb (approved):

BryanFauble (approved):
🔥 LGTM!

linglp:
close this PR in favor of: close this PR in favor of https://github.com/Sage-Bionetworks/schematic/pull/1452

PR #1452, opened 2024-06-27
feat: set annotations in an async way when submitting a manifest -- new changes added

linglp:
## Context
See original PR for context and benchmarking result: https://github.com/Sage-Bionetworks/schematic/pull/1440

## Changes: 
Related to: https://sagebionetworks.jira.com/browse/FDS-2179
Related to: https://sagebionetworks.jira.com/browse/FDS-2181

* created the decorator that works for async format_row_annotation to accommodate the current expected behavior: when trying to add annotations to an entity id, and that entity id is in the trash can, return a warning instead of an error. 
* added unit tests for the following: 
1) added a test for `format_row_annotation` to make sure that when entity id is in trash can, the function would return None and a warning will be logged. This is the current behavior. If not followed, integration test: test_upsertTable would fail. 
2) added a test for `process_store_annos` to make sure that storing annotations is not triggered when annotations are empty
3) update the code to accommodate different variants of "EntityId" and "Id"
4) Added tests to ensure that variations of "EntityId" and "Id" get handled. 

Note:  format_row_annotations is still under tested because of technical debt in previous development of schematic (independent of async). Please see FDS 2138 for further testing and refactoring this function 

## Tests
I re-ran all the tests, including all the integration tests to make sure that they are all working.


linglp:
@andrewelamb @GiaJordan I ran all tests again after merging with the latest develop, and they all passed.

andrewelamb (approved):

PR #1456, opened 2024-07-10
Feature: `filenameExists` Validation Rule

GiaJordan:
This PR introduces the `filenameExists` metadata validation rule that checks for parity between the file paths in a metadata manifest (and the associated entityIds) and the file paths in the master file view on Synapse.

Another associated change is that schematic's handling of paths for all files on synapse has been changed; paths now include the project name at the beginning of the path to match the format of paths in synapse file views.


schematic/models/validate_attribute.py
```diff
@@ -731,6 +769,31 @@ class ValidateAttribute(object):
...
 
+    def _login(
+        self,
+        access_token: Optional[str] = None,
+        project_scope: Optional[list[str]] = None,
+        columns: Optional[list] = None,
+        where_clauses: Optional[list] = None,
+    ):
+        # if the ValidateAttribute object already has a SynapseStorage object, just requery the fileview, if not then login
+        if hasattr(self, "synStore"):
```
andrewelamb, line 780: Why this and not `if self.synStore:`?

andrewelamb: This really should be functionality in synapse store class.

andrewelamb (approved):

schematic/models/validate_attribute.py
GiaJordan: > This really should be functionality in synapse store class.

@andrewelamb can you explain what you're envisioning for implementing this functionality in the `SynapseStorage` class? I just want to make sure we're on the same page. Thanks.

andrewelamb: > > This really should be functionality in synapse store class.
> 
> @andrewelamb can you explain what you're envisioning for implementing this functionality in the `SynapseStorage` class? I just want to make sure we're on the same page. Thanks.

This whole method is kind weird. I don't think I understand the code well enough to offer an alternative, but alof of this seems like code that should live in the Synapse Store class not here.

GiaJordan: I think given the scope of this PR and new feature its best to leave it as is. I had explored adding similar functionality to the `SynapseStorage` class as part of the refactor, so if necessary it can be added then.

GiaJordan: > Why this and not `if self.synStore:`

@andrewelamb because if the `ValidateAttribute` class does not have a `synStore` attribute then it will raise an error

linglp, line 780: Under what situations that the if statement here:  `if hasattr(self, "synStore")` would get triggered?

GiaJordan: If the `ValidateAttribute` object already has a `synStore` attribute because it was used during the validation of another column

schematic/store/synapse.py
```diff
@@ -512,7 +580,10 @@ def getFilesInStorageDataset(
 
                     if fullpath:
                         # append directory path to filename
-                        filename = (dirpath[0] + "/" + filename[0], filename[1])
+                        filename = (
+                            project_name + "/" + dirpath[0] + "/" + filename[0],
```
linglp, line 584: I think this is not the expected behavior @GiaJordan @andrewelamb 
Please see ticket: https://sagebionetworks.jira.com/browse/FDS-2275

linglp: I am thinking now that you might need to include project_name for the validation to work. Could you make file validation work without changing the Filename column?

PR #1458, opened 2024-07-11
fix: check existing great expectation suite before adding a new one

linglp:
Related to: https://sagebionetworks.jira.com/browse/FDS-2155

## Problem
Sometimes our GH workflows failed with great expectation suite error: 
```
(1424 durations < 0.005s hidden.  Use -vv to show these durations.)
=========================== short test summary info ============================
FAILED tests/test_validation.py::TestManifestValidation::test_rule_combinations[int-IsNA] - great_expectations.exceptions.exceptions.ExpectationSuiteError: An ExpectationSuite named Manifest_test_suite already exists.
====== 1 failed, 715 passed, 691 warnings, 4 rerun in 1449.84s (0:24:09) =======
```
I think this problem occurred because the code adds an expectation suite without checking if it exists or not. See the original code below: 
```
self.suite = self.context.add_expectation_suite(
            expectation_suite_name=self.expectation_suite_name,
        )
```

Link of failed GH action: https://github.com/Sage-Bionetworks/schematic/actions/runs/9893404399/attempts/1

## Solution
Check if an expectation suite already exists before adding a new expectation suite. Based on the documentation [here](https://docs.greatexpectations.io/docs/oss/guides/expectations/how_to_create_and_edit_expectations_with_instant_feedback_from_a_sample_batch_of_data/#create-a-validator), you could use function `list_expectation_suite_name` to check if an expectation suite has been created. If it does, then we could retrieve the current suite (see documentation [here](https://docs.greatexpectations.io/docs/oss/guides/expectations/how_to_edit_an_existing_expectationsuite/#retrieve-an-existing-expectation-suite))

### DIFF
--- a/tests/test_ge_helpers.py
+++ b/tests/test_ge_helpers.py
@@ -1,0 +1,54 @@
+from unittest.mock import MagicMock
+
+import pandas as pd
+import pytest
+
+from schematic.models.GE_Helpers import GreatExpectationsHelpers
+
+
+@pytest.fixture(scope="class")
+def mock_ge_helpers(helpers):
+    dmge = helpers.get_data_model_graph_explorer(path="example.model.jsonld")
+    unimplemented_expectations = ["url"]
+    test_manifest_path = helpers.get_data_path("mock_manifests/Valid_Test_Manifest.csv")
+    manifest = pd.read_csv(test_manifest_path)
+
+    ge_helpers = GreatExpectationsHelpers(
+        dmge=dmge,
+        unimplemented_expectations=unimplemented_expectations,
+        manifest=manifest,
+        manifestPath=test_manifest_path,
+    )
+    yield ge_helpers
+
+
+class TestGreatExpectationsHelpers:
+    def test_add_expectation_suite_if_not_exists_does_not_exist(self, mock_ge_helpers):
+        # mock context provided by ge_helpers
+        mock_ge_helpers.context = MagicMock()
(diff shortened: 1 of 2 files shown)


linglp:
This is the same issue as github action workflow run failure here: https://github.com/Sage-Bionetworks/schematic/actions/runs/10014376516/job/27683959671

tests/test_ge_helpers.py
```diff
@@ -0,0 +1,72 @@
...
+
+@pytest.fixture(scope="class")
+def mock_ge_helpers(
+    helpers: Helpers,
+) -> Generator[GreatExpectationsHelpers, None, None]:
+    """Fixture for creating a GreatExpectationsHelpers object"""
+    dmge = helpers.get_data_model_graph_explorer(path="example.model.jsonld")
+    unimplemented_expectations = ["url"]
+    test_manifest_path = helpers.get_data_path("mock_manifests/Valid_Test_Manifest.csv")
+    manifest = pd.read_csv(test_manifest_path)
```
GiaJordan, line 19: The `get_data_frame` fixture from the conftest file should be used here to be uniform with how manifests are read in elsewhere in the tests

GiaJordan (approved):
Just one small comment to address before merge but otherwise looks good!
As an aside at some point we should probably switch to using the `GX` abbreviation to match how GreatExpectations refers to themselves but that's not pressing at all

PR #1460, opened 2024-07-19
Fix FDS-2251 NF annotation upload

afwillia:
This fixes a bug described in FDS-2251. The user was trying to submit with `annotation_keys=display_label` and got `keyError` errors. 

The solution is to update `format_row_annotations` to call `get_node_validation_rules` with `node_label` when `annotation_keys=class_label` and `node_display_name` when `annotation_keys=display_label`.

This PR also fixes an issue discovered during testing where the annotation key, `keySyn` does not get assigned and `metadataSyn[keySyn] = v` fails.

Tests for `format_row_annotations` are added to test/test_store.py.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1404,11 +1404,14 @@ class SynapseStorage(BaseStorage):
                 elif (
                     isinstance(anno_v, str)
                     and re.fullmatch(csv_list_regex, anno_v)
-                    and rule_in_rule_list(
-                        "list", dmge.get_node_validation_rules(anno_k)
-                    )
                 ):
-                    annos[anno_k] = anno_v.split(",")
+                    rule_list = []
+                    if annotation_keys == "class_label":
+                        rule_list = dmge.get_node_validation_rules(node_label=anno_k)
+                    else:
+                        rule_list = dmge.get_node_validation_rules(node_display_name=anno_k)
+                    if rule_in_rule_list("list", rule_list):
+                        annos[anno_k] = anno_v.split(",")
                 else:
                     annos[anno_k] = anno_v
 
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1367,7 +1367,7 @@ class SynapseStorage(BaseStorage):
         blacklist_chars = ["(", ")", ".", " ", "-"]
 
         for k, v in row.to_dict().items():
-            keySyn = None
+            keySyn = str(k)
             if annotation_keys == "display_label":
                 keySyn = str(k).translate({ord(x): "" for x in blacklist_chars})
             elif annotation_keys == "class_label":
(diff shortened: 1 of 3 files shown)


schematic/store/synapse.py
```diff
@@ -1367,15 +1367,16 @@ def format_row_annotations(
         blacklist_chars = ["(", ")", ".", " ", "-"]
 
         for k, v in row.to_dict().items():
+            keysyn = str(k)
             if annotation_keys == "display_label":
-                keySyn = str(k).translate({ord(x): "" for x in blacklist_chars})
+                kesSyn = str(k).translate({ord(x): "" for x in blacklist_chars})
```
linglp, line 1372: is this a typo? `kesSyn`? This should be caught by unit test. If not, I am thinking that the unit test needs to be more robust

schematic/store/synapse.py
```diff
@@ -1367,15 +1367,16 @@ def format_row_annotations(
         blacklist_chars = ["(", ")", ".", " ", "-"]
 
         for k, v in row.to_dict().items():
+            keysyn = str(k)
```
linglp, line 1370: why is this line needed?

schematic/store/synapse.py
```diff
@@ -1401,14 +1402,16 @@ def format_row_annotations(
...
                     annos[anno_k] = ""
-                elif (
-                    isinstance(anno_v, str)
-                    and re.fullmatch(csv_list_regex, anno_v)
-                    and rule_in_rule_list(
-                        "list", dmge.get_node_validation_rules(anno_k)
-                    )
-                ):
-                    annos[anno_k] = anno_v.split(",")
+                elif isinstance(anno_v, str) and re.fullmatch(csv_list_regex, anno_v):
```
linglp, line 1405: this is too many layers of elif and if else underneath. I would prefer this part to be refactored to be more maintainable.

schematic/store/synapse.py
```diff
@@ -1401,14 +1402,16 @@ def format_row_annotations(
...
+                    rule_list = []
+                    if annotation_keys == "class_label":
+                        rule_list = dmge.get_node_validation_rules(node_label=anno_k)
+                    else:
+                        rule_list = dmge.get_node_validation_rules(
```
linglp, line 1410: there's also no unit test for `get_node_validation_rules`. Can we also have a ticket for that?

andrewelamb:
@afwillia I ahven't gone in and looked at the code yet but:

- You'll need to handle the merge conflicts
- The automated tests don't seem to be running, do you knwo why that is?

linglp:
@afwillia After looking at the code more, I think i understand the issue that you are trying to fix! Thanks for bringing this up.

thomasyu888:
I'm going to close this, because I read in the Jira ticket that another ticket was opened and this is the PR: #1478

PR #1464, opened 2024-08-03
add csv data model to unit tests

lakikowolfe:
Include testing of csv data model in addition to jsonld

Related Jira ticket: https://sagebionetworks.jira.com/browse/FDS-1132

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -75,11 +75,12 @@ def test_manifest_json(helpers):
     )
     yield test_manifest_path
 
-
-@pytest.fixture(scope="class")
-def data_model_jsonld():
-    data_model_jsonld = "https://raw.githubusercontent.com/Sage-Bionetworks/schematic/develop/tests/data/example.model.jsonld"
-    yield data_model_jsonld
+@pytest.fixture(scope="class",
+                params=["example.model.jsonld",
+                        "example.model.csv"])
+def data_model_url(request):
+    data_model_url = "https://raw.githubusercontent.com/Sage-Bionetworks/schematic/develop/tests/data/" + request.param
+    yield data_model_url
 
 
 @pytest.fixture(scope="class")
@@ -320,9 +321,9 @@ class TestSynapseStorage:
 @pytest.mark.schematic_api
 class TestMetadataModelOperation:
     @pytest.mark.parametrize("as_graph", [True, False])
-    def test_component_requirement(self, client, data_model_jsonld, as_graph):
+    def test_component_requirement(self, client, data_model_url, as_graph):
         params = {
-            "schema_url": data_model_jsonld,
+            "schema_url": data_model_url,
             "source_component": "BulkRNA-seqAssay",
             "as_graph": as_graph,
         }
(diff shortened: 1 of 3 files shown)


GiaJordan (changes requested):

GiaJordan (changes requested):
Looking through this again, it seems like the tests in `test_manifest.py` and `test_validation.py` are still only using the `.jsonld` could you update to test with the CSVs where the schemas module is concerned as well?

linglp:
@lakikowolfe can ticket be linked to PR description? Also, when I queried the logs of `manifest/generate`, I didn't see any projects that are using .csv url when generating a manifest.. I am curious where the "ask" comes from...

lakikowolfe:
@linglp this is the ticket: https://sagebionetworks.jira.com/browse/FDS-1132

I believe the motivation behind this ticket is to update the tests to include csv because it is a supported file type and we'd want to know if for some reason csv input was not working.

lakikowolfe:
@GiaJordan My understanding is that this is low priority so I am focusing on the HTAN bug ticket that I have rn and will circle back to this when I can. Please let me know if I am misunderstanding the priority of this ticket and we can discuss.

GiaJordan:
@lakikowolfe yeah your understanding is correct

GiaJordan:
@lakikowolfe yeah from my perspective your understanding is correct, but I'll loop in kevin so he can make the final determination

GiaJordan:
~~Looking through this again, it seems like the tests in test_manifest.py and test_validation.py are still only using the .jsonld could you update to test with the CSVs where the schemas module is concerned as well?~~
@lakikowolfe sorry about this earlier comment, looking at the current `test_schemas.py` I see it's already using both the csv and jsonld. you can disregard

GiaJordan (approved):

GiaJordan:
Approving pending resolution of merge conflicts

tests/test_api.py
```diff
@@ -320,9 +321,9 @@ def test_if_in_assetview(self, request_headers, client, entity_id):
 @pytest.mark.schematic_api
 class TestMetadataModelOperation:
     @pytest.mark.parametrize("as_graph", [True, False])
-    def test_component_requirement(self, client, data_model_jsonld, as_graph):
+    def test_component_requirement(self, client, data_model_url, as_graph):
```
thomasyu888, line 324: This is neat @lakikowolfe , just for my education, does this mean that the tests run twice for both data_model_urls?

tests/test_api.py
```diff
@@ -75,11 +75,12 @@ def test_manifest_json(helpers):
     )
     yield test_manifest_path
 
-
-@pytest.fixture(scope="class")
-def data_model_jsonld():
-    data_model_jsonld = "https://raw.githubusercontent.com/Sage-Bionetworks/schematic/develop/tests/data/example.model.jsonld"
-    yield data_model_jsonld
+@pytest.fixture(scope="class",
```
thomasyu888, line 78: @lakikowolfe Thanks for the great work here!  As noted below, I really like this solution.  I just wanted to understand your rationale behind  why you chose this over adding these to `pytest.parametrize`

thomasyu888 (approved):
Great review @GiaJordan , and thanks for the work @lakikowolfe .  Please answer some  of my comments, but I'm going to go ahead and approve.

tests/test_api.py
lakikowolfe: Yep that's correct

tests/test_api.py
lakikowolfe: hmm I don't have a rationale except for when I was looking into fixtures (I'm new to pytest) I came across this method. I don't know the difference between the two but will look into it tomorrow.

thomasyu888: No worries, a quick browse of this documentation: https://docs.pytest.org/en/7.1.x/how-to/parametrize.html and some of the tests below may provide some guidance. 

That said, this was more of a self-learning exercise.  A quick summary is that it would've required more code changes, but I'll leave it to you to see why.

PR #1467, opened 2024-08-16
Update existing file paths in manifests at generation to conform to new convention

GiaJordan:
This PR changes the functionality of file based manifest generation to also update the paths of existing files to account for the change in file path convention in schematic. Also includes more helpful error messages in cases where filepath validation can not be completed because the fileview needs to be updated.

## Tests:
`test_get_manfiest_with_files` has been updated to check file paths for file based manifests and to ensure that the order of files and the paths themselves are as expected when generating a manifest when new files have been added to the dataset and the existing manifest has paths that need updating.

`test_submit_filebased_manifest` has been added as a complement to `test_submit_metadata_manifest` to actually test submission to synapse in addition to functionality of the method. This is to ensure that manifests with the new paths are able to be submitted.

`test_fill_in_entity_id_filename` has been updated to match how one of the utilized methods now returns values.

`test_view_query_exception` has been added to ensure that the appropriate exceptions are raised when the fileview can't be queried


tests/test_store.py
```diff
@@ -432,27 +466,50 @@ def test_getDatasetManifest(self, synapse_store, downloadFile):
...
+                [
+                    {
+                        "Filename": ["existing_mock_file_path", "mock_file_path"],
+                        "entityId": ["existing_mock_entity_id", "mock_entity_id"],
+                    },
+                    {
+                        "Filename": ["mock_file_path"],
+                        "entityId": ["mock_entity_id"],
+                    },
+                ],
```
linglp, line 500: Also, later in the test, you have: 
```

            if not existing_manifest_df.empty:
                expected_df = pd.DataFrame(
                    {
                        "Filename": ["existing_mock_file_path", "mock_file_path"],
                        "entityId": ["existing_mock_entity_id", "mock_entity_id"],
                    }
                )
            else:
                expected_df = pd.DataFrame(
                    {"Filename": ["mock_file_path"], "entityId": ["mock_entity_id"]}
                )
```
I think the "expected_df" can be in @pytest.mark.parametrize. So you should just need to do: 
`assert_frame_equal(new_manifest, expected_df)` without doing if else

schematic/store/synapse.py
```diff
@@ -788,18 +799,42 @@ def fill_in_entity_id_filename(
...
             new_files = pd.DataFrame(new_files)
+
+            # update manifest so that it contains new dataset files
             manifest = (
                 pd.concat([manifest, new_files], sort=False)
                 .reset_index()
                 .drop("index", axis=1)
             )
 
+            manifest_reindex = manifest.set_index("entityId")
```
linglp, line 819: there's no comment from line 819 - line 837. Could you add some comments to breakdown some of the steps?

tests/test_store.py
```diff
@@ -214,6 +220,34 @@ def test_view_query(
...
 
+    @pytest.mark.parametrize(
+        "asset_view,columns,expectation",
+        [
+            ("syn62339865", ["path"], pytest.raises(ValueError)),
+            ("syn62340177", ["id"], pytest.raises(ValueError)),
+            ("syn23643253", ["id", "path"], does_not_raise()),
+        ],
+    )
+    def test_view_query_exception(
```
linglp, line 231: I am seeing both `test_view_query` and this test. Could you change the both of their name to `test_query_fileview` and test_query_fileview_exception`? I think that will make the test more searchable

linglp, line 231: In addition, for `test_view_query`, I am seeing the logic of "only perform the query if it is different from the previous query" not being tested. Can that also be added to this PR? If not, can an additional ticket be created?

linglp, line 231: when I read `test_view_query`, I think there are certain cases that are not covered: 
1) when project scope is None but columns, where_clauses are provided
2) when columns is None but where_clauses and project scope are provided. 

Should they also be covered?

tests/test_store.py
```diff
@@ -214,6 +220,34 @@ def test_view_query(
...
+        expectation: str,
+    ) -> None:
+        project_scope = ["syn23643250"]
+
+        # set to use appropriate test view and set scope
+        CONFIG.synapse_master_fileview_id = asset_view
+        synapse_store_special_scope.project_scope = project_scope
+
+        # ensure approriate exception is raised
+        with expectation:
```
linglp, line 245: it is a small thing, but I imagine you could do: 
```
    try: 
          with expectation: 
                  XXX
    finally:
        CONFIG.synapse_master_fileview_id = "syn23643253"
```
More about using `finally`: https://www.w3schools.com/python/ref_keyword_finally.asp#:~:text=It%20defines%20a%20block%20of,objects%20and%20clean%20up%20resources.

tests/test_store.py
```diff
@@ -214,6 +220,34 @@ def test_view_query(
...
+            ("syn62339865", ["path"], pytest.raises(ValueError)),
+            ("syn62340177", ["id"], pytest.raises(ValueError)),
+            ("syn23643253", ["id", "path"], does_not_raise()),
+        ],
+    )
+    def test_view_query_exception(
+        self,
+        synapse_store_special_scope: SynapseStorage,
+        asset_view: str,
+        columns: list,
```
linglp, line 235: a small thing but here for columns it is a list of string right? so `list[str]`?

linglp (changes requested):

linglp:
@GiaJordan Thanks for the fix. If my understanding is correct, even without "path" being enabled in the asset view, when users generate a file-based manifest, "Filename" column is now being updated too to include project name in the path? If my understanding is correct, should we add a message and let users know that we are adding "project name" to the path and they could now enable "path" column by themselves in synapse too? But this might not be super important since DCA users can't see most of the schematic log.

GiaJordan:
> @GiaJordan Thanks for the fix. If my understanding is correct, even without "path" being enabled in the asset view, when users generate a file-based manifest, "Filename" column is now being updated too to include project name in the path? 

that's correct

> should we add a message and let users know that we are adding "project name" to the path and they could now enable "path" column by themselves in synapse too?

From my perspective, it's more important to communicate this kind of thing to the users directly, as well as include in the release notes. Especially since most users will be using the DCA. From my presentation/recommendations and our discussion with @AmyHeiser during the team meeting earlier this week, this is the current plan.

schematic/store/synapse.py
```diff
@@ -291,8 +291,19 @@ def query_fileview(
                 self.storageFileviewTable = self.syn.tableQuery(
                     query=self.fileview_query,
                 ).asDataFrame()
-            except SynapseHTTPError:
-                raise AccessCredentialsError(self.storageFileview)
+            except SynapseHTTPError as exc:
+                exception_text = str(exc)
+                if "Unknown column path" in exception_text:
+                    raise ValueError(
+                        "The path column has not been added to the fileview. Please make sure that the fileview is up to date. You can add the path column to the fileview by follwing the instructions in the validation rules documentation (https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/2645262364/Data+Model+Validation+Rules#Filename-Validation)."
```
linglp, line 298: hmm should the link of documentation be included in our code? I feel like this might be hard to maintain in the future? If the documentation moves we might forget to update the code. Also, if we are not expecting users to see it, then we probably shouldn't include it? External users can't click on to Sage's confluence page anyway

tests/test_metadata.py
```diff
@@ -143,3 +149,20 @@ def test_submit_metadata_manifest(
...
                 assert mock_manifest_id == "mock manifest id"
+
+    def test_submit_filebased_manifest(self, helpers):
+        meta_data_model = metadata_model(helpers, "class_label")
+
+        manifest_path = helpers.get_data_path(
+            "mock_manifests/filepath_submission_test_manifest.csv"
+        )
+
+        with does_not_raise():
```
thomasyu888, line 160: Could you explain the rationale behind this?

tests/test_metadata.py
```diff
@@ -143,3 +149,20 @@ def test_submit_metadata_manifest(
...
+            manifest_id = meta_data_model.submit_metadata_manifest(
+                manifest_path=manifest_path,
+                dataset_id="syn62276880",
+                manifest_record_type="file_only",
+                restrict_rules=False,
+                file_annotations_upload=True,
+            )
+            assert manifest_id == "syn62280543"
```
thomasyu888, line 168: Nit (nitpicky): Just so I'm understanding, is this an integration test?  If so, since we know we want to make those "quality of life" changes, I'd go ahead and start adding tests in this format

```
.
└── tests
    ├── data
    │   ├── mock_manifests
    │   └── test_configs
    ├── integration
			data (Moved from above)
			test_configs (Moved from above)
    └── unit
			data (Moved from above)
			test_configs (Moved from above)
```

Without worrying about the older tests.  Just a thought.

tests/test_store.py
```diff
@@ -214,6 +220,35 @@ def test_view_query(
         # tests that the query was valid and successful, that a view subset has actually been retrived
         assert synapse_store_special_scope.storageFileviewTable.empty is False
 
+    @pytest.mark.parametrize(
+        "asset_view,columns,expectation",
+        [
+            ("syn62339865", ["path"], pytest.raises(ValueError)),
+            ("syn62340177", ["id"], pytest.raises(ValueError)),
+            ("syn23643253", ["id", "path"], does_not_raise()),
```
thomasyu888, line 228: Nit: When you're doing tests, you ideally want to split off the testing the "error" vs the non errors.  For example:

- ```def test_view_query_exception_errors()``` <- This would contain all the errors
- ```def test_view_query()``` <- This would be the success.  This would allow you to not have to do `does_not_raise()` function.

Separately, I understand from reading the code and the tests that you're catching for these two conditions

* "Unknown column path"
* "Unknown column"

You may do something like this
```pytest.raises(ValueError, match=r"The path column has not been added to the fileview. Please .*"```

I encourage these tests to be split up, because then youre parametrize, could take `error_type` and `message` as parameters and you could include in the test itself `pytest.raises`

Comment (thomasyu888) [reply tests/test_store.py:228]: Side note, when I review code, the first thing I'm looking at is the tests, because it _should_ tell me exactly what the code's behavior is.

View #1469 for an example of what I personally think are well formatted tests.

File: tests/test_store.py [hunk 10]
```diff
@@ -214,6 +220,35 @@ def test_view_query(
...
+        CONFIG.synapse_master_fileview_id = asset_view
+        synapse_store_special_scope.project_scope = project_scope
+
+        # ensure approriate exception is raised
+        try:
+            with expectation:
+                synapse_store_special_scope.query_fileview(columns)
+        finally:
+            # reset config to default fileview
+            CONFIG.synapse_master_fileview_id = "syn23643253"
```
thomasyu888, line 250: Nit: Why is it important to reset the config to default fileview?  That, in my opinion, is more important to write as a comment, because we can read that you're resetting the `synapse_master_fileview_id` back to another synapse id.

schematic/store/synapse.py
```diff
@@ -788,18 +799,47 @@ def fill_in_entity_id_filename(
...
+            # If all the paths do not match, update the manifest with the filepaths from synapse
+            if not file_paths_match.all():
+                manifest_reindex.loc[
+                    ~file_paths_match, "Filename"
+                ] = all_files_reindex_like_manifest.loc[~file_paths_match, "Filename"]
+
+                # reformat manifest for further use
+                manifest = manifest_reindex.reset_index()
+                entityIdCol = manifest.pop("entityId")
+                manifest.insert(len(manifest.columns), "entityId", entityIdCol)
```
thomasyu888, line 841: I'm coming at this from a naive angle, but it seems what is on synapse is the source of truth. Instead of spending the compute and doing a comparison, why not just do a join and replace with what's on Synapse?  This would probably simplify the code, but I'm most likely missing something.

schematic/store/synapse.py
GiaJordan: Yeah that's a good point; I'll remove

tests/test_metadata.py
GiaJordan: Yeah for sure, it was in the `pytest` documentation for testing that an exception is raised or not raised under certain conditions. For this test the important thing was to make sure the `submit_metadata_manifest` was able to complete without raising an exception. I added the `with does_not_raise` to try and be more explicit about that. I wanted to avoid having a test that just called the method and did nothing else. I guess with the assertion with `with` statement kind of unnecessary but that was added later and isn't the main point of the test and I wasn't sure what the best combination was in this case.
@thomasyu888

schematic/store/synapse.py
```diff
@@ -777,18 +788,47 @@ def fill_in_entity_id_filename(
         # the columns Filename and entityId are assumed to be present in manifest schema
         # TODO: use idiomatic panda syntax
         if dataset_files:
```
BryanFauble, line 790: You could flip this `if` check and lower the congnative complexity of this function. Like:

```
  if not dataset_files:
      manifest = manifest.fillna("")
      return dataset_files, manifest

# The rest of the function
.
.
.
.
manifest = manifest.fillna("")
return dataset_files, manifest
    
```

The idea here is that you are failing fast with the negative conditions ie: If you are providing things that I am not expecting I am going to return as soon as possible. That way you don't have significantly nested code; In the rest of the function you can guarantee that `dataset_files` has a truthy value in it.

GiaJordan: That makes a lot of sense @BryanFauble thank you

schematic/store/synapse.py
GiaJordan: @thomasyu888 this is me being conservative from a user experience standpoint. The files that are returned from synapse are sorted in alphabetical order. What I'm thinking of is the case where a user has already uploaded a manifest and annotated files, and later more files are uploaded that may not alphabetically come after the existing files on synapse. 

In this case, if we just take and copy what is on synapse, the rows for the new files that need to be annotated would be scattered separately amongst the already annotated files. What I'm trying to cover by the reindexing, matching, and the resetting is users being confused as to where the rows are for the new files they have to annotate. Matching based on entityId ensures that the files are in the same order so we can do a safer filename swap (as opposed to the case where files are in a different order in the manifest and on synapse) and then the reindexing preserves the old behavior where new files always come after the already annotated files that exists on synapse.

tests/test_store.py
GiaJordan: @thomasyu888 that makes sense, I can explain that more clearly

schematic/store/synapse.py
thomasyu888: Makes sense! Thanks for the thoughtful explanation!

tests/test_metadata.py
thomasyu888: Thanks!  I also could've read the pytest documentation: https://docs.pytest.org/en/6.2.x/example/parametrize.html#parametrizing-conditional-raising. 😅

schematic/store/synapse.py
GiaJordan: @thomasyu888 yeah no problem!

tests/test_metadata.py
GiaJordan: @thomasyu888 Yeah you're welcome!

tests/test_store.py
GiaJordan: @thomasyu888 that makes sense and thanks for the linked example. Both were really helpful!

tests/test_store.py
GiaJordan: @linglp yes thats a good point thanks!

GiaJordan: Thanks for catching! @linglp

tests/test_manifest.py
```diff
@@ -763,39 +763,77 @@ def test_create_manifests(
...
+        expected_rows,
+        expected_files,
+    ):
         """
-        Test to ensure that when generating a record based manifset that has files in the dataset that the files are not added to the manifest as well
+        Test to ensure that
+            when generating a record based manifset that has files in the dataset that the files are not added to the manifest as well
+            when generating a file based manifest from a dataset thathas had files added that the files are added correctly
         """
+        # GIVEN the example data model
```
thomasyu888, line 801: Fantastic usage of what's learned from a PR.  Love to see it!

GiaJordan: Thanks @thomasyu888 !

GiaJordan:
@thomasyu888 @BryanFauble @linglp 
This PR is now ready for y'all to review again

schematic/store/synapse.py
```diff
@@ -291,8 +291,19 @@ def query_fileview(
...
+                    raise ValueError(
+                        "The path column has not been added to the fileview. Please make sure that the fileview is up to date. You can add the path column to the fileview by follwing the instructions in the validation rules documentation."
+                    ) from exc
+                elif "Unknown column" in exception_text:
+                    missing_column = exception_text.split("Unknown column ")[-1]
+                    raise ValueError(
+                        f"The columns {missing_column} specified in the query do not exist in the fileview. Please make sure that the column names are correct and that all expected columns have been added to the fileview."
+                    ) from exc
+                else:
+                    raise AccessCredentialsError(self.storageFileview)
```
BryanFauble, line 306: Does this also need the `from exc` clause?

tests/data/mock_manifests/filepath_submission_test_manifest.csv
```diff
@@ -0,0 +1,3 @@
+Filename,Sample ID,File Format,Component,Genome Build,Genome FASTA,Id,entityId
```
BryanFauble, line 1: Should we also start to put these into: `tests/integration/data`?

tests/integration/test_metadata_model.py
```diff
@@ -0,0 +1,33 @@
...
+
+        with does_not_raise():
+            manifest_id = meta_data_model.submit_metadata_manifest(
+                manifest_path=manifest_path,
+                dataset_id="syn62276880",
+                manifest_record_type="file_only",
+                restrict_rules=False,
+                file_annotations_upload=True,
+            )
+            assert manifest_id == "syn62280543"
```
BryanFauble, line 33: Is there other items that we can assert on here? I see that it only have a single return, but perhaps you might also want to assert that some nested functions were called for the specific logic you are testing.

Here is an example where I am spying on a function I expect to be called:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/tests/integration/synapseclient/core/test_download.py#L27-L73

From there you can use all the usual suspects like `assert_called_once`, `assert_called_once_with`, ect...

tests/test_store.py
```diff
@@ -31,6 +32,11 @@
 logger = logging.getLogger(__name__)
 
 
+@contextmanager
+def does_not_raise():
```
BryanFauble, line 36: Could we place this out into a utility class so it doesn't need to be copied over to everywhere it's used?


I wonder if pytest might also have something like this that we can use without needing to write it ourselves.

tests/test_store.py
```diff
@@ -196,17 +227,62 @@ def test_view_query(
...
+        synapse_store_special_scope.project_scope = project_scope
+
+        # WHEN the query is built and run
+        try:
+            # THEN it should raise a ValueError with the appropriate message
+            with pytest.raises(ValueError, match=message):
+                synapse_store_special_scope.query_fileview(columns)
+        finally:
+            # AND the fileview should be reset to the default so the other tests are not affected regardless of the outcome of the query
+            CONFIG.synapse_master_fileview_id = "syn23643253"
```
BryanFauble, line 285: If you are updating a global value and setting it back after you have the potential to cause side affects while the test is running before the global value was set back.

Rather than modifying the global value I would recommend to use some object matching here. That way any changes are only applicable within the `with` statement managed context.

I think you'll just need to patch the Configuration function for this property:
```
    @property
    def synapse_master_fileview_id(self) -> str:
        """
        Returns:
            str:
        """
        return self._synapse_config.master_fileview_id
```

Unfortunately mocking in Python can be a bit messy, but it's a trade-off to prevent side affects for other tests.

BryanFauble (approved):
I'll go ahead and approve the changes as they are now. I noted a few items, but it should not affect any logic you've written.

tests/test_store.py
GiaJordan: @BryanFauble Yeah, would it be appropriate to move to the `conftest.py`?

tests/data/mock_manifests/filepath_submission_test_manifest.csv
GiaJordan: @BryanFauble I'd rather wait until we do the bigger structure change or do that move separate from this PR as the helper functions will need to be updated as well

schematic/store/synapse.py
GiaJordan: I was trying to be more complete and include the context from synapse, but if you think it's not necessary then I can remove @BryanFauble

tests/test_store.py
GiaJordan: @BryanFauble yeah that makes sense

schematic/store/synapse.py
BryanFauble: What I meant was, does this line:
`raise AccessCredentialsError(self.storageFileview)`
need to be
`raise AccessCredentialsError(self.storageFileview) from exc`
?

exception handling in python is really un-intuitive, and I'm not sure if this is needed to properly maintain the context from the original exception. This is what you have doing in the 2 ValueErrors

tests/test_store.py
GiaJordan: @BryanFauble I'm encountering more issues with trying to mock the config object. Are this change and the others you noted with you approval necessary for the PR in the context of getting our fixes out quickly? I ask because I have other issues outstanding as well

GiaJordan: @thomasyu888 thanks for that! Yes I was concerned because as I was working on this PR to implement all the feedback it ended up taking much longer than I had estimated at our sprint planning and along with all the other coordination work affected my other issues. But I would appreciate the having the time to be able to actually implement things in a correct and robust way.

thomasyu888: As for challenges, I trust everyone puts in their effort, but don't be afraid to reach out for help you need it. What issues are you encountering?

schematic/store/synapse.py
GiaJordan: @BryanFauble that makes sense

tests/test_store.py
GiaJordan: @thomasyu888 I was running into issues with how the mocked config was being used during the tests, but I think I was actually able to figure it out since I had the space to. The tests are passing locally now with the change so I think it's all good

tests/integration/test_metadata_model.py
GiaJordan: @BryanFauble thanks, I added on to the tests and in the process realized that there was another issue going on that's present on `develop`, so doing the extra checks were really helpful in catching that as well

thomasyu888 (approved):
LGTM! thanks for all your hard work on this! I think someone on fair should take a last pass on this

thomasyu888:
@linglp, @GiaJordan , I cloned this branch and did

```
docker build -t test -f schematic_api/Dockerfile .
```

Then I ran schematic profiler

```
INFO: [2024-08-29 20:40:34] manifest-validate - Monitoring manifest validation
INFO: [2024-08-29 20:40:34] httpx - HTTP Request: POST http://localhost:3001/v1/model/validate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&data_type=Patient&restrict_rules=True "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:40:35] httpx - HTTP Request: POST http://localhost:3001/v1/model/validate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&data_type=Patient "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:41:13] httpx - HTTP Request: POST http://localhost:3001/v1/model/validate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2Fncihtan%2Fdata-models%2Fmain%2FHTAN.model.jsonld&data_type=Biospecimen "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:41:13] manifest-submit - Monitoring manifest submission
INFO: [2024-08-29 20:41:58] httpx - HTTP Request: POST http://localhost:3001/v1/model/submit?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&dataset_id=syn51376664&asset_view=syn51376649&restrict_rules=True&use_schema_label=True&data_model_labels=class_label&table_manipulation=replace&manifest_record_type=table_and_file "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:42:10] httpx - HTTP Request: POST http://localhost:3001/v1/model/submit?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&dataset_id=syn51376664&asset_view=syn51376649&restrict_rules=True&use_schema_label=True&data_model_labels=class_label&table_manipulation=replace&manifest_record_type=file_only "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:42:22] httpx - HTTP Request: POST http://localhost:3001/v1/model/submit?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fdata_flow_config%2Fmain%2FHTAN%2Fdataflow_component.csv&dataset_id=syn51376664&asset_view=syn51376649&restrict_rules=True&use_schema_label=True&data_model_labels=class_label&table_manipulation=replace&manifest_record_type=file_only "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:42:24] manifest-storage - Monitoring storage endpoints
INFO: [2024-08-29 20:42:30] httpx - HTTP Request: GET http://localhost:3001/v1/storage/assets/tables?asset_view=syn23643253&return_type=json "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:42:36] httpx - HTTP Request: GET http://localhost:3001/v1/storage/project/datasets?asset_view=syn23643253&project_id=syn26251192 "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:43:03] httpx - HTTP Request: GET http://localhost:3001/v1/storage/project/datasets?asset_view=syn20446927&project_id=syn32596076 "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:43:03] manifest-generator - Monitoring manifest generation
INFO: [2024-08-29 20:43:18] httpx - HTTP Request: GET http://localhost:3001/v1/manifest/generate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&title=example&data_type=Patient "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:43:30] httpx - HTTP Request: GET http://localhost:3001/v1/manifest/generate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&title=example&data_type=Patient&output=excel "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:43:52] httpx - HTTP Request: GET http://localhost:3001/v1/manifest/generate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&title=example&data_type=Patient&output=excel&dataset_id=syn51078367&asset_view=syn23643253 "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:44:36] httpx - HTTP Request: GET http://localhost:3001/v1/manifest/generate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2Fncihtan%2Fdata-models%2Fmain%2FHTAN.model.jsonld&title=example&data_type=Patient "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:44:36] manifest-storage - Monitoring storage endpoints
INFO: [2024-08-29 20:44:42] httpx - HTTP Request: GET http://localhost:3001/v1/storage/assets/tables?asset_view=syn23643253&return_type=json "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:44:48] httpx - HTTP Request: GET http://localhost:3001/v1/storage/project/datasets?asset_view=syn23643253&project_id=syn26251192 "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:45:14] httpx - HTTP Request: GET http://localhost:3001/v1/storage/project/datasets?asset_view=syn20446927&project_id=syn32596076 "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:45:14] manifest-generator - Monitoring manifest generation
INFO: [2024-08-29 20:45:26] httpx - HTTP Request: GET http://localhost:3001/v1/manifest/generate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&title=example&data_type=Patient "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:45:38] httpx - HTTP Request: GET http://localhost:3001/v1/manifest/generate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&title=example&data_type=Patient&output=excel "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:46:01] httpx - HTTP Request: GET http://localhost:3001/v1/manifest/generate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&title=example&data_type=Patient&output=excel&dataset_id=syn51078367&asset_view=syn23643253 "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:46:45] httpx - HTTP Request: GET http://localhost:3001/v1/manifest/generate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2Fncihtan%2Fdata-models%2Fmain%2FHTAN.model.jsonld&title=example&data_type=Patient "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:46:45] manifest-submit - Monitoring manifest submission
INFO: [2024-08-29 20:47:29] httpx - HTTP Request: POST http://localhost:3001/v1/model/submit?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&dataset_id=syn51376664&asset_view=syn51376649&restrict_rules=True&use_schema_label=True&data_model_labels=class_label&table_manipulation=replace&manifest_record_type=table_and_file "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:47:42] httpx - HTTP Request: POST http://localhost:3001/v1/model/submit?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&dataset_id=syn51376664&asset_view=syn51376649&restrict_rules=True&use_schema_label=True&data_model_labels=class_label&table_manipulation=replace&manifest_record_type=file_only "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:47:55] httpx - HTTP Request: POST http://localhost:3001/v1/model/submit?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fdata_flow_config%2Fmain%2FHTAN%2Fdataflow_component.csv&dataset_id=syn51376664&asset_view=syn51376649&restrict_rules=True&use_schema_label=True&data_model_labels=class_label&table_manipulation=replace&manifest_record_type=file_only "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:47:57] manifest-validate - Monitoring manifest validation
INFO: [2024-08-29 20:47:57] httpx - HTTP Request: POST http://localhost:3001/v1/model/validate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&data_type=Patient&restrict_rules=True "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:47:58] httpx - HTTP Request: POST http://localhost:3001/v1/model/validate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2FSage-Bionetworks%2Fschematic%2Fdevelop%2Ftests%2Fdata%2Fexample.model.jsonld&data_type=Patient "HTTP/1.1 200 OK"
INFO: [2024-08-29 20:48:34] httpx - HTTP Request: POST http://localhost:3001/v1/model/validate?schema_url=https%3A%2F%2Fraw.githubusercontent.com%2Fncihtan%2Fdata-models%2Fmain%2FHTAN.model.jsonld&data_type=Biospecimen "HTTP/1.1 200 OK"
```

linglp:
@thomasyu888 thank you so much! This is amazing to see.

linglp (approved):
Thanks for your hard work on this Gianna. It looks good to me, and I am happy to see that the new test gets moved to integration folder as well.

GiaJordan:
@linglp @thomasyu888 @BryanFauble thank you all so much for your reviews, the comments you left were very helpful and I appreciate the time you took to give them

PR #1469, opened 2024-08-23
[FDS-2127] Update URL validation to use requests.options to verify connectivity

BryanFauble:
**Problem:**

1. Valid URLs like `https://doi.org/10.1158/0008-5472.can-23-0128` were being rejected as the `urlopen` was not connecting to the server hosting this webpage.

**Solution:**

1. Swap to `requests.options` that will verify the webserver in question is reachable. Read more about the options method here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS

**Testing:**

1. I set up a few integration tests around the changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ section. **Note:** You must place the file at the root of the project like
 
 <strong>Configure config.yml File</strong>
 
-``config_example.yml`` avaiable in the root of this repo. It contains an example repoistory
+There are some defaults in schematic that can be configured. These fields are in ``config_example.yml``:
 
 ```text
 
--- a/README.md
+++ b/README.md
@@ -2,17 +2,28 @@
 [![Build Status](https://img.shields.io/endpoint.svg?url=https%3A%2F%2Factions-badge.atrox.dev%2FSage-Bionetworks%2Fschematic%2Fbadge%3Fref%3Ddevelop&style=flat)](https://actions-badge.atrox.dev/Sage-Bionetworks/schematic/goto?ref=develop) [![Documentation Status](https://readthedocs.org/projects/sage-schematic/badge/?version=develop)](https://sage-schematic.readthedocs.io/en/develop/?badge=develop) [![PyPI version](https://badge.fury.io/py/schematicpy.svg)](https://badge.fury.io/py/schematicpy)
 
 # Table of contents
+- [Schematic](#schematic)
+- [Table of contents](#table-of-contents)
 - [Introduction](#introduction)
 - [Installation](#installation)
   - [Installation Requirements](#installation-requirements)
   - [Installation guide for Schematic CLI users](#installation-guide-for-schematic-cli-users)
   - [Installation guide for developers/contributors](#installation-guide-for-developerscontributors)
+    - [Development environment setup](#development-environment-setup)
+    - [Development process instruction](#development-process-instruction)
+    - [Example For REST API ](#example-for-rest-api-)
+      - [Use file path of `config.yml` to run API endpoints:](#use-file-path-of-configyml-to-run-api-endpoints)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -2,17 +2,28 @@
 [![Build Status](https://img.shields.io/endpoint.svg?url=https%3A%2F%2Factions-badge.atrox.dev%2FSage-Bionetworks%2Fschematic%2Fbadge%3Fref%3Ddevelop&style=flat)](https://actions-badge.atrox.dev/Sage-Bionetworks/schematic/goto?ref=develop) [![Documentation Status](https://readthedocs.org/projects/sage-schematic/badge/?version=develop)](https://sage-schematic.readthedocs.io/en/develop/?badge=develop) [![PyPI version](https://badge.fury.io/py/schematicpy.svg)](https://badge.fury.io/py/schematicpy)
 
 # Table of contents
+- [Schematic](#schematic)
```
BryanFauble, line 5: A have a markdown plugin in VSCODE that automatically creates/updates these. If you don't wish to have this update I will delete it from the PR.

tests/conftest.py
```diff
@@ -142,3 +142,9 @@ def temporary_file_copy(request, helpers: Helpers) -> Generator[str, None, None]
     # Teardown
     if os.path.exists(temp_csv_path):
         os.remove(temp_csv_path)
+
+
+@pytest.fixture(name="dmge", scope="function")
```
BryanFauble, line 147: I moved this up to conftest so I could use it in my new tests as well

tests/integration/test_validate_attribute.py
```diff
@@ -0,0 +1,73 @@
+import pandas as pd
```
BryanFauble, line 1: This is the start of moving some files into an `tests/integration` directory. We can start slow and add new items into the appropriate locations while we slowly move items over in a strangler fig pattern: https://shopify.engineering/refactoring-legacy-code-strangler-fig-pattern

schematic/models/validate_attribute.py
```diff
@@ -1207,7 +1209,7 @@ def url_validation(
                             errors.append(vr_errors)
                         if vr_warnings:
                             warnings.append(vr_warnings)
-                    if valid_url == True:
+                    if valid_url is True:
```
thomasyu888, line 1212: Nit: could we this just be 

```
if valid_url
```

tests/integration/test_validate_attribute.py
thomasyu888: I appreciate the additive nature of these changes

schematic/models/validate_attribute.py
```diff
@@ -1186,10 +1185,13 @@ def url_validation(
                     try:
                         # Check that the URL points to a working webpage
                         # if not log the appropriate error.
-                        request = Request(url)
-                        response = urlopen(request)
                         valid_url = True
-                        response_code = response.getcode()
+                        response = requests.options(url, allow_redirects=True)
```
thomasyu888, line 1189: For the FAIR team, was there a specific reason to use `urllib` other than the fact that it is a builtin package?

I _think_ something like this works too
```
from urllib import request

req = request.Request(url, method='OPTIONS')
response = request.urlopen(req)
print(response.status, response.headers)
```

thomasyu888 (approved):
🔥 Thanks for hopping on this so quick, I'm going to pre-approve but will leave it up to the FAIR team to approve.

schematic/models/validate_attribute.py
linglp: I see. I also prefer using requests package. What Brian has there seems perfect to me.

tests/integration/test_validate_attribute.py
linglp: Thanks Bryan! I like the change.

BryanFauble:
@thomasyu888 @linglp I moved this to a non-forked branch/PR: https://github.com/Sage-Bionetworks/schematic/pull/1472

PR #1472, opened 2024-08-26
[FDS-2127] Update URL validation to use requests.options to verify connectivity

BryanFauble:
**Problem:**

1. Valid URLs like `https://doi.org/10.1158/0008-5472.can-23-0128` were being rejected as the `urlopen` was not connecting to the server hosting this webpage.

**Solution:**

1. Swap to `requests.options` that will verify the webserver in question is reachable. Read more about the options method here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS

**Testing:**

1. I set up a few integration tests around the changes.
2. All tests (Except for 1) have passed: https://github.com/Sage-Bionetworks/schematic/actions/runs/10563739475/job/29264599046 - The one not passing was going to be addressed in https://github.com/Sage-Bionetworks/schematic/pull/1473

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ section. **Note:** You must place the file at the root of the project like
 
 <strong>Configure config.yml File</strong>
 
-``config_example.yml`` avaiable in the root of this repo. It contains an example repoistory
+There are some defaults in schematic that can be configured. These fields are in ``config_example.yml``:
 
 ```text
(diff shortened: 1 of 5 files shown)


thomasyu888 (commented):
🚄 LGTM! Thanks for doing this!

PR #1471, opened 2024-08-26
fix: updated test after updating to synapse python client 4.4.0 and updated test after test resources get updated on synapse

linglp:
## Problem: 
I noticed the following test failed in develop: 
```
FAILED tests/test_store.py::TestSynapseStorage::test_store_async_annotation - AssertionError: expected call not found.
```
And also this test failed: 
```
FAILED tests/test_manifest.py::TestManifestGenerator::test_get_manifest_with_files[File based] 
```
Related to: https://sagebionetworks.jira.com/browse/FDS-2304

## Reason 
left over from: https://github.com/Sage-Bionetworks/schematic/pull/1468. After updating to synapse python client, this test needs to be updated

And also the following tests were skipped in github action [run](https://github.com/Sage-Bionetworks/schematic/actions/runs/10550260545/job/29226221198): 
```
[gw0] [ 86%] SKIPPED tests/test_store.py::TestSynapseStorage::test_format_row_annotations_entity_id_trash_can[True-display_label] 
tests/test_store.py::TestSynapseStorage::test_format_row_annotations_entity_id_trash_can[False-display_label] 
[gw0] [ 86%] SKIPPED tests/test_store.py::TestSynapseStorage::test_format_row_annotations_entity_id_trash_can[False-display_label] 
tests/test_store.py::TestSynapseStorage::test_format_row_annotations_entity_id_trash_can[True-class_label] 
[gw0] [ 86%] SKIPPED tests/test_store.py::TestSynapseStorage::test_format_row_annotations_entity_id_trash_can[True-class_label] 
tests/test_store.py::TestSynapseStorage::test_format_row_annotations_entity_id_trash_can[False-class_label] 
[gw0] [ 86%] SKIPPED tests/test_store.py::TestSynapseStorage::test_format_row_annotations_entity_id_trash_can[False-class_label] 
tests/test_store.py::TestSynapseStorage::test_get_files_metadata_from_dataset 
[gw0] [ 86%] PASSED tests/test_store.py::TestSynapseStorage::test_get_files_metadata_from_dataset 
tests/test_store.py::TestSynapseStorage::test_get_async_annotation 
[gw0] [ 87%] SKIPPED tests/test_store.py::TestSynapseStorage::test_get_async_annotation 
tests/test_store.py::TestSynapseStorage::test_store_async_annotation 
[gw0] [ 87%] SKIPPED tests/test_store.py::TestSynapseStorage::test_store_async_annotation 
[...]

PR #1473, opened 2024-08-26
Update `test_get_manifest_with_files` to match changes on synapse

GiaJordan:
Update test_get_manifest_with_files to match changes on synapse so the test doesn't fail on other branches. Other changes will be included in #1467

### DIFF
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -798,4 +798,4 @@ class TestManifestGenerator:
             assert n_rows == 4
         elif component == "BulkRNA-seqAssay":
             assert filename_in_manifest_columns
-            assert n_rows == 3
+            assert n_rows == 4


tests/test_manifest.py
```diff
@@ -798,4 +798,4 @@ def test_get_manifest_with_files(self, helpers, component, datasetId):
             assert n_rows == 4
         elif component == "BulkRNA-seqAssay":
             assert filename_in_manifest_columns
-            assert n_rows == 3
+            assert n_rows == 4
```
thomasyu888, line 801: Nit: Does it make sense to map to specific fileview snapshot version so that this doesn't happen? 

In the long term future, it would be helpful to think about tests with setup and teardown methods so that they are stable.

thomasyu888:
Also separately, is there a Jira ticket associated with this?

tests/test_manifest.py
thomasyu888: The overall concern is that you want to know explicitly _why_ a test failed. It's not ideal to have to figure out that the test failing is due to something updated on Synapse not directly related to the code.

linglp: Yeah I completely agree with what Tom said. I could create a separate ticket for this: https://sagebionetworks.jira.com/browse/FDS-2306, and this would be a good way to drive to cleaner and more organized code.

linglp (approved):
approve this as an urgent fix. More improvement will be in separate ticket: https://sagebionetworks.jira.com/browse/FDS-2306

GiaJordan:
Closing, #1471 will address

PR #1474, opened 2024-08-26
[FDS-2294] Prevent including project name twice in walked path

BryanFauble:
**Problem:**

1. When the project structure was being walked the name of the project was being included twice: `schematic - main/schematic - main/DataTypeX/Sample_A.txt`
2. Lots of small issues flagged by sonar

**Solution:**

1. Add in a small check to determine if the project name is already being included in the directory structure, and not include the duplicated data if it's the case. I was not certain if I could totally remove adding the project name manually, so that is why I made the check in this way.

**Testing:**

1. I verified I could run the tests in question around this logic (Note: This does not contain all of the tests noted in the jira, just a few):
image

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -547,7 +547,8 @@ class SynapseStorage(BaseStorage):
         Raises:
             ValueError: Dataset ID not found.
         """
-        # select all files within a given storage dataset folder (top level folder in a Synapse storage project or folder marked with contentType = 'dataset')
+        # select all files within a given storage dataset folder (top level folder in
+        # a Synapse storage project or folder marked with contentType = 'dataset')
         walked_path = synapseutils.walk(
             self.syn, datasetId, includeTypes=["folder", "file"]
         )
@@ -557,25 +558,36 @@ class SynapseStorage(BaseStorage):
         file_list = []
 
         # iterate over all results
-        for dirpath, dirname, filenames in walked_path:
+        for dirpath, _, path_filenames in walked_path:
             # iterate over all files in a folder
-            for filename in filenames:
-                if (not "manifest" in filename[0] and not fileNames) or (
-                    fileNames and filename[0] in fileNames
+            for path_filename in path_filenames:
+                if ("manifest" not in path_filename[0] and not fileNames) or (
+                    fileNames and path_filename[0] in fileNames
                 ):
-                    # don't add manifest to list of files unless it is specified in the list of specified fileNames; return all found files
+                    # don't add manifest to list of files unless it is specified in the
+                    # list of specified fileNames; return all found files
                     # except the manifest if no fileNames have been specified
                     # TODO: refactor for clarity/maintainability
 
                     if fullpath:
                         # append directory path to filename
-                        filename = (
-                            project_name + "/" + dirpath[0] + "/" + filename[0],
-                            filename[1],
-                        )
+                        if dirpath[0].startswith(f"{project_name}/"):
+                            path_filename = (
+                                dirpath[0] + "/" + path_filename[0],
+                                path_filename[1],
+                            )
+                        else:
+                            path_filename = (
+                                project_name
+                                + "/"
+                                + dirpath[0]
+                                + "/"
+                                + path_filename[0],
+                                path_filename[1],
+                            )
 
                     # add file name file id tuple, rearranged so that id is first and name follows
-                    file_list.append(filename[::-1])
+                    file_list.append(path_filename[::-1])
 
         return file_list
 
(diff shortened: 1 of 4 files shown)


schematic/store/synapse.py
```diff
@@ -557,25 +558,36 @@ def getFilesInStorageDataset(
...
 
                     if fullpath:
                         # append directory path to filename
-                        filename = (
-                            project_name + "/" + dirpath[0] + "/" + filename[0],
-                            filename[1],
-                        )
+                        if dirpath[0].startswith(f"{project_name}/"):
+                            path_filename = (
+                                dirpath[0] + "/" + path_filename[0],
```
thomasyu888, line 576: Nit: Does it make sense to use os.path.join since we are making this change?

BryanFauble: No. If forward slash is the expected format then it'll be broken if the code runs on windows.

schematic_api/api/routes.py
```diff
@@ -596,7 +585,9 @@ def get_storage_projects_datasets(asset_view, project_id):
     return sorted_dataset_lst
 
 
-def get_files_storage_dataset(asset_view, dataset_id, full_path, file_names=None):
+def get_files_storage_dataset(
```
thomasyu888, line 588: Nit: may want to add type hints if you are touching this return type

tests/test_api.py
```diff
@@ -3,98 +3,85 @@
...
 
 
 @pytest.fixture(scope="class")
-def test_manifest_submit(helpers):
+def test_manifest_submit(helpers) -> str:
     test_manifest_path = helpers.get_data_path(
         "mock_manifests/example_biospecimen_test.csv"
     )
-    yield test_manifest_path
+    return test_manifest_path
```
thomasyu888, line 52: May I ask why switch to return?  The yield is like a "teardown" of resources

schematic/store/synapse.py
```diff
@@ -557,25 +558,36 @@ def getFilesInStorageDataset(
...
         # iterate over all results
-        for dirpath, dirname, filenames in walked_path:
+        for dirpath, _, path_filenames in walked_path:
             # iterate over all files in a folder
-            for filename in filenames:
-                if (not "manifest" in filename[0] and not fileNames) or (
-                    fileNames and filename[0] in fileNames
+            for path_filename in path_filenames:
+                if ("manifest" not in path_filename[0] and not fileNames) or (
+                    fileNames and path_filename[0] in fileNames
```
thomasyu888, line 565: Just making sure I have this right, this is to make sure `fileNames is not None` then it checks `path_filename[0] in fileNames`

tests/test_api.py
```diff
@@ -3,98 +3,85 @@
...
-from schematic.schemas.data_model_relationships import DataModelRelationships
-
+from schematic.schemas.data_model_parser import DataModelParser
 from schematic_api.api import create_app
 
-
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+BENCHMARK_DATA_MODEL_JSON_LD = "https://raw.githubusercontent.com/Sage-Bionetworks/schematic/develop/tests/data/example.single_rule.model.jsonld"
```
thomasyu888, line 23: Great simplification of tests by defining global variables

tests/test_api.py
BryanFauble: yield is preferred if it needs to be context managed or there is code after the yield to clean up the resources. https://docs.pytest.org/en/stable/how-to/fixtures.html#teardown-cleanup-aka-fixture-finalization covers it pretty well. Since in this case we are doing neither of these it makes the return type of the function a `str` instead of a `Generator[str, None, None]`. Just a bit clearer. Though if there are other reasons to use yield I don't mind either way.

schematic/store/synapse.py
BryanFauble: There are 2 things occurring here, either:

1. We include all files that were walked (Except for any files that contain the word "manifest"
2. We include only those files in the `fileNames` list

thomasyu888 (approved):
🤖 LGTM - but will leave it to FAIR for final review

schematic/store/synapse.py
```diff
@@ -3024,6 +3036,8 @@ def _fix_int_columns(self):
         for col in int_columns:
             # Coercing to string because NaN is a floating point value
             # and cannot exist alongside integers in a column
-            to_int_fn = lambda x: "" if np.isnan(x) else str(int(x))
+            def to_int_fn(x):
```
linglp, line 3039: here the function was defined but you didn't call the function again?

linglp, line 3039: when you define small in-line function like this, do you also add tests to this function?

schematic/store/synapse.py
```diff
@@ -557,25 +558,36 @@ def getFilesInStorageDataset(
...
-                        filename = (
-                            project_name + "/" + dirpath[0] + "/" + filename[0],
-                            filename[1],
-                        )
+                        if dirpath[0].startswith(f"{project_name}/"):
+                            path_filename = (
+                                dirpath[0] + "/" + path_filename[0],
+                                path_filename[1],
+                            )
+                        else:
```
linglp, line 579: I read unit test `test_getFilesInStorageDataset` again. I think that test only covers the "else" logic over here. Since the "if" logic gets introduced, could you also help modify `test_getFilesInStorageDataset`? Thank you!

linglp: to be more specific, the `mock_return` in `test_getFilesInStorageDataset` mock the return of `walk`. But now it 
only cover cases where the project names are not included. Could you add cases where the project names are included? so that the "if" statement will get triggered/covered in the test.

BryanFauble: Added https://github.com/Sage-Bionetworks/schematic/pull/1474/commits/6642bf7e0048d115466cd40995ebaa4398004c96

linglp (approved):
Thanks @BryanFauble! LGTM

PR #1476, opened 2024-08-27
[FDS-2294] Run schematic_api tests

BryanFauble:
**Problem:**

1. `schematic_api` tests were being ignored.
2. One of the integration tests was out of date for the new logic.
3. There are issues with the `tangled_tree` logic that was causing test failures. It was tracked down, context was added to the comment: https://github.com/Sage-Bionetworks/schematic/pull/1476#discussion_r1733406612

**Solution:**

1. unignore `schematic_api` tests
2. Ignore the benchmark tests which are long-running tests that should not really be included with the normal integration test runs
3. Update the logic around `tangled_tree`.

**Testing:**

1. I verified the logic with the return of the API data as well as manually via breakpoints and inspecting the data.
2. Will wait for a fully passing integration test suite.

image

Related to this ticket: FDS-2294

### DIFF
--- a/schematic/visualization/tangled_tree.py
+++ b/schematic/visualization/tangled_tree.py
@@ -3,10 +3,10 @@
 # pylint: disable=logging-fstring-interpolation
 # pylint: disable=too-many-lines
 
+import ast
 import json
 import logging
 import os
-import re
 from io import StringIO
 from os import path
 from typing import Literal, Optional, TypedDict, Union
@@ -243,25 +243,6 @@ class TangledTree:  # pylint: disable=too-many-instance-attributes
             cond_req = cond_req_new.replace("If", "").lstrip().rstrip()
         return cond_req
 
-    def _extract_content_within_quotes(self, strings: list[str]) -> list[str]:
-        """
-        Extract content within quotes from a list of strings.
-
-        Args:
-            strings (list): list of strings
-
-        Returns:
-            list: list of strings with content within single quotes
-        """
-        pattern = r"\'(.*?)\'"
-        extracted_content = []
-
-        for string in strings:
-            matches = re.findall(pattern, string)
-            extracted_content.extend(matches)
-
-        return extracted_content
-
     def _get_ca_alias(self, conditional_requirements: list[str]) -> dict[str, str]:
         """Get the alias for each conditional attribute.
 
@@ -277,9 +258,11 @@ class TangledTree:  # pylint: disable=too-many-instance-attributes
                 value: attribute
         """
         ca_alias: dict[str, str] = {}
-        extracted_conditional_requirements = self._extract_content_within_quotes(
-            conditional_requirements
-        )
+        extracted_conditional_requirements = []
+        for conditional_requirement in conditional_requirements:
+            extracted_conditional_requirements.extend(
+                ast.literal_eval(node_or_string=conditional_requirement)
+            )
 
         # clean up conditional requirements
         conditional_requirements = [
--- a/schematic/visualization/tangled_tree.py
+++ b/schematic/visualization/tangled_tree.py
@@ -3,24 +3,24 @@
 # pylint: disable=logging-fstring-interpolation
 # pylint: disable=too-many-lines
 
-from io import StringIO
 import json
 import logging
 import os
+import re
+from io import StringIO
 from os import path
-from typing import Optional, Literal, TypedDict, Union
-from typing_extensions import assert_never
+from typing import Literal, Optional, TypedDict, Union
 
 import networkx as nx  # type: ignore
-from networkx.classes.reportviews import NodeView, EdgeDataView  # type: ignore
 import pandas as pd
+from networkx.classes.reportviews import EdgeDataView, NodeView  # type: ignore
+from typing_extensions import assert_never
 
-from schematic.visualization.attributes_explorer import AttributesExplorer
-from schematic.schemas.data_model_parser import DataModelParser
 from schematic.schemas.data_model_graph import DataModelGraph, DataModelGraphExplorer
+from schematic.schemas.data_model_parser import DataModelParser
 from schematic.utils.io_utils import load_json
 from schematic.utils.schema_utils import DisplayLabelType
-
+from schematic.visualization.attributes_explorer import AttributesExplorer
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


tests/test_api.py
```diff
@@ -773,9 +773,9 @@ def test_generate_manifest_file_based_annotations(
 
         # make sure Filename, entityId, and component get filled with correct value
         assert google_sheet_df["Filename"].to_list() == [
-            "TestDataset-Annotations-v3/Sample_A.txt",
-            "TestDataset-Annotations-v3/Sample_B.txt",
-            "TestDataset-Annotations-v3/Sample_C.txt",
+            "schematic - main/TestDataset-Annotations-v3/Sample_A.txt",
```
BryanFauble, line 776: The code ends up calling to `getFilesInStorageDataset` where the file names are retrieved. In that function we are including the project within the file name.

schematic/visualization/tangled_tree.py
```diff
@@ -258,11 +277,14 @@ def _get_ca_alias(self, conditional_requirements: list[str]) -> dict[str, str]:
                 value: attribute
         """
         ca_alias: dict[str, str] = {}
+        extracted_conditional_requirements = self._extract_content_within_quotes(
```
BryanFauble, line 280: Problem this is addressing:
```
  File "/home/bfauble/BryansGreatWorkspace/schematic/schematic/visualization/tangled_tree.py", line 292, in _get_ca_alias
    attr, ali = req.split(" is ")
ValueError: too many values to unpack (expected 2)
```

I was seeing some test failures pop up in the logic `attr, ali = req.split(" is ")` below.

The reason is that in an integration test `.split` was splitting and returning back more than 2 values on this input:

- `['[\'File Format is "CRAM"\', \'File Format is "BAM"\', \'File Format is "CSV/TSV"\']', '[\'File Format is "CRAM"\']']`

**There are a few problems for this put:**

1. It contained multiple strings of `" is "`, causing the immediate exception to be thrown.
2. Nothing in this code was properly formatting the content out. The result return statement of this function was:

**After this change, the output of the function is:**

Input: `['[\'Diagnosis is "Cancer"\']', '[\'Diagnosis is "Cancer"\']']`
Output: `{'Cancer': 'Diagnosis'}`

Input: `['[\'File Format is "CSV/TSV"\', \'File Format is "BAM"\', \'File Format is "CRAM"\']', '[\'File Format is "CRAM"\']']`
Output: `{'CSV/TSV': 'FileFormat', 'BAM': 'FileFormat', 'CRAM': 'FileFormat'}`


**Example of malformed content:**
Input: `['[\'Diagnosis is "Cancer"\']', '[\'Diagnosis is "Cancer"\']']`
Output: `{'"Cancer"\']': "['Diagnosis"}`

schematic/visualization/tangled_tree.py
```diff
@@ -243,6 +243,25 @@ def _remove_unwanted_characters_from_conditional_statement(
...
+        """
+        Extract content within quotes from a list of strings.
+
+        Args:
+            strings (list): list of strings
+
+        Returns:
+            list: list of strings with content within single quotes
+        """
+        pattern = r"\'(.*?)\'"
```
thomasyu888, line 256: Do we need to account for double quotes?

thomasyu888 (approved):
🔥LGTM - the code looks good, but ill leave it to FAIR team for the logic

schematic/visualization/tangled_tree.py
BryanFauble: Not from all of the tests that ran, they were all consistent in that the content was all within an escaped single quote, like: `['[\'Diagnosis is "Cancer"\']', '[\'Diagnosis is "Cancer"\']']`


If I go further up in the code where the list is originally created:
```
            conditional_requirements = list(
                attributes_df[
                    (attributes_df["Cond_Req"])
                    & (attributes_df["Component"] == component_name)
                ]["Conditional Requirements"]
            )
```

It seems like this is how pandas is extracting this content out.

BryanFauble: I swapped to using `ast.literal_eval`: https://docs.python.org/3/library/ast.html#ast.literal_eval

I am using it within the synapse python client for a similar purpose when translating CSV content into python types.

schematic/visualization/tangled_tree.py
```diff
@@ -258,11 +258,16 @@ def _get_ca_alias(self, conditional_requirements: list[str]) -> dict[str, str]:
                 value: attribute
         """
         ca_alias: dict[str, str] = {}
+        extracted_conditional_requirements = []
```
linglp, line 261: Thank you so much for the change @BryanFauble ! I am curious if this part should be wrapped in a function? If this part gets wrapped in a function, we could test it with different conditional requirements string and make sure they are all formatted correctly.

linglp: I tested it with some conditional requirements from HTAN data model, and they all worked!

linglp (approved):
Thanks! I am going to pre-approve too!

tests/test_api.py
```diff
@@ -773,9 +773,9 @@ def test_generate_manifest_file_based_annotations(
```
linglp, line 773: Also, since we now have "integration" folder for all the integration test, should we move this file there too?

linglp:
@BryanFauble I also mentioned the Jira ticket number in the PR description so that this PR is automatically links to the ticket.

linglp:
> > @BryanFauble I also mentioned the Jira ticket number in the PR description so that this PR is automatically links to the ticket.
> 
> @linglp DPE adds the Jira ticket at the Github PR name `[FDS-2294] Run schematic_api tests` and that adds the Jira linkage and it its also in the branch name

thanks for clarifying. I just noticed that

schematic/visualization/tangled_tree.py
linglp: I added unit test for `_get_ca_alias` function: https://github.com/Sage-Bionetworks/schematic/pull/1480
Maybe that will help too?

BryanFauble: Thanks - I'll merge this in

tests/test_api.py
BryanFauble: There are a lot of interconnected pieces with these tests. I would need to spend some time to pull out specific components from this file and move everything over. I think it would be better to have some tech debt items that focus on moving items over.

I do agree though that (most) new tests should be created in the respective directories.

PR #1478, opened 2024-08-27
fix: call get_node_validation_rules with appropriate parameters when storing annotations

GiaJordan:
# Problem
When storing annotations when user use option: `-ak display_label`, the `display_label` option wasn't got passed correctly in schematic. Thus the following error was triggered:
```
ERROR:schematic_api.api:Traceback (most recent call last):
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/store/synapse.py", line 1845, in _process_store_annos
    annos = completed_task.result()
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/store/synapse.py", line 404, in wrapper
    return await method(*args, **kwargs)
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/store/synapse.py", line 1588, in format_row_annotations
    "list", dmge.get_node_validation_rules(anno_k)
  File "/Users/lpeng/Documents/schematic-git2/schematic/schematic/schemas/data_model_graph.py", line 792, in get_node_validation_rules
    node_validation_rules = self.graph.nodes[node_label]["validationRules"]
  File "/Users/lpeng/Library/Caches/pypoetry/virtualenvs/schematicpy-9OomxyhV-py3.10/lib/python3.10/site-packages/networkx/classes/reportviews.py", line 194, in __getitem__
    return self._nodes[n]
KeyError: 'individualID'
```
The root cause is here when get_node_validation_rules function gets called: 
```
                elif (
                    isinstance(anno_v, str)
                    and re.fullmatch(csv_list_regex, anno_v)
                    and rule_in_rule_list(
                        "list", dmge.get_node_validation_rules(anno_k)
                    )
                ):
                    annos[anno_k] = anno_v.split(",")
```
This `dmge.get_node_validation_rules(anno_k)`  function has two parameters: node_label and node_display_name. But in this case, only node_label parameter was used. 

# Solution 
Make sure that `display_label` option gets passed down correctly when calling `format_row_annotations`

# Changes:
* updated get_node_validation_rules to be called with the right parameter
* wrapped a long loop from the original format_row_annotations function and created a new function called `process_row_annotations`
* cleaned up the if else statement in the original format_row_annotations function
* added tests

Note: there is an issue when submitting the manifest using the file paths that don't include the project name. This issue will be addressed in https://sagebionetworks.jira.com/browse/FDS-2378


schematic/store/synapse.py
```diff
@@ -1861,7 +1871,17 @@ async def add_annotations_to_entities_files(
             # Merge dataframes to add entityIds
             manifest = manifest.merge(
                 file_df, how="left", on="Filename", suffixes=["_x", None]
-            ).drop("entityId_x", axis=1)
+            )
+
+            # drop the duplicate entity column with NA values
+            col_to_drop = "entityId_x"
+            if manifest.entityId.isnull().all():
+                col_to_drop = "entityId"
```
thomasyu888, line 1879: Will there ever be a time when there is a mixture of NA and non-NA values within this column?

schematic/store/synapse.py
```diff
@@ -1537,14 +1540,21 @@ async def format_row_annotations(
                 annos.pop(anno_k) if anno_k in annos.keys() else annos
             # Otherwise save annotation as approrpriate
             else:
+                if annotation_keys == "display_label":
```
thomasyu888, line 1543: You may have picked this up as you've gone on your writing test journey, is that one big aspect of writing code is to write the code in a way that is more easily testable.  

There are probably some strategies to make these parts of code testable or at least this logic you have added.

BryanFauble, line 1543: I think you'll find this video very interesting @GiaJordan https://www.youtube.com/watch?v=CFRhGnuXG-4

They take it to an extreme, however, the principals are solid. I find that there is a good in-between where you are extracting out nested logic into it's own functional blocks, but not being so strict that simple things need to have the extra boilerplate code.

Clarity of the code helps debugging quite a lot. It can also lead to identifying better/different solutions to the problem.

GiaJordan:
@thomasyu888 @BryanFauble Thank you so much for the feedback!
I hadn't had the chance to mark this PR as draft because of the other issues that have come up but it is still very much a work in progress. The code as it is right now was just because I wanted to verify that that would fix the issue the user described, I plan on going back to format it differently so that it's easier to test and read. I had just noticed that the test that currently covers this method was being skipped at every run and I wanted to solve the issue of being able to run that test so that I could begin testing the changes here better.

schematic/store/synapse.py
GiaJordan: From my understanding of how `pd.df.merge` works, no. It should always be one or the other completely blank

GiaJordan:
@linglp

schematic/store/synapse.py
```diff
@@ -1903,7 +1913,17 @@ async def add_annotations_to_entities_files(
             # Merge dataframes to add entityIds
             manifest = manifest.merge(
                 file_df, how="left", on="Filename", suffixes=["_x", None]
-            ).drop("entityId_x", axis=1)
+            )
+
+            # drop the duplicate entity column with NA values
+            col_to_drop = "entityId_x"
```
GiaJordan, line 1919: To add context @linglp , after the noted issue was addressed I ran into another issue with the user's manifest where the column containing the entityIds was dropped after the merge, this ensures that other groups wont run into that issue. @thomasyu888 should I open a ticket to track that work as well?

linglp: From my perspective, the problem of "the user's manifest where the column containing the entityIds were dropped after the merge" is a separate problem than get_node_validation_rules wasn't called with appropriate parameters. If resolving FDS-2268 means resolving both, we could use the same ticket. But if the merging problem is separate from FDS-2268, there should be a separate ticket.

schematic/store/synapse.py
```diff
@@ -1517,6 +1520,59 @@ async def store_async_annotation(self, annotation_dict: dict) -> Annotations:
...
+            annotation_keys (str): display_label/class_label
+
+        Returns:
+            dict: annotations as a dictionary
+        """
+        for anno_k, anno_v in metadataSyn.items():
+            # Remove keys with nan or empty string values from dict of annotations to be uploaded
+            # if present on current data annotation
+            if hideBlanks and (
+                (isinstance(anno_v, str) and anno_v.strip() == "")
```
linglp, line 1549: this is new in comparison with the old one. Here I added functionality to handle long white space

schematic/store/synapse.py
linglp: This turns out to be a different issue. Here's the new ticket: https://sagebionetworks.jira.com/browse/FDS-2378

thomasyu888: Thanks @GiaJordan and @linglp!  Yes, agreed that it should be tracked in a separate ticket and we can bring it up during sprint planning for prioritization.

schematic/store/synapse.py
```diff
@@ -1517,6 +1517,60 @@ async def store_async_annotation(self, annotation_dict: dict) -> Annotations:
...
+        annotation_keys: str,
+    ) -> Dict[str, Any]:
+        """processes metadata annotations
+
+        Args:
+            dmge (DataModelGraphExplorer): data model graph explorer
+            metadata_syn (dict): metadata used for Synapse storage
+            hideBlanks (bool): if true, does not upload annotation keys with blank values.
+            csv_list_regex (str):
+            annos (dict):
```
andrewelamb, line 1536: Could you complete the docstring? It's missing some descriptions

schematic/store/synapse.py
```diff
@@ -1517,6 +1517,60 @@ async def store_async_annotation(self, annotation_dict: dict) -> Annotations:
...
+            dict: annotations as a dictionary
+        """
+        for anno_k, anno_v in metadata_syn.items():
+            # Remove keys with nan or empty string values from dict of annotations to be uploaded
+            # if present on current data annotation
+            if hide_blanks and (
+                (isinstance(anno_v, str) and anno_v.strip() == "")
+                or (isinstance(anno_v, float) and np.isnan(anno_v))
+            ):
+                annos.pop(anno_k) if anno_k in annos.keys() else annos
```
andrewelamb, line 1549: I don't understand this line, could you include a comment?

schematic/store/synapse.py
```diff
@@ -1517,6 +1517,60 @@ async def store_async_annotation(self, annotation_dict: dict) -> Annotations:
...
+
+        Returns:
+            dict: annotations as a dictionary
+        """
+        for anno_k, anno_v in metadata_syn.items():
+            # Remove keys with nan or empty string values from dict of annotations to be uploaded
+            # if present on current data annotation
+            if hide_blanks and (
+                (isinstance(anno_v, str) and anno_v.strip() == "")
+                or (isinstance(anno_v, float) and np.isnan(anno_v))
```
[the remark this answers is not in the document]
linglp: I kept this part almost the same as the original function (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L1581))

schematic/store/synapse.py
linglp: Good catch. I will add the description.

schematic/store/synapse.py
linglp: the comment was added in line 1543: https://github.com/Sage-Bionetworks/schematic/pull/1478/files#diff-7b756ab3c7de9e192e644befb74f09e376c916bb68bc3aefb66989510371eba3R1543

I also kept this part the same as the original function: https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L1583

schematic/store/synapse.py
```diff
@@ -1517,6 +1517,60 @@ async def store_async_annotation(self, annotation_dict: dict) -> Annotations:
...
+    def process_row_annotations(
+        self,
+        dmge: DataModelGraphExplorer,
+        metadata_syn: Dict[str, Any],
+        hide_blanks: bool,
+        csv_list_regex: str,
+        annos: Dict[str, Any],
+        annotation_keys: str,
+    ) -> Dict[str, Any]:
+        """processes metadata annotations
```
BryanFauble, line 1529: Could you update the docstring with a small bullet point list on what it means to "processes metadata annotations"?

schematic/store/synapse.py
```diff
@@ -1517,6 +1517,60 @@ async def store_async_annotation(self, annotation_dict: dict) -> Annotations:
...
+            ):
+                annos.pop(anno_k) if anno_k in annos.keys() else annos
+                continue
+
+            # Otherwise save annotation as approrpriate
+            if isinstance(anno_v, float) and np.isnan(anno_v):
+                annos[anno_k] = ""
+                continue
+
+            # Handle strings that match the csv_list_regex and pass the validation rule
```
BryanFauble, line 1557: Comments like this might be better within the docstring. This is a decent breakdown in the difference in purpose of the two and I found:



| Feature   | Comments                                                                                                                                  | Docstrings                                                                                                                                                                                                 |
|-----------|-------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Purpose   | Explains code purpose, provides context, adds reminders, etc.                                                                             | Programmatically documents code; should explain what the code does, can list inputs, outputs, and returns.                                                                                                 |
| Syntax    | Single-line comments: <br> `#` followed by a blank space. <br> Multi-line comments: <br> text is enclosed by `"""`, or each line is preceded by a `#` followed by a space. | Both multi-line and one-liner docstrings are enclosed by `"""`.                                                                                                                                            |
| Location  | Any place that is not immediately 1) after another docstring or, 2) the initial string following the definition of a module, function, or class is a place where a comment may be inserted. | The initial literal string that follows the definition of a module, function, or class and the first string after some other docstring is considered a docstring.                                          |
| Scope     | Only present in Python source code, removed at compile-time.                                                                               | Can be accessed by various automated documentation generators in both source code and bytecode.                                                                                                            |

tests/integration/test_store_synapse.py
```diff
@@ -0,0 +1,125 @@
...
+from schematic.store.synapse import SynapseStorage
+from schematic.utils.validate_utils import comma_separated_list_regex
+from tests.conftest import Helpers
+
+
+@pytest.fixture(name="dmge", scope="function")
+def DMGE(helpers: Helpers) -> DataModelGraphExplorer:
+    """Fixture to instantiate a DataModelGraphExplorer object."""
+    dmge = helpers.get_data_model_graph_explorer(path="example.model.jsonld")
+    return dmge
```
BryanFauble, line 16: This does not need to be copied into individual files. I have included this code within the `conftest.py` file

BryanFauble (approved):
A few formatting comments, but otherwise the logic looks good to me!

andrewelamb:
@linglp Could you resolve the comments that are alreayd here? Other than that I think it's good to go, and since I'll be out the rest of the week I'll approve.

andrewelamb (approved):

schematic/store/synapse.py
linglp: Thanks for the suggestion Bryan! This makes sense. The if else logic in the original code was pretty crazy and hard to understand. I will provide a mermaid diagram for this function so that the logic is clearer.

tests/integration/test_store_synapse.py
linglp: Thanks Bryan. I will remove it.

schematic/store/synapse.py
linglp: I am resolving this conversation to let this PR gets merged. But we are not dealing with the "merge" issue in this PR any more.

PR #1480, opened 2024-08-28
fix: add unit test for function get_ca_alias

linglp:
Problem: 
I noticed that get_ca_alias doesn't have a unit test.

Solution: 
added unit test for get_ca_alias and make sure different situations get covered.

### DIFF
--- a/schematic/visualization/tangled_tree.py
+++ b/schematic/visualization/tangled_tree.py
@@ -276,15 +276,12 @@ class TangledTree:  # pylint: disable=too-many-instance-attributes
                 attr = "".join(attr.split())
                 ali = self._strip_double_quotes(ali)
                 ca_alias[ali] = attr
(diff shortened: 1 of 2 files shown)


linglp:
Note: @milen-sage I also looked into HTAN data model because they have a lot of conditional requirements.. but interestingly, I didn't find anything that could potentially [trigger the "else" statement](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/visualization/tangled_tree.py#L274) in _get_ca_alias.. that made me wonder what was the initial edge cases that make the else statement exists.. is it multi-select?

BryanFauble (approved):
This is awesome! Thanks for adding these.

milen-sage:
But whether we have a data model that runs in this case now, it's a separate question. We can check if our example data model does when visualized.

PR #1483, opened 2024-08-28
fix: added unit test for get_node_validation_rules function 

linglp:
Problem: 
unit test for get_node_validation_rules was never added after schema refactor.. and due to Mialy's departure, the work of adding the unit test for it was delayed. But this is an important function being used for get validation rules for nodes in multiple parts of the code base. 

Also, I noticed that if node_label is not valid, currently the code is likely to raise a key error which is confusing. I changed the error message so that a value error can be raised. 

See issue here: https://sagebionetworks.jira.com/browse/FDS-2324

Solution: 
- updated the code to add tests
- provided better error handling when node label and display label are both not provided
- provided better error handling when node label doesn't exist in the graph


tests/test_schemas.py
```diff
@@ -616,8 +616,82 @@ def test_get_node_range(self):
     def test_get_node_required(self):
         return
 
-    def test_get_node_validation_rules(self):
-        return
+    @pytest.mark.parametrize(
```
BryanFauble, line 619: If it isn't too much trouble to extract this out to the `tests/integration` let's move forward on adding new items in that directory.

tests/test_schemas.py
```diff
@@ -616,8 +616,82 @@ def test_get_node_range(self):
...
+            ),
+        ],
+    )
+    def test_get_node_validation_rules_valid(
+        self,
+        helpers,
+        data_model,
+        node_label,
+        node_display_name,
+        expected_validation_rule,
```
BryanFauble, line 656: nit: typing for arguments

schematic/schemas/data_model_graph.py
```diff
@@ -780,16 +779,26 @@ def get_node_validation_rules(
...
+            # try search node label using display name
             node_label = self.get_node_label(node_display_name)
 
         if not node_label:
             return []
 
-        node_validation_rules = self.graph.nodes[node_label]["validationRules"]
+        try:
+            node_validation_rules = self.graph.nodes[node_label]["validationRules"]
+        except:
```
BryanFauble, line 798: Is there a smaller more targeted exception that can be caught here rather than a catch all?

tests/test_schemas.py
linglp: hmm I think this is a unit test since I am only testing `get_node_validation_rules` function, and integration tests typically involve more than one function?

tests/test_schemas.py
linglp: Thanks. Just finished adding type hinting.

schematic/schemas/data_model_graph.py
linglp: Yeah. I just added `KeyError`

tests/test_schemas.py
BryanFauble: Unit tests need to be able to run on an air-gapped machine. Meaning that it does not depend on any outside connection to the world. Within the python client we are enforcing this by disabling networking: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/tests/unit/conftest.py#L18-L35


For this - If it does connect to any other services then it is an integration test.

If it does not connect to any other services then it's a bit subjective and does fall into your idea of "involves more than one function", but not always.

BryanFauble (approved):
1 open comment chain, but I will go ahead and approve as the logic we want is being tested.

tests/test_schemas.py
linglp: @BryanFauble Thanks for pointing this out. I tried adding the following: 
```
        from pytest_socket import SocketBlockedError, disable_socket
        disable_socket(allow_unix_socket=True)
```
into the middle of my test `test_get_node_validation_rules_valid`, and it still passed. I think this means that this test is indeed a unit test :)

linglp: Let me add this into the `unit` folder then.

PR #1485, opened 2024-08-29
update how annotations are accessed

GiaJordan:
This PR updates how annotations from synapse are accessed by schematic. In `SynapaseStorage.format_row_annotations` we switched from using `syn.get_annotations` to using `SynapseStorage.get_async_annotations`. The latter gets the annotations from the API and the former gets them from the client.

Using the client annotations are formated:
`{"ID":[], <file annotations as key value pairs>}`
Using the API the annotations are formatted:
`{'annotations':{'id': value,... 'annotations': {<file annotations as key value pairs>}}}`

Schematic has been looking for annotations in the first level of the dict, this PR updates it to look for them in the nested dict where they are located now

----

Jenny edits:

I've wrapped up this PR with some minor changes:

    Merge with the new develop
    Added a new manifest file input - filepath_submission_test_manifest_sampleidx10.csv - which annotates the same synIDs but changes the SampleID annotation. This serves the purpose of ensuring that the annotations we see on Synapse are indeed the annotations that were uploaded during the current test run, and not a relic of a previous successful run.
    Re-introduce validating the uploaded annotations with the one from the mock_manifests manifest CSV
    Updated the Sample ID column to SampleID. The extra space was preventing being able to call the column by its name using the dataframe object, so it was being called with e.g. df._2. Now we can call with df.SampleID
    Implemented asyncio logic to retrieve the annotations from Synapse concurrently. This adds code complexity without improving the duration of the test by more than 2 seconds, so I lean towards removing it before merge, but I'm open to discussion.
    (out of scope) - upload-artifacts@v2 is no longer supported so I updated to v4 (more info [here](https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/))
    (out of scope) - The logic that caches the virtual environment was removed from test.yml


tests/integration/test_metadata_model.py
```diff
@@ -25,31 +27,45 @@ def test_submit_filebased_manifest(self, helpers, mocker: MockerFixture):
...
 
-            # AND the manifest should be uploaded as a CSV
-            spy_upload_file_as_csv.assert_called_once()
-            # AND annotations should be added to the files
-            spy_add_annotations.assert_called_once()
+        # AND the annotations should have the correct metadata
+        for row in manifest.itertuples():
+            entityId = row.entityId
+            expected_sample_id = row._2
+            annos = synapse_store.syn.get_annotations(entityId)
```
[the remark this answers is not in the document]
jaymedina: Hi @BryanFauble thanks for the suggestion, I went ahead and implemented it [here](https://github.com/Sage-Bionetworks/schematic/pull/1485/commits/6d76a93478e6b12e50a3f422e6593251fd5de501), however I found that it doesn't speed up the test by much - at most it seems to have shaved off 2 seconds from the 26 second duration. It seems the bottleneck is at the upload part which cannot be multi-threaded.

With the amount of complexity this new logic adds to the test, I don't think it's worth it to keep this as an async test, but let me know what you think.

jaymedina:
I've wrapped up this PR with some minor changes:

1. Merge with the new `develop`
2. Added a new manifest file input - `filepath_submission_test_manifest_sampleidx10.csv` - which annotates the same synIDs but changes the `SampleID` annotation. This serves the purpose of ensuring that the annotations we see on Synapse are indeed the annotations that were uploaded during the current test run, and not a relic of a previous successful run.
3. Re-introduce validating the uploaded annotations with the one from the `mock_manifests` manifest CSV
4. Updated the `Sample ID` column to `SampleID`. The extra space was preventing being able to call the column by its name using the dataframe object, so it was being called with e.g. `df._2`. Now we can call with `df.SampleID`
5. Implemented `asyncio` logic to retrieve the annotations from Synapse concurrently. This adds code complexity without improving the duration of the test by more than 2 seconds, **so I lean towards removing it before merge, but I'm open to discussion**.

tests/data/mock_manifests/filepath_submission_test_manifest.csv
```diff
@@ -1,3 +1,3 @@
-Filename,Sample ID,File Format,Component,Genome Build,Genome FASTA,Id,entityId
+Filename,SampleID,File Format,Component,Genome Build,Genome FASTA,Id,entityId
```
thomasyu888, line 1: May I ask why the change to `SampleID`?

tests/integration/test_metadata_model.py
```diff
@@ -71,14 +95,34 @@ def test_submit_filebased_manifest(
                 dataset_scope=dataset_scope,
             )
 
-            # AND the manifest should be submitted to the correct place
-            assert manifest_id == expected_manifest_id
+        # AND the files should be annotated
+        spy_add_annotations.assert_called_once()
+
+        import asyncio
```
thomasyu888, line 101: Nit: Should we move this import statement up top?

tests/data/mock_manifests/filepath_submission_test_manifest.csv
jaymedina: I updated the Sample ID column to SampleID because extra space was preventing being able to call the column by its name using the dataframe object, so it was being called with e.g. df._2. Now we can call with df.SampleID

andrewelamb: Yeah I'm worried this is going to break other tests...

jaymedina: Hi @andrewelamb - you raise a good point, however this test is the only one that references `filepath_submission_test_manifest.csv`, which is why it's best to change it now in case future integration tests reference it
image

This is what the search for that filename came up on GitHub

thomasyu888: All the tests seemed to pass via CI/CD, but can we triple check "Sample ID" is not referenced in the test.

Separately, if it was being referenced as `df._2`, were you seeing that in the tests?

andrewelamb: Awesome thanks for checking!

jaymedina: > can we triple check "Sample ID" is not referenced in the test?

I can confirm `Sample ID` is not referenced in this integration test in its current state.

>  if it was being referenced as df._2, were you seeing that in the tests?

You can view [this older commit ](https://github.com/Sage-Bionetworks/schematic/pull/1485/commits/48f10c2a69bd9a159e10ed8ad342b780262eb69d)which shows show the `Sample ID` column was referenced previously (line 60).

The `_2` is a byproduct of using `itertuples` on the dataframe. Here is a simple CSV file I made to recreate this:

image

andrewelamb: I want to get @linglp 's take on this before any approval. We allow users to have spaces in their manifest column names, so I want to make sure we aren't messing that up.

tests/integration/test_metadata_model.py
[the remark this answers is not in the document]
BryanFauble: No worries then! Thanks for trying it

BryanFauble (approved):
The logic is the changes looks great! I'll let fair folks give their final review though on the testing changes :D

tests/data/mock_manifests/filepath_submission_test_manifest.csv
jaymedina: Sounds good! Separately, I can add another integration test that covers for cases where the CSV files have column names with spaces.

jaymedina:
There's a new test that's breaking due to a change on the Synapse end. The test that's breaking is [test_validate.py/test_filename_manifest](https://github.com/Sage-Bionetworks/schematic/blob/b038bf63db016f5ae2afd43fca22f1d1a748036c/tests/test_validation.py#L682)

Here is the expected error ([source](https://github.com/Sage-Bionetworks/schematic/blob/b038bf63db016f5ae2afd43fca22f1d1a748036c/tests/test_validation.py#L716)):

```
assert (
            GenerateError.generate_filename_error(
                val_rule="filenameExists",
                attribute_name="Filename",
                row_num="4",
                invalid_entry="schematic - main/MockFilenameComponent/txt5.txt",
                error_type="path does not exist",
                dmge=dmge,
            )[0]
            in errors
        )
```

Here is the actual error I'm getting:
```
In [7]: errors
Out[7]: 
[['3',
  'Filename',
  "The entityId for file path 'schematic - main/MockFilenameComponent/txt4.txt' on row 3 does not match the entityId for the file in the file view",
  'schematic - main/MockFilenameComponent/txt4.txt'],
 ['4',
  'Filename',
  "The entityId for file path 'schematic - main/MockFilenameComponent/txt5.txt' on row 4 does not match the entityId for the file in the file view",
  'schematic - main/MockFilenameComponent/txt5.txt']]
  ```

The file was added yesterday afternoon (my time) so the timing checks out with when this test started breaking ([link](https://www.synapse.org/Synapse:syn61682648)):

image

I can patch this up, but will do so on another PR to avoid scope creep on this one.

We might want to create standalone projects for some of these major test areas to avoid unexpected changes on Synapse. Or add logic to `helpers` that tears down and rebuilds for every test (or chunk of tests) so we have a blank slate each time. @BryanFauble I remember you started talking about this on Slack, what are your thoughts on this?

**Comment by andrewelamb** [comment]
@jaymedina I was just looking at this myself

**Comment by andrewelamb** [comment]
@jaymedina No I haven't I think you're jsut a bit ahead of me :) I don't quite understand what's going on here, could you elaborate?

**Comment by jaymedina** [comment]
It looks like this test was written back when `txt5.txt` in this [dataset_scope](https://www.synapse.org/Synapse:syn61682648) did not exist, so one of the expected errors from `validateModelManifest` is a `"path does not exist"` error type. However, because this file exists as of yesterday, I am not getting this error when I re-create the logic of this test. Rather, I'm getting another mismatched entity error:

image

**Comment by thomasyu888** [comment]
Cc @BWMac on this, as I think he might be working on this logic related to filename/entity ids

**Comment by andrewelamb** [comment]
So to summarize, the txt5.txt file was added by you yesterday, but there was a test that was specifically checking for the lack of a txt5.txt file, is that correct?

**Comment by andrewelamb** [comment]
> My understanding of how these errors are generated is limited, but I know changing the expected error for `txt5.txt` to a `mismatched entityId` fixes the problem:

I'd rather not chnage this test. @BWMac Can you work around this test as-is for your work on filename/entity ids?

**Comment by jaymedina** [comment]
Another suggestion that I think would work is to keep the `txt5.txt` file, but name it something else.

**Comment by thomasyu888** [comment]
> Another suggestion that I think would work is to keep the `txt5.txt` file, but name it something else. But I realize these are band-aid fixes to the real problem which is tests communicating with the same entities on Synapse which are subject to change

See this: https://sagebionetworks.slack.com/archives/C07HLTBS5D2/p1726156665980609?thread_ts=1726000362.049489&cid=C07HLTBS5D2

File: tests/data/mock_manifests/filepath_submission_test_manifest.csv (hunk 2 above)
Comment (linglp) [reply tests/data/mock_manifests/filepath_submission_test_manifest.csv:1]: Apologies for the delayed response; I was out sick. we can't change the column name from Sample ID to SampleID. From what I understand, this manifest was generated using the example data model with the BulkRNA-seqAssay component data type. When I ran the manifest/generate function to create a manifest, the column headers I saw included "Sample ID" with the space. In this case, our test manifest should align with the output from manifest/generate, we need to keep the column name as is.

The issue that you refer @jaymedina , is because of this line: 
```
        manifest = helpers.get_data_frame(
            manifest_path, preserve_raw_input=False, **load_args
        )
```
If you change preserve_raw_input to True, and you also change how manifest was looped: 
```
for index, row in manifest.iterrows():
```
Does that solve the issue?

**Review by linglp — CHANGES_REQUESTED** [review]

File: tests/data/mock_manifests/filepath_submission_test_manifest_sampleidx10.csv [hunk 4]
```diff
@@ -0,0 +1,3 @@
+Filename,SampleID,File Format,Component,Genome Build,Genome FASTA,Id,entityId
```
linglp, line 1: @jaymedina , similarly here, since the test manifest is using component `BulkRNA-seqAssay`, the column name should be kept as "Sample ID"

tests/data/mock_manifests/ValidFilenameManifest.csv
```diff
@@ -1,5 +1,5 @@
-Component,Filename,Id,entityId
-MockFilename,schematic - main/TestSubmitMockFilename/txt1.txt,3c4c384b-5c49-4a7c-90a6-43f03a5ddbdc,syn62822369
-MockFilename,schematic - main/TestSubmitMockFilename/txt2.txt,3b45f5f3-408f-47ff-945e-9badf0a43195,syn62822368
-MockFilename,schematic - main/TestSubmitMockFilename/txt3.txt,2bbc898f-2651-4af3-834a-10c506de0fbd,syn62822366
-MockFilename,schematic - main/TestSubmitMockFilename/txt4.txt,5a2d3816-436e-458f-9887-cb8355518e23,syn62822364
+Component,SampleID,Filename,Id,entityId
```
linglp, line 1: here, since the component is `MockFilename`, I am fine with this column be kept as "SampleID".

tests/integration/test_metadata_model.py
```diff
@@ -22,6 +32,13 @@ class TestMetadataModel:
                 "syn62280543",
                 None,
             ),
+            (
```
linglp, line 35: here i am seeing three test cases here. Could you add some comments to explain what each test case is meant to test?

BryanFauble:
> We might want to create standalone projects for some of these major test areas to avoid unexpected changes on Synapse. Or add logic to helpers that tears down and rebuilds for every test (or chunk of tests) so we have a blank slate each time. @BryanFauble I remember you started talking about this on Slack, what are your thoughts on this?

I would like us to eventually pivot to use the dev instance of Synapse to table all of these tests. At that point in time we can separate out all of the tests to prevent the usage of shared resources. In theory we should allow for each and every test to run at the exact same moment in time, however, I am aware there are many race conditions present in the code that modify shared resources - So it's not exactly realistic right now.

@linglp @andrewelamb I created this ticket to track this - Probably not high priority, but something to keep in mind:
https://sagebionetworks.jira.com/browse/FDS-2414

tests/integration/test_metadata_model.py
```diff
@@ -54,14 +72,20 @@ def test_submit_filebased_manifest(
...
 
-        # WHEN the manifest it submitted
+        # WHEN the manifest it submitted and files are annotated
         # THEN submission should complete without error
         with does_not_raise():
             manifest_id = meta_data_model.submit_metadata_manifest(
-                manifest_path=manifest_path,
+                manifest_path=manifest_full_path,
                 dataset_id=dataset_id,
                 manifest_record_type="file_and_entities",
```
linglp, line 90: I am only seeing `manifest_record_type = "files_and_entities"` got tested. Can we also have `manifest_record_type="table_and_file"` got tested somewhere too? 

when submitting a manifest with parameter`manifest_record_type="table_and_file"`, this function `upload_manifest_as_table` (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L1946)) would get triggered. And this function would do the following in terms of adding annotations: 
* adding annotations to entities listed in the manifest (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L1997C22-L1997C55)) 
* setting annotations for the file itself (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L2017)) 
* setting annotations for the table (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L2017))

From my perspective, ideally the test case should test that annotations are added for all these three cases. Can we have this added in the PR, if not, can we open a new ticket? 

Also, when `manifest_record_type = "files_and_entities"`, when submitting a manifest, this function [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L2042) got called. And this function would do the following in terms of adding annotations: 
* adding annotations to entities listed in the manifest (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L2073))
* setting annotations for the file itself (see [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L2093))

By reading your test, I am confident that bullet point 1 is tested, but bullet point 2 is not. Can we add test for bullet point 2?

linglp:
@BWMac could you remove the txt5.txt here: https://www.synapse.org/Synapse:syn63106789
This is causing the test in develop branch to fail too: 
```
=============================================================== short test summary info ===============================================================
FAILED tests/test_validation.py::TestManifestValidation::test_filename_manifest - assert ['4', 'Filename', "The file path 'schematic - main/MockFilenameComponent/txt5.txt' on row 4 does not exist in the file view.", 'schematic -...
```
If this doesn't get resolved, other PRs also can't merge to develop. If you need to experiment, could you use a separate project or dataset? 

cc @jaymedina , @andrewelamb

BWMac:
Sorry for not chiming in sooner, my email notifications for this PR weren't working for some reason.

I have removed text5.txt, but I created the file to add new tests while continuing the naming convention for testing `filename_validation`, not simply to experiment.

tests/data/mock_manifests/filepath_submission_test_manifest.csv
jaymedina: Hi @linglp , yes switching to `iterrows()` and preserving the raw input solved this issue without needing to alter `Sample ID`, thanks!

tests/integration/test_metadata_model.py
linglp: Hi Jenny! Sorry for the late reply. I was out yesterday. Your understanding that each manifest file could only contain one component type is correct 👍 . 

For your second question, I just noticed that we don't have a unit test for it (which I just create a ticket for it: https://sagebionetworks.jira.com/browse/FDS-2421). From my understanding, this function `format_manifest_annotations` handles adding annotations for manifest file itself and also the table. Here's how it works: 
* when manifest_record_type = "file_only", The code (see the "if" statement there) in `format_manifest_annotation` makes sure that component in the manifest gets captured in the annotations. In this case, Component = `BulkRNA-seqAssay` got added as an annotation key pair. As you could see below, the manifest csv got annotated like below eventually: 
Screenshot 2024-09-17 at 3 33 10 PM


* when manifest_record_type = "table_and_file", we add annotations to table: 
Screenshot 2024-09-17 at 3 38 14 PM
But as you could see, it only contains etag and id

jaymedina: @linglp thank you for the breakdown, that makes sense. It looks like [the logic of adding the `Component` annotation](https://github.com/Sage-Bionetworks/schematic/blob/5721c77d35ce14b45c7bf768af133c6d81eb4abe/schematic/store/synapse.py#L1618) is only triggered under `is_file` and not `is_table`, which is why there's no `Component` annotation in the tables. 

What is the purpose of adding this `Component` annotation to the manifest file itself? My first assumption would be: because it makes it easier to see what model is being used to annotate the entities, but then that raises the question of: why are the tables not annotated the same way?

tests/integration/test_metadata_model.py
```diff
@@ -12,36 +25,112 @@
...
+        mocker,
+        synapse_store,
+        manifest_path,
+        dataset_id,
+        expected_manifest_id,
+        expected_table_id,
+        manifest_record_type,
+        validate_component=None,
+        dataset_scope=None,
+        download_dir=os.path.join(os.getcwd(), "temp"),
```
linglp, line 131: we probably shouldn't download them to a fixed directory because what if one workflow
creates this temp directory, and another concurrent workflow attempts to delete it?  What do you think @BryanFauble ?

tests/integration/test_metadata_model.py
```diff
@@ -54,31 +143,70 @@ def test_submit_filebased_manifest(
...
+
+        # AND the annotations in the manifest table should reflect the ones in the file, if applicable
+        if manifest_record_type == "table_and_file":
+            table_annotations = synapse_store.syn.tableQuery(
+                f"select * from {expected_table_id}", downloadLocation=download_dir
+            ).asDataFrame()
+
+            table_columns = table_annotations.columns
+            manifest_columns = [col.replace(" ", "") for col in manifest.columns]
+            assert set(table_columns) == set(manifest_columns)
```
linglp, line 196: Thanks for the update Jenny! I think here you make sure that the table being submitted has the same columns as the columns in the manifest. But my previous comment also mentioned about the annotations of the table: 

Screenshot 2024-09-17 at 4 09 21 PM

Do you have time to add a test here and make sure the following?  
* the table gets annotated with the correct entity id 
* etag is not empty

If not, no worry. We could open an additional ticket 👍🏼

linglp (approved):
This overall looks pretty good. I left some comments with some additional thoughts. I am going to approve it if the downloading manifest to a fixed directory part is good after confirming with Bryan.

tests/integration/test_metadata_model.py
jaymedina: [Added](https://github.com/Sage-Bionetworks/schematic/pull/1485/commits/f32e756e40efdb440ccefb6c8f5bf8f9b84c6e1c)!

tests/integration/test_metadata_model.py
```diff
@@ -54,31 +143,70 @@ def test_submit_filebased_manifest(
...
+        for index, row in manifest.iterrows():
+            entityId = row["entityId"]
+            expected_sample_id = row["Sample ID"]
+            annos = synapse_store.syn.get_annotations(entityId)
+            sample_id = annos["SampleID"][0]
+            assert str(sample_id) == str(expected_sample_id)
+
+        # AND the annotations on the manifest file itself should have the correct metadata
+        manifest_annos = synapse_store.syn.get_annotations(manifest_id)
+        for annotation in manifest_annos.keys():
```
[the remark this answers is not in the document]
jaymedina: Hi @linglp - sounds good! To clarify before I move forward: I generalized this logic to validate the annotations for the manifest file/table because [this docstring ](https://github.com/Sage-Bionetworks/schematic/blob/5721c77d35ce14b45c7bf768af133c6d81eb4abe/schematic/store/synapse.py#L1606) was implying that, in the future, the manifest files/tables will be annotated with more annotations beyond `Component`.

With that in mind, I'll be keeping that generalized logic so that developers don't have to change this test in the future, but I will patch the other checks for `eTag` and `entityId` so it's easier to read. Thanks!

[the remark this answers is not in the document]
linglp: Oh I see. Thanks for pointing that out. What you proposed makes sense. Thank you!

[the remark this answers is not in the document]
jaymedina: [Updated](https://github.com/Sage-Bionetworks/schematic/pull/1485/commits/e3da3d2cf25a6e31e1a17be2d91878d01aa70264)

tests/integration/test_metadata_model.py
jaymedina: This is a good point @linglp, I've re-worked this logic to use a native python package called `tempfile` instead which builds and tears down a unique directory each time: https://github.com/Sage-Bionetworks/schematic/pull/1485/commits/e3da3d2cf25a6e31e1a17be2d91878d01aa70264

linglp: This is looking very good Jenny! Thank you again! I appreciate a lot!

jaymedina:
Hi @linglp!

After the merge of https://github.com/Sage-Bionetworks/schematic/pull/1478 I made a few final changes to the scripts in this branch that needed to be updated. Namely:
1. I needed to update `process_row_annotations(...)` to move Gianna's solution in that method
2. I updated the tests to `process_row_annotations`, which are in `test_store_synapse.py` to have annotation dictionaries with the same structure as the annotations retrieved from Synapse

The change to these dictionary structures in the tests are best explained by Gianna's comment in the description:
>Using the client annotations are formated:
>{"ID":[], <file annotations as key value pairs>}
>Using the API the annotations are formatted:
>{'annotations':{'id': value,... 'annotations': {<file annotations as key value pairs>}}}

Since we are now retrieving the annotations using the API, I formatted the testing annotations accordingly

3. Lastly, I added a check in the `test.yaml` that removes corrupt venvs to prevent this error from breaking the builds:

```
The virtual environment found in /home/runner/work/schematic/schematic/.venv seems to be broken.
Recreating virtualenv schematicpy in /home/runner/work/schematic/schematic/.venv
``` 
Whether or not this will stay in the PR is contingent on if we decide to keep the virtual environment cache-ing in `test.yml`.

----

I'm going to tag you in the spots where the changes were made so it's easier to sift through. Thanks again for your time

schematic/store/synapse.py
```diff
@@ -1578,12 +1578,14 @@ def process_row_annotations(
                 (isinstance(anno_v, str) and anno_v.strip() == "")
                 or (isinstance(anno_v, float) and np.isnan(anno_v))
             ):
-                annos.pop(anno_k) if anno_k in annos.keys() else annos
+                annos["annotations"]["annotations"].pop(anno_k) if anno_k in annos[
```
jaymedina, line 1581: Here is the start of the changes to `process_row_annotations` @linglp

tests/integration/test_store_synapse.py
```diff
@@ -34,12 +34,16 @@ def test_process_row_annotations_hide_blanks(
             "CancerType": "   ",  # Blank value (whitespace string)
         }
         annos = {
-            "PatientID": "old_value1",
-            "Sex": "old_value2",
-            "Diagnosis": "old_value3",
-            "FamilyHistory": "old_value4",
-            "YearofBirth": "old_value5",
-            "CancerType": "old_value6",
+            "annotations": {
```
jaymedina, line 37: @linglp

linglp:
@jaymedina Thanks Jenny. It looks good to me. Could you highlight the changes of workflow in the PR description?

jaymedina:
@linglp Sure thing - [Here](https://github.com/Sage-Bionetworks/schematic/pull/1485#issue-2495487646) is the new description. Thanks! I'll merge this in once the final integration test on py 3.10 passes

PR #1490, opened 2024-09-04
[INSPIRATION ONLY] - pseudo-implement asynchronous request-reply design [DO NOT MERGE]

thomasyu888:
**Problem**
There are endpoints that take > 1 second to return a response, it is best practice for API endpoints for these to return fast

**Solution**
Use celery + redis and the async request reply design pattern. https://learn.microsoft.com/en-us/azure/architecture/patterns/async-request-reply.  NOTE: this PR does not fully work, but it meant to showcase some tech that can be used.

What didn't work about this PR was the connection between redis and celery in the docker containers locally.

Here's a rough sequence diagram that may not be fully correct (ie the status codes, etc)

image


linglp:
@thomasyu888 very interesting. I looked into turning schematic APIs to async a long time ago when we were still doing tiger team, and also saw `celery + redis` pattern. I could do more research into this too.

afwillia:
@thomasyu888 how did you make this diagram?

thomasyu888:
Thanks both for looking at this!

@linglp before you dive deeper into this, lets focus on the tasks at hand and we can align as a group what gets tackled next

@afwillia See [this](https://www.mermaidchart.com/play?utm_source=mermaid_live_editor&utm_medium=banner_ad&utm_campaign=teams#pako:eNqlVO9r2zAQ_VeOfEkDSZb1o2EdXZqGstGEuGFfCkHIZ1dEljz9CJTS_32yJcfOapeQ-YuN7969u6d3ehtQmeAgGmj8Y1FQvGMkUyR_FuCegijDKCuIMDDnDIX5-P-eE72HGNUBVQcKOapX-C3Vviu8wYTpZ-EDnmFyc9MuGcF6FT_Bl5wIlqI2kwwFKmKYFHCl6QvmZGcVH4PVuCNCSFPF9Bim0-nIF27Xc-Ur0ghiIxWCKUNEJIDCCWARmAhNd2Ad2DcZwfXsGm4pxcJgAldllR1LRvUorg0ExbIXAzKFGuPfoJAiOzhYQNUgLmUBa8k5E5n_84kqy0VLlO-h0rdjxRrdPfoGjWJ4CNNrJ5nVDaZKmnxgfKpzsdKLiVQ2GMKNT2Aa1ovHu4fHZRP8XMcZrH66oywLRzAM4OEYqFWqSpmNwbhj5RF8HYdmy0QUiRPKnfJw1FAh13hsxB1loWSmUOsLmjl2EOhr7j62eDufL-L4srED2I3turXc9JHc3z782m4Wl5EE8LCl4kIpqSB3CpEMT4QUwUXVx_mu1mCLhJjg7qO3Wg7_Qeg-U9I6D62VpI761O_tG6PPsqmSOVT72g_85xbxVB5-7sXRsUEnLG4N_W2EUK9iXz_NJNtKn7Y8LfdvVsuN90G5XxG8dXvwfdSxelTmBcdSem1pOWxqOX89tcolTZ3lzZQwjsn_kzUe9QJgadBdMGiPPwfvfwFg9irs).  That said, you can write a GPT prompt like this:

```
I am a software architect and I would like to generate the markdown code for mermaid sequence diagram for the below code (paste the code)
```

BryanFauble:
⚠️  To be aware of if looking to implement this ⚠️ 

If we implement an async job queue in front of schematic. Now schematic would intentionally have request-to-request persistence. We would likely want to match the sub in the JWT claim to the status of the job they're requesting to make sure they match. This is to ensure we are properly having an authorization check in place to prevent user "A" from requesting information about the job of user "B".

PR #1491, opened 2024-09-04
Change how dataset to validate against is specified for `filenameExists` rule

GiaJordan:
Specify the dataset with files that should be validated against as a parameter in the cli and api instead of in the data model

### DIFF
--- a/schematic/models/commands.py
+++ b/schematic/models/commands.py
@@ -4,7 +4,7 @@ import logging
 import sys
 from gc import callbacks
 from time import perf_counter
-from typing import get_args
+from typing import Optional, get_args
 
 import click
 import click_log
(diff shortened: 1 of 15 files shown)


schematic/models/commands.py
```diff
@@ -150,6 +155,7 @@ def submit_manifest(
     hide_blanks,
     restrict_rules,
     project_scope,
+    dataset_scope,
```
BryanFauble, line 158: nit: Could add in some argument types to the function

schematic/models/validate_attribute.py
```diff
@@ -2027,13 +2028,18 @@ def filename_validation(
             errors: list[str] Error details for further storage.
             warnings: list[str] Warning details for further storage.
         """
+
+        if dataset_scope is None:
```
BryanFauble, line 2032: If this cannot be none I would expect that the argument of the function is not optional (ie: No `= None`).

Could you also add this to the Args of the docstring?

GiaJordan: @BryanFauble yeah that makes sense thanks for catching

GiaJordan:
At the time of writing the tests have passed, comments have been addressed, and this PR can be re-reviewed. If that happens while I'm out then one of the reviewers can merge to `develop`

BryanFauble (approved):
LGTM!

PR #1493, opened 2024-09-05
[FDS-2373] Update jwt verification

BryanFauble:
https://sagebionetworks.jira.com/browse/FDS-2373

**Problem:**

1. At the code level nothing is doing an authentication verification on the JWT passed into schematic.

**Solution:**

1. Using the Synapse public JWKs to verify tokens before entering into schematic code

**Testing:**

1. I ran testing locally to verify that I needed to pass along a valid production synapse token.


schematic_api/api/security_controller_.py
```diff
@@ -1,14 +1,48 @@
...
+            key=signing_key.key,
+            algorithms=["RS256"],
+            options={"verify_aud": False},
+        )
+
+        return data
+    except InvalidTokenError:
+        logger.exception("Error decoding authentication token")
+        # When the return type is None the web framework will return a 401 OAuthResponseProblem exception
+        return None
```
BryanFauble, line 48: https://github.com/Sage-Bionetworks/schematic/blob/fds-2373-jwt-verification/tests/test_api.py#L130-L138

is testing that an invalid header is returning back the expected HTTP 401 error.

Every other test is handling the other case where valid auth headers are needed.

schematic_api/api/security_controller_.py
```diff
@@ -1,14 +1,48 @@
...
+    Args:
+        token (str): Bearer token.
+
+    Returns:
+        dict: Decoded token information.
     """
-    return {"uid": "user_id"}
+    try:
+        jwks_client.get_signing_key_from_jwt(token)
+        signing_key = jwks_client.get_signing_key_from_jwt(token)
```
BWMac, line 36: Why are both `get_signing_key_from_jwt` calls needed?

BWMac (commented):
A couple more questions:
- Is the extra `_` in `security_controller_.py` intentional? 
- Is there any need to handle expired tokens?

schematic_api/api/security_controller_.py
BryanFauble: That was a mistake. I remove if

BryanFauble:
> A couple more questions:
> 
>     * Is the extra `_` in `security_controller_.py` intentional?
> 
>     * Is there any need to handle expired tokens?

`_` im not sure if it's needed, but I also am not going to change it at the moment.

This automatically handles expired tokens and will fail the decode. If the token has an "exp" field it will be considered.

schematic_api/api/security_controller_.py
```diff
@@ -1,14 +1,48 @@
...
+        data = decode(
+            jwt=token,
+            key=signing_key.key,
+            algorithms=["RS256"],
+            options={"verify_aud": False},
+        )
```
brucehoff, line 41: The new code lacks tests.
What if the token is expired?
The returned 'data' doesn't appear to be used in any way. 
The old code returned a key 'uid' but the new code does not.
The comments say "The token claims are decoded and returned." but the code doesn't do that.
The code returns the raw token, signing key, etc. Why?
If you want to return the signing algorithm, don't hard code it, get it from the OIDC discovery document, https://repo-prod.prod.sagebase.org/auth/v1/oauth2/jwks

brucehoff (changes requested):

schematic_api/api/security_controller_.py
BryanFauble: > The new code lacks tests.

The code is implicitly being tested via the existing integration tests, but I agree. I will add some narrower tests for this logic specifically.

> What if the token is expired?

The pyjwt library will raise an exception:
```
    def _validate_exp(
        self,
        payload: dict[str, Any],
        now: float,
        leeway: float,
    ) -> None:
        try:
            exp = int(payload["exp"])
        except ValueError:
            raise DecodeError("Expiration Time claim (exp) must be an integer.")

        if exp <= (now - leeway):
            raise ExpiredSignatureError("Signature has expired")
```

> The returned 'data' doesn't appear to be used in any way.

The data can be returned in each API route via specifying a `token_info` argument. I am planning to make updates in a further PR to add the user ID to each open telemetry trace so we can easily identify what user produced a trace.

https://connexion.readthedocs.io/en/stable/request.html#context

> The old code returned a key 'uid' but the new code does not.

Thats correct. The old code could have returned anything (Except None) and I think the web framework would have been fine with it.

> The comments say "The token claims are decoded and returned." but the code doesn't do that.
> The code returns the raw token, signing key, etc. Why?

The code calls the `decode` function (This is Python syntax). It is returning the decoded JWT payload.
```
# The import:
from jwt import PyJWKClient, decode
.
.
.
        data = decode(
            jwt=token,
            key=signing_key.key,
            algorithms=["RS256"],
            options={"verify_aud": False},
        )
```

> If you want to return the signing algorithm, don't hard code it, get it from the OIDC discovery document, https://repo-prod.prod.sagebase.org/auth/v1/oauth2/jwks

The code is accomplishing this via this code:
```
jwks_client = PyJWKClient(
    uri=syn.authEndpoint + "/oauth2/jwks", headers=syn._generate_headers()
)
```

As a part of the JWT library it handles the logic needed to parse the document.

brucehoff: My mistake, I didn't see "decode()", sorry.  While it will work to hard code `RS256` (since that's what Synapse uses), it would be best to get it from Synapse's `jwks` endpoint.

thomasyu888 (approved):
🔥 Just going to pre-approve - thanks for being so quick on this.

schematic_api/api/security_controller_.py
BryanFauble: The synapse `jwks` endpoint does not provide the `alg` so I wouldn't be able to grab it from there

brucehoff: It's the `kty` (key type) field here: https://repo-prod.prod.sagebase.org/auth/v1/oauth2/jwks.
The algorithm for getting the type is:
1. get the Key ID (kid) from the token;
2. look up the kid in the list of public keys returned by the jwks endpoint.
3. for the matching record, take the kid field.

BryanFauble: > There may be a Python library that does all this for you, e.g. you give it the token and the jwks endpoint and it, validates the token and returns the payload (if valid).

The python library I am using is what is handling this for me, they did not have it in their examples but I was able find out that I could grab the algorithm used like:

```
        signing_key = jwks_client.get_signing_key_from_jwt(token)
        data = decode(
            jwt=token,
            key=signing_key.key,
            algorithms=[signing_key.algorithm_name],
            options={"verify_aud": False},
        )
```

poetry.lock
```diff
@@ -4385,13 +4470,13 @@ Jinja2 = ">=2.0"
 
 [[package]]
 name = "synapseclient"
-version = "4.4.0"
+version = "4.4.1"
```
linglp, line 4473: Do we need to update synapse python client to 4.4.1 for this feature?

linglp (approved):
Looks good to me too. I do not mind upgrading synapse python client to v4.4.1, but this should probably be in PR description or somewhere more visible like a Jira ticket. Synapse python client is an important dependencies, so we want to make sure that people know when we update it to a higher version.

poetry.lock
BryanFauble: For what I wrote yes. It's using this boolean I introduced here: https://github.com/Sage-Bionetworks/synapsePythonClient/compare/v4.4.0...v4.4.1-rc#diff-848a7a4238d8f0a7792848097e227f8e3aa284aeadf15e1fa3d890f17791b89cR280

It also contains a minor improvement to remove 1 HTTP call during login.

schematic_api/api/security_controller.py
```diff
@@ -0,0 +1,47 @@
...
+    configPath=CONFIG.synapse_configuration_path,
+    cache_client=False,
+)
+jwks_client = PyJWKClient(
+    uri=syn.authEndpoint + "/oauth2/jwks", headers=syn._generate_headers()
+)
+
+
+def info_from_bearer_auth(token: str) -> Dict[str, Union[str, int]]:
+    """
```
brucehoff, line 22: Client code should not treat an access token as a JWT.

PR #1496, opened 2024-09-06
[FDS-1725] Missing `entityId` handling testing

BWMac:
**Description:**
This PR follows up on #1456, where the `filenameExists` validation rule was implemented. This rule handles cases where a file path present in a manifest does not exist in a dataset and cases where the entityId provided in a manifest row does not match with its corresponding file path. 

This PR adds additional error handling for situations including:
- The entityId field in a manifest row is empty
- The entityId provided in a manifest row does not exist in the dataset

Previously, an outer join was used in the validation rule followed by a line that handled cases where there were more rows in the dataset than the manifest. The outer join implementation did not always maintain the order of manifest rows resulting in incorrect errors on manifest rows. I changed this outer join to a left join on the manifest to handle both issues.

The row number indexing for `filename_validation` is also updated to match the other validation rules for consistency.

I have also updated `generate_filename_error` to handle the new error cases.

**Testing:**
- New integration test cases for `filename_validation` have been added as needed.
- Unit tests for `filename_validation` and `generate_filename_error` have been added


BryanFauble (approved):
🔥

tests/data/mock_manifests/InvalidFilenameAndEntityIDManifest.csv
```diff
@@ -2,5 +2,5 @@ Component,Filename,entityId
 MockFilename,schematic - main/MockFilenameComponent/txt1.txt,syn61682653
 MockFilename,schematic - main/MockFilenameComponent/txt2.txt,syn61682660
 MockFilename,schematic - main/MockFilenameComponent/txt3.txt,syn61682662
-MockFilename,schematic - main/MockFilenameComponent/txt4.txt,syn6168265
-MockFilename,schematic - main/MockFilenameComponent/txt5.txt,
+MockFilename,schematic - main/MockFilenameComponent/txt4.txt,syn1234567
```
thomasyu888, line 5: One note here, there may have been a reason for a manifest to have a empty synapse id as well. Should confirm with FAIR

tests/data/mock_manifests/InvalidFilenameAndEntityIDManifest.csv
```diff
@@ -2,5 +2,5 @@ Component,Filename,entityId
 MockFilename,schematic - main/MockFilenameComponent/txt1.txt,syn61682653
 MockFilename,schematic - main/MockFilenameComponent/txt2.txt,syn61682660
 MockFilename,schematic - main/MockFilenameComponent/txt3.txt,syn61682662
-MockFilename,schematic - main/MockFilenameComponent/txt4.txt,syn6168265
-MockFilename,schematic - main/MockFilenameComponent/txt5.txt,
```
GiaJordan, line 6: So in the original feature request, the two conditions that needed to be covered for validation were
>- if filename/path in a manifest does not exist in the corresponding Synapse dataset folder: Provide error indicating which row(s) do not have valid filename/path(s)

which was addressed by line 6 in the manifest here
and
>- if filename/path exists but is associated with a different entity ID than the one provided in the manifest: Provide error indicating which row(s) have filename/path(s) which do not match the entity ID provided in the manifest

which was addressed by line 5 in the manifest here

to test the entityID validation you should add new rows or a separate manifest/dataset combination

GiaJordan (dismissed):

GiaJordan:
For the new cases where we're checking if entityIds don't exist at all or are missing from the manifest we should also add validation error messages to [`generate_filename_error`](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/models/validate_attribute.py#L482) that specify to the user what the issue with their metadata was. And then the validation integration test should be updated to check for those error messages for the test cases you add to the manifest

tests/data/mock_manifests/InvalidFilenameAndEntityIDManifest.csv
BWMac: Ah, my bad I misinterpreted that manifest CSV file and test. I will revert these changes and then circle back to this on Monday.

BWMac:
Thanks for the comments @GiaJordan! I am going to revert this back to being a draft for now and work on those updates.

schematic/models/validate_attribute.py
```diff
@@ -2098,7 +2110,7 @@ def filename_validation(
             vr_errors, vr_warnings = GenerateError.generate_filename_error(
                 val_rule=val_rule,
                 attribute_name="Filename",
-                row_num=str(index),
+                row_num=str(index + 2),
```
thomasyu888, line 2113: May I ask why the +2 here?

BWMac: This is how the row number indexing is done everywhere else in `validate_attribute.py`. [Example](https://github.com/Sage-Bionetworks/schematic/blob/045d8c2004f7615778f47f9be3b931de95d012cc/schematic/models/validate_attribute.py#L959)

So I did this to make it consistent and to avoid confusion like [this](https://sagebionetworks.slack.com/archives/C07HLTBS5D2/p1726156665980609?thread_ts=1726000362.049489&cid=C07HLTBS5D2).

thomasyu888: May want to add a comment about that in the code

tests/unit/test_validate_attribute.py
```diff
@@ -99,6 +99,53 @@
     }
 )
 
+TEST_DF_FILEVIEW = DataFrame(
```
linglp, line 102: Thanks for all the test cases here. I am curious: what if the entity id is missing and also the manifest file path does not exist, what would happen? I think based on the current design, only one message gets raised?

linglp (approved):
This is looking very good! I left a few comments. But I have no problems having it pre-approved

tests/unit/test_validate_attribute.py
BWMac: That is correct, only one message can be raised at present. In this case, the `missing entityId` error would be raised. [We give precedence](https://sagebionetworks.slack.com/archives/C07HLTBS5D2/p1726001699040209?thread_ts=1726000362.049489&cid=C07HLTBS5D2) to the `entityId` for these cases.

linglp: Thanks for the clarification Brad. Can we ask you to document all these expected behavior somewhere with examples? I will help coordinate and see where would be a good place to document.

linglp:
@BWMac sorry one more comment. Since there's no test related to CLI, I could test manually. Give me some time :) I will report back shortly.

tests/unit/test_validate_attribute.py
BWMac: Sure! Just let me know where would be best.

tests/unit/test_validate_attribute.py
```diff
@@ -108,6 +155,12 @@ def fixture_va_obj(
     yield ValidateAttribute(dmge)
 
 
+@pytest.fixture(name="mock_dmge")
```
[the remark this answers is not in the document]
BWMac: Thanks! Sorry I missed that

linglp:
@BWMac thanks for your patience. I tested the validation CLI with my own test manifest and it works. And I also realized that we don't have "dataset scope" parameter added for manifest submission so we can't use the file name validation during submission. But this issue is definitely outside of your PR! Thanks for your hard work. After the tests are all passing, I think you could merge 👍🏼

tests/unit/test_validate_attribute.py
linglp: Let me know if you need a new ticket for the documentation work.

BWMac: It's OK I'll do it as part of this ticket

BWMac: I have updated the documentation

PR #1497, opened 2024-09-07
[FDS-2381] Remove call to change file metadata

BryanFauble:
TODO

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1430,13 +1430,15 @@ class SynapseStorage(BaseStorage):
     @tracer.start_as_current_span("SynapseStorage::upload_manifest_file")
     def upload_manifest_file(
         self,
-        manifest,
-        metadataManifestPath,
+        manifest: pd.DataFrame,
+        metadataManifestPath: str,
         datasetId,
         restrict_manifest,
         component_name="",
     ):
         # Update manifest to have the new entityId column
+        # TODO: Update this code to allow saving the manifest to disk via an in memory buffer
+        # so that we can avoid writing to disk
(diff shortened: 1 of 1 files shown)


schematic/store/synapse.py
```diff
@@ -1430,13 +1430,15 @@ def buildDB(
...
+        manifest: pd.DataFrame,
+        metadataManifestPath: str,
         datasetId,
         restrict_manifest,
         component_name="",
     ):
         # Update manifest to have the new entityId column
+        # TODO: Update this code to allow saving the manifest to disk via an in memory buffer
+        # so that we can avoid writing to disk
         manifest.to_csv(metadataManifestPath, index=False)
```
BryanFauble, line 1442: This is optional (Feel free to remove the TODO). It's something that would need to be implemented within synpy and is quite a large change to make this happen.

schematic/store/synapse.py
```diff
@@ -1463,22 +1465,29 @@ def upload_manifest_file(
...
-            syn=self.syn,
-            entity=manifest_synapse_file_id,
-            downloadAs=file_name_new,
-            forceVersion=False,
-        )
+        try:
+            # Rename the file to file_name_new
+            original_file_path = metadataManifestPath
+            new_file_path = os.path.join(os.path.dirname(metadataManifestPath), file_name_new)
+            os.rename(original_file_path, new_file_path)
```
BryanFauble, line 1472: To investigate: Do we need to write the file to disk with `metadataManifestPath`? Look at the code that is generating this and determine if we can instead just go ahead and write the CSV out to `file_name_new` instead.

BryanFauble:
I will be handling this is another PR

PR #1499, opened 2024-09-11
PR suggestion: Remove graph_data_model parameter in create_manifests to simplify the code

linglp:
# Problem: 
* the graph_data_model parameter was added to create_manifests function. But this parameter is not necessary

# Changes: 
* Remove graph_data_model parameter in create_manifests function 
* Have create_manifests function to handle reading pickle files

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -151,7 +151,7 @@ jobs:
 
 
       - name: Upload pytest test results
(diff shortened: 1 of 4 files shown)


andrewelamb:
@linglp Thanks for all the work on this! Since the DCA is frozen and the Schematic API will soon be sunsetted, I'm going to close this PR.

PR #1502, opened 2024-09-13
[FDS-2415] Fixes `pytest` CI Python Versions

BWMac:
While working on #1496, I noticed that our `pytest` runs in the CI workflow were only being executed in python 3.9, rather than in 3.9 AND 3.10. This is due to the `actions/cache` GH action not having a specification for which python version it should be using. This PR adds specific Python versions where necessary and also bumps up the versions of all GH actions we are using where needed (some were 2 or 3 versions behind the latest releases).

### DIFF
--- a/.github/workflows/api_test.yml
+++ b/.github/workflows/api_test.yml
@@ -51,7 +51,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: .venv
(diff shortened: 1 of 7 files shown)


BryanFauble (approved):
🔥

linglp (approved):
LGTM! Thanks for the fix.

PR #1504, opened 2024-09-16
Fixes failing `pdoc` deployment

BWMac:
This is a follow-up to #1503 which should have fixed the `pdoc` deployment. The problem with the last change was that it created two artifacts with the names `schematic-docs-3.9` and `schematic-docs-3.10`, when the `deploy` step expects an artifact called `github-pages`. The previous fix required an update to the naming because the newer version of `upload-pages-artifact` does not allow duplicate names. So therefore we now cannot have the `build` part of the workflow run on both Python 3.9 and 3.10 and also have the `deploy` step work properly.

This PR adjusts the `build` step to use only Python 3.10 and then to produce a single artifact called `github-pages` which can then (hopefully) be deployed. 

**Note:** It is difficult to test this step in a feature branch because there is a branch protection policy in place preventing that from happening.

### DIFF
--- a/.github/workflows/pdoc.yml
+++ b/.github/workflows/pdoc.yml
@@ -78,7 +78,7 @@ jobs:
       - uses: actions/upload-pages-artifact@v3
         with:
           path: docs/schematic
-          name: github-pages-${{ env.PYTHON_VERSION }}
+          name: github-pages
 
   # Deploy the artifact to GitHub pages.
   # This is a separate job so that only actions/deploy-pages has the necessary permissions.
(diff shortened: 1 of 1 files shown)


linglp (approved):
Thanks for the fix @BWMac . This makes sense to me.

linglp:
@BWMac this failure of integration test is due to this issue that Bryan is working on: https://sagebionetworks.jira.com/browse/FDS-2240. Since the error is inconsistent, I think you could rerun the workflow to avoid being blocked. But we should definitely work on/prioritize FDS-2240.

PR #1505, opened 2024-09-17
[FDS-2386] Synapse entity tracking and code concurrency updates

BryanFauble:
I will be adding comments to specific areas of the code that should be looked at and carefully considered. Below covers the overall gist of what is going on here:

**Problem:**

1. When an entity is retrieved from the synapse python client it is not stored for further use within the context of a request hitting schematic. The entity is re-retrieved several times throughout the lifecycle of the request causing significant bottlenecks when retrieving data out of Synapse.
2. When downloading manifests a temporary directory is being used when running on AWS, but not locally, or for integration tests. As a result concurrent requests in those contexts were causing manifest files to be overwritten for new tests and causing previous tests to fail.
3. A few concepts for great expectations (suites and checkpoints) were constantly being created and torn down, even during requests where the checkpoint/suite was not created. This leads to more concurrency type issues where a checkpoint/suite could be torn down in the middle of a request, or at a point in time it isn't expected.

**Solution:**

1. Creating a simple "pull through cache" that can be added via composition to the classes that interact with Synapse. This is acting as a small (albeit hacky) shim layer between schematic and SYNPY. The usage of this entity tracker allows the code to function largely as it has been without the need to more in-depth refactoring. This means that an entity for a user session is pulled out of memory before it is pulled from Synapse again.
2. Writing manifests to a temporary directory in all cases except for when being used as a CLI tool.
3. Updating the logic for tearing down the great expectation suite and checkpoints to only consider tearing down the checkpoints that itself created.

**Testing:**

1. I have been stepping through the code with the integration tests that have been written to verify functionality.
2. Will need to have a verbose testing plan in place to verify that schematic continues to function as expected. @linglp @andrewelamb Are you two available to help with the concepts of a plan for helping to test all of these changes?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,9 +54,7 @@ pygsheets = "^2.0.4"
 PyYAML = "^6.0.0"
 rdflib = "^6.0.0"
 setuptools = "^66.0.0"
-#synapseclient = "4.4.1"
-# GIT dependency for synapseclient
-synapseclient = { git = "https://github.com/Sage-Bionetworks/synapsePythonClient.git", branch = "fds-2386-allow-global-client-cache-disable" }
+synapseclient = "4.5.1"
 tenacity = "^8.0.1"
 toml = "^0.10.2"
 great-expectations = "^0.15.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,9 @@ pygsheets = "^2.0.4"
 PyYAML = "^6.0.0"
 rdflib = "^6.0.0"
 setuptools = "^66.0.0"
-synapseclient = "4.4.1"
+#synapseclient = "4.4.1"
+# GIT dependency for synapseclient
+synapseclient = { git = "https://github.com/Sage-Bionetworks/synapsePythonClient.git", branch = "fds-2386-allow-global-client-cache-disable" }
 tenacity = "^8.0.1"
 toml = "^0.10.2"
 great-expectations = "^0.15.0"
@@ -65,7 +67,6 @@ openpyxl = "^3.0.9"
 pdoc = "^14.0.0"
 dateparser = "^1.1.4"
 pandarallel = "^1.6.4"
-schematic-db = {version = "0.0.41", extras = ["synapse"]}
 pyopenssl = {version = "^23.0.0", optional = true}
 dataclasses-json = "^0.6.1"
 pydantic = "^1.10.4"
@@ -75,13 +76,14 @@ Flask-Cors = {version = "^3.0.10", optional = true}
 uWSGI = {version = "^2.0.21", optional = true}
 Jinja2 = {version = ">2.11.3", optional = true}
 asyncio = "^3.4.3"
-jaeger-client = {version = "^4.8.0", optional = true}
-flask-opentracing = {version="^2.0.0", optional = true}
 PyJWT = "^2.9.0"
+opentelemetry-api = {version = ">=1.21.0", optional = true}
+opentelemetry-sdk = {version = ">=1.21.0", optional = true}
 opentelemetry-exporter-otlp-proto-grpc = {version="^1.0.0", optional = true}
+opentelemetry-instrumentation-flask = {version=">=0.48b0 ", optional = true}
 
 [tool.poetry.extras]
-api = ["connexion", "Flask", "Flask-Cors", "Jinja2", "pyopenssl", "jaeger-client", "flask-opentracing", "opentelemetry-exporter-otlp-proto-grpc"]
+api = ["connexion", "Flask", "Flask-Cors", "Jinja2", "pyopenssl", "opentelemetry-api", "opentelemetry-sdk", "opentelemetry-exporter-otlp-proto-grpc", "opentelemetry-instrumentation-flask"]
 aws = ["uWSGI"]
 
 
@@ -98,12 +100,6 @@ mypy = "^1.4.1"
 pylint = "^2.16.1"
 pytest-xdist = "^3.5.0"
 pre-commit = "^3.6.2"
-
-[tool.poetry.group.aws]
-optional = true
-
-[tool.poetry.group.aws.dependencies]
-
 
 [tool.black]
 line-length = 88
@@ -134,42 +130,4 @@ testpaths = [
 ]
 filterwarnings = [
     "ignore::DeprecationWarning"
-]
-markers = [
-    """\
-    google_credentials_needed: marks tests requiring \
-    Google credentials (skipped on GitHub CI) \
-    """,
-    """\
-    submission: tests that involve submitting manifests
-    """,
-    """\
-    not_windows: tests that don't work on on windows machine
-    """,
-    """\
-    schematic_api: marks tests covering \
-    API functionality (skipped on regular GitHub CI test suite)   
-    """,
-    """\
-    rule_combos: marks tests covering \
-    combinations of rules that aren't always necessary \
-    and can add significantly to CI runtime (skipped on GitHub CI unless prompted to run in commit message)   
-    """,   
-    """\
-    table_operations: marks tests covering \
-    table operations that pass locally \
-    but fail on CI due to interactions with Synapse (skipped on GitHub CI)   
-    """,    
-    """\
-    rule_benchmark: marks tests covering \
-    validation rule benchmarking
-    """,
-    """\
-    synapse_credentials_needed: marks api tests that \
-    require synapse credentials to run
-    """,
-    """\
-    empty_token: marks api tests that \
-    send empty credentials in the request
-    """
 ]
--- a/schematic/__init__.py
+++ b/schematic/__init__.py
@@ -1,2 +1,114 @@
+import logging
+import os
+import time
+from typing import Dict, List
+
+from opentelemetry import trace
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.instrumentation.flask import FlaskInstrumentor
+from opentelemetry.sdk.resources import SERVICE_NAME, Resource
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import (
+    BatchSpanProcessor,
+    ConsoleSpanExporter,
+    SimpleSpanProcessor,
+    Span,
+)
+from opentelemetry.sdk.trace.sampling import ALWAYS_OFF
+from synapseclient import Synapse
+from werkzeug import Request
+
 from schematic.configuration.configuration import CONFIG
 from schematic.loader import LOADER
+from schematic_api.api.security_controller import info_from_bearer_auth
+
+Synapse.allow_client_caching(False)
+logger = logging.getLogger(__name__)
+
+
+# borrowed from: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/tests/integration/conftest.py
+class FileSpanExporter(ConsoleSpanExporter):
+    """Create an exporter for OTEL data to a file."""
+
+    def __init__(self, file_path: str) -> None:
+        """Init with a path."""
+        self.file_path = file_path
+
+    def export(self, spans: List[Span]) -> None:
+        """Export the spans to the file."""
+        with open(self.file_path, "a", encoding="utf-8") as f:
+            for span in spans:
+                span_json_one_line = span.to_json().replace("\n", "") + "\n"
+                f.write(span_json_one_line)
+
+
+def set_up_tracing() -> None:
+    """Set up tracing for the API."""
+    tracing_export = os.environ.get("TRACING_EXPORT_FORMAT", None)
+    if tracing_export is not None and tracing_export:
+        Synapse.enable_open_telemetry(True)
+        tracing_service_name = os.environ.get("TRACING_SERVICE_NAME", "schematic-api")
+
+        trace.set_tracer_provider(
+            TracerProvider(
+                resource=Resource(attributes={SERVICE_NAME: tracing_service_name})
+            )
+        )
+        FlaskInstrumentor().instrument(
+            request_hook=request_hook, response_hook=response_hook
+        )
+
+    if tracing_export == "otlp":
+        trace.get_tracer_provider().add_span_processor(
+            BatchSpanProcessor(OTLPSpanExporter())
+        )
+    elif tracing_export == "file":
+        timestamp_millis = int(time.time() * 1000)
+        file_name = f"otel_spans_integration_testing_{timestamp_millis}.ndjson"
+        file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), file_name)
+        processor = SimpleSpanProcessor(FileSpanExporter(file_path))
+        trace.get_tracer_provider().add_span_processor(processor)
+    else:
+        trace.set_tracer_provider(TracerProvider(sampler=ALWAYS_OFF))
+
+
+def request_hook(span: Span, environ: Dict) -> None:
+    """
+    Request hook for the flask server to handle setting attributes in the span. If
+    anything here happens to fail we do not want to stop the request from being
+    processed so we catch all exceptions and log them.
+
+    Args:
+        span: The span object to set attributes in
+        environ: The environment variables from the request
+    """
+    if not span or not span.is_recording():
+        return
+    try:
+        if auth_header := environ.get("HTTP_AUTHORIZATION", None):
+            split_headers = auth_header.split(" ")
+            if len(split_headers) > 1:
+                token = auth_header.split(" ")[1]
+                user_info = info_from_bearer_auth(token)
+                if user_info:
+                    span.set_attribute("user.id", user_info.get("sub"))
+    except Exception:
+        logger.exception("Failed to set user info in span")
+
+    try:
+        if (request := environ.get("werkzeug.request", None)) and isinstance(
+            request, Request
+        ):
+            for arg in request.args:
+                span.set_attribute(key=f"schematic.{arg}", value=request.args[arg])
+    except Exception:
+        logger.exception("Failed to set request info in span")
+
+
+def response_hook(span: Span, status: str, response_headers: List) -> None:
+    """Nothing is implemented here yet, but it follows the same pattern as the
+    request hook."""
+    pass
+
+
+set_up_tracing()
--- a/schematic/manifest/commands.py
+++ b/schematic/manifest/commands.py
@@ -1,25 +1,21 @@
-import json
-import os
-import pandas as pd
 import logging
-from pathlib import Path
+import os
 import sys
-from typing import get_args, List
+from pathlib import Path
+from typing import List, get_args
+
 import click
 import click_log
 
-from schematic.schemas.data_model_parser import DataModelParser
-from schematic.schemas.data_model_graph import DataModelGraph, DataModelGraphExplorer
-from schematic.manifest.generator import ManifestGenerator
-
-from schematic.utils.schema_utils import DisplayLabelType
-from schematic.utils.cli_utils import log_value_from_config, query_dict, parse_syn_ids
-from schematic.utils.google_api_utils import export_manifest_csv
-
-from schematic.help import manifest_commands
-
-from schematic.store.synapse import SynapseStorage
 from schematic.configuration.configuration import CONFIG
+from schematic.help import manifest_commands
+from schematic.manifest.generator import ManifestGenerator
+from schematic.schemas.data_model_graph import DataModelGraph, DataModelGraphExplorer
+from schematic.schemas.data_model_parser import DataModelParser
+from schematic.store.synapse import SynapseStorage
+from schematic.utils.cli_utils import log_value_from_config, parse_syn_ids, query_dict
+from schematic.utils.google_api_utils import export_manifest_csv
+from schematic.utils.schema_utils import DisplayLabelType
 
 logger = logging.getLogger("schematic")
 click_log.basic_config(logger)
@@ -343,14 +339,15 @@ def migrate_manifests(
 )
 @click.pass_obj
 def download_manifest(ctx, dataset_id, new_manifest_name):
-    master_fileview = CONFIG["synapse"]["master_fileview"]
-
     # use Synapse Storage
     store = SynapseStorage()
 
     # download existing file
     manifest_data = store.getDatasetManifest(
-        datasetId=dataset_id, downloadFile=True, newManifestName=new_manifest_name
+        datasetId=dataset_id,
+        downloadFile=True,
+        newManifestName=new_manifest_name,
+        use_temporary_folder=False,
     )
 
     if not manifest_data:
--- a/schematic/models/GE_Helpers.py
+++ b/schematic/models/GE_Helpers.py
@@ -1,21 +1,16 @@
 import logging
 import os
-import re
-from statistics import mode
-from tabnanny import check
+import uuid
 
 # allows specifying explicit variable types
-from typing import Any, Dict, List, Optional, Text
-from urllib import error
-from urllib.parse import urlparse
-from urllib.request import HTTPDefaultErrorHandler, OpenerDirector, Request, urlopen
+from typing import Dict, List
 
 import numpy as np
-from attr import attr
 from great_expectations.core import ExpectationSuite
 from great_expectations.core.expectation_configuration import ExpectationConfiguration
 from great_expectations.data_context import BaseDataContext
 from great_expectations.data_context.types.base import (
+    AnonymizedUsageStatisticsConfig,
     DataContextConfig,
     DatasourceConfig,
     FilesystemStoreBackendDefaults,
@@ -23,8 +18,10 @@ from great_expectations.data_context.types.base import (
 from great_expectations.data_context.types.resource_identifiers import (
     ExpectationSuiteIdentifier,
 )
-from great_expectations.exceptions.exceptions import GreatExpectationsError
-from ruamel import yaml
+from great_expectations.exceptions.exceptions import (
+    DataContextError,
+    GreatExpectationsError,
+)
 
 import great_expectations as ge
 from schematic.models.validate_attribute import GenerateError
@@ -119,6 +116,8 @@ class GreatExpectationsHelpers(object):
             },
         }
 
+        anonymous_usage_statistics = AnonymizedUsageStatisticsConfig(enabled=False)
+
         # create data context configuration
         data_context_config = DataContextConfig(
             datasources={
@@ -136,6 +135,7 @@ class GreatExpectationsHelpers(object):
             store_backend_defaults=FilesystemStoreBackendDefaults(
                 root_directory=os.path.join(os.getcwd(), "great_expectations")
             ),
+            anonymous_usage_statistics=anonymous_usage_statistics,
         )
 
         # build context and add data source
@@ -152,29 +152,22 @@ class GreatExpectationsHelpers(object):
             saves expectation suite and identifier to self
         """
         self.expectation_suite_name = "Manifest_test_suite"
-        # Get a list of all expectation suites
-        suite_names = self.context.list_expectation_suite_names()
-        # Get a list of all checkpoints
-        all_checkpoints = self.context.list_checkpoints()
 
-        # if the suite exists, delete it
-        if self.expectation_suite_name in suite_names:
-            self.context.delete_expectation_suite(self.expectation_suite_name)
-
-            # also delete all the checkpoints associated with the suite
-            if all_checkpoints:
-                for checkpoint_name in all_checkpoints:
-                    self.context.delete_checkpoint(checkpoint_name)
-
-        self.suite = self.context.add_expectation_suite(
-            expectation_suite_name=self.expectation_suite_name,
-        )
+        try:
+            expectation_suite = self.context.get_expectation_suite(
+                expectation_suite_name=self.expectation_suite_name
+            )
+        except DataContextError:
+            expectation_suite = self.context.add_expectation_suite(
+                expectation_suite_name=self.expectation_suite_name,
+            )
+        self.suite = expectation_suite
 
         return self.suite
 
     def build_expectation_suite(
         self,
-    ):
+    ) -> None:
         """
         Purpose:
             Construct an expectation suite to validate columns with rules that have expectations
@@ -374,6 +367,7 @@ class GreatExpectationsHelpers(object):
         suite_identifier = ExpectationSuiteIdentifier(
             expectation_suite_name=self.expectation_suite_name
         )
+        # TODO: Do we need to build data docs here during normal operation?
         self.context.build_data_docs(resource_identifiers=[suite_identifier])
         ##Webpage DataDocs opened here:
         # self.context.open_data_docs(resource_identifier=suite_identifier)
@@ -421,7 +415,7 @@ class GreatExpectationsHelpers(object):
             adds checkpoint to self
         """
         # create manifest checkpoint
-        self.checkpoint_name = "manifest_checkpoint"
+        self.checkpoint_name = f"manifest_checkpoint_{uuid.uuid4()}"
         checkpoint_config = {
             "name": self.checkpoint_name,
             "config_version": 1,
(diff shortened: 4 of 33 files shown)


pyproject.toml
```diff
@@ -54,7 +54,9 @@ pygsheets = "^2.0.4"
 PyYAML = "^6.0.0"
 rdflib = "^6.0.0"
 setuptools = "^66.0.0"
-synapseclient = "4.4.1"
+#synapseclient = "4.4.1"
+# GIT dependency for synapseclient
+synapseclient = { git = "https://github.com/Sage-Bionetworks/synapsePythonClient.git", branch = "fds-2386-allow-global-client-cache-disable" }
```
BryanFauble, line 59: I have a few fixes I've found and added for SYNPY that will need to be included. For now I am pointing to the GIT branch that has the changes, but will plan to release this before I merged this into develop.

pyproject.toml
```diff
@@ -75,13 +76,14 @@ Flask-Cors = {version = "^3.0.10", optional = true}
 uWSGI = {version = "^2.0.21", optional = true}
 Jinja2 = {version = ">2.11.3", optional = true}
 asyncio = "^3.4.3"
-jaeger-client = {version = "^4.8.0", optional = true}
-flask-opentracing = {version="^2.0.0", optional = true}
```
BryanFauble, line 79: flask/jaeger are deprecated. Swapping over to the otel-api/sdk libraries instead.

pyproject.toml
```diff
@@ -135,41 +131,3 @@ testpaths = [
 filterwarnings = [
     "ignore::DeprecationWarning"
 ]
-markers = [
```
BryanFauble, line 138: These were moved to `pytest.ini` to stop a bunch of warning from being emitted.

schematic/__init__.py
```diff
@@ -1,2 +1,114 @@
...
+        tracing_service_name = os.environ.get("TRACING_SERVICE_NAME", "schematic-api")
+
+        trace.set_tracer_provider(
+            TracerProvider(
+                resource=Resource(attributes={SERVICE_NAME: tracing_service_name})
+            )
+        )
+        FlaskInstrumentor().instrument(
+            request_hook=request_hook, response_hook=response_hook
+        )
```
BryanFauble, line 59: Most all the code in this script was just moved except for this `FlaskInstruement` and the request/response hook. This is coming from this library: https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-flask - Which allows auto instrumentation of the flask server that is running here

schematic/manifest/commands.py
```diff
@@ -343,14 +339,15 @@ def migrate_manifests(
...
     # use Synapse Storage
     store = SynapseStorage()
 
     # download existing file
     manifest_data = store.getDatasetManifest(
-        datasetId=dataset_id, downloadFile=True, newManifestName=new_manifest_name
+        datasetId=dataset_id,
+        downloadFile=True,
+        newManifestName=new_manifest_name,
+        use_temporary_folder=False,
```
BryanFauble, line 350: When running in CLI mode we will not be downloading manifest files to a temporary folder, instead continue to use the defined manifest folder from the config.

schematic/models/GE_Helpers.py
```diff
@@ -119,6 +116,8 @@ def build_context(self):
             },
         }
 
+        anonymous_usage_statistics = AnonymizedUsageStatisticsConfig(enabled=False)
+
```
BryanFauble, line 120: This removes extra calls from leaving schematic.

schematic/models/GE_Helpers.py
```diff
@@ -151,30 +151,17 @@ def add_expectation_suite_if_not_exists(self) -> ExpectationSuite:
...
-            if all_checkpoints:
-                for checkpoint_name in all_checkpoints:
-                    self.context.delete_checkpoint(checkpoint_name)
-
-        self.suite = self.context.add_expectation_suite(
+        self.expectation_suite_name = f"Manifest_test_suite_{uuid.uuid4()}"
+        expectation_suite = self.context.add_expectation_suite(
             expectation_suite_name=self.expectation_suite_name,
         )
+        self.suite = expectation_suite
```
BryanFauble, line 158: I found issues when tests were running concurrently where great expectation test suites were stepping over eachother, causing tests to break in Un-expected places. With this change the JSON file that is created for each test suite run is unique, and cleaned up after the test suite has ran.

No more will another test clean these up.

BryanFauble: The same notes about concurrency also apply when this code is running as an API server. If 2 users were sending in a request at the same time this is likely to have caused issues if they lined up just perfectly.

schematic/models/GE_Helpers.py
```diff
@@ -421,7 +409,7 @@ def build_checkpoint(self):
             adds checkpoint to self
         """
         # create manifest checkpoint
-        self.checkpoint_name = "manifest_checkpoint"
+        self.checkpoint_name = f"manifest_checkpoint_{uuid.uuid4()}"
```
BryanFauble, line 412: Using a unique checkpoint name to prevent name collisions with other GE runs.

schematic/models/GE_Helpers.py
```diff
@@ -374,6 +361,7 @@ def build_expectation_suite(
         suite_identifier = ExpectationSuiteIdentifier(
             expectation_suite_name=self.expectation_suite_name
         )
+        # TODO: Do we need to build data docs here during normal operation?
         self.context.build_data_docs(resource_identifiers=[suite_identifier])
```
BryanFauble, line 365: @andrewelamb @linglp @milen-sage - Is this code needed? It builds a bunch of HTML files, but I couldn't understand if they were actually used anywhere?

schematic/models/metadata.py
```diff
@@ -273,10 +273,6 @@ def validateModelManifest(
 
             return errors, warnings
 
-        # check if suite has been created. If so, delete it
-        if os.path.exists("great_expectations/expectations/Manifest_test_suite.json"):
-            os.remove("great_expectations/expectations/Manifest_test_suite.json")
-
```
BryanFauble, line 279: No longer cleaning up the test suite run here. Instead relying on the cleanup to happen in the `finally` block after the expectation runs.

schematic/models/validate_manifest.py
```diff
@@ -174,7 +160,8 @@ def validate_manifest_rules(
         warnings = []
 
         if not restrict_rules:
-            t_GE = perf_counter()
+            if logger.isEnabledFor(logging.DEBUG):
+                t_GE = perf_counter()
```
BryanFauble, line 164: Some extra code to help guard against extra things from being tracked in memory (It's super minor)

schematic/models/validate_manifest.py
```diff
@@ -193,14 +180,16 @@ def validate_manifest_rules(
                     checkpoint_name=ge_helpers.checkpoint_name,
                     batch_request={
                         "runtime_parameters": {"batch_data": manifest},
-                        "batch_identifiers": {"default_identifier_name": "manifestID"},
+                        "batch_identifiers": {
+                            "default_identifier_name": f"manifestID_{uuid.uuid4()}"
+                        },
```
BryanFauble, line 185: Unique identifier to prevent name collisions

schematic/models/validate_manifest.py
```diff
@@ -321,12 +315,21 @@ def validate_manifest_values(
...
+                    sorted_error.path[0]
+                    if len(sorted_error.path) > 0
+                    else "Wrong schema"
+                )
+                errorMsg = sorted_error.message[0:500]
+                errorVal = (
+                    sorted_error.instance
+                    if len(sorted_error.path) > 0
+                    else "Wrong schema"
+                )
```
BryanFauble, line 332: This is just automatically re-formatted code. I did not make any changes here.

schematic/store/synapse_tracker.py
```diff
@@ -0,0 +1,103 @@
+"""This script is responsible for creating a 'pull through cache' class that can be
```
BryanFauble, line 1: This file is the main star of all these changes. Please read carefully.

schematic/store/synapse.py
```diff
@@ -81,31 +85,93 @@ class ManifestDownload(object):
...
+            return manifest_data
+
         if "SECRETS_MANAGER_SECRETS" in os.environ:
             temporary_manifest_storage = "/var/tmp/temp_manifest_download"
-            # clear out all the existing manifests
-            if os.path.exists(temporary_manifest_storage):
-                shutil.rmtree(temporary_manifest_storage)
+            cleanup_temporary_storage(
+                temporary_manifest_storage, time_delta_seconds=3600
+            )
```
BryanFauble, line 126: Instead of blanket removing previously downloaded manifest files instead this will have a 1 hour delay to clean them up. This was implemented to prevent a manifest from getting cleaned up that might be actively being used, or used for a successful cache pull (ie: Not downloading the manifest file again).

schematic/store/synapse.py
```diff
@@ -81,31 +85,93 @@ class ManifestDownload(object):
...
+            os.rename(manifest_data.path, manifest_original_name_and_path)
+            manifest_data.path = manifest_original_name_and_path
+            self.syn.cache.add(
+                file_handle_id=manifest_data.dataFileHandleId,
+                path=manifest_original_name_and_path,
+                md5=manifest_data._file_handle.contentMd5,
+            )
+
         return manifest_data
```
BryanFauble, line 175: This is a potentially impactful change here @andrewelamb @linglp @milen-sage that could use a good look. Basically what this is solving:

(When NOT running on AWS)
1. When manifest were downloaded the `downloadAs` name, or the name of the actual file, could conflict with a file that was already in the location being downloaded to. This would overwrite the file and potentially cause the test, or whatever kick off that code to fail.

When running on either AWS or not

1. When downloading an entity where we already have a local copy of it (Think of a file that exists within the synapse cachemap) it will copy that file over to the requested location. The issue however is that the file that is copied retains whatever name it has been renamed to (Some tests do this, and code in schematic does this). Because of this I have the `os.rename` logic in place to set the name of the copied file back to the expected file name so that any subsequent code behaves as expected.

schematic/store/synapse.py
```diff
@@ -490,8 +584,8 @@ def getStorageProjects(self, project_scope: List = None) -> list[tuple[str, str]
         # prepare a return list of project IDs and names
         projects = []
         for projectId in storageProjects:
-            projectName = self.syn.get(projectId, downloadFile=False).name
-            projects.append((projectId, projectName))
+            project_name_from_project_header = project_id_to_name_dict.get(projectId)
+            projects.append((projectId, project_name_from_project_header))
```
BryanFauble, line 588: There are a few things to note here on these changes:

1. I have added a change to SYNPY (https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1130) that stored the ID of the user profile that logged into Synapse. This removes the need to call to `getUserProfile()` and saves an extra HTTP call.
2. I am storing the result of getting the paginated list of project headers into the `synapse_entity_tracker`. It is really slow to crawl this information and it occurs several times. By doing this the content is re-used in the context of the request.
3. I removed the need to call `syn.get()` for each of the projects as the paginated list of project headers contained all the data we needed.

schematic/store/synapse.py
```diff
@@ -1382,34 +1493,27 @@ def buildDB(
...
-        table_info = self.get_table_info(datasetId=datasetId)
-        # Put table manifest onto synapse
-        schema = Schema(
-            name=table_name,
-            columns=col_schema,
-            parent=self.getDatasetProject(datasetId),
+        table_parent_id = self.getDatasetProject(datasetId=datasetId)
+        existing_table_id = self.syn.findEntityId(
+            name=table_name, parent=table_parent_id
         )
```
BryanFauble, line 1499: Swapping to `findEntityId` prevents the need to walk the project to find the applicable table ID. Instead we can utilize the Synapse Rest API to find the ID for us.

BryanFauble: This will return `None` if not found

schematic/store/synapse.py
```diff
@@ -1423,11 +1527,18 @@ def buildDB(
...
+            )
+            annos = OldAnnotations(
+                id=table_entity.id,
+                etag=table_entity.etag,
+                values=table_entity.annotations,
+            )
             annos["primary_key"] = table_manifest["Component"][0] + "_id"
             annos = self.syn.set_annotations(annos)
+            table_entity.etag = annos.etag
+            table_entity.annotations = annos
```
BryanFauble, line 1541: In this file 2 sets of Annotations from SYNPY are being used:
```
from synapseclient.models.annotations import Annotations # This is the new set of Annotations that has Async support
from synapseclient import Annotations as OldAnnotations # This is the old set of Annotations that works with the current Entity objects
```
Since all of the Entities returned from `syn.get()` still use `OldAnnotations` I am using it in this context.

schematic/store/synapse.py
```diff
@@ -1467,24 +1578,89 @@ def upload_manifest_file(
...
+            manifest_synapse_file_id = manifest_synapse_file.id
+
+        finally:
+            # Revert the file name back to the original
+            os.rename(new_file_path, original_file_path)
+
+            if manifest_synapse_file:
+                manifest_synapse_file.path = original_file_path
+
+        return manifest_synapse_file_id
```
BryanFauble, line 1607: Doing this prevents the need to call `changeFileMetaData` as the file will be uploaded with the expected name. That function was/is super slow so this is a big gain.

schematic/store/synapse.py
```diff
@@ -1584,17 +1794,17 @@ async def format_row_annotations(
...
                     and re.fullmatch(csv_list_regex, anno_v)
                     and rule_in_rule_list(
                         "list", dmge.get_node_validation_rules(anno_k)
                     )
                 ):
-                    annos[anno_k] = anno_v.split(",")
+                    annos["annotations"]["annotations"][anno_k] = anno_v.split(",")
                 else:
-                    annos[anno_k] = anno_v
+                    annos["annotations"]["annotations"][anno_k] = anno_v
```
BryanFauble, line 1806: I pulled in some changes from: https://github.com/Sage-Bionetworks/schematic/pull/1485 so I could make sure everything was working. I will handle any merge conflicts if that code goes first.

schematic/store/synapse.py
```diff
@@ -2803,7 +3072,7 @@ def _get_auth_token(
         # Try getting creds from .synapseConfig file if it exists
         # Primarily useful for local users. Seems to correlate with credentials stored in synaspe object when logged in
         if os.path.exists(CONFIG.synapse_configuration_path):
-            config = self.synStore.syn.getConfigFile(CONFIG.synapse_configuration_path)
+            config = get_config_file(CONFIG.synapse_configuration_path)
```
BryanFauble, line 3074: Moved to the non-deprecated function (Same behavior)

schematic_api/api/routes.py
```diff
@@ -43,77 +31,10 @@
...
-                    span.set_attribute(name, value)
-                # Call the actual function
-                result = func(*args, **kwargs)
-                return result
-
-        return wrapper
-
-    return decorator
-
-
```
BryanFauble, line 116: This is no longer needed with the request hooks that I added for the `FlaskInstrumentor`

schematic_api/api/routes.py
```diff
@@ -182,7 +103,7 @@ def convert_df_to_csv(self, df, file_name):
         """
 
         # convert dataframe to a temporary csv file
-        temp_dir = tempfile.gettempdir()
+        temp_dir = create_temp_folder(path=tempfile.gettempdir())
```
BryanFauble, line 106: This was failing with high concurrency as `tempfile.gettempdir()` only returned back `/tmp`, and for tests where the name of the uploaded file was the same was causing errors. By doing this we are guaranteed to get back a unique temporary directory.

schematic/store/database/README.md
thomasyu888, line 1: Thanks for adding this readme!

.github/workflows/api_test.yml
```diff
@@ -43,21 +43,11 @@ jobs:
...
-      #       load cached venv if cache exists
-      #----------------------------------------------
-      - name: Load cached venv
-        id: cached-poetry-dependencies
-        uses: actions/cache@v4
-        with:
-          path: .venv
-          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}-${{ matrix.python-version }}
-
+
```
BryanFauble, line 46: I pulled the changes from https://github.com/Sage-Bionetworks/schematic/pull/1507 - As my build started to fail. I will pull in develop when this is merged in.

schematic/models/GE_Helpers.py
thomasyu888: Should this be a comment?

schematic/store/synapse_tracker.py
```diff
@@ -0,0 +1,140 @@
...
+            current_user_id: profile id for the user whose projects we want to get.
+
+        Returns:
+            A list of dictionaries matching <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/ProjectHeader.html>
+        """
+        project_headers = self.project_headers.get(current_user_id, None)
+        if project_headers:
+            return project_headers
+
+        all_results = syn.restGET(
```
thomasyu888, line 117: Nit: There is a syn._GET_Paginated if that makes it easier

schematic_api/api/routes.py
thomasyu888: I wonder if this impacts deployment, not sure if the /tmp directory is mounted in any way (probably not)

thomasyu888 (commented):
I did a quick initial scan but I'll leave it to @BWMac / @jaymedina / @linglp / @andrewelamb / @GiaJordan for further comments before I go deeper into it.

I found your self review very helpful!

schematic/store/synapse_tracker.py
```diff
@@ -0,0 +1,140 @@
...
+            retrieve_if_not_present: If True, retrieve the entity if it is not present in the cache.
+            download_location: The location to download the file to.
+            if_collision: The action to take if there is a collision when downloading the file.
+
+        Returns:
+            The Synapse entity.
+        """
+        entity = self.synapse_entities.get(synapse_id, None)
+
+        if entity is None or (download_file and not entity.path):
```
linglp, line 50: I found the logic here a bit difficult to follow without a comment. Could you add: "Returns None if `retrieve_if_not_present` is False and the entity is not found in the cache" as a comment?

schematic/store/synapse_tracker.py
```diff
@@ -0,0 +1,140 @@
...
+    ) -> None:
+        """Adds a Synapse entity to the cache.
+
+        Args:
+            synapse_id: The Synapse ID of the entity to add.
+            entity: The Synapse entity to add.
+        """
+        self.synapse_entities.update({synapse_id: entity})
+
+    def remove(self, synapse_id: str) -> None:
```
linglp, line 73: Thanks for adding this @BryanFauble ! I am curious what do you envision the use case for this?

schematic/store/synapse.py
```diff
@@ -1467,24 +1585,89 @@ def upload_manifest_file(
...
+            new_file_path (str): The path to the new manifest file
+            dataset_id (str): The Synapse ID of the dataset the manifest is associated with
+            existing_file_name (str): The name of the existing file
+            file_name_new (str): The name of the new file
+            restrict_manifest (bool): Whether the manifest should be restricted
+
+        Returns:
+            File: The stored manifest file
+        """
+        local_tracked_file_instance = (
```
linglp, line 1640: I want to make sure that I understand the logic here correctly. Is the following correct? : 
1)  locate a file whose name matches existing_file_name
2) if the first step is None, then locate a file whose name matches file_name_new
3) if both are None, then local_tracked_file_instance is None

schematic/store/synapse.py
```diff
@@ -2082,22 +2297,28 @@ def upload_manifest_as_table(
             )
         # Load manifest to synapse as a CSV File
         manifest_synapse_file_id = self.upload_manifest_file(
-            manifest,
-            metadataManifestPath,
-            datasetId,
-            restrict,
+            manifest=manifest,
```
linglp, line 2300: I am seeing that you made similar changes in other parts of the code base by calling out the name of the parameter function. Is that a good practice?

schematic/utils/io_utils.py
```diff
@@ -31,12 +32,50 @@ def export_json(json_doc: Any, file_path: str) -> None:
...
     return load_json(schema_org_path)
+
+
+def cleanup_temporary_storage(
+    temporary_storage_directory: str, time_delta_seconds: int
+) -> None:
+    """Handles cleanup of temporary storage directory. The usage of the
+    `time_delta_seconds` parameter is to prevent deleting files that are currently
+    being used by other requests. In production we will be deleting those files
+    which have not been modified for more than 1 hour.
```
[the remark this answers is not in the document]
BryanFauble: It's this code:

```
                if os.path.isfile(file_path) and os.path.getmtime(file_path) < (
                    time.time() - time_delta_seconds
                ):
```

And `3600` (1 hour) is being passed as `time_delta_seconds`

schematic/store/synapse_tracker.py
```diff
@@ -0,0 +1,140 @@
...
+        if_collision: str = None,
+    ) -> Union[Entity, Project, File, Folder, Schema]:
+        """Retrieves a Synapse entity from the cache if it exists, otherwise downloads
+        the entity from Synapse and adds it to the cache.
+
+        Args:
+            synapse_id: The Synapse ID of the entity to retrieve.
+            syn: A Synapse object.
+            download_file: If True, download the file.
+            retrieve_if_not_present: If True, retrieve the entity if it is not present in the cache.
```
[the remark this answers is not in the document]
BryanFauble: Keeping both leaves the door open to use the function to retrieve metadata about an entity without the possibility of downloading that file. Consider these cases:

```
synapse_entity_tracker.get(
    synapse_id="syn123",
    syn=self.syn,
    download_file=False,
    retrieve_if_not_present=False,
)
```

```
synapse_entity_tracker.get(
    synapse_id="syn123",
    syn=self.syn,
    download_file=True,
    retrieve_if_not_present=True,
)
```

```
synapse_entity_tracker.get(
    synapse_id="syn123",
    syn=self.syn,
    download_file=False,
    retrieve_if_not_present=True,
)
```

The following case is the only one not valid as it would not download the file:

```
synapse_entity_tracker.get(
    synapse_id="syn123",
    syn=self.syn,
    download_file=True,
    retrieve_if_not_present=False,
)
```

schematic/store/synapse_tracker.py
BryanFauble: Updated docstring to read:
```
        """Retrieves a Synapse entity from the cache if it exists, otherwise downloads
        the entity from Synapse and adds it to the cache.

        Args:
            synapse_id: The Synapse ID of the entity to retrieve.
            syn: A Synapse object.
            download_file: If True, download the file.
            retrieve_if_not_present: If True, retrieve the entity if it is not present 
                in the cache. If not found in the cache, and this is False, return None.
            download_location: The location to download the file to.
            if_collision: The action to take if there is a collision when downloading 
                the file. May be "overwrite.local", "keep.local", or "keep.both". A 
                collision occurs when a file with the same name already exists at the 
                download location.

        Returns:
            The Synapse entity if found. When retrieve_if_not_present is False and the
            entity is not found in the local cache, returns None. If
            retrieve_if_not_present is True and the entity is not found in the local
            cache, retrieve the entity from Synapse and add it to the cache.
        """
```

schematic/store/synapse_tracker.py
BryanFauble: See https://sagebionetworks.jira.com/browse/PLFM-8605 - We must remove entities from the local cache when the table data is updated.

schematic/store/synapse.py
BryanFauble: Yes that's correct

schematic/store/synapse.py
BryanFauble: Yes. IMO Positional arguments are only good when there are 1, maybe 2 arguments you can pass into a function. Any more then it makes it significantly more difficult to skim the code and understand what is going on.

schematic/store/synapse_tracker.py
BryanFauble: That, nor `_POST_paginated` would appear to work in this case as `nextPageToken` is a query parameter, instead of a piece of the body.

The return type of:
https://rest-docs.synapse.org/rest/GET/projects/user/principalId.html

Is different in it's format to the `_GET_Paginated` expected type of:
https://rest-docs.synapse.org/rest/GET/teamMembers/id.html

schematic/models/GE_Helpers.py
BryanFauble: Gotcha @GiaJordan What are your thoughts if I only call `self.context.build_data_docs(resource_identifiers=[suite_identifier])` if requested by the person running the code? This means when schematic is running as an API server it will never build these HTML files.

This is extra processing currently occurring in production schematic that is not needed. It's small, but non-zero.

On a side note: Do we have any environment variables/flags I could re-use that area already used in other areas of the code?

schematic/manifest/commands.py
BryanFauble: Added! @linglp

schematic/models/GE_Helpers.py
BryanFauble: Added!

schematic/models/GE_Helpers.py
GiaJordan: @BryanFauble yeah that's fine. And there isn't one we've made that can be reused, but I think it would make sense to tie it to the verbosity requested for the command and include both lines like
```
if logger.isEnabledFor(logging.DEBUG):
    self.context.build_data_docs(resource_identifiers=[suite_identifier])
    #Webpage DataDocs opened here:
    self.context.open_data_docs(resource_identifier=suite_identifier)
```
It won't log to the logger, but `open_data_docs` will open the html file with the useful information in a browser and won't interrupt the rest of the run

schematic/store/synapse_tracker.py
```diff
@@ -0,0 +1,140 @@
...
+            download_location: The location to download the file to.
+            if_collision: The action to take if there is a collision when downloading the file.
```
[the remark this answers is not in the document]
BryanFauble: Added!

schematic/store/synapse_tracker.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def get(
+        self,
+        synapse_id: str,
+        syn: synapseclient.Synapse,
+        download_file: bool = False,
+        retrieve_if_not_present: bool = True,
+        download_location: str = None,
+        if_collision: str = None,
+    ) -> Union[Entity, Project, File, Folder, Schema]:
```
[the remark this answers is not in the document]
BryanFauble: Added!

schematic/models/GE_Helpers.py
BryanFauble: Perfect @GiaJordan I updated this code to:

```

        if logger.isEnabledFor(logging.DEBUG):
            self.context.build_data_docs(resource_identifiers=[suite_identifier])
            ##Webpage DataDocs opened here:
            # self.context.open_data_docs(resource_identifier=suite_identifier)
```

I am going to leave `open_data_docs` commented out like it was.

schematic/store/synapse.py
linglp: Got it Bryan! Thanks for the explanation.

schematic/store/synapse.py
linglp: Thanks Bryan!

schematic/manifest/commands.py
linglp: Thanks Bryan!

schematic/store/synapse_tracker.py
linglp: Hi @BryanFauble ! Thanks for providing more context. I read the Jira ticket that you linked but I don't think that I understand. Based on the ticket and the comment, I am guessing that `store` function is not returning the etag of table because UploadToTableResult.etag is not the etag of the table. But why is it related to removing entities from the local cache?

BryanFauble: @linglp Let's recap the concept of eTag and what it's used for in Synapse:

> Synapse employs an Optimistic Concurrency Control (OCC) scheme to handle concurrent updates. Since the E-Tag changes every time an entity is updated it is used to detect when a client's current representation of an entity is out-of-date.

We can take a simple example where suppose we create a Table and it's assigned an etag of `1`, I go ahead and add an annotation to the table and that Table now has a new eTag `2`. When I updated those annotations we will want to make sure that we capture in the local cache:

1. The current state of the entity (This includes all of the traditional fields about the Entity, AND, the annotations for that entity)
2. The updated eTag as a result of saving those annotations.

The above works because when we make the change to add those annotations the Synapse API gives us back the eTag of the updated entity and we can store that back into our local in-memory cache.


In the case of updates to the data or schema of a table the eTag is also getting updated. Let's suppose we add data to a table, Synapse will now update that eTag to `3`, however, us as the caller have no way of knowing that Synapse updated the eTag to `3`, we only ever know it was `2`, and we never got the updated eTag as a result of submitting our transaction. 

Ask: What would occur if I try to submit another transaction on that Table and pass along my known eTag of `2`? Synapse would reject the transaction because we on the client side do not have the most up to date version of that entity that Synapse has. This is the root of what the ticket is talking about/addressing.


Why do we need to remove the entity from our local in-memory cache? It's the result of the `Ask` above. Since we do not have the most up to date eTag of that entity, the next time we read for that entity in order to perform some operation, that read MUST be sent to Synapse. We cannot trust our local in-memory cache anymore because we know we do not have the up to date eTag of the entity. That is what the `.remove()` operation is for.

schematic/store/synapse.py
```diff
@@ -77,35 +81,98 @@ class ManifestDownload(object):
...
     syn: an object of type synapseclient.
     manifest_id: id of a manifest
+    synapse_entity_tracker: Tracker for a pull-through cache of Synapse entities
     """
 
     syn: synapseclient.Synapse
     manifest_id: str
+    synapse_entity_tracker: SynapseEntiyTracker = field(
+        default_factory=SynapseEntiyTracker
+    )
```
andrewelamb, line 91: Entity has a typo

schematic/store/synapse.py
```diff
@@ -77,35 +81,98 @@ class ManifestDownload(object):
...
+            )
+            os.rename(manifest_data.path, manifest_original_name_and_path)
+            manifest_data.path = manifest_original_name_and_path
+            self.syn.cache.add(
+                file_handle_id=manifest_data.dataFileHandleId,
+                path=manifest_original_name_and_path,
+                md5=manifest_data._file_handle.contentMd5,
+            )
+
         return manifest_data
```
andrewelamb, line 176: I see some use of protected attibutes in the File Enntity class. Are these somethign that could be made pubic in the Python client in the near future?

schematic/store/synapse.py
andrewelamb: I'm curious why renaming the file is a concern. Is that something schematic was doing?

schematic/store/synapse_tracker.py
andrewelamb: This question is probably due to how little time I ahve spent in this module. I'm curious why this was needed in the first place? Was Schematic repeatedly downloading the same entity in the same API call?

schematic/store/synapse.py
BryanFauble: The intention is that these FileEntities are going to be deprecated/removed to be replaced by this interface:
https://python-docs.synapse.org/reference/oop/models_async/#synapseclient.models.File

schematic/store/synapse.py
BryanFauble: lol. I can't believe I hadn't noticed this before :D

Patched!

schematic/store/synapse_tracker.py
BryanFauble: > Was Schematic repeatedly downloading the same entity in the same API call?

Yes. Early in the script metadata about an entity was retrieved from Synapse. We can take the manifest FileEntity for example. Every time that manifest needed to be referenced the metadata from Synapse was re-retrieved.

In the benchmarking that I had done it cut down a significant number of calls to Synapse. In the screenshot I ran all tests prefixed with `test_submit_%` and the total number of HTTP calls went from 972 -> 594.
image

andrewelamb (approved):

PR #1506, opened 2024-09-18
Fix `pdoc` Python Version

BWMac:
Apparently, the `PYTHON_VERSION` needs to be a [string](https://github.com/Sage-Bionetworks/schematic/actions/runs/10928918669/job/30338417623). Apologies for all of the PRs stemming from the `pdoc` issue!


BWMac:
I added the string `"3.10"` in #1496

PR #1512, opened 2024-10-03
[FDS-2491] Integration tests for Schematic API Test plan

BryanFauble:
**Problem:**

1. Running through the test plan is a largely manual process: https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/3055779846/Schematic+API+test+plan
2. We do not have integration tests covering this functionality, and those tests we do have do not verify everything that the manual test would test for.
3. Some items cannot be verified in code and still need manual verification. Those items will be marked as such.

**Solution:**

1. Creating integration tests starting with manifest generation.
2. Updating the verification of the tests to assert as much information about the generated excel/google sheets as we can.
3. Writing out the excel file to a directory where files can be manually checked for the portions that cannot be verified in code.

**Testing:**

1. I verified that while running the Schematic API server locally I could run these tests.

**TODO:**

- [x] Mark any tests with pytest mark that require manual verification
- [x] Create some sort of feature flag that is only set when running the tests manually (IE: If tests are running in the CI pipeline, skip writing files out for manual verification)
- [x] Structure the tests to allow using a flask server passed in as a pytest fixture, OR, the schematic API server running locally, in dev, or in staging.
- [x] Document (Likely in confluence) the tests that need manual verification along with what needs verification - In addition to the docstring of the tests themselves.


tests/integration/test_manifest_generation.py
```diff
@@ -0,0 +1,503 @@
...
+        # Load the response content into memory
+        content = BytesIO(response.content)
+        workbook = load_workbook(content)
+        sheet1 = workbook["Sheet1"]
+
+        # Track column positions
+        columns = {cell.value: cell.column_letter for cell in sheet1[1]}
+
+        # AND the content of the first sheet is as expected
+        assert sheet1[f"{columns['Patient ID']}1"].value == "Patient ID"
```
BWMac, line 77: @BryanFauble Rather than having all of these assert statements with manually entered values, would it make more sense to use the provided example files and directly compare the generated manifest file to the example where possible? Like
```
    for col in test_sheet.iter_cols(1, test_sheet.max_column):
        for cell in col:
            assert cell.value == example_sheet[cell.coordinate].value
```

BryanFauble: Let's consider a few possibilities:

1. Example files stored outside of source control are the "Source of truth" for the tests that are running. If those example files are modified and tests are not updated "In lockstep" then the tests are going to fail for some amount of time.
2. Example files stored inside source control, as example files are updated, so too, are the tests that are using them.
3. There are no example files stored anywhere and we are doing the verbose verification of the content of those files like what is being done here.

@linglp @andrewelamb @GiaJordan I would like you opinion on the direction that we should take here.

BryanFauble: Something else to think about as well is that the final file to compare against is not the file as it is downloaded from the schematic API. It is the state of the file after 1 or more manipulations have been done to it (See the case where the test plan mentions changing drop downs to verify conditional formatting).

Those change would need to be made in the code in order to have the "same" file to compare.

This is one of the things that is point me towards handling all of the verification in the code so we have 1 single spot to maintain everything. Yes, it is a bit more maintenance, but it's nothing that an LLM cannot do for us.

tests/integration/test_manifest_submission.py
```diff
@@ -0,0 +1,97 @@
...
+            "hide_blanks": "false",
+            "asset_view": "syn63561606",
+            "table_manipulation": "replace",
+            "table_column_names": "class_label",
+            "annotation_keys": "class_label",
+            "file_annotations_upload": "false",
+        }
+
+        syn = Synapse()
+        syn.login(authToken=syn_token, silent=True)
```
BryanFauble, line 54: @linglp Should this go into conftest.py instead? Or come from the `synapse_store: SynapseStorage` instance that you can grab for the test?

tests/integration/test_manifest_submission.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        manifest_submitted_df = pd.read_csv(manifest_file_path)
+        assert "entityId" in manifest_submitted_df.columns
+        assert "Id" in manifest_submitted_df.columns
+
+        # make sure that Id column is not empty
+        assert manifest_submitted_df["Id"].notnull().all()
+
+        # clean up
+        schedule_for_cleanup(CleanupItem(manifest_id))
```
BryanFauble, line 97: @linglp 
Can you add the cleanup directly after getting the ID of the resource. The way this is setup now is that if the assertion fails this line won't be hit, nor the resource cleaned up.

tests/integration/test_manifest_submission.py
```diff
@@ -0,0 +1,97 @@
...
+    if not os.path.exists(temporary_manifest_storage):
+        os.makedirs(temporary_manifest_storage)
+    full_parent_path = os.path.abspath(temporary_manifest_storage)
+
+    download_location = create_temp_folder(full_parent_path)
+    yield download_location
+
+    # Cleanup after tests have used the temp folder
+    if os.path.exists(download_location):
+        shutil.rmtree(download_location)
```
BryanFauble, line 29: @linglp 
Should this go into conftest.py so it can be re-used? The name of the storage location doesn't matter quite so much, so it could be anything.

Also to note, a session scope with a temp folder that does not have a unique name is problematic. The reason is that when xdist runs tests it starts up new processes to run the code. Each process has a unique session scope that is not shared. When you use a non-unique folder to store content, which later gets cleaned up, it is introducing a race condition in the code. If the same worker picks up each tests that uses this location generator - Great no issues. If different workers pick up tests that share this, whoever finishes the test first is going to delete the directory tree and break the test that might still be running.

https://github.com/Sage-Bionetworks/schematic/pull/1505/files#diff-664ac28bfa7dd824a516813cb7f560aa7207c6a7bb54b7053cc9dcfd9c157d46 includes the changes I had introduced to solve this non-unique problem (However, that code has not been merged)

tests/integration/test_manifest_submission.py
linglp: Yes of course.

tests/integration/test_manifest_submission.py
linglp: I could add this to conftest.py!

tests/integration/test_manifest_generation.py
```diff
@@ -318,10 +346,17 @@ def test_single_manifest_generation_google_sheet(
         assert columns["key_datetime"] is not None
         assert columns["entityId"] is not None
 
-        assert (
-            sheet1[f"{columns['Filename']}2"].value
-            == "Manifest generation - Manual test - generate an existing manifest/test dataset/test dataset 1/sample A.txt"
-        )
+        if local_flask_instance:
```
BryanFauble, line 349: This was a little odd that there is a difference here between the local instance and the deployed instance, anyone have ideas here?

thomasyu888: I haven't looked closely, the dev instance is not the most up to date release yet - so there may be differences.

tests/integration/test_manifest_submission.py
linglp: Thanks for the feedback @BryanFauble . I updated to move download_location fixture to conftest and use uuid as folder name. Does that work?

tests/integration/test_manifest_generation.py
linglp: @BryanFauble Tom is right. The file path feature has not yet been deployed (so there are differences). After we deploy, this ` if local_flask_instance:` can be removed.

BryanFauble: When is the dev instance being deployed? Shouldn't this already be deployed if the ask is that we are running through the test plan?

Isn't everything a moot point for testing against the dev instance of schematic if we aren't running the version of schematic in dev we want to test against?

tests/integration/test_manifest_submission.py
BryanFauble: > I updated to move download_location fixture to conftest

This is great!

> use uuid as folder name. Does that work?

This does not work. I added a commit to this branch, take a look: https://github.com/Sage-Bionetworks/schematic/pull/1512/commits/9fc7fd1b2d1d9b33f8764f218c4984a0ceac2c7e


In the end out goal is that these temporary directories are being make in the temp directory for the applicable OS that is running the test. The solution that you proposed was creating the temp directory at the working directory where the code was being ran.

linglp: I see. Thanks for explaining Bryan. To make sure that I understand, the difference is that tempfile.gettempdir() would create temporary folders in the system's temp directory and my approach was creating unique folders at the working directory?

BryanFauble: Yes, that's correct @linglp

linglp: Thanks Bryan. I will mark this comment as resolved :)

tests/integration/test_manifest_generation.py
linglp: Thanks for the questions @BryanFauble . We can deploy to AWS dev any time (there's no technical blockers). But based on FAIR data team's current release process, we go through test plans before we release to PYPI. Usually after PYPI release, we would release to aws dev environment. However, because of the bug related to nested data structure Schematic v24.10.1 was deleted in PYPI and we never deploy to AWS dev. 

But you are right! I agree that we could expedite deploying to AWS dev since that will make testing changes on AWS sooner.

tests/integration/test_manifest_submission.py
```diff
@@ -0,0 +1,318 @@
...
+
+        Args:
+            response (Any): The response containing the manifest ID.
+            syn (Synapse): An instance of the Synapse client.
+            data_type (str): The data type used in manifest.
+            download_location (str): path to download location
+            schedule_for_cleanup (Callable[[CleanupItem], None]): Returns a closure that takes an item that should be scheduled for cleanup.
+            testing_config (ConfigurationForTesting): Confiugration for testing
+        """
+        # Get the manifest ID from the response
```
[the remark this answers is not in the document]
BryanFauble: Your suggestion works!

linglp:
@BryanFauble see PR here: https://github.com/Sage-Bionetworks/schematic-infra/pull/154. I deployed the latest develop to AWS dev.

schematic/utils/general.py
```diff
@@ -224,6 +224,9 @@ def create_temp_folder(path: str) -> str:
         path(str): a directory path where all the temporary files will live
     Returns: returns the absolute pathname of the new directory.
     """
+    if not os.path.exists(path):
```
thomasyu888, line 227: Is there any reason we would want an error to be thrown if the path doesn't exist?

BryanFauble: Not in this case I don't think

thomasyu888 (approved):
🔥 LGTM!  Incredible work here

tests/integration/test_manifest_generation.py
andrewelamb: Sorry, I must have missed this when I first got tagged. My preference would be to use a "source of truth" file to test all cell values. Something like

```
test_df = pnadas.read_excel(response.content) # not sure of pandas cna do this
expected_df = pandas.read_csv("expected.csv")

for col in expected_df.columns:
  assert expected_df[col].values.to_list() == test_df[col].values.to_list()
```
But it's nto a strong perefence and doesn't need to be adressed in this PR.

tests/integration/test_manifest_generation.py
```diff
@@ -0,0 +1,1217 @@
...
+            assert sheet1[f"{columns[col]}1"].fill.start_color.index == LIGHT_BLUE
+
+        for col in [
+            "Genome Build",
+            "Genome FASTA",
+            "entityId",
+        ]:
+            assert sheet1[f"{columns[col]}1"].fill.start_color.index == GRAY
+
+        # AND conditional formatting is functioning as expected (MANUAL VERIFICATION)
```
andrewelamb, line 712: This comment is a little confusing to me. Is it referencing the tests right below it?

andrewelamb (approved):

BWMac (approved):

tests/integration/test_manifest_generation.py
BryanFauble: It is setting a value right below this line:
`workbook["Sheet1"][f"{columns['Diagnosis']}2"].value = "Cancer"`

that must be manually verified for conditional formatting. The library does not re-render the fill color of cells so it's a part of the instructions in the docstring of things to verify. Since we wanted to make manual verification as simple as "Open this and look at this", we're setting the values ahead of time so no one needs to mess around with changing values in the cells themselves. The test does it for them.

PR #1515, opened 2024-10-03
Adds Manifest Generation Tests to Common Testing Branch

BWMac:
This PR adds tests for the second (`test_single_manifest_generation_google_sheet_no_annotations`) and fourth (`test_manifest_generation_multiple_blank_google_sheets`) use cases in the [test plan](https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/3055779846/Schematic+API+test+plan). I followed the example set in the existing tests and updated the new tests to use the logic for testing on different deployments/local flask server.

### DIFF
--- a/tests/integration/test_manifest_generation.py
+++ b/tests/integration/test_manifest_generation.py
@@ -833,7 +833,7 @@ class TestManifestGeneration:
         - When File Format = "CRAM", [Genome Build, Genome FASTA] is Light Blue (Required)
         - When File Format = "FASTQ", [Genome Build] is White (Optional)
         """
-        url = "http://localhost:3001/v1/manifest/generate"
+        url = f"{schematic_api_server_url}/v1/manifest/generate"
         # GIVEN a valid request to the Schematic API to generate two blank Google Sheets manifests
         params = {
             "schema_url": "https://raw.githubusercontent.com/Sage-Bionetworks/schematic/develop/tests/data/example.model.jsonld",
--- a/tests/integration/test_manifest_generation.py
+++ b/tests/integration/test_manifest_generation.py
@@ -543,12 +543,13 @@ class TestManifestGeneration:
         assert sheet2["G1"].value is None
 
         # AND a copy of the Excel file is saved to the test directory for manual verification
-        workbook.save(
-            os.path.join(
-                manual_test_verification_path,
-                "TestManifestGeneration_test_single_manifest_generation_google_sheet_with_annotations.xlsx",
+        if manual_test_verification_enabled:
+            workbook.save(
+                os.path.join(
+                    manual_test_verification_path,
+                    "TestManifestGeneration_test_single_manifest_generation_google_sheet_with_annotations.xlsx",
+                )
             )
-        )
 
     @pytest.mark.manual_verification_required
     def test_single_manifest_generation_google_sheet_no_annotations(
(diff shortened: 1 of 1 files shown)


tests/integration/test_manifest_generation.py
```diff
@@ -498,6 +498,670 @@ def test_single_manifest_generation_google_sheet(
...
+        sheet2 = workbook["Sheet2"]
+
+        # AND the column headers in "Sheet1" are as expected
+        assert sheet1["A1"].value == "Filename"
+        assert sheet1["B1"].value == "Sample ID"
+        assert sheet1["C1"].value == "File Format"
+        assert sheet1["D1"].value == "Component"
+        assert sheet1["E1"].value == "Genome Build"
+        assert sheet1["F1"].value == "Genome FASTA"
+        assert sheet1["G1"].value == "entityId"
```
BryanFauble, line 575: One of the things I had run into was that the column ordering can change. You have that `track column positions` code below, but these assertions may fail one day. From what @andrewelamb had mentioned, let's operate under the assumption that column order may change, and worry about ensuring a consistent order later on.

tests/integration/test_manifest_generation.py
```diff
@@ -498,6 +498,670 @@ def test_single_manifest_generation_google_sheet(
...
+        """
+        Download a manifest without annotations from the Schematic API and verify that it is a valid Google
+        Sheet. We are validating the following:
+
+        - The first row of the Google Sheet contains the column headers
+        - The first row is locked on scroll
+        - Each cell A-F in the first row has a comment "TBD"
+        - Cell G in the first row does not have a comment
+        - The cell C2 in "Sheet1" has a dropdown list with values from "Sheet2!C2:C5"
+        - The cell E2 in "Sheet1" has a dropdown list with values from "Sheet2!E2:E5"
```
BryanFauble, line 517: I also need to update my docstring, but if we operate under the assumption the order may change, these comments are outdated.

tests/integration/test_manifest_generation.py
```diff
@@ -498,6 +498,670 @@ def test_single_manifest_generation_google_sheet(
...
+        # AND the second sheet is hidden
+        assert sheet2.sheet_state == "hidden"
+
+        # AND the values in "Sheet2" are as expected
+        assert sheet2["A1"].value == "Filename"
+        assert sheet2["B1"].value == "Sample ID"
+        assert sheet2["C1"].value == "File Format"
+        assert sheet2["D1"].value == "Component"
+        assert sheet2["E1"].value == "Genome Build"
+        assert sheet2["F1"].value == "Genome FASTA"
```
BryanFauble, line 698: Could this ordering change 🤔 I would need to update my code as well. I haven't gotten any failures for this second sheet changing, so we're probably ok to assume it won't change.

tests/integration/test_manifest_generation.py
```diff
@@ -543,10 +543,683 @@ def test_single_manifest_generation_google_sheet(
...
-                os.path.join(
-                    manual_test_verification_path,
-                    "TestManifestGeneration_test_single_manifest_generation_google_sheet.xlsx",
-                )
+        workbook.save(
+            os.path.join(
+                manual_test_verification_path,
+                "TestManifestGeneration_test_single_manifest_generation_google_sheet_with_annotations.xlsx",
+            )
+        )
```
BryanFauble, line 551: I assume this was revert because of a merge conflict. Could you update this to only save the workbook when the flag is set to True

BWMac: Thanks for catching that. Updated in latest commit!

tests/integration/test_manifest_generation.py
```diff
@@ -547,6 +547,683 @@ def test_single_manifest_generation_google_sheet(
...
+        - "Sheet2" is hidden
+        - The values in "Sheet1" are as expected
+        - The values in "Sheet2" are as expected
+
+        Manual verification steps:
+        - When File Format = "BAM", [Genome Build] is Light Blue (Required)
+        - When File Format = "CRAM", [Genome Build, Genome FASTA] is Light Blue (Required)
+        - When File Format = "FASTQ", [Genome Build] is White (Optional)
+        """
+        url = "http://localhost:3001/v1/manifest/generate"
```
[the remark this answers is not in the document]
BryanFauble: `url = f"{schematic_api_server_url}/v1/manifest/generate"`

BryanFauble (approved):
1 comment for the URL, otherwise looks great!

tests/integration/test_manifest_generation.py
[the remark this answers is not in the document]
BWMac: Ah thanks

PR #1516, opened 2024-10-04
[FDS-2500] Add Integration Tests for: Manifest Validation

jaymedina:
## problem

The new integration test suite that verifies communications with the `schematic` API endpoints did not include tests for  the `/manifest/validate` endpoint.

## solution

- [x] Introduce tests for the `/manifest/validate` endpoint based on Lingling's test plan [here](https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/3055779846/Schematic+API+test+plan#Manifest-validation) and Tom's prototype work [here](https://github.com/thomasyu888/playground/blob/play-with-locust/locust/test_schematic.py).
- [x] Ensure these tests can run locally, or in the dev/staging/prod environments of schematic. This is done by integrating the `conftest.py` updates in #1512.

## testing


tests/integration/test_manifest_validation.py
```diff
@@ -0,0 +1,486 @@
...
+import pytest
+import requests
+from tests.conftest import Helpers, testing_config, ConfigurationForTesting
+
+
+SYNAPSE_PAT = os.getenv("SYNAPSE_PAT")
+if not SYNAPSE_PAT:
+    raise EnvironmentError("SYNAPSE_PAT environment variable is not set.")
+
+HEADERS = {"Authorization": f"Bearer {SYNAPSE_PAT}"}
```
BryanFauble, line 17: This should come from the `syn_token` test fixture instead:

```

@pytest.fixture(scope="class")
def syn_token(config: Configuration):
    synapse_config_path = config.synapse_configuration_path
    config_parser = configparser.ConfigParser()
    config_parser.read(synapse_config_path)
    # try using synapse access token
    if "SYNAPSE_ACCESS_TOKEN" in os.environ:
        token = os.environ["SYNAPSE_ACCESS_TOKEN"]
    else:
        token = config_parser["authentication"]["authtoken"]
    return token
```

tests/integration/test_manifest_validation.py
```diff
@@ -0,0 +1,486 @@
...
+
+@pytest.fixture(scope="module")
+def setup_api(testing_config: ConfigurationForTesting) -> str:
+    url = testing_config.schematic_api_server_url
+
+    # Make a request to the API and make sure it is responsive
+    response = requests.get(url)
+
+    assert response.status_code == 200, f"Failed to connect to API: {response.text}"
+    return url
```
BryanFauble, line 28: Some complexity here is shown in this commit that I made to other parts of the code:
https://github.com/Sage-Bionetworks/schematic/pull/1512/commits/21d95f5bb60d8ca94e52c0332a36b20679d5c405

Basically in 1 case: When the API server is running on localhost separately from the integration test run, OR, deployed to DEV we can use the requests library.

In the second case: We need to continue using the `flask_client: FlaskClient` pytest fixture. This is for if the integration test is running in the GH pipeline, or we directly run the tests with pytest without starting a local schematic API server.

Unfortunately, the way to invoke `.get()` and retrieves the response differs between the 2 libraries, however, the code I linked shows show I am doing this.


To test out one or the other will depend on what you've set in your `.env` file. The levers that can be flipped are detailed here: https://github.com/Sage-Bionetworks/schematic/blob/21d95f5bb60d8ca94e52c0332a36b20679d5c405/env.example#L15-L30

tests/integration/test_manifest_validation.py
```diff
@@ -0,0 +1,486 @@
...
+            warnings = data.get("warnings", [])
+
+            for idx, expected_idx in zip(warnings, expected_warnings):
+                assert idx == expected_idx
+
+        # TODO: Turn this into a class attribute in case we need the Synapse instance
+        # in other places.
+        import synapseclient
+
+        syn = synapseclient.login()
```
BryanFauble, line 350: Instead of creating this yourself this is available as a fixture from contest.py:

```
@pytest.fixture(scope="class")
def syn(syn_token) -> Synapse:
    syn = Synapse()
    syn.login(authToken=syn_token, silent=True)
    return syn
```

tests/integration/test_manifest_validation.py
```diff
@@ -0,0 +1,486 @@
...
+
+        # WHEN any previous files are deleted
+        files_to_delete = syn.getChildren("syn63582792")
+        for file_to_delete in files_to_delete:
+            syn.delete(file_to_delete["id"])
```
BryanFauble, line 355: Is this a continuation of the test or should these be added to the finalizer that handles cleanup?

`schedule_for_cleanup(CleanupItem(synapse_id=file_to_delete["id"]))`

This is a fixture from contest that can be added as an argument to the test: `schedule_for_cleanup: Callable[[CleanupItem], None]`

tests/integration/test_manifest_validation.py
```diff
@@ -0,0 +1,487 @@
...
+
+            # AND with no expected errors
+            assert len(response_json.get("errors")) == 0
+
+    @pytest.mark.parametrize(
+        ("input_data_type", "input_file_name"),
+        [
+            (
+                "Biospecimen",
+                "mock_manifests/Invalid_Biospecimen_Missing_Column_Manifest.csv",
```
linglp, line 91: @jaymedina if this is not too much work, could you switch to use the invalid manifest that I provided in the test plan. I saw that this manifest: Invalid_Biospecimen_Missing_Column_Manifest.csv is already being used by test: `test_validation.py::TestManifestValidation::test_missing_column`.. and this error message was already tested there. I think if you use the invalid manifest that you provided in the test plan, it could also help us better "map" the manual tests with integration tests. 

Thank you!

BryanFauble: Updated!

tests/data/mock_manifests/MockComponent-cross-manifest-1.csv
```diff
@@ -0,0 +1,15 @@
...
+MockComponent,"a,m,f","ab,cd","ab,ef",ab,"ab,cd","ab,cd,ef","d,e","a,c","d,e",orange,"a,b,c,d",8,8,15.06666667,8,orange,https://www.google.com/,orange,2000,300,300,990109,980109,,6571,109,59,01/31/2005,1
+MockComponent,"a,m,f","ab,cd","ab,ef",ef,"ab,cd","ab,cd,ef","d,e","a,c","d,e",test,"a,b,c,d",9,9,16.71666667,9,test,https://www.google.com/,test,2100,300,300,990110,980110,,6571,110,60,02/01/2005,1
+MockComponent,"a,m,f","ab,cd","ab,ef",cd,"ab,cd","ab,cd,ef","d,e","a,c","d,e",test,"a,b,c,d",0,0,18.36666667,0,test,https://www.google.com/,test,2200,300,300,990111,980111,,6571,111,61,02/02/2005,1
+MockComponent,"a,m,f","ab,cd","ab,ef",ab,"ab,cd","ab,cd,ef","d,e","a,c","d,e",test,"a,b,c,d",1,1,20.01666667,1,test,https://www.google.com/,test,2300,300,300,990120,980120,,6571,120,62,02/03/2005,1
+MockComponent,"a,m,f","ab,cd","ab,ef",ef,"ab,cd","ab,cd,ef","d,e","a,c","d,e",apple,"a,b,c,d",2,2,21.66666667,2,apple,https://www.google.com/,apple,2400,300,300,990130,980130,,6571,130,63,02/04/2005,1
```
BryanFauble, line 15: @linglp When you get a chance please review the difference in this CSV to what you had in the test plan. I updated the values of these 2 columns: `Check Match None,Check Match None values` - The values that were in the test plan did not seem to work.

tests/integration/test_manifest_validation.py
```diff
@@ -0,0 +1,530 @@
...
+
+        # AND an invalid file
+        file = helpers.get_data_path(input_file_name)
+
+        # WHEN we make a POST request to validate the file
+        response = (
+            requests.post(
+                url,
+                params=params,
+                files={"file_name": open(file, "rb")},
```
jaymedina, line 132: Commenting here so we don't run into branch conflicts: These files aren't closed at the end of their test. Could this lead to resource issues if multiple branches are running CI/CD tests with open files that are never closed within their environments? @BryanFauble

BryanFauble: I see the concern, this is how other tests were setup so I used this for consistency.

> Could this lead to resource issues if multiple branches are running CI/CD tests with open files that are never closed within their environments?

No. The file system is not shared between GH runners.

tests/integration/test_manifest_validation.py
```diff
@@ -0,0 +1,530 @@
...
+        response_json = (
+            response.json()
+            if testing_config.use_deployed_schematic_api_server
+            else response.json
+        )
+        assert "warnings" in response_json.keys()
+        assert "errors" in response_json.keys()
+
+        # AND with no expected errors
+        assert len(response_json.get("errors")) == 0
```
GiaJordan, line 91: Could we also assert that there are no warnings?

BryanFauble: Added! https://github.com/Sage-Bionetworks/schematic/pull/1516/commits/0828e340c65709d8bee4eaebe05d1af44de2a72a

tests/integration/test_manifest_validation.py
jaymedina: I see, thanks! NPB :) but I think this would still pose an issue as we continue adding tests to `test_manifest_validation`, since all these tests are executed sequentially in the same GH runner. I can just make a note of it when I'm working on [this ticket](https://sagebionetworks.jira.com/browse/FDS-2510).

jaymedina:
I just went through the last commits and tested the latest branch against a local run of the API and it succeeded. Let me know if you have any further edits for this and I can tackle as they come, otherwise, feel free to merge this into your branch @BryanFauble. Thanks!

tests/integration/test_manifest_validation.py
```diff
@@ -0,0 +1,536 @@
...
+        ("input_data_type", "input_file_name"),
+        [
+            ("Biospecimen", "mock_manifests/example_biospecimen_test.csv"),
+            (
+                "Patient",
+                "mock_manifests/TestManifestValidation_test_manifest_validation_basic_valid.csv",
+            ),
+        ],
+    )
+    def test_manifest_validation_basic_valid(
```
thomasyu888, line 38: @BryanFauble do we think we should add the same documentation that you've added in your test to document the tests here?

thomasyu888 (approved):
🔥 LGTM! Going to pre-approve, but let the team be final reviewer.

tests/integration/test_manifest_validation.py
BryanFauble: I think that's a good idea. @jaymedina Is this something you'd be comfortable adding in?

An example that I did in another test:
```
        """
        Download a manifest from the Schematic API and verify that it is a valid Excel
        file. We are validating the following:
        - The first row of the Excel file contains the column headers
        - The first row is locked on scroll
        - Each cell in the first row has a comment/tooltip "TBD"
        - The "Family History" cell has a comment/tooltip with the expected message
        - The cell corresponding to "Sex" in Sheet1 has a dropdown list with values from Sheet2!B2:B4
        - The cell corresponding to "Diagnosis" in Sheet1 has a dropdown list with values from Sheet2!D2:D3
        - The cell corresponding to "Cancer Type" in Sheet1 has a dropdown list with values from Sheet2!F2:F6
        - The workbook contains two sheets: "Sheet1" and "Sheet2"
        - "Sheet2" is hidden
        - The values in "Sheet2" are as expected
        - The fill colors of the first row cells are as expected
        - Conditional formatting is functioning as expected
        Manual verification steps:
        - Open the Excel file prefixed with TestManifestGeneration_test_single_manifest_generation_excel.xlsx
        - When Diagnosis = "Cancer", [Cancer Type, Family History] is Light Blue (Required)
        """
```

Let's see if we can keep as much context about what each test is doing in the docstring.


@linglp had the great idea to be able to throw this information onto the pdoc site (https://sage-bionetworks.github.io/schematic/) so that we have:

1. A place to organize all of this information in 1 place
2. All changes are back by git, so we get history and tracking

jaymedina: Of course! That's trivial, so no problem. I'll add it in the next commit. BTW are we planning to sunset the readthedocs documentation in place of pdocs?

BryanFauble: @andrewelamb do you know anything about the readthedocs page and the plans for it?

andrewelamb: I don't unfortunately. I'll ask Milen, he shoudl atleast know who to ask.

BryanFauble (approved):
Thanks for the latest round of updates!

tests/integration/test_manifest_validation.py
```diff
@@ -0,0 +1,582 @@
...
+the schematic API test plan found here: <https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/3055779846/Schematic+API+test+plan>.
+"""
+
+import json
+from typing import Dict
+
+import pytest
+import requests
+from flask.testing import FlaskClient
+from synapseclient.client import Synapse
```
BWMac, line 12: NIT: remove unused import

tests/integration/test_manifest_validation.py
```diff
@@ -0,0 +1,582 @@
...
+    @pytest.mark.parametrize(
+        ("input_data_type", "input_file_name"),
+        [
+            (
+                "Patient",
+                "mock_manifests/TestManifestValidation_test_patient_manifest_invalid.csv",
+            ),
+        ],
+    )
+    def test_manifest_validation_basic_invalid(
```
BWMac, line 118: Any particular reason for this to be parameterized with an input of length 1?

BWMac (approved):
Great work! Just one NIT and a question, but pre-approving :fire:

tests/integration/test_manifest_validation.py
BryanFauble: @jaymedina ^

---
I like the idea that we should be able to throw another file into these tests if we wanted to expand on them, but I'm not sure if that was the original intention on this.

andrewelamb (approved):

tests/integration/test_manifest_validation.py
jaymedina: Yes, the original intention was to have a "framework" for future devs to just add into the existing parametrization.

jaymedina:
Thank you guys for taking this to the finish line :) 🚀

PR #1517, opened 2024-10-04
Fds 2333 validate attribute unit tests

andrewelamb:
- Fixes [FDS-2333]
- Adds unit tests for Methods of the ValidateAttribute class
- Cleans up documentation for ValidateAttribute class

[FDS-2333]: https://sagebionetworks.jira.com/browse/FDS-2333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


schematic/models/validate_attribute.py
```diff
@@ -1022,13 +1025,15 @@ def regex_validation(
...
-        errors = []
-        warnings = []
+        errors: list[list[str]] = []
+        warnings: list[list[str]] = []
 
         validation_rules = self.dmge.get_node_validation_rules(
             node_display_name=manifest_col.name
         )
-
+        # It seems like this statement can ever be true
```
BryanFauble, line 1034:
```suggestion
        # It seems like this statement can never be true
```

BryanFauble: Does this need a follow up tech debt ticket? Probably not something we would care to go back and fix if we'll sunset all this code anyways.

schematic/models/validate_attribute.py
```diff
@@ -1147,6 +1152,7 @@ def type_validation(
                     )
                     if vr_errors:
                         errors.append(vr_errors)
+                    # It seems impossible to get warnings with type rules
```
BryanFauble, line 1155: Should this have any follow-up tech debt ticket?

tests/unit/test_validate_attribute.py
```diff
@@ -1574,3 +1558,644 @@ def test__get_column_names(
     ) -> None:
         """Tests for ValidateAttribute._get_column_names"""
         assert va_obj._get_column_names(DataFrame(input_dict)) == expected_dict
+
+    ##############
+    # get_no_entry
+    ##############
```
BryanFauble, line 1564: Instead of grouping by comment like this should this be a new class?

```
class TestGetNoEntry:
```

schematic/models/validate_attribute.py
andrewelamb: Yes I think so. I wanted to mention it here to see if I misunderstood something first.

BryanFauble (approved):
Great additions to the tests! A few comments, but they're all trivial concerns.

tests/unit/test_validate_attribute.py
andrewelamb: Yeah I'm not sure. Do other egineers have preferences for this?

schematic/models/validate_attribute.py
GiaJordan: @andrewelamb I'm not sure where I'm seeing where it's impossible, could you explain in case I missed something? The default message for all the type rules is `error` but for every validation rule a user could specify in their data model to specifically raise `warnings` or `error` regardless of the default level

andrewelamb: @GiaJordan that's doesn't appear to be true. See [here.](https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/2645262364/Data+Model+Validation+Rules#Type-Validation-Type)

"Use the type checker by setting the ‘Validation Rules’  to one of the types listed below. No additional specifiers are allowed. "

If you look at the code in the validator above that is confirmed. Unless the val_rule is one of ["num", "str", "int", "float"] you'll just get empty errors and warnigns back.

GiaJordan: @andrewelamb that does look like the current behavior, I don't think that's intended though, but that would probably be another discussion. Message level specifications are allowed in the rule definitions in `validation_rule_info()`

andrewelamb: I'll start a jira issue then.

andrewelamb: [issue](https://sagebionetworks.jira.com/browse/FDS-2506)

PR #1518, opened 2024-10-04
[FDS-2502] Walk back up directory tree to get location in project

BryanFauble:
**Problem:**

1. When a nested file is used as the datasetId the walk function was only returning back the path where the nested file was located and below. It did not consider where it was in the project if it was in a nested folder.

**Solution:**

1. Walk back up the directory tree until we get back to the project, this was we know the prefix of where we started the walk function and can return back the correct location.

**Testing:**

1. There is at least 2 integration tests covering this.


tests/test_api.py
```diff
@@ -780,9 +780,9 @@ def test_generate_manifest_file_based_annotations(
 
         # make sure Filename, entityId, and component get filled with correct value
         assert google_sheet_df["Filename"].to_list() == [
-            "schematic - main/TestDataset-Annotations-v3/Sample_A.txt",
-            "schematic - main/TestDataset-Annotations-v3/Sample_B.txt",
-            "schematic - main/TestDataset-Annotations-v3/Sample_C.txt",
+            "schematic - main/TestDatasets/TestDataset-Annotations-v3/Sample_A.txt",
```
thomasyu888, line 783: I think this integration test is enough, it probably should've been these values to begin with

BryanFauble: I agree - This along with other tests (I verified we have coverage on the new changes) should be sufficient.

thomasyu888: I guess one thought is should we add a test if the datasetId is the projectid so that it triggers: `if dirpath[0].startswith(f"{project_name}/")`

BryanFauble: It does exist - I verified that when running `tests/test_api.py::TestSynapseStorage::test_get_dataset_files[None-True]` it hits this path

schematic/store/synapse.py
```diff
@@ -585,17 +629,16 @@ def getFilesInStorageDataset(
                     if fullpath:
                         # append directory path to filename
                         if dirpath[0].startswith(f"{project_name}/"):
+                            path_without_project_prefix = (
+                                dirpath[0] + "/"
+                            ).removeprefix(f"{project_name}/")
                             path_filename = (
-                                dirpath[0] + "/" + path_filename[0],
+                                prefix + path_without_project_prefix + path_filename[0],
```
thomasyu888, line 636: Does this need `/`?

thomasyu888 (approved):
🔥 LGTM! Not sure if we just wanted to add the test in that branch as well, but going to pre-approve.

schematic/store/synapse.py
BryanFauble: It does not with how this is written.

thomasyu888:
@BryanFauble I think one of the tests failed.

tests/test_api.py
```diff
@@ -1281,6 +1285,58 @@ def test_submit_manifest_table_and_file_upsert(
         )
         assert response_csv.status_code == 200
 
+    @pytest.mark.synapse_credentials_needed
+    @pytest.mark.submission
+    def test_submit_nested_manifest_table_and_file_upsert(
```
linglp, line 1290: `upsert` in schematic means appending rows in an existing table, which is not what this function is doing. Could you remove the word "upsert" here in the function?

BryanFauble: Good catch! I updated this

tests/test_api.py
```diff
@@ -1281,6 +1285,57 @@ def test_submit_manifest_table_and_file_upsert(
...
+            "manifest_record_type": "table_and_file",
+            "asset_view": "syn63646213",
+            "dataset_id": "syn63646197",
+            "table_manipulation": "replace",
+            "data_model_labels": "class_label",
+            "table_column_names": "display_name",
+        }
+
+        # AND a test manifest with a nested file entity
+        test_upsert_manifest_path = helpers.get_data_path(
```
GiaJordan, line 1311: Along the lines of @linglp's comment, this is still a table replace operation right?

tests/test_api.py
```diff
@@ -1281,6 +1285,57 @@ def test_submit_manifest_table_and_file_upsert(
...
+            "mock_manifests/TestManifestOperation_test_submit_nested_manifest_table_and_file_replace.csv"
+        )
+
+        # AND a randomized annotation we can verify was added
+        df = pd.read_csv(test_upsert_manifest_path)
```
GiaJordan, line 1316: There's a method in the `conftest.py`: `get_data_frame` that we used to load dataframes for the tests to so that the specific util function we use in the main schematic codebase is used in the tests as well, because of the specific way data in the manifests are handled

GiaJordan (changes requested):

tests/test_api.py
```diff
@@ -1281,6 +1285,57 @@ def test_submit_manifest_table_and_file_upsert(
...
+            "http://localhost:3001/v1/model/submit",
+            query_string=params,
+            data={"file_name": (csv_file, "test.csv")},
+            headers=request_headers,
+        )
+
+        # THEN the submission should be successful
+        assert response_csv.status_code == 200
+
+        # AND the file should be uploaded to Synapse with the new annotation
```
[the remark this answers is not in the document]
BryanFauble: > Do we also need to ensure that the manifest file on synapse gets cleaned up and the table gets deleted afterwards?


I do not believe so since as a part of the test the table needs to exist on Synapse with a static ID. As far as the manifest file being deleted, I do not think so either. I was adding the randomized annotation to verify that it went through the annotation update process; In conjunction with verifying the downloadAs attribute of the file should meet exactly what we need.

If the test modifies the name of the file, or does not update the annotation then we know that the logic was broken.

tests/test_api.py
BryanFauble: This was addressed

tests/test_api.py
BryanFauble: I updated to use that func

tests/test_api.py
```diff
@@ -1281,6 +1285,57 @@ def test_submit_manifest_table_and_file_upsert(
         )
         assert response_csv.status_code == 200
 
+    @pytest.mark.synapse_credentials_needed
+    @pytest.mark.submission
+    def test_submit_nested_manifest_table_and_file_replace(
```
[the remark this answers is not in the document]
BryanFauble: Moved over!

tests/test_api.py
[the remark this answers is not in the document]
BryanFauble: > manifest file named synapse_storage_manifest_bulkrna-seqassay.csv gets added to the dataset folder

I added this logic.

> table was created

I am not sure what you mean for me to verify here. The assert_view has to be given in the params of the API, so by virtue of having a Synapse ID I can statically commit to source control the view exists already.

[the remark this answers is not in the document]
BryanFauble: Thanks @linglp I added this in the latest commit!

GiaJordan (approved):
@BryanFauble thank you so much for your help with all of this! Looks good to me!

PR #1519, opened 2024-10-08
Attach additional telemetry data to OTEL traces

BryanFauble:
**Problem:**

1. We did not capture the size of the manifest (Bytes) or the version of schematic that produced the trace.
2. We did not capture the User ID on the trace.
3. We could not export logs out of schematic when running it as a regular server
4. I was encoding the environment into the service name

**Solution:**

1. Attach the package version to the resource which is applied to all of the traces
5. Attach the size of the manifest to the span that covers the download of the manifest
6. Attach the user ID to the trace
7. Move code out of conftest for sending logs next to the trace setup, allowing the configuration to work for both integration tests and the code when deployed as a server.
8. Update to use the `DEPLOYMENT_ENVIRONMENT` attribute to specify where the application is deployed instead of encoding this into the name of the service. This allows for filtering by service and environment separately.

**Testing:**

1. I verified the functionality of the exported trace data:
image

image

image


thomasyu888:
Thanks for the quick PR here.

tests/conftest.py
```diff
@@ -227,27 +222,3 @@ def wrap_with_otel(request):
...
-        )
-
-        logger_provider = LoggerProvider(resource=resource)
-        set_logger_provider(logger_provider=logger_provider)
-
-        # TODO: Add support for secure connections
-        exporter = OTLPLogExporter(insecure=True)
-        logger_provider.add_log_record_processor(BatchLogRecordProcessor(exporter))
-        handler = LoggingHandler(level=logging.NOTSET, logger_provider=logger_provider)
-        logging.getLogger().addHandler(handler)
```
BryanFauble, line 253: This was moved to sit next to the setup for traces, allowing logs and traces to be exported when running integration tests or during normal operations if the env vars are set

thomasyu888 (commented):
Looks good to me so far! Thanks for adding all of this.

pyproject.toml
```diff
@@ -54,7 +54,7 @@ pygsheets = "^2.0.4"
 PyYAML = "^6.0.0"
 rdflib = "^6.0.0"
 setuptools = "^66.0.0"
-synapseclient = "4.5.1"
+synapseclient = "4.6.0"
```
BryanFauble, line 57: This is including a bug fix that we put out a while ago with the 4.5.1 version that had broken downloads from google buckets.

thomasyu888 (approved):
🔥 LGTM!

BryanFauble:
> 🔥 LGTM! May want to add some documentation on how to export signoz traces and deploy signoz locally, but going to approve here.

@thomasyu888 I've added instructions around this in this commit:

https://github.com/Sage-Bionetworks/schematic/pull/1527/commits/4de4b240a4220eb2109274cedfd9265765a8a3f6

My thought is that I do not want folks to have to setup and run SigNoz locally - When I was testing it a while ago it takes a lot of system resources to run all of the components locally. What I wanted to do instead was to let people send their data into the dev instance of SigNoz. That is what the instructions I made above include.

My concerns with this approach:

1. Each developer will be given a unique client ID/client secret that they will use to export telemetry data. **Off-boarding individuals was my concern here** as this system is separate from something like Jumpcloud. It's an additional thing to be aware of as developers come and go.
2. We are given a quota of 1,000 access token that can be requested per month. If we use up all tokens for development, it might mean that schematic running in AWS cannot export it's telemetry data. This was the main reason why I am suggesting in the documentation that developers running locally manually create a token and use it, instead of the oauth2 client credential exchange that the deployed schematic server (Or github actions) would be using.

PR #1522, opened 2024-10-21
[FDS-2466] Update references to schema location in config.yml

jaymedina:
## problem

A user reported lack of clarity in the expected location of the model schema in the Confluence documentation (see [ticket](https://sagebionetworks.jira.com/browse/FDS-2466) for more context)

## solution

The `README` documentation in the schematic repository is now the single source of truth, so I went ahead and moved sections around to improve the flow of the doc. 

Changes:
- [x] **Moved the Set up configuration files section below Google credentials section**. Now, schematic users will set up their Google credentials and download their relevant schema file before setting up the `config.yaml`, at which point the instructions in section 6 are clear for where the schema file should live.
- [x] **Changed `config.yaml` update example** so that users can see how they would modify the path to their model schema in the `config.yaml`.
- [x] **Moved the IMPORTANT banner in section 4. Get your data model as a JSON-LD schema file to section 6** because it flowed better with the new example from above.

## testing
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Here are some example repos with schema files:
 * https://github.com/nf-osi/nf-metadata-dictionary/
 
 > [!IMPORTANT]
-> Your local working directory would typically have `model.jsonld` and `config.yml` side-by-side. The path to your data model should match what is in `config.yml`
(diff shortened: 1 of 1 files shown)


jaymedina:
I've cancelled the **test (3.10)** check since these changes only affect the `README`.

PR #1523, opened 2024-10-21
[FDS-2347] Add unit tests to DataModelValidator class

andrewelamb:
- [FDS-2347] 
- Add unit tests to DataModelValidator class
- Refactors code from class into functions for ease of testing
- Data moved from class to globals
- Most methods changed to private
- Unused methods removed

[FDS-2347]: https://sagebionetworks.jira.com/browse/FDS-2347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


schematic/schemas/data_model_validator.py
```diff
@@ -10,8 +12,22 @@
...
 
-class DataModelValidator:
+
+NodeList = list[Tuple[Any, dict]]
+"""
+A list of node tuples.
+The first item is the name of the node.
+The second item is a dict of its fields.
+This object is gotten from doing nx.MultiDiGraph.nodes(data=True)
+"""
```
BryanFauble, line 27: Rather than `Tuple`, I much prefer the usage of classes to hold data. Maybe something like:

```
from typing import Any, Dict, List, Tuple

@dataclass
class Node:
    name: Any
    """Name of the node."""
    
    fields: Dict
    """Fields of the node"""
```

Then all the return types can be a List[Node]

schematic/schemas/data_model_validator.py
```diff
@@ -143,72 +134,190 @@ def check_blacklisted_characters(self) -> list[str]:
...
-            for node in self.graph.nodes
-            for name in self.reserved_names
-            if name.lower() == node.lower()
+        reserved_names_found = match_node_names_with_reserved_names(
+            self.graph.nodes, RESERVED_NAMES
+        )
+        return create_reserve_name_error_messages(reserved_names_found)
+
+
+def get_node_labels_from(input_dict: dict) -> list:
```
BryanFauble, line 153: Is there a more specific return typing than list? Maybe

```
from typing into List

def get_node_labels_from(input_dict: dict) -> List[str]:
```

andrewelamb: I don't think so, for the same reason the input dict is not more specific. The dict is from networkx.Multidigraph.nodes. And it looks like the values can be anything, and the nested dicts cna be anything as well. In the context of Schematic, the "node_label" field has so far always been a string, but I don't think we should gauruntee it in the return type.

schematic/schemas/data_model_validator.py
```diff
@@ -27,6 +27,30 @@
...
+
+        Raises:
+            ValueError: If the fields don't contain 'displayName'
+
+        Returns:
+            (str): The display name of the node
+        """
+        if "displayName" not in self.fields:
+            raise ValueError(f"Node: {str(self.name)} missing displayName field")
+        return str(self.fields["displayName"])
```
BryanFauble, line 51: An alternative that could possibly be done here is to create an attribute on `Node` for `display_name` that gets set in `__post__init__`:
https://docs.python.org/3/library/dataclasses.html#post-init-processing

The idea is that at the time of creation of the class instance we can apply the validation then and there. Then all downstream logic knows the data has been validated to exist in the way we're expecting it to.

What you have here works, the different being when an exception can be raised, and where it's being raised. IMO If every node is expected to have a `displayName` in `fields`, it might make sense to raise the exception as soon as possible (Ie: When the `Node` instance is being constructed`).

That way the code blows up earlier; you can have error handling further up. The way it is now means that anything calling `get_node_display_name` may need to add it's error handling of how to deal with a ValueError exception.

andrewelamb: If we did that, we'd proably want to follow that concept to its logical conclusion, and do the same for The DatModelValidator class as well.

Currently:

```
class DataModelValidator:  # pylint: disable=too-few-public-methods
    """
    Check for consistency within data model.
    """

    def __init__(
        self,
        graph: nx.MultiDiGraph,
    ):
        """
        Args:
            graph (nx.MultiDiGraph): Graph representation of the data model.
        """
        self.graph = graph
        self.dmr = DataModelRelationships()


    def _get_node_info(self) -> list[Node]:
        """Gets the complete node information form the graph

        Returns:
            list[Node]: A list of nodes from the graph
        """
        return [Node(node[0], node[1]) for node in self.graph.nodes(data=True)]
```

proposed:

```
class DataModelValidator:  # pylint: disable=too-few-public-methods
    """
    Check for consistency within data model.
    """

    def __init__(
        self,
        graph: nx.MultiDiGraph,
    ):
        """
        Args:
            graph (nx.MultiDiGraph): Graph representation of the data model.
        """
        self.graph = graph
        self.node_info = [Node(node[0], node[1]) for node in self.graph.nodes(data=True)]
        self.dmr = DataModelRelationships()

```

BryanFauble: I like the idea!

BryanFauble (approved):
LGTM!

PR #1526, opened 2024-10-25
[FDS-2506] Fix type Validation

andrewelamb:
- Fixes [FDS-2506]
- Fixes type validation so that the msg level parameter now works as intended

[FDS-2506]: https://sagebionetworks.jira.com/browse/FDS-2506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schematic/models/validate_attribute.py
+++ b/schematic/models/validate_attribute.py
@@ -1130,11 +1130,13 @@ class ValidateAttribute(object):
         if len(val_rule_components) > 2:
             raise ValueError("val_rule must contain no more than two components.")
         val_rule_type = val_rule_components[0]
-        if val_rule_type not in ['float', 'int', 'num', 'str']:
-            raise ValueError((
-                f"val_rule first component: {val_rule_type} must be one of "
-                "['float', 'int', 'num', 'str']"
-            ))
+        if val_rule_type not in ["float", "int", "num", "str"]:
+            raise ValueError(
+                (
+                    f"val_rule first component: {val_rule_type} must be one of "
+                    "['float', 'int', 'num', 'str']"
+                )
+            )
 
         specified_type = {
             "num": (int, np.int64, float),
--- a/schematic/models/validate_attribute.py
+++ b/schematic/models/validate_attribute.py
@@ -1105,7 +1105,7 @@ class ValidateAttribute(object):
     ) -> tuple[list[list[str]], list[list[str]]]:
         """
         Purpose:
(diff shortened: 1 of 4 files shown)


schematic/models/validate_attribute.py
```diff
@@ -1121,6 +1121,10 @@ def type_validation(
         TODO:
             Convert all inputs to .lower() just to prevent any entry errors.
         """
+
+        val_rule_components = val_rule.split(" ")
+        val_rule_type = val_rule_components[0]
+
```
BryanFauble, line 1127: Should we add a small pre-condition check here, like?

```
val_rule_components = val_rule.split(" ")
if len(val_rule_components) == 0:
    raise ValueError("val_rule must contain at least one component.")
val_rule_type = val_rule_components[0]
```


I was looking at the google style guide for guidance here: https://google.github.io/styleguide/pyguide.html#244-decision

andrewelamb: Yeah I'm torn here. I really hate how this whole module is designed. In an ideal world there would be one place that does this type of checking since all the rules follow a similar format (just the number of parameter vary), instead of doing this check in each rule.

BryanFauble (approved):
1 small comment, otherwise changes look good to me!

schematic/models/validate_attribute.py
BryanFauble: Agreed. The validation should occur at the moment in time any of this is first read into memory.

andrewelamb: I added this: but I'm not sure how you can get a result of len 0 after splitting
```
       val_rule_components = val_rule.split(" ")
        if len(val_rule_components) == 0:
            raise ValueError("val_rule must contain at least one component.")
        if len(val_rule_components) > 2:
            raise ValueError("val_rule must contain no more than two components.")
        val_rule_type = val_rule_components[0]
        if val_rule_type not in ['float', 'int', 'num', 'str']:
            raise ValueError((
                f"val_rule first component: {val_rule_type} must be one of "
                "['float', 'int', 'num', 'str']"
            ))
```

thomasyu888 (approved):
🔥 LGTM.  At some point, we want to think about auto generating validation rules documentation to capture all of this

PR #1527, opened 2024-10-29
[FDS-2525] Authenticated export of telemetry data

BryanFauble:
**Problem:**

1. When creating telemetry data where an open-telemetry collector is not available to perform an oauth2 token exchange the only way to attach an auth token on outbound requests was to define a static header an environment variables.

**Solution:**

1. Adding the `requests-oauth2client` library that will handle the oauth2 token exchange with Auth0 for us if the environment variables are set.
2. Attach the auth mechanism on the `requests.Session()` object that the opentelemetry exporter uses to send data out.
3. This is the first portion of https://sagebionetworks.jira.com/browse/FDS-2525 - Which allows developers to export telemetry data locally. It also verifies that all we need to do for FDS-2525 is set a few environment variables and allow schematic running within AWS to export it's data to the OTEL collector, which will then forward the data to the DPE K8s cluster.

**Testing:**

1. I ran this code locally with the environment variables set and confirmed that both traces AND log were exported and viewable within SigNoz on the sandbox kubernetes cluster.
2. I also verified I could point this at the `https://dev.sagedpe.org` URL too

**TODO:**

- [x] Remove hacky code around `verify=False` for the HTTP POST logic when non-self signed certs can be used.


schematic/__init__.py
```diff
@@ -26,52 +36,116 @@
...
+            "Using static OTLP headers set in environment variable `OTEL_EXPORTER_OTLP_HEADERS`."
+        )
+        return session
+
+    client_id = os.environ.get("TELEMETRY_EXPORTER_CLIENT_ID", None)
+    client_secret = os.environ.get("TELEMETRY_EXPORTER_CLIENT_SECRET", None)
+    client_token_endpoint = os.environ.get(
+        "TELEMETRY_EXPORTER_CLIENT_TOKEN_ENDPOINT", None
+    )
+    client_audience = os.environ.get("TELEMETRY_EXPORTER_CLIENT_AUDIENCE", None)
```
BryanFauble, line 68: I wrote this before we were 100% certain we were going to have an opentelemetry collector deployed to the schematic VPC where schematic API is running.

During normal operations we probably won't be using this OAuth2Client logic, however, if we ever need to configure the application to work like this it will be available for us to use. It also helped me to prove out the external telemetry system that is receiving these logs.

schematic/__init__.py
```diff
@@ -26,52 +36,116 @@
...
+    if not (tracing_export or logging_export):
+        return None
+
+    session = requests.Session()
+    static_otlp_headers = os.environ.get("OTEL_EXPORTER_OTLP_HEADERS", None)
+    if static_otlp_headers:
+        logger.info(
+            "Using static OTLP headers set in environment variable `OTEL_EXPORTER_OTLP_HEADERS`."
+        )
+        return session
```
BryanFauble, line 61: This section is how developers running schematic locally will be able to export telemetry data with an auth header. The `OTEL_EXPORTER_OTLP_HEADERS` is built into the OTEL library. The library handles adding whatever is set in that environment variable to each outgoing request. Because of this i am just returning an initialized `requests.Session()` here without any additional configuration.

BryanFauble: See README.md for how developers would be using this.

thomasyu888 (approved):
🔥 LGTM!  Thanks for doing this.  Make sense to me, I'll leave it to at least one other developer to review this.

GiaJordan:
@BryanFauble The changes look good to me! 
I just see that a couple tests are failing and there are a couple merge conflicts.

PR #1528, opened 2024-10-30
[FDS-2238, FDS-2236] Automate CLI test plan

andrewelamb:
- Fixes [FDS-2238]
- Fixes [FDS-2236]
- Automates the CLI test plan

[FDS-2238]: https://sagebionetworks.jira.com/browse/FDS-2238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FDS-2236]: https://sagebionetworks.jira.com/browse/FDS-2236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/data/test_configs/CLI_test_config.yml
```diff
@@ -0,0 +1,8 @@
+# This config is used to change the fileview for certain CLI tests
+asset_store:
+  synapse:
+    config: "../../../.synapseConfig"
```
thomasyu888, line 4: Is this synapse config actually used?

thomasyu888: Hmm I'm assuming that there's somewhere in the CI pipeline that fills in the credentials?

tests/integration/test_commands.py
```diff
@@ -1,20 +1,965 @@
...
+        assert result.output.split("\n")[4].startswith("error: 'Random' is not one of")
+        assert result.output.split("\n")[5].startswith("error: 'Random' is not one of")
+        assert result.output.split("\n")[6].startswith("[['2', 'Family History',")
+
+
+class TestManifestCommand:
+    """Tests the schematic/manifest/commands validate manifest command"""
+
+    def test_generate_empty_csv_manifest(self, runner: CliRunner) -> None:
+        """Generate two empty csv manifests"""
```
thomasyu888, line 110: I'll leave it to @BryanFauble for final review, but did we want to follow similar paradigm as the API testing documentation to list exactly what is being tested?

also are there any manual things that need to be checked in the manifests here?

andrewelamb: Yep it was a lot fo C+P and the modifying for the CLI tests hwere needed. If there's anythign I missed that I should include please let me know :)

thomasyu888: C+P? I'll leave it to your responsibility to ensure that you've captured everything that the manual test plan outlined.

As for the documentation, I'd be great if there was heavier documentation in the doc strings like:

```
        Download a manifest from the Schematic API and verify that it is a valid Excel
        file. We are validating the following:
        - The first row of the Excel file contains the column headers
        - The first row is locked on scroll
        - Each cell in the first row has a comment/tooltip "TBD"
        - The "Family History" cell has a comment/tooltip with the expected message
        - The cell corresponding to "Sex" in Sheet1 has a dropdown list with values from Sheet2!B2:B4
        - The cell corresponding to "Diagnosis" in Sheet1 has a dropdown list with values from Sheet2!D2:D3
        - The cell corresponding to "Cancer Type" in Sheet1 has a dropdown list with values from Sheet2!F2:F6
        - The workbook contains two sheets: "Sheet1" and "Sheet2"
        - "Sheet2" is hidden
        - The values in "Sheet2" are as expected
        - The fill colors of the first row cells are as expected
        - Conditional formatting is functioning as expected
       
        Manual verification steps:
        - Open the Excel file prefixed with TestManifestGeneration_test_single_manifest_generation_excel.xlsx
        - When Diagnosis = "Cancer", [Cancer Type, Family History] is Light Blue (Required)
```

If there are manul verification steps required for the manifest itself it should follow the same paradigm that Bryan set up

andrewelamb:
Converting this to draft, going to work on some more.

tests/integration/test_commands.py
```diff
@@ -1,20 +1,965 @@
...
+        # Assert these files were created:
+        assert os.path.isfile("tests/data/example.Biospecimen.manifest.csv")
+        assert os.path.isfile("tests/data/example.Biospecimen.schema.json")
+        assert os.path.isfile("tests/data/example.Patient.manifest.csv")
+        assert os.path.isfile("tests/data/example.Patient.schema.json")
+        # Remove created files:
+        os.remove("tests/data/example.Biospecimen.manifest.csv")
+        os.remove("tests/data/example.Biospecimen.schema.json")
+        os.remove("tests/data/example.Patient.manifest.csv")
+        os.remove("tests/data/example.Patient.schema.json")
```
BryanFauble, line 131: This should exist within a `finally:` block to ensure that an `assert` statement does not prevent the files from being cleaned up

tests/integration/test_commands.py
BryanFauble: +1 My hope is that the docstrings are the single source of great documentation that we can add to the pdocs site.

andrewelamb:
The test for the submit command still needs to be done, after it has been fixed, so this is still a draft for now. I can start gettign feedback on the other parts though.

thomasyu888 (commented):
I took a look at the test failures and there's something that changed that's also causing the synapse storage tests to fail...


```
FAILED tests/test_store.py::TestSynapseStorage::test_init - AssertionError: assert 'syn28559058' == 'syn23643253'
```

Related to the other testing problem:
hmm, would it matter if we ran `tests/integration/test_manifest_generation.py` outside of all the other tests? (it's not ideal, but maybe the ordering of this DOES matter)?

.github/workflows/test.yml
```diff
@@ -136,14 +136,14 @@ jobs:
...
 
-      - name: Run integration tests single process
-        if: ${{ contains(fromJSON('["3.10"]'), matrix.python-version) }}
-        env:
-          SYNAPSE_ACCESS_TOKEN: ${{ secrets.SYNAPSE_ACCESS_TOKEN }}
-          SERVICE_ACCOUNT_CREDS: ${{ secrets.SERVICE_ACCOUNT_CREDS }}
-        run: >
-          poetry run pytest --durations=0 --cov-append --cov-report=term --cov-report=html:htmlcov --cov-report=xml:coverage.xml --cov=schematic/
-          -m "single_process_execution" --reruns 4 --ignore=tests/unit
+      #- name: Run integration tests single process
```
thomasyu888, line 139: Did you mean to leave this commented out?

thomasyu888 (commented):
🔥 Thanks for adding these tests!  I'll leave it to someone on the team for final review

tests/integration/test_commands.py
```diff
@@ -1,20 +1,1289 @@
...
+
+        Patient google sheet:
+        - first row of 'Family History has its own comment
+
+        Manual tests:
+        - Open 'CLI_TestManifestCommand_google_sheet_empty_patient.xlsx'
+        - Select 'Diagnosis' to be 'cancer' in the first row:
+            - 'Cancer Type' and 'Family History' cells in first row should be light blue.
+        - Select 'Diagnosis' to NOT be 'cancer' in the first row:
+            - 'Cancer Type' and 'Family History' cells in first row should be white
```
BryanFauble, line 240: For any tests that need manual verification please use the pytest mark `manual_verification_required`

https://github.com/Sage-Bionetworks/schematic/blob/778bf54db9c5b4de0af334c4efe034b3dde0b348/pytest.ini#L17

tests/integration/test_commands.py
```diff
@@ -1,20 +1,1289 @@
...
+        - first row comments are 'TBD'
+        - first row of 'Family History has its own comment
+
+
+        Manual tests:
+        - Open 'CLI_TestManifestCommand_excel_empty_patient.xlsx'
+        - Select 'Diagnosis' to be 'cancer' in the first row:
+            - 'Cancer Type' and 'Family History' cells in first row should be light blue.
+        - Select 'Diagnosis' to NOT be 'cancer' in the first row:
+            - 'Cancer Type' and 'Family History' cells in first row should be white
```
BryanFauble, line 477: What I was doing in some other tests was to do the selection of data ahead of time so no one has to edit the excel files.

On my machine I don't have an office365 subscription so I cannot actually edit any excel files. But I can still open them to verify the content of them. Please update the data in the files before writing to disk.

BryanFauble: https://github.com/Sage-Bionetworks/schematic/blob/778bf54db9c5b4de0af334c4efe034b3dde0b348/tests/integration/test_manifest_generation.py#L180-L181

tests/integration/test_commands.py
```diff
@@ -1,20 +1,1289 @@
...
+        """
+        try:
+            result = runner.invoke(
+                manifest, ["--config", "config_example.yml", "get", "--sheet_url"]
+            )
+            # Assert these files were created:
+            assert os.path.isfile("tests/data/example.Biospecimen.manifest.csv")
+            assert os.path.isfile("tests/data/example.Patient.manifest.csv")
+            assert os.path.isfile("tests/data/example.Biospecimen.schema.json")
+            assert os.path.isfile("tests/data/example.Patient.schema.json")
```
BryanFauble, line 251: If the names of the files overlap with other tests this is likely to have issues when running the tests concurrently. Where one test is going to create files and remove them in places that might break other tests.

Is it possible to write these to a unique directory that applies to the specific test run, and/or get the cli runner to use a unique file name?

tests/integration/test_commands.py
```diff
@@ -1,20 +1,1289 @@
...
+                ],
+            )
+            # Assert these files were created:
+            assert os.path.isfile("tests/data/example.MockComponent.schema.json")
+            assert os.path.isfile("test-example.xlsx")
+            workbook = load_workbook("test-example.xlsx")
+        finally:
+            # Remove created files:
+            os.remove("tests/data/example.MockComponent.schema.json")
+            os.remove("test-example.xlsx")
```
BryanFauble, line 1161: I see in the TestDownloadManifest (Below) it was using this logic:

```
        try:
            os.remove(expected_manifest_file)
        except Exception:
            pass
```

If we run into issues we might need to do the same here. Or put the logic into a reusable function so we don't have a million try/except blocks.

BryanFauble (changes requested):
Take a look at comments and re-request review with changes

tests/integration/test_commands.py
andrewelamb: So after lookign at the documentation, and some testing, the answer is kind of.

For the csvs: yes with `--output_csv`, but all csv files will be saved with this name, so the second one wil overwrite the first. This is ok for this test, since we are testing the google sheets.

Fopr the jsons: no

tests/integration/test_commands.py
andrewelamb: Will do

tests/integration/test_commands.py
andrewelamb: Will do

tests/integration/test_commands.py
andrewelamb: So for the ones I could set a unique path for the output files I did so. For the others I put the removal of the file in an `if` block in case another test removes them first. I've also created this [issue](https://sagebionetworks.jira.com/browse/SCHEMATIC-209) that would allow for doing thsi in the future.

thomasyu888 (approved):
🔥 Thanks for all the work here!  I'll leave it to Bryan for final review!  I glanced and saw that you made the changes that he requested, but I didn't look super close

BryanFauble (approved):
Thanks Andrew!

PR #1529, opened 2024-10-30
[SCHEMATIC-183] Use paths from file view for manifest generation

GiaJordan:
Resolves [SCHEMATIC-183](https://sagebionetworks.jira.com/browse/SCHEMATIC-183) and switches to using the `path` column from synapse file views for file paths when generating manifests and getting files in a dataset.

Tests are provided in a couple of formats per conversations with @thomasyu888 

[SCHEMATIC-183]: https://sagebionetworks.jira.com/browse/SCHEMATIC-183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

depends on #1554


tests/test_manifest.py
```diff
@@ -763,13 +767,11 @@ def test_create_manifests(
             assert all_results == expected_result
 
     @pytest.mark.parametrize(
-        "component,datasetId,expected_file_based,expected_rows,expected_files",
+        "component, datasetId, expected_rows, expected_files, expected_annotations",
         [
-            ("Biospecimen", "syn61260107", False, 4, None),
```
GiaJordan, line 768: `TestManifestGenerator.test_get_manifest_first_time` already tests record based manifest generation, so it didn't seem necessary to duplicate that here. However, for file based manifests it only tested the nested dataset case without the `contentType:dataset` annotation.
The checks in that test have been replicated here and test cases are included for the top-level folder structure and nested dataset structure. Other dataset structures, ie `contentType:dataset` annotation datasets with nested files, and datasets with doubly nested files, can be added in the future but are tested in `TestStoreSynapse.test_getFilesInStorageDataset`

tests/integration/test_store_synapse.py
```diff
@@ -125,3 +132,337 @@ def test_process_row_annotations_get_validation(
...
+            "doubly nested data files",
+        ],
+    )
+    @pytest.mark.test_this
+    @pytest.mark.parametrize(
+        "filenames",
+        [None, ["txt1.txt", "txt2.txt"]],
+        ids=["no file filtering", "file filtering"],
+    )
+    def test_getFilesInStorageDataset(
```
GiaJordan, line 248: `test_api.TestManifestOperation.test_generate_manifest_file_based_annotations` tests manifest generation through the api. It tests the nested file dataset structure in a way and sends requests to synapse. The `dataset_id` used in the test is not a dataset, but the `schematic - main` project. This contains other datasets and by our definition is not a valid dataset.
So to also have this case covered after the api migration I included this test and also set up valid datasets on synapse with different file structures.

schematic/utils/general.py
```diff
@@ -332,3 +332,14 @@ def normalize_path(path: str, parent_folder: str) -> str:
...
+
+def syn_id_regex() -> str:
+    """
+    Regex pattern to match synapse ids
+    Args:
+        None
+    Returns:
+        str: synID regex pattern
+    """
+    return r"(syn\d+\,?)+"
```
andrewelamb, line 345: Why a function and not a global?

thomasyu888 (commented):
Great work here!  I added some comments, but hope you have a good weekend!

schematic/store/synapse.py
```diff
@@ -659,105 +670,36 @@ def getFilesInStorageDataset(
...
-                download_file=False,
-            )
-            return walk_back_to_project(
-                current_location=current_location,
-                location_prefix=updated_prefix,
-                skip_entry=False,
-            )
+        # The query will also be ammended to include everything containted in all the subdirectories of the dataset
+        for subfolder, _, path in folders:
+            dataset_clause += f"OR parentId='{subfolder[1]}' "
```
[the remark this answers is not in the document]
thomasyu888: I did a test

```
import synapseclient
syn = synapseclient.login()
# AD Knowledge portal file view (254744 files/folders)
parents = syn.tableQuery("SELECT distinct parentId FROM syn11346063")
parents_df = parents.asDataFrame()
len(parents_df)
# 8905 folders

parents_str = "','".join(parents_df['parentId'].tolist())
total_query = syn.tableQuery(
    f"select * from syn11346063 where parentId in ('{parents_str}')"
)
temp = total_query.asDataFrame()
len(temp)
# 253744 files/folders
```

I would hope we don't go beyond 8000 folders?

schematic/utils/general.py
GiaJordan: Yeah I can change it to a global

schematic/store/synapse.py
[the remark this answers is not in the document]
GiaJordan: @thomasyu888 Thanks for the suggestion, I can see how that would be better. Are you also saying to get the folders to query from a query of the view too? Or was that just to facilitate the example here?
`path`, `id`, `name`, `parentId`, `projectId` and `type` I believe, though we should probably document that for users somewhere. These columns are all added by default now; `partentId` has been for a while and we've already been using it elsewhere in schematic for operations so this wouldn't be a new additional requirement

schematic/store/synapse.py
```diff
@@ -659,105 +670,36 @@ def getFilesInStorageDataset(
...
-                                prefix + path_without_project_prefix + path_filename[0],
-                                path_filename[1],
-                            )
-                        else:
-                            path_filename = (
-                                prefix + dirpath[0] + "/" + path_filename[0],
-                                path_filename[1],
-                            )
+        # Exclude manifest files
+        non_manifest_files = self.storageFileviewTable.loc[
```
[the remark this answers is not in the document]
GiaJordan: @thomasyu888 I think we could note that higher up in the docstring instead of under one of the parameters and in the description of the `storage/dataset/files` endpoint.
as an aside, do we know of any users that don't use the default basename and change it in their configs? I guess we could also just check for both

schematic/store/synapse.py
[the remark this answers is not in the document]
thomasyu888: @GiaJordan 
> Are you also saying to get the folders to query from a query of the view too? Or was that just to facilitate the example here?

That was just an example because that won't work without recursive SQL queries - I just wanted to get the MAX parentId's I could from a fileview.

Thanks - this is helpful, I'll add that to the documentation I've been working on.

schematic/store/synapse.py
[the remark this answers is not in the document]
thomasyu888: @GiaJordan good point, for now, I searched "basename" in the data_curator_config: https://github.com/search?q=repo%3ASage-Bionetworks%2Fdata_curator_config%20basename&type=code and didn't get anything returned, but we'll probably want to watch for that eventually to pull from the basename config instead of using "manifest" because people can technically change that.

Can you help create a ticket for that to track?

schematic/store/synapse.py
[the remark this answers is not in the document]
GiaJordan: Okay, understood. And yeah no problem!

schematic/store/synapse.py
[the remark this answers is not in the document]
GiaJordan: Yes I can. It could also be technically changed locally in the `config.yml` but I'm not aware of projects that change that parameter currently

schematic/store/synapse.py
```diff
@@ -680,105 +691,47 @@ def getFilesInStorageDataset(
...
+
+            matching_files = non_manifest_files["path"].str.contains(
+                filename_regex, case=False, regex=True
+            )
+
+            non_manifest_files = non_manifest_files.loc[matching_files, :]
+
+        # Truncate path if necessary
+        if not fullpath:
+            non_manifest_files.path = non_manifest_files.path.apply(os.path.basename)
```
BryanFauble, line 731: Since I know schematic will always write these with a forward slash I also double checked this logic in linux & windows to verify. It passed!

```
import os
import pandas as pd

# Build a df with a `path` column
df = pd.DataFrame(
    {"path": ["/tmp/tmpnin5mnj1/1", "/tmp/tmpnin5mnj1/2", "/tmp/tmpnin5mnj1/3"]})
print(df.path)
df.path = df.path.apply(os.path.basename)
print(df.path)
```
Result:
```
0    /tmp/tmpnin5mnj1/1
1    /tmp/tmpnin5mnj1/2
2    /tmp/tmpnin5mnj1/3
Name: path, dtype: object
0    1
1    2
2    3
Name: path, dtype: object
```

tests/integration/test_store_synapse.py
BryanFauble, line 248: nit: snake_case_for_functions should be preferred over pascalCase

BryanFauble (approved):
No significant changes should be needed here. Just a few things that could be improved.

Thanks for these updates, If you make any new changes feel free to re-request a review and I'll take a look.

schematic/store/synapse.py
GiaJordan: @BryanFauble thanks!

tests/integration/test_commands.py
```diff
@@ -665,18 +665,18 @@ def test_generate_bulk_rna_google_sheet_manifest(
             # Reset config to it's default values
             CONFIG.load_config("config_example.yml")
 
-        assert result.output.split("\n")[7] == (
+        assert result.output.split("\n")[8] == (
```
thomasyu888, line 668: Ah this is to account for the new data model URL column added right?

GiaJordan: No,  the changes in this PR change the output text because it add lines for displaying progress bars for the file view query necessary for getting the files. The test is checking specific lines of the text output for content and now that content is on a different line. This test wasn't present when I started developing this PR but was pulled in with `develop` later in the process.

schematic/store/synapse.py
```diff
@@ -659,105 +670,36 @@ def getFilesInStorageDataset(
...
-        walked_path = synapseutils.walk(
-            self.syn, datasetId, includeTypes=["folder", "file"]
-        )
+        file_list = []
 
-        current_entity_location = self.synapse_entity_tracker.get(
-            synapse_id=datasetId, syn=self.syn, download_file=False
-        )
+        # Identify all folders nested under the dataset folder
+        folders = synapseutils.walk(self.syn, datasetId, includeTypes=["folder"])
```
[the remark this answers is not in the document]
GiaJordan: >  Would building a temporary view based on the dataset folder be helpful here instead of this?

I think it could be used here as well. I think it might be useful to explore building a temporary file view to use for `SynapseStorage` objects in cases where `dataset_scope` is used instead of utilizing the entire view as it exists on synapse. We'd then have all the information we need for the operation locally and wouldn't have to do repeated queries, and in this case we'd be able to avoid walking the directory and querying the view again to get the files by just using the temporary view.

In the past we've only scoped by project; now that `dataset_scope` is in production as an option we might want to make the switch to leverage the ability to set the scope of the temporary view since that isn't an option when we just query the existing complete view and we have to do these workarounds for folders and files. But this all seemed like a separate issue from changing the source of the paths. I'd love to know y'all's thoughts.

> One other thought if you had the fileview, is using the path and do something like this ..
>
> select * from .... where path like 'myproject/myfolder/path/%'
>
> That should get you all the files under a folder

I had considered a solution like that too, I thought gathering the files based on direct parent would help avoid anything weird that could happen with the paths but I think both are viable.

GiaJordan:
@BryanFauble I've made some changes and addressed the comments, if you want to review it again it's ready now. The tests that failed previously are passing locally now so there should only be minimal changes if any going forward

schematic/store/synapse.py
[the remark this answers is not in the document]
thomasyu888: Before merging, we do need to discuss this.  I think there are projects with over 100 nested folders that this current function would break.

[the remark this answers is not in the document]
BryanFauble: > In the past we've only scoped by project; now that dataset_scope is in production as an option we might want to make the switch to leverage the ability to set the scope of the temporary view since that isn't an option when we just query the existing complete view and we have to do these workarounds for folders and files.

From Toms initial investigation the performance is quickly flips over to being more efficient for the temporary view. What I am not a fan of is that this means we need to:

1. Build a view (in some project)
2. Wait for an index of that view to occur to update the synapse project
3. Query it
4. tear down the view



Querying an existing file view that contains the dataset you're looking at seems like the easier thing to do here, and executing the `select * from .... where path like 'myproject/myfolder/path/%'` query. This is closer to how SQL is expected to be used. Us hacking around needing to create temporary views I feel like is a "brittle" solution, with a lot of moving parts where if any individual piece fails, fails the entire transaction.

GiaJordan:
@BryanFauble have you seen an error like the ones from the latest run of the integration tests action?
```
--- Logging error ---
  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/urllib3/connection.py", line 199, in _new_conn
    sock = connection.create_connection(
  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/urllib3/util/connection.py", line 60, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
  File "/opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/socket.py", line 967, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

The above exception was the direct cause of the following exception:

  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/urllib3/connectionpool.py", line 789, in urlopen
    response = self._make_request(
  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/urllib3/connectionpool.py", line 490, in _make_request
    raise new_e
  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/urllib3/connectionpool.py", line 466, in _make_request
    self._validate_conn(conn)
  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/urllib3/connectionpool.py", line 1095, in _validate_conn
    conn.connect()
  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/urllib3/connection.py", line 693, in connect
    self.sock = sock = self._new_conn()
  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/urllib3/connection.py", line 206, in _new_conn
    raise NameResolutionError(self.host, self, e) from e
urllib3.exceptions.NameResolutionError: <urllib3.connection.HTTPSConnection object at 0x7f1ccf6fc8e0>: Failed to resolve 'dev.sagedpe.org' ([Errno -2] Name or service not known)

The above exception was the direct cause of the following exception:

  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/requests/adapters.py", line 667, in send
    resp = conn.urlopen(
  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/urllib3/connectionpool.py", line 843, in urlopen
    retries = retries.increment(
  File "/home/runner/work/schematic/schematic/.venv/lib/python3.10/site-packages/urllib3/util/retry.py", line 519, in increment
    raise MaxRetryError(_pool, url, reason) from reason  # type: ignore[arg-type]
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='dev.sagedpe.org', port=[44](https://github.com/Sage-Bionetworks/schematic/actions/runs/11942387272/job/33289323317?pr=1529#step:11:45)3): Max retries exceeded with url: /telemetry/v1/traces (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7f1ccf6fc8e0>: Failed to resolve 'dev.sagedpe.org' ([Errno -2] Name or service not known)"))

During handling of the above exception, another exception occurred:
```

GiaJordan:
> Yeah I am already working to resolve this.
>
> It's probably not the root cause of the test failure, but it's obfuscating what the real reason of the test failures are. I will re-run the GH action when I get this resolved.

Okay, thanks for the heads up! @BryanFauble

thomasyu888 (changes requested):
```
tests/test_api.py::TestSynapseStorage::test_get_dataset_files
```

This test fails locally.  I took a look and here's the issue:

```
  File "/Users/tyu/anaconda3/envs/schematic/lib/python3.10/site-packages/flask/app.py", line 1517, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/tyu/anaconda3/envs/schematic/lib/python3.10/site-packages/flask/app.py", line 1503, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/Users/tyu/anaconda3/envs/schematic/lib/python3.10/site-packages/connexion/decorators/decorator.py", line 68, in wrapper
    response = function(request)
  File "/Users/tyu/anaconda3/envs/schematic/lib/python3.10/site-packages/connexion/security/security_handler_factory.py", line 388, in wrapper
    return function(request)
  File "/Users/tyu/anaconda3/envs/schematic/lib/python3.10/site-packages/connexion/decorators/uri_parsing.py", line 149, in wrapper
    response = function(request)
  File "/Users/tyu/anaconda3/envs/schematic/lib/python3.10/site-packages/connexion/decorators/validation.py", line 399, in wrapper
    return function(request)
  File "/Users/tyu/anaconda3/envs/schematic/lib/python3.10/site-packages/connexion/decorators/parameter.py", line 120, in wrapper
    return function(**kwargs)
  File "/Users/tyu/sage/schematic/schematic_api/api/routes.py", line 533, in get_files_storage_dataset
    file_lst = store.getFilesInStorageDataset(
  File "/Users/tyu/anaconda3/envs/schematic/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/Users/tyu/sage/schematic/schematic/store/synapse.py", line 710, in getFilesInStorageDataset
    dataset_path = f"'{self.storageFileviewTable.loc[self.storageFileviewTable['id']==datasetId,'path'][0]}/%'"
  File "/Users/tyu/anaconda3/envs/schematic/lib/python3.10/site-packages/pandas/core/series.py", line 1118, in __getitem__
    return self._values[key]
IndexError: index 0 is out of bounds for axis 0 with size 0
```

schematic/store/synapse.py
```diff
@@ -680,105 +704,39 @@ def getFilesInStorageDataset(
...
-        walked_path = synapseutils.walk(
-            self.syn, datasetId, includeTypes=["folder", "file"]
-        )
+        file_list = []
 
-        current_entity_location = self.synapse_entity_tracker.get(
-            synapse_id=datasetId, syn=self.syn, download_file=False
-        )
+        # Get path to dataset folder from fileview to avoid building a new fileview and walking to determine folders and files within
+        dataset_path = f"'{self.storageFileviewTable.loc[self.storageFileviewTable['id']==datasetId,'path'][0]}/%'"
```
thomasyu888, line 710: Ah, I'm glad we had that test, what we didn't think about was the fact that the "scope" id's aren't necessarily included as an "id".  In this case, the `datasetId` in the project is the scope of the fileview, which fails.

schematic/store/synapse.py
```diff
@@ -680,105 +704,44 @@ def getFilesInStorageDataset(
...
-        current_entity_location = self.synapse_entity_tracker.get(
-            synapse_id=datasetId, syn=self.syn, download_file=False
-        )
+        # Get path to dataset folder from fileview to avoid building a new fileview and walking to determine folders and files within
+        child_path = self.storageFileviewTable.loc[
+            self.storageFileviewTable["parentId"] == datasetId, "path"
+        ][0]
+        parent = child_path.split("/")[0]
+        dataset_path = f"{parent}/%"
+        dataset_path = "'" + dataset_path + "'"
```
thomasyu888, line 715:
```suggestion
        dataset_path = f"'{parent}/%'"
```

Sorry I meant things like this is a bit awkward, but I think you can have single quotes like the above
```
f"{json['foo']}/mytext"
```

GiaJordan: Oh okay, I thought you meant to specifically avoid doing that. I understand now thanks

GiaJordan:
Okay with that last comment I think that's everything provided the tests pass. 
@BryanFauble and @thomasyu888 I've re-requested your review

schematic/store/synapse.py
```diff
@@ -693,28 +706,35 @@ def getFilesInStorageDataset(
...
-
-        # The query will also be ammended to include everything containted in all the subdirectories of the dataset
-        for subfolder, _, path in folders:
-            dataset_clause += f"OR parentId='{subfolder[1]}' "
-        dataset_clause = f"({dataset_clause})"
+        # Get path to dataset folder from fileview to avoid building a new fileview and walking to determine folders and files within
+        child_path = self.storageFileviewTable.loc[
+            self.storageFileviewTable["parentId"] == datasetId, "path"
+        ][0]
+        parent = child_path.split("/")[0]
```
BryanFauble, line 713: Should we guard against index errors if the .loc returns an empty list? It would occur if the datasetId is not in the file view table. I'd think we could add a guard condition and raise an exception is not found.

BryanFauble: Will doing a split on "/" and taking the first of the list work for nested datasets? I was thinking of this case where a dataset is at

Root/sub/dir/dataset

I'd think it would return "Root" in this case, which isn't really looking at the path we want to get files at.

Should this instead be to split and use everything but the last of the split list, then join them back with a "/"?

thomasyu888: Ah, good catch @BryanFauble , yes, indeed that is the case.  We want to remove just the last element after the split.

schematic/store/synapse.py
```diff
@@ -706,11 +706,21 @@ def getFilesInStorageDataset(
...
         file_list = []
 
-        # Get path to dataset folder from fileview to avoid building a new fileview and walking to determine folders and files within
+        # Get path to dataset folder by using childern to avoid cases where the dataset is the scope of the view
         child_path = self.storageFileviewTable.loc[
             self.storageFileviewTable["parentId"] == datasetId, "path"
-        ][0]
-        parent = child_path.split("/")[0]
+        ]
+        if child_path.empty:
```
thomasyu888, line 713: You'll want to add a test case for this to ensure this error is raised (no need for integration test, just a mocked test)

thomasyu888:
I took a deeper look at the `test_store.py`, please take a look: https://github.com/Sage-Bionetworks/schematic/pull/1554

thomasyu888:
> @thomasyu888 @GiaJordan 
> if my understanding is correct, we are changing the logic of obtaining file paths by doing the two steps below:  
> 1. Retrieve the folder name associated with the dataset ID.
> 2. Execute a query to find all files matching the dataset name and having a file type of "file".
> I am curious why we have to run query_fileview in `getFilesInStorageDataset` again? Can we "filter" the result from `self.storageFileviewTable` view table since that table already contains all the paths that we need for a given dataset folder? 

@linglp / @GiaJordan / @andrewelamb 

Before reviewing this PR again, please review: https://github.com/Sage-Bionetworks/schematic/pull/1554.

Point number 2 is probably caused by behavior of fileviews.

schematic/store/synapse.py
```diff
@@ -680,105 +704,53 @@ def getFilesInStorageDataset(
...
-        walked_path = synapseutils.walk(
-            self.syn, datasetId, includeTypes=["folder", "file"]
-        )
+        file_list = []
 
-        current_entity_location = self.synapse_entity_tracker.get(
-            synapse_id=datasetId, syn=self.syn, download_file=False
-        )
+        # Get path to dataset folder by using childern to avoid cases where the dataset is the scope of the view
+        child_path = self.storageFileviewTable.loc[
```
linglp, line 710: Should we account for cases where `self.storageFileviewTable` is empty? If self.storageFileviewTable is empty, then it would trigger a key error.

linglp: In addition, I am curious can we assume that all file views have "path" column? I took a look at file view: syn52156826, and noticed that it doesn't have "path" column. If we can't assume that all file views have "path" column, can we add some logic here to better handle this situation?

thomasyu888: Thanks for the review, before going further, please review all comments.

1. I like the idea of handling empty fileview
2. There is error handling in the store/synapse.py for fileviews missing the path column (I think)

tests/integration/test_store_synapse.py
```diff
@@ -125,3 +132,306 @@ def test_process_row_annotations_get_validation(
             "value2",
             "value3",
         ]
+
+    @pytest.mark.parametrize(
```
linglp, line 136: can we also add the situations where no files are expected?

schematic/store/synapse.py
linglp: @thomasyu888 Thanks for the feedback! Are you referring to this line: https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L407? This line is only triggered if self.syn.tableQuery fails. However, even when the "path" column is missing, the query can still run successfully, so the exception here doesn’t catch that error. Apologies, I missed the JIRA comment [here](https://sagebionetworks.jira.com/browse/SCHEMATIC-183?focusedCommentId=224847). If we can ensure that all the file views we're currently using include the "path" column, and we don't need to consider legacy file views, then we shouldn't have to worry too much about the comment that I mentioned above.

schematic/store/synapse.py
```diff
@@ -680,105 +704,58 @@ def getFilesInStorageDataset(
...
-            if (
-                "concreteType" in current_location
-                and current_location["concreteType"] == PROJECT_ENTITY
-            ):
-                return updated_prefix
-            current_location = self.synapse_entity_tracker.get(
-                synapse_id=current_location["parentId"],
-                syn=self.syn,
-                download_file=False,
+        if self.storageFileviewTable.empty:
```
thomasyu888, line 709: @GiaJordan thanks for adding this, but lets take a step back and think about where this should be added?  Should this be added here or within `synapse.py`?

schematic/store/synapse.py
thomasyu888: @linglp right, so you're asking if we can first validate if the fileview even has that column before running the query?  I think that could be a separate ticket if we end up tackling that.

schematic/store/synapse.py
GiaJordan: > Should this be added here or within synapse.py?
I can add it to a different part of the `SynapseStorage` class in `synapse.py` instead? Like after the query is performed, unless you meant a different file?  @thomasyu888

GiaJordan: @thomasyu888 I originally put it there because I didn't want to prohibit any functionality that could be valid with an empty file view. For instance later in this method the case where someone was trying to generate a manifest for a new dataset without files and no manifest already existing. 
The intention for `query_fileview` was that we could use it throughout schematic to query only the parts of the view we need for the operation initially to reduce the amount of subsequent queries we'd need to make (there wasn't time to do this kind of scoping within this ticket) but I thought that as we moved away from just taking everything in the view there would be more cases where an empty view is a valid result and tells us something useful for an operation as well.
I don't recall ever having issues with completely empty views, possibly because how users have to manually set those up.
Just a note that I see these cases as separate from when a view is queried after files have been recently modified but the view hasn't been updated; we're using [tenacity](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L2883) to cover that case.

thomasyu888: Thanks @GiaJordan for the explanation and context.  I think we can keep this here as is.  The chances of having an empty view is probably low in the grand scheme of things anyhow. 

I would add a unit test to make sure this exception is caught, but after reviewal of https://github.com/Sage-Bionetworks/schematic/pull/1554 since there's some tests implemented in this PR that won't pass due to the HACK i implemented there.

linglp: Thanks Tom and Gianna. Adding the code here to deal with empty file view makes sense to me.

tests/integration/test_store_synapse.py
GiaJordan: @linglp that's a good point, thanks!

GiaJordan: @linglp so that's actually covered by the case discussed [here](https://github.com/Sage-Bionetworks/schematic/pull/1529#discussion_r1854383443), when there are no files a `LookupError` is raised
cc @BryanFauble

linglp: Sorry for the late reply. Thanks for pointing out Bryan's comment. I noticed that we have this empty dataset id: syn64319379 in schematic main.Should we use this dataset in the integration test to verify that the error message is returned correctly? 

I tested the API, and it correctly returned the expected error: `LookupError: Dataset syn64319379 could not be found in fileview syn23643253.` However, I think it would be beneficial to include this as a test case in the integration test as well.

thomasyu888: Thanks @linglp, couple of thoughts here and I'll let the two of you decide if we should add it

1. If we add it, it should be a separate integration test to not complicate the test function with an if/else.
2. As mentioned, there is a test case in test_store.py.  It doesn't hurt to add an integration test, but we would be testing Synapse and the Python client's functionality to return a fileview without that dataset id.  Note, the beauty of the test in `test_store.py` is that it returns a representation of the python client/Synapse should return in the hope that the python client is tested.

GiaJordan: > Sorry for the late reply. Thanks for pointing out Bryan's comment. I noticed that we have this empty dataset id: syn64319379 in schematic main.Should we use this dataset in the integration test to verify that the error message is returned correctly?

No worries @linglp! That was a new dataset I created to use for a test in response to your comment.

> If we add it, it should be a separate integration test to not complicate this test function with an if/else.

I agree.

> I'll let @GiaJordan do a final call on whether we should or shouldn't add it and why
> As mentioned, there is a test case in test_store.py. It doesn't hurt to add an integration test, but we would be partially testing Synapse and the Python client's functionality to return a fileview without that dataset id. Note, the beauty of the test in test_store.py is that it returns a representation of the python client/Synapse should return in the hope that the python client is tested. That said, there might be a small risk of overwriting the query string for this error to not be raised, but with all the other existing integration tests, it might be difficult to achieve that.

@thomasyu888 I'd prefer to add an integration test. I know there's been focus recently on the speed of the test suite but I think here there's value in ensuring that given known user actions (ie dataset setup) the interactions with synapse are verified to go as we expect.

thomasyu888: Thanks both!  I'm on board with that ✅

GiaJordan: Sounds good. And just to note, the 2 new tests take about 11 seconds to run sequentially locally.

thomasyu888 (approved):
🔥 The tests look great!  feel free to merge once the CI/CD passes!

GiaJordan:
> 🔥 The tests look great! feel free to merge once the CI/CD passes!

@thomasyu888 sounds good thanks!

PR #1530, opened 2024-10-31
[FDS-2449] Lock `sphinx` version and update `poetry.lock`

jaymedina:
## problem
The ReadTheDocs page for `schematic` was not rendering its contents properly. I was able to re-create the error when starting a new environment from scratch with `sphinx` v7.4.7. I found that there was no version locked for `sphinx` which led to the `poetry.lock` file updating with v7.4.7. This version caused the issue with the documentation build. While the reason isn't clear, I suspect it has something to do with v7.4.7 not working with v4.4.0 of `sphinx-click`.

## solution

- [x] Specify a `sphinx` version in the `pyproject.toml` file, which is used to create the `poetry.lock` file
- [x] Lock the `4.4.0` version of `sphinx-click` in the `pyproject.toml` since we know that works with the new specified version of `sphinx`. This is the package that the `ci_references.rtd` needs in order to render properly.
- [x] Rerun the lock file after

## testing

Creating a fresh environment with the new `poetry.lock` file and building the docs locally now results in a properly rendered CLI Reference page.

Here are the steps to build the docs locally off these changes:

1. **Create a fresh environment**

```
pyenv virtualenv 3.9 rtd-test-env
pyenv activate rtd-test-env
```

2. **Follow the installation steps from the last passing [RTD build](https://readthedocs.org/projects/schematicpy/builds/26119314/)** (do this in the root of the `schematic` repository while checked into this branch)

```
python -m pip install --upgrade --no-cache-dir pip setuptools
pip install poetry==1.3.0
poetry config virtualenvs.create false
poetry install --with doc
```

3. **Build the documentation** (the command below assumes you're in the root of the repository)

```
sphinx-build -b html docs/source _build/html
``` 

Result:

image


pyproject.toml
```diff
@@ -37,7 +37,7 @@ schematic = "schematic.__main__:main"
 
 [tool.poetry.dependencies]
 python = ">=3.9.0,<3.11"
-
+sphinx = "7.3.7"
```
thomasyu888, line 40: Question for all - should these be part of the dev dependencies?

thomasyu888:
This should fix this trivy issue: https://github.com/Sage-Bionetworks/schematic/pull/1531

pyproject.toml
jaymedina: cc @GiaJordan @andrewelamb

BryanFauble: Yes that makes sense since one installing/using this as a CLI would not need sphinx

BryanFauble (approved):
Thanks for the update!

pyproject.toml
jaymedina: The original `README` instructions assumed only developers/contributors would be using `poetry`, so I think that's why the sphinx libraries were added to the main dependencies. I'll create a separate group for documentation-related librarie, move everything there, and update the `README` accordingly.

thomasyu888 (changes requested):
You will have to change the installation instructions in `.readthedocs.yml` for this to work

README.md
```diff
@@ -270,12 +270,17 @@ poetry debug info
 
 Before you begin, make sure you are in the latest `develop` of the repository.
 
-The following command will install the dependencies based on what we specify in the `poetry.lock` file of this repository. If this step is taking a long time, try to go back to Step 2 and check your version of `poetry`. Alternatively, you can try deleting the lock file and regenerate it by doing `poetry install` (Please note this method should be used as a last resort because this would force other developers to change their development environment)
+The following command will install the dependencies based on what we specify in the `poetry.lock` file of this repository (which is generated from the libraries listed in the `pyproject.toml` file). If this step is taking a long time, try to go back to Step 2 and check your version of `poetry`. Alternatively, you can try deleting the lock file and regenerate it by doing `poetry lock` (Please note this method should be used as a last resort because this would force other developers to change their development environment).
 
 ```
-poetry install --all-extras
+poetry install --dev,doc
```
thomasyu888, line 276: poetry install --with dev,doc

jaymedina:
> Do you have to change the installation instructions in .readthedocs.yml for this to work?

No, I changed it so I'd have the opportunity to explain the groups being downloaded with the command. I thought this would be better than doing an `--all-extras`, which makes the process feel more opaque (you don't know what these extras are, or what they're for).

thomasyu888:
> No, I changed it so I'd have the opportunity to explain the groups being downloaded with the command. I thought this would be better than doing an --all-extras, which makes the process feel more opaque (you don't know what these extras are, or what they're for).

@jaymedina Makes sense, lets confirm that this builds: https://schematicpy.readthedocs.io/en/fds-2449-fix-rtd/cli_reference.html.  You can log into readthedocs to view the schematicpy builds and logs.  We may want to add dependencies then to the `doc.dependencies` section if there are still ones missing.

thomasyu888 (approved):
🔥 LGTM!

thomasyu888:
@jaymedina Can you make sure the tests run locally?  My concern is that there are dependency version updates, so the 3.10 tests not passing is unfortunate..

PR #1536, opened 2024-11-09
[SCHEMATIC-30, SCHEMATIC-200] Add version to click cli / use pathlib.Path module for checking cache size

thomasyu888:
## Problem
* It was not easy to surface the version of the cli or library used
* we were using `du -h` system to check for cache size which isn't accurate AND may fail on windows

## solution
Add version to cli

```
schematic -h

Options:
  --version            Show the version and exit.

Commands:

schematic --version
schematic, version 24.12.1
```

```
import schematic
schematic.__version__
Out[5]: '24.12.1'
```

This confirms that it will pull from the pyproject.toml, so that needs to be updated

```
docker build . -t test
schematic, version 24.12.1
```


schematic/schemas/data_model_parser.py
```diff
@@ -101,7 +101,6 @@ def parse_model(self) -> dict[str, dict[str, Any]]:
           off base model.
         """
         # base_model = self.parse_base_model()
-
         # Call appropriate data model parser and return parsed model.
```
thomasyu888, line 104:
```suggestion

        # Call appropriate data model parser and return parsed model.
```

schematic/utils/general.py
```diff
@@ -129,40 +129,19 @@ def calculate_datetime(
...
-        float: returns size of .synapsecache directory in bytes
+        float: size of .synapsecache directory in bytes
     """
-    # Note: this command might fail on windows user.
-    # But since this command is primarily for running on AWS, it is fine.
-    command = ["du", "-sh", directory]
-    output = subprocess.run(command, capture_output=True, check=False).stdout.decode(
-        "utf-8"
+    total_size = sum(
+        f.stat().st_size for f in Path(directory).rglob("*") if f.is_file()
```
thomasyu888, line 142: There may be some minor discrepancies between how this would operate on windows, mac, and ubuntu

tests/unit/test_utils_general.py
```diff
@@ -0,0 +1,66 @@
...
+@pytest.mark.parametrize(
+    "directory, file_sizes, expected_size",
+    [
+        ("", [1024], 1024),  # Default directory with 1 KB file
+        ("/custom/directory", [2048], 2048),  # Custom directory with 2 KB file
+        ("", [], 0),  # Empty directory
+        ("", [1024, 2048], 3072),  # Directory with multiple files (1 KB + 2 KB)
+    ],
+)
+def test_check_synapse_cache_size(mocker, directory, file_sizes, expected_size):
```
thomasyu888, line 21: There are separate tests in the `test_utils.py`, should I move them here?

schematic/__main__.py
```diff
@@ -24,6 +25,7 @@
 # invoke_without_command=True -> forces the application not to show aids before losing them with a --h
 @click.group(context_settings=CONTEXT_SETTINGS, invoke_without_command=True)
 @click_log.simple_verbosity_option(logger)
+@click.version_option(version=__version__, prog_name="schematic")
```
BryanFauble, line 28: Double check that this works in a docker container. I ran into issues using `__version__` where on my local machine it worked fine, but in docker it would crash because the package was not found.

schematic/utils/general.py
BryanFauble: This doesn't work on ubunutu:

```
  File "/home/bfauble/BryansGreatWorkspace/schematic/junk.py", line 7, in <module>
    print(check_synapse_cache_size())
  File "/home/bfauble/BryansGreatWorkspace/schematic/schematic/utils/general.py", line 141, in check_synapse_cache_size
    total_size = sum(
  File "/home/bfauble/BryansGreatWorkspace/schematic/schematic/utils/general.py", line 141, in <genexpr>
    total_size = sum(
  File "/home/bfauble/.pyenv/versions/3.10.14/lib/python3.10/pathlib.py", line 1047, in rglob
    for p in selector.select_from(self):
  File "/home/bfauble/.pyenv/versions/3.10.14/lib/python3.10/pathlib.py", line 405, in select_from
    if not is_dir(parent_path):
  File "/home/bfauble/.pyenv/versions/3.10.14/lib/python3.10/pathlib.py", line 1305, in is_dir
    return S_ISDIR(self.stat().st_mode)
  File "/home/bfauble/.pyenv/versions/3.10.14/lib/python3.10/pathlib.py", line 1097, in stat
    return self._accessor.stat(self, follow_symlinks=follow_symlinks)
PermissionError: [Errno 13] Permission denied: '/root/.synapseCache'
```

```
from schematic.utils.general import check_synapse_cache_size

print(check_synapse_cache_size())
```

BryanFauble: I guess before it probably wouldn't have either since the default is to use `/root`. It does work though with how it's implemented in code:

```
from schematic.utils.general import check_synapse_cache_size

synapse_store = SynapseStorage()

print(check_synapse_cache_size(synapse_store.root_synapse_cache))
```

tests/unit/test_utils_general.py
BryanFauble: Yes please, those under `class TestGeneral:` look like good candidates to move over here

schematic/__main__.py
BryanFauble: For this I had just used `docker build .` and did a `docker run` on the built image. If it starts up we would just want to confirm that we could print the version to console.

I had tried the same things you did here:

https://github.com/Sage-Bionetworks/schematic/pull/1527/commits/a096639da83a4024d8f3c3874e5d9f6a1c91f279

https://github.com/Sage-Bionetworks/schematic/pull/1527/commits/2e3b3cf594d3d00d858ab4a95603935737215de2

PR #1537, opened 2024-11-11
Update for integration test failures

BryanFauble:
**Problem:**

1. Integration tests are failing due to modified state in-between tests.

**Solution:**

1. Try out updating the scope of the synapse storage object to be function scoped.
2. @andrewelamb had some great changes to reset config files back to initial values for tests that modify the config files.

**Testing:**

1. Will see results in integration test runs


thomasyu888 (approved):
🔥 LGTM!

PR #1538, opened 2024-11-11
[SCHEMATIC-1] BugFix: manifest submission when data previously annotated without manifest upload

GiaJordan:
## Issue
Resolves [SCHEMATIC-1](https://sagebionetworks.jira.com/browse/SCHEMATIC-1) which occurs when users:
- Have a dataset/top level folder on synapse with 50 or more files
- generate a manifest with schematic/DCA and fill it out
  - possibly validate the manifest with schematic
- Use the manifest to annotate the data with a client other than schematic ie. R
- Try to generate a new manifest with schematic using the existing annotations from the data on synapse

## Changes

- when a temporary file view is created with `DatasetFileView` and the resulting view is "tidied", any existing `eTag`  columns will be removed before the new values are stored in that column.
- Test resources on synapse have been created to replicate the issue and an integration test has been added. It was not necessary to include 50+ files as we're testing `DatasetFileView` directly instead of through manifest generation

## Notes

- Technically using `--use_annotations` to generate a manifest for a dataset where there isn't already a manifest present is not supported, but this provides a way for users to get back on the recommended data flow path instead of having to start all over. 
- `DatasetFileView` is only used in this one case, gathering annotations for files in a dataset with more than 50 files. With the ability we have now to scope fileview queries in the `SynapseStorage` object and get the annotations for all files in a dataset we should consider whether we should deprecate the `DatasetFileView` class.
- We should also consider whether there's anything else preventing us from officially supporting using the `--use_annotations` feature when there is no manifest in the dataset since this change enables it in at least this case.

[SCHEMATIC-1]: https://sagebionetworks.jira.com/browse/SCHEMATIC-1?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -2825,7 +2825,7 @@ class SynapseStorage(BaseStorage):
             try:
                 logger.info("Trying batch mode for retrieving Synapse annotations")
                 table = self.getDatasetAnnotationsBatch(datasetId, dataset_file_ids)
-            except (SynapseAuthenticationError, SynapseHTTPError):
+            except (SynapseAuthenticationError, SynapseHTTPError, ValueError):
                 logger.info(
                     f"Unable to create a temporary file view bound to {datasetId}. "
                     "Defaulting to slower iterative retrieval of annotations."
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -2825,7 +2825,7 @@ class SynapseStorage(BaseStorage):
             try:
                 logger.info("Trying batch mode for retrieving Synapse annotations")
                 table = self.getDatasetAnnotationsBatch(datasetId, dataset_file_ids)
-            except (SynapseAuthenticationError, SynapseHTTPError):
+            except:
                 logger.info(
                     f"Unable to create a temporary file view bound to {datasetId}. "
                     "Defaulting to slower iterative retrieval of annotations."
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -3508,6 +3508,12 @@ class DatasetFileView:
         # Rename ROW_ETAG column to eTag and place at end of data frame
         if "ROW_ETAG" in self.table:
             row_etags = self.table.pop("ROW_ETAG")
+
+            # eTag column may already present if users annotated data without submitting manifest
+            # we're only concerned with the new values and not the existing ones
+            if "eTag" in self.table:
+                del self.table["eTag"]
+
             self.table.insert(len(self.table.columns), "eTag", row_etags)
 
         return self.table
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -2825,7 +2825,7 @@ class SynapseStorage(BaseStorage):
             try:
                 logger.info("Trying batch mode for retrieving Synapse annotations")
                 table = self.getDatasetAnnotationsBatch(datasetId, dataset_file_ids)
-            except:
+            except (SynapseAuthenticationError, SynapseHTTPError):
                 logger.info(
                     f"Unable to create a temporary file view bound to {datasetId}. "
                     "Defaulting to slower iterative retrieval of annotations."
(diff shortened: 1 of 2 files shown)


schematic/store/synapse.py
```diff
@@ -2825,7 +2825,7 @@ def getDatasetAnnotations(
             try:
                 logger.info("Trying batch mode for retrieving Synapse annotations")
                 table = self.getDatasetAnnotationsBatch(datasetId, dataset_file_ids)
-            except (SynapseAuthenticationError, SynapseHTTPError):
+            except:
```
thomasyu888, line 2828: This is not great practice - do you know what other exceptions you ran into here?

thomasyu888 (commented):
Great work @GiaJordan , I'll defer to @BryanFauble and/or @andrewelamb for final review

schematic/store/synapse.py
GiaJordan: The one in this case was a `ValueError`

thomasyu888: @GiaJordan and everyone else, we can add this multiblock exception: https://rollbar.com/blog/python-catching-multiple-exceptions/.

if it makes sense to add `ValueError` we should

thomasyu888: I should also add that thanks for being proactive about fixing that.  I agree that we can fix things on the fly so every commit to the codebase is better!

GiaJordan: So `valueError` should no longer be raised, that's what this fix is for and the bug the user experienced. However, the previous behavior that the user experienced was that the program would stop since `ValueError` was not one of `SynapseAuthenticationError` or `SynapseHTTPError`. 
Since none of the other methods used later in the `try` block raise exceptions themselves, my thinking was that if there were any issue with using batch mode, then sequential mode should be used and if another exception was raised during that process it would be surfaced to the user like normal, but that it would allow the user to actually generate a manifest in cases where there isn't an issue with the sequential method instead of repeatedly getting errors and being unable to proceed. At this point I don't believe we need to also catch `ValueError` given this fix, I was just worried that the two listed were too restricted given the behavior the user experienced in this case.
@thomasyu888

thomasyu888: Thanks @GiaJordan , I'm not opposed to add more exceptions, but accepting all exceptions is a bit scary (imo)

so something like
```
except (SynapseAuthenticationError, SynapseHTTPError, ValueError)
```

andrewelamb (approved):

schematic/store/synapse.py
GiaJordan: > accepting all exceptions is a bit scary (imo)

Yeah I can see that and I agree. I was weighing the behavior of not blocking generation vs being specific because at this point I'm not sure which other exceptions could be raised depending on what users do (like how in this case this user flow wasn't expected.

In the end though we can add `ValueError` to be safe, though I don't think it's terribly necessary with this change, and we can add to the list of exceptions (and possibly add other blocks) as necessary in the future.

Thanks for this discussion! @thomasyu888

schematic/store/synapse.py
```diff
@@ -2825,7 +2825,7 @@ def getDatasetAnnotations(
             try:
                 logger.info("Trying batch mode for retrieving Synapse annotations")
                 table = self.getDatasetAnnotationsBatch(datasetId, dataset_file_ids)
-            except (SynapseAuthenticationError, SynapseHTTPError):
+            except (SynapseAuthenticationError, SynapseHTTPError, ValueError):
```
thomasyu888, line 2828: Thanks for adding this.  Not to add to your work, but we may want a unit test to mock that this logic is working as expected to catch the different exceptions.

Let me know if you want to meet to discuss how to add a test specifically for `getDatasetAnnotations` function

It's a bit more work, but if you mock the particular calls, then we aren't actually using any test projects.

BryanFauble (approved):
LGTM!

PR #1540, opened 2024-11-13
[SCHEMATIC-157] Make some dependencies required to avoid `schematic CLI` commands from potentially erroring when doing a pip install

thomasyu888:
## Problem
Tagging as SCHEMATIC-157 as it is related to the `--all-extras` we had to tack on to make the CLI documentation to show up.

When doing 
```
docker build -t test .
  File "/usr/local/bin/schematic", line 3, in <module>
    from schematic.__main__ import main
  File "/usr/src/app/schematic/__init__.py", line 10, in <module>
    from opentelemetry.instrumentation.flask import FlaskInstrumentor
ModuleNotFoundError: No module named 'opentelemetry.instrumentation.flask'
```

## Solution

Add in dependencies as non-optional (which does blow up the dependencies that people have to install when using the CLI...)

In particular,  made "connexion", "Flask", "Flask-Cors", "opentelemetry-instrumentation-flask", "requests-oauth2client" non-extras

```
docker build -t test .
docker run test schematic --help                 

Options:

Commands:
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -275,7 +275,7 @@ lxml = ["lxml"]
 name = "binapy"
 version = "0.8.0"
 description = "Binary Data manipulation, for humans."
-optional = true
+optional = false
 python-versions = ">=3.8"
 files = [
     {file = "binapy-0.8.0-py3-none-any.whl", hash = "sha256:8af1e1e856900ef8b79ef32236e296127c9cf26414ec355982ff7ce5f173504d"},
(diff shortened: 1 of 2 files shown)


BryanFauble (approved):

andrewelamb:
@thomasyu888 Why are we doing this? This is going to force all users to install those packages, including CLI and library only users.

thomasyu888:
> @thomasyu888 Why are we doing this? This is going to force all users to install those packages, including CLI and library only users.

@andrewelamb that's a good question - unfortunately, this is due to the fact that we have open telemetry on the library code to enable tracing and observability - this then requires us to have these until we can have clearer separation between the CLI and library.

PR #1541, opened 2024-11-14
Add precommit and run on all files

thomasyu888:
Adds precommit so that development goes easier


.github/workflows/test.yml
GiaJordan, line 1: Should we also enforce `black` for all files now with this change?

.pre-commit-config.yaml
```diff
@@ -1,21 +1,68 @@
...
+      - id: check-xml
+      - id: check-yaml
+        exclude: ^mkdocs\.yml$
+      - id: debug-statements
+      - id: end-of-file-fixer
+      - id: requirements-txt-fixer
+      - id: mixed-line-ending
+        args: ["--fix=auto"] # replace 'auto' with 'lf' to enforce Linux/Mac line endings or 'crlf' for Windows
+
+  # - repo: https://github.com/astral-sh/ruff-pre-commit
```
GiaJordan, line 21: Are the comments hooks we plan to turn on in the future? are we no longer using `isort` as well?

.github/workflows/test.yml
thomasyu888: The `.precommit.config..` has this (buried under the sea of file changes) ha.

```
  - repo: https://github.com/psf/black
    rev: 23.12.1
    hooks:
      - id: black
        language_version: python3
```

So it does enforce black with this change.

.pre-commit-config.yaml
thomasyu888: ^ so isort seems to cause a weird error with the code, so I turned it off for now. Will turn it back on later.

GiaJordan: that makes sense, sounds good

.github/workflows/test.yml
GiaJordan: Okay cool, so this will now run the precommit hook in the action itself and modify the files instead of just checking that the files are already formatted correctly?

thomasyu888: So once this is merged in there will be a new development workflow.  Everyone will run precommit install locally and whenever you commit, you'll something like this:

```
git commit -am "Use pre-commit"
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
trim trailing whitespace.................................................Passed
check for added large files..............................................Passed
check python ast.....................................(no files to check)Skipped
check for merge conflicts................................................Passed
check xml............................................(no files to check)Skipped
check yaml...............................................................Passed
debug statements (python)............................(no files to check)Skipped
fix end of files.........................................................Passed
fix requirements.txt.................................(no files to check)Skipped
mixed line ending........................................................Passed
black................................................(no files to check)Skipped
[add-precommit d1741be0] Use pre-commit
 1 file changed, 1 insertion(+), 5 deletions(-)
```

It will automatically run these for you during commit.  This pre-commit hook is just an added layer and will fail if pre-commit isn't installed locally.

BryanFauble (approved):
Hopefully isort is an easy fix, I like not needing to think about adding imports in certain locations - No biggie though.


Thanks for running this on all files, current PR's are going to run into merge conflicts because of the sheer number of changes, but, it's a bandage we need to rip off at some point.

thomasyu888:
> Hopefully isort is an easy fix, I like not needing to think about adding imports in certain locations - No biggie though.
> 
> Thanks for running this on all files, current PR's are going to run into merge conflicts because of the sheer number of changes, but, it's a bandage we need to rip off at some point.

I'll wait for all your PRs to go in, and I'm happy to fix the merge conflicts since it'll be easier to fix on my end

.pre-commit-config.yaml
thomasyu888: Ya, we can certainly look into it!  My concern is always that `ruff` may clash against `black` or other formatting tools, but thats a concern i have about every formatting tool. 😓

GiaJordan (approved):

BWMac:
#1560 is merged, good to go when ready.

BryanFauble (approved):

PR #1542, opened 2024-11-14
[SCHEMATIC-30, SCHEMATIC-200] Add version to click cli / use pathlib.Path module for checking cache size

thomasyu888:
## Problem
* It was not easy to surface the version of the cli or library used
* we were using `du -h` system to check for cache size which isn't accurate AND may fail on windows

## solution
Add version to cli

```
schematic -h

Options:
  --version            Show the version and exit.

Commands:

schematic --version
schematic, version 24.12.1
```

```
import schematic
schematic.__version__
Out[5]: '24.12.1'
```

This confirms that it will pull from the pyproject.toml, so that needs to be updated

```
docker build . -t test
schematic, version 24.12.1
```


.github/workflows/scan_repo.yml
```diff
@@ -12,6 +12,9 @@ jobs:
   trivy:
     name: Trivy
     runs-on: ubuntu-latest
+    env:
```
thomasyu888, line 15: Fixes trivy issue

tests/unit/test_utils_general.py
```diff
@@ -0,0 +1,66 @@
...
+from schematic.utils.general import (
+    check_synapse_cache_size,
+)
+
+import pytest
+from unittest.mock import MagicMock
+from unittest import mock
+
+
+@pytest.mark.parametrize(
```
thomasyu888, line 12: Need to move tests from test_utils.py

schematic/utils/general.py
```diff
@@ -129,40 +129,19 @@ def calculate_datetime(
     return date_time_result
 
 
-def check_synapse_cache_size(
-    directory: str = "/root/.synapseCache",
-) -> float:
-    """use du --sh command to calculate size of .synapseCache.
+def check_synapse_cache_size(directory: str = "/root/.synapseCache") -> float:
```
thomasyu888, line 132: Need to potentially look into the the default values

tests/unit/test_utils_general.py
thomasyu888: Done here: https://github.com/Sage-Bionetworks/schematic/pull/1542/commits/ffa9498fb55b602182beb28e60210734a24a8bbd

schematic/utils/general.py
thomasyu888: I'm not actually sure... if we should refactor this in this PR, but if we were, I probably would set the default value to `None` and then see if `/root/.synapseCache` exists.  If not... then default it to the system default's synapse cache, but I'm not sure if this will be an issue.

BryanFauble (approved):
Lgtm! 


Assuming the version works fine id like to look into attaching it to the telemetry data created. That way we can know the version that produced the telemetry data.

schematic/utils/general.py
GiaJordan: I've seen two versions of the default: `~/.synapseCache` is the default in the `.synapseConfig` and then in `SynapseStorage.__init__` it says it defaults to `None`

schematic/version.py
```diff
@@ -1,4 +1,4 @@
 """Sets the version of the package"""
 import importlib.metadata
 
-__version__ = importlib.metadata.version("schematic")
+__version__ = importlib.metadata.version("schematicpy")
```
GiaJordan, line 4: Going forward will we still want to manually update the version in the `pyproject.toml`?

PR #1543, opened 2024-11-14
V24.11.1

andrewelamb:


thomasyu888:
1. Does the pre release tag trigger a test release to pypi? (It's ok if it doesn't, just trying to fill in gaps of knowledge)
2. Does the pre release tag trigger a docker image creation? - if so: we could update staging API with this docker tag.

If both of the answers are "no", we may want to consider adding those practices in the near future, but it's ok for now to merge into main.

andrewelamb:
Publishing to pypi is triuggered when a tag is pushed to main via this [workflow](https://github.com/Sage-Bionetworks/schematic/blob/develop/.github/workflows/publish.yml)

- No test release
- Currently there is no difference between release and pre-release in github workflows (AFAIK)

GiaJordan (approved):

PR #1550, opened 2024-11-18
[SCHEMATIC-193] Support exporting telemetry data from GH integration test runs

BryanFauble:
**Problem:**

1. In order for telemetry data to be exported form schematic during the integration test runs we need to be able to grab an oauth2 token from Auth0. We need to keep the number of token requests low to prevent hitting a monthly cap, to accomplish this we are grabbing a token once per run of the pipeline.

**Solution:**

1. Grab a token per run of the pipeline and pass from one step to the next step of the GH run.
2. Using the add-mask feature described in this example: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#example-masking-a-generated-output-within-a-single-job

**Testing:**

Traces and logs are showing up in SigNoz as expected:
image

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -127,11 +127,30 @@ jobs:
       #----------------------------------------------
       #              run integration test suite
       #----------------------------------------------
+
+      - name: Retrieve telemetry access token from IDP
+        if: ${{ contains(fromJSON('["3.10"]'), matrix.python-version) }}
+        id: retrieve-telemetry-access-token
+        run: |
+          response=$(curl --request POST \
+            --url ${{ vars.TELEMETRY_AUTH_CLIENT_URL }} \
+            --header 'content-type: application/json' \
+            --data '{"client_id":"${{ vars.TELEMETRY_AUTH_CLIENT_ID }}","client_secret":"${{ secrets.TELEMETRY_AUTH_CLIENT_SECRET }}","audience":"${{ vars.TELEMETRY_AUTH_AUDIENCE }}","grant_type":"client_credentials"}')
+          access_token=$(echo $response | jq -r .access_token)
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM!

.github/workflows/test.yml
```diff
@@ -127,11 +127,30 @@ jobs:
...
       #----------------------------------------------
+
+      - name: Retrieve telemetry access token from IDP
+        if: ${{ contains(fromJSON('["3.10"]'), matrix.python-version) }}
+        id: retrieve-telemetry-access-token
+        run: |
+          response=$(curl --request POST \
+            --url ${{ vars.TELEMETRY_AUTH_CLIENT_URL }} \
+            --header 'content-type: application/json' \
+            --data '{"client_id":"${{ vars.TELEMETRY_AUTH_CLIENT_ID }}","client_secret":"${{ secrets.TELEMETRY_AUTH_CLIENT_SECRET }}","audience":"${{ vars.TELEMETRY_AUTH_AUDIENCE }}","grant_type":"client_credentials"}')
```
GiaJordan, line 138: What does the `audience` parameter control here?

GiaJordan (approved):
Also LGTM, thanks!

.github/workflows/test.yml
BryanFauble: > The audience (presented as the aud claim in the access token) defines the intended consumer of the token. This is typically the resource server (API, in the dashboard) that a client (Application) would like to access.

https://community.auth0.com/t/what-is-the-audience/71414

GiaJordan: Ahh I see, thanks!

PR #1551, opened 2024-11-21
[SCHEMATIC-163] Catch error when manifest is generated and existing one doesn't have `entityId`

BWMac:
**Description:**

When a user generates a manifest and existing manifests are used to populate information in the new one, the manifests must have the `entityId` column. Recently a user encountered a nondescript `KeyError` due to a missing `entityId` column in a manifest. This was most likely due to their manually creating a manifest and uploading it manually or by generating the manifest with `schematic`, removing the `entityId` column and uploading it without `schematic`. Going forward, we should include a more detailed error message in these cases so that users can be helped back into the correct workflow of using `schematic`.

This PR adds error handling in `SynapseStorage._get_file_entityIds` for these cases and a detailed error message. I also added unit tests for `_get_file_entityIds` because there weren't any previously. 

**Notes:**
- I reproduced the error described in the [ticket](https://sagebionetworks.jira.com/browse/SCHEMATIC-163), introduced the error handling and tested it again, confirming that the new error type and message are received in this case.

**Update (11/26):**

I have updated this PR to include a snippet of code that will catch if the manifest provided does not have the `entityId` column in the `SynapseStorage.updateDatasetManifestFiles` function and trigger a new manifest to be generated in this case. 

**Testing:**

I used the instructions in the ticket to reproduce the error, and then applied this latest change:

When I run
```
schematic manifest -c test_flat_config.yml get -dt BulkRNA-seqAssay -d syn64109617 -s -a
```
I get a [manifest](https://docs.google.com/spreadsheets/d/1hIDhAr0y8L2DPJi1tDmkaK9ugfyUDliBxuQNIIA7NRE) with annotations filled out.

When I run
```
schematic manifest -c test_flat_config.yml get -dt BulkRNA-seqAssay -d syn64109617 -s
```
I get a [manifest](https://docs.google.com/spreadsheets/d/1uPjFKxnkIQwq7PeVDYNdH1k9xhcIieeXP10FiGZ-ZyU) without annotations filled out.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1056,7 +1056,7 @@ class SynapseStorage(BaseStorage):
 
         Returns:
             Synapse ID of updated manifest and Pandas dataframe containing the updated manifest.
-            If there is no existing manifest return None
+            If there is no existing manifest or if the manifest does not have an entityId column, return None
         """
 
         # get existing manifest Synapse ID


BryanFauble (approved):
LGTM!

schematic/store/synapse.py
```diff
@@ -1119,6 +1119,13 @@ def _get_file_entityIds(
                     "No manifest was passed in, a manifest is required when `only_new_files` is True."
                 )
 
+            if "entityId" not in manifest.columns:
```
thomasyu888, line 1122: Just because I'm a worry-wort, can you tell me when the `_get_file_entityIds` function is called?  Otherwise, this is great!

thomasyu888 (approved):
I'll leave for @GiaJordan to leave any comments in case we miss anything

schematic/store/synapse.py
```diff
@@ -1119,6 +1119,13 @@ def _get_file_entityIds(
...
+                raise ValueError(
+                    "The manifest in your dataset and/or top level folder must contain the 'entityId' column. "
+                    "Please generate an empty manifest without filling in the annotations, "
+                    "add files to your manifest and try again."
```
GiaJordan, line 1126: I think having the last part of the message say `Please generate an empty manifest without annotations, manually add annotations to the appropriate files in the manifest, and then try again.` Would be a little more clear since when they regenerate the manifest the files and entityIds will be added automatically but otherwise it looks good!

GiaJordan (approved):
Just the one comment but otherwise looks good!

schematic/store/synapse.py
thomasyu888: I think that is ok. I got feedback from this slack thread: https://sagebionetworks.slack.com/archives/C01ANC02U59/p1732144795107819?thread_ts=1732144795.107819&cid=C01ANC02U59.

But I do think what you have stated is more clear.

schematic/store/synapse.py
BWMac: It is called when:
- `add_annotations_to_entities_files` is called - not relevant here because it is called with `only_new_files=False`
- `_get_files_metadata_from_dataset` is called - not relevant here because it is called with `only_new_files=False`
- `fill_in_entity_id_filename` is called - twice, once with `only_new_files=True`
     - This is in turn called when a file-based manifest is updated (`updateDatasetManifestFiles`)

That being said, I don't foresee any unintended consequences (famous last words) to this change because it is the line before a `KeyError` will always be raised if the manifest doesn't have that column.

schematic/store/synapse.py
thomasyu888: @BWMac I think the text is incorrect but we can wait for final feedback

GiaJordan: >@GiaJordan do you think that we should even suggest that they can fill out entityId themselves? For example, super users who know what they're doing may want to fill in the entityId column without having to fill out an entirely blank sheet again. I know that this is what the HTAN community has been doing instead of generating an empty manifest.

@thomasyu888 we could. The risk is when files/entities(folders) are annotated it's based on the entityId. If there was a mismatch then the wrong annotations would be added to the files.

schematic/store/synapse.py
thomasyu888: @GiaJordan Hmmm, so what you're saying is there's two scenarios that people may hit this issue:
1. generating an empty manifest when there's a manifest in the dataset folder
2. generating a manifest when there's a manifest in the dataset folder and we want annotations filled?

schematic/store/synapse.py
thomasyu888: Thanks @GiaJordan ^ That's true, should we include that in the error message like:

```
The manifest in your dataset and/or top level folder must contain the 'entityId' column. Please generate an empty manifest without annotations, manually add annotations to the appropriate files in the manifest, and then try again. For super users, you may fill out entityId's yourself, but you risk adding annotations to the wrong File entity if you fill out the wrong entityId"

schematic/store/synapse.py
thomasyu888: Oops. accidentally resolved it... @GiaJordan

GiaJordan: Yes, schematic always looks to see if a manifest exists in the dataset (if an id is provided) when generating a manifest. In this case because a manifest exists but its formatted incorrectly it would cause issues when determining how to order the existing/new files that exist on the dataset.

GiaJordan: > unless our recommendation of generating an empty manifest will fail.

This is what I'm saying will happen

> Correct me if I'm wrong, but if creation of a new blank manifest is unable to occur, that was already true before this change.

No you're correct, this is about the recommended suggestion to get users back into the flow, now raising the exception itself

thomasyu888: Yikes. We definitely don't want creation of an empty manifest to fail if there's an existing manifest. 

So then if we add this, will we need that error message?

GiaJordan: > what happens when submitting a manifest without the entityId column? That gets filled in if the Filename is correct right?

No, entityId are required to annotate the files. There are some cases where the ids are present and then erroneously dropped, but there are other cases where they weren't present in the first place, like this one. When submitting the raised error is the same iirc @thomasyu888

thomasyu888: We will need to test both of these solutions for when there isn't a manifest in the folder
This is for when there is an existing manifest in the folder
```schematic manifest -c test_flat_config.yml get -dt BulkRNA-seqAssay -d syn64109617 -s```
^ This will also cause the same error, so we don't want this to error and we want an empty manifest returned (ideally with entityId) filled out

```schematic manifest -c test_flat_config.yml get -dt BulkRNA-seqAssay -d syn64109617 -s -a```
^ We want to throw an error

OR

Product research done on asking if people are ok with blank manifests returned even with the `-a` parameter (if entityId) is NOT present.

BWMac: I was messing around with this and I think the simplest way to trigger the new manifest to generate is just to add a check for the `entityId` column in `SynapseStorage.updateDatasetManifestFiles` directly instead of doing some error handling which is a bit more complicated due to the nested nature of the functions. I added this in my latest [commit](https://github.com/Sage-Bionetworks/schematic/pull/1551/commits/6795c4cb947936ff58b2181e123f4e4420a23b44). Thoughts @GiaJordan @thomasyu888? Any consequences of doing this I haven't thought of?

andrewelamb (approved):

PR #1552, opened 2024-11-22
[SCHEMATIC-212] Prevent traces from being combined

BryanFauble:
**Problem:**

1. When integration tests were run the telemetry data in SigNoz was combining a bunch of traces under the same trace as shown in the screenshot:

Untitled

**Solution:**

1. Un-Instrument the flask auto tracing that causes issues when running a local flask test server.
2. Adding in logic to add the github branch to the created traces/logs in order to filter out specific test/branch runs

**Testing:**

1. The run from this github action looks good!


tests/conftest.py
```diff
@@ -42,6 +43,8 @@
 TESTS_DIR = os.path.dirname(os.path.abspath(__file__))
 DATA_DIR = os.path.join(TESTS_DIR, "data")
 
+FlaskInstrumentor().uninstrument()
+
```
BryanFauble, line 47: This doesn't make us lose much. Mostly just the hooks for tracking what arguments were sent into the flask server, but isn't that important since we can just look at the integration test.

thomasyu888 (commented):
🔥 LGTM! I'll have someone else do a final review for cross training efforts.

BWMac (approved):
LGTM

PR #1553, opened 2024-11-22
[SCHEMATIC-214] Wrap pandas functions to support not including `None` with the NA values argument

BryanFauble:
**Problem:**

1. When the `None` string is included with a manifest the pandas function was causing it to be converted over to a float not a number (NaN). This is a change in Pandas 2.0 release: "Added "None" to default na_values in read_csv() (GH 50286)"

**Solution:**

1. Grab the existing list of `na_value` objects and remove the `None` value from the list. Pass that list back into the function and replace the default `na_value` objects with this new list.

**Testing:**

1. Unit/Integration testing
2. Tom tested with an HTAN manifest that was broken, confirmed that this branch processed the manifest without issue


schematic/utils/df_utils.py
```diff
@@ -4,17 +4,58 @@
...
 import numpy as np
+import pandas as pd
 from pandarallel import pandarallel  # type: ignore
 
+# pylint:disable=no-name-in-module
+from pandas._libs.parsers import STR_NA_VALUES  # type: ignore
+
+STR_NA_VALUES_FILTERED = deepcopy(STR_NA_VALUES)
+
+try:
```
thomasyu888, line 21: May I ask why this is necessary?

BryanFauble: When using `.remove` on a set it will raise a KeyError if not found. I wanted to add this in and future proof the code and the super slim chance that some version that a user has of pandas doesn't have "None" in the set.

I know the project toml has pandas 2.x.x set, but if you were to manually go back to 1.x.x this code would break without the try/except

thomasyu888 (approved):
🔥 LGTM!  let's wait for Gianna/Andrew to review this when they're back before merging in case there are things we aren't thinking of.

andrewelamb (approved):

schematic/utils/df_utils.py
```diff
@@ -4,17 +4,58 @@
...
+    STR_NA_VALUES_FILTERED.remove("None")
+except KeyError:
+    pass
+
 logger = logging.getLogger(__name__)
 
 
+def read_csv(
```
linglp, line 29: should we give this function a slightly different name to avoid confusion? i.e. `read_csv_no_none` or `read_csv_filter_none`?

linglp (commented):
The changes here make sense to me. But I am thinking if we could add `None` as a valid value here in the data model: https://github.com/Sage-Bionetworks/schematic/blob/develop/tests/data/example.model.csv#L9 and then create a manifest with string `None` to better test the changes? Currently, we can't test with our existing data model because no attribute has "None" as a valid value.

schematic/utils/df_utils.py
BryanFauble: I'm not sure if we should change the name of the function. Depending on the `keep_default_na` bool is what determines if `None` is going to be filtered out. It would be a bit awkward to call a function called `read_csv_no_none` if you aren't using that functionality.

BryanFauble:
> The changes here make sense to me. But I am thinking if we could add `None` as a valid value here in the data model: https://github.com/Sage-Bionetworks/schematic/blob/develop/tests/data/example.model.csv#L9 and then create a manifest with string `None` to better test the changes? Currently, we can't test with our existing data model because no attribute has "None" as a valid value.

I added to an existing test for this, let me know if this covers what you had in mind @linglp 

https://github.com/Sage-Bionetworks/schematic/pull/1553/commits/6eeacd5ddbfcd62aad94e301859a2d59c0b7def6

tests/data/example_test_nones.model.csv
```diff
@@ -6,7 +6,7 @@ Sex,,,"Female, Male, Other",,,TRUE,DataProperty,,
 Year of Birth,,,,,,FALSE,DataProperty,,
 Diagnosis,,,"Healthy, Cancer",,,TRUE,DataProperty,,
 Cancer,,,,"Cancer Type, Family History",,FALSE,ValidValue,,
-Cancer Type,,,"Breast, Colorectal, Lung, Prostate, Skin",,,TRUE,DataProperty,,
+Cancer Type,,,"Breast, Colorectal, Lung, Prostate, Skin, None",,,TRUE,DataProperty,,
```
GiaJordan, line 9: `Cancer Type` needs to be added to the `Depends On` column of the `MockComponent` component; without it, the jsonschema validation won't actually validate the `Cancer Type` column of the test manifest because it's not a part of the component and the valid values won't be verified. Alternatively, a `Patient` type manifest could also be used.

GiaJordan (dismissed):

tests/data/example_test_nones.model.csv
BryanFauble: I spent some time and added the suggestion, but then no matter what I add into the `Cancer Type` of the `MockCompontent` manifest nothing happens. Would I be expecting that `validate_manifest_rules` would be triggering any errors or warning if a value isn't one in the list that it could be?

I also tried modifying one of the CSV files in this test and I couldn't get anything to fail for any random data I add into the CSV: https://github.com/Sage-Bionetworks/schematic/blob/schematic-210-pandas-remove-none/tests/test_validation.py#L73-L81

Would you be willing to add a commit with the changes you're suggesting so I can learn what needs to be done?

GiaJordan: @BryanFauble sure, I opened #1555 to address

tests/integration/test_validate_manifest.py
thomasyu888, line 1: Can these tests be moved into `test_manifest_validation.py`?

tests/integration/test_validate_manifest.py
```diff
@@ -0,0 +1,76 @@
...
+
+        vm = ValidateManifest(errors, manifest_df, manifest_path, dmge, json_schema)
+        errors, warnings = vm.validate_manifest_values(manifest_df, json_schema, dmge)
+        assert not warnings
+        assert errors
+
+        # Both manifests will have errors of type "<xxx> is not of type 'array'
+        # Only the invalid manifest will have errors of type "<xxx is not one of
+        #  ['Breast', 'None', 'Prostate', 'Colorectal', 'Skin', 'Lung']" in the
+        #  "Cancer Type" column.
```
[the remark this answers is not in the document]
thomasyu888: @GiaJordan I printed out the errors:

```[['2', 'Check List Enum', "'ab,cd' is not of type 'array'", 'ab,cd'], ['2', 'Check List Enum Strict', "'ab,cd' is not of type 'array'", 'ab,cd'], ['2', 'Check List Like Enum', "'ab' is not of type 'array'", 'ab'], ['3', 'Check List Enum', "'ab,cd' is not of type 'array'", 'ab,cd'], ['3', 'Check List Enum Strict', "'ab,cd' is not of type 'array'", 'ab,cd'], ['3', 'Check List Like Enum', "'' is not of type 'array'", ''], ['4', 'Check List Enum', "'ab,cd' is not of type 'array'", 'ab,cd'], ['4', 'Check List Enum Strict', "'ab,cd' is not of type 'array'", 'ab,cd'], ['4', 'Check List Like Enum', "'' is not of type 'array'", ''], ['5', 'Check List Enum', "'ab,cd' is not of type 'array'", 'ab,cd'], ['5', 'Check List Enum Strict', "'ab,cd' is not of type 'array'", 'ab,cd'], ['5', 'Check List Like Enum', "'ab,cd' is not of type 'array'", 'ab,cd']]```

Something is odd about this, can you help look into this as I think Andrew is off?

File: tests/data/example_test_nones.model.jsonld
Comment (thomasyu888) [inline tests/data/example_test_nones.model.jsonld:1]: Should this file be modified?

File: tests/data/example_test_nones.model.csv
Comment (thomasyu888) [inline tests/data/example_test_nones.model.csv:1]: I think this csv was used for something else.  Mialy contributed this for testing "nones" but I don't think it had anything to do with "none" string values.

File: tests/integration/test_validate_manifest.py [hunk 5]
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        dmge = helpers.get_data_model_graph_explorer(path=model)
+
+        data_model_js = DataModelJSONSchema(jsonld_path=model_path, graph=dmge.graph)
+        json_schema = data_model_js.get_json_validation_schema(
+            root_node, root_node + "_validation"
+        )
+
+        manifest_df = load_df(
```
thomasyu888, line 56: I'm confused, is there a function within schematic that takes a filepath and loads everything?  I feel like the columns that are lists aren't being read in correctly.  Is that part of `load_df`'s job?

tests/data/mock_manifests/Valid_Test_Manifest_with_nones.csv
thomasyu888, line 1: Same note as above, this test wasn't initially created for the "none" values

tests/integration/test_validate_manifest.py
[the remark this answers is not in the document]
GiaJordan: Yep!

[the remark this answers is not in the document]
thomasyu888: If I'm reading the code correctly, it might be safest to just run the `validate_all` function.  I didn't even realize that `validate_manifest_rules` returns a dataframe.  Does it subset it by columns the jsonschema cannot validate?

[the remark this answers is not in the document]
GiaJordan: I think that's it. List validation does modify the dataframe; it parses the string to a list of strings when the rule is applied to a column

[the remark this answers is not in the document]
thomasyu888: Thanks for the confirmation, I'll count on the two of you to resolve this - I agree that a valid manifest should have 0 errors.

tests/integration/test_validate_manifest.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Both manifests will have errors of type "<xxx> is not of type 'array'
+        # Only the invalid manifest will have errors of type "<xxx is not one of
+        #  ['Breast', 'None', 'Prostate', 'Colorectal', 'Skin', 'Lung']" in the
+        #  "Cancer Type" column.
+        error_attributes = [error[1] for error in errors]
+        if manifest == "mock_manifests/Valid_Test_Manifest_with_nones.csv":
+            assert "Cancer Type" not in error_attributes
+        else:
+            assert "Cancer Type" in error_attributes
```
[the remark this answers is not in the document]
thomasyu888: `"Cancer Type" in error_attributes` is risky because I could exclude the Cancer Type column and it would cause an error message with "Cancer Type" in the error message.

thomasyu888 (commented):
@andrewelamb, thanks for adding the integration test - please see my comments.

tests/integration/test_validate_manifest.py
GiaJordan: > I'm confused, is there a function within schematic that takes a filepath and then validates everything?

Yes, that would be the function you listed. I think @andrewelamb was trying to skip running all the validation rules and just perform the jsonschema validation but I think in this case it's safer to continue to use `validateModelManifest` like we have in `test_validation.py`

tests/integration/test_validate_manifest.py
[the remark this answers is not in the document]
andrewelamb: "List validation does modify the dataframe; it parses the string to a list of strings when the rule is applied to a column" Is that supposed to happen?

tests/data/example_test_nones.model.csv
GiaJordan: I already was of both minds on using this model and component, I think we should switch both @thomasyu888

GiaJordan:
I've opened up #1556 to address the data model and component concerns. That PR just updates the existing data models and adds new test manifests. We can revert #1555 or modify the `test_nones` data models in this branch to what they were before.
@andrewelamb The valid test manifest used in that pr raises no warnings or errors and the invalid manifest only raises the one expected error message. You can use those for your tests instead of the ones currently being used.
cc: @thomasyu888

tests/integration/test_validate_manifest.py
GiaJordan: @andrewelamb going through `validateModelManifest` will be less of an issue with the Biospecimen manifest, there are no validation rules defined, its just the valid values in one attribute.

thomasyu888 (approved):
To aid in the upcoming release due to long testing times, I did this: https://github.com/Sage-Bionetworks/schematic/pull/1553/commits/d0a8e1577d439eb487370d31d565f71f1ddfbe95.

1. Nit: Thanks for adding this in an existing test module, but please be sure to modify the docstring to fit the extra testing functions
2. Nit: In my opinion, parametrize is best used when it's testing the same condition.  For example, if there were many valid manifests to be tested, that is when I would personally use it to add all the valid manifest paths.  In my opinion, it's more readable than trying to figure out the if-else AND parametrize around two testing conditions.

Great work everybody!

cc @andrewelamb.

PR #1554, opened 2024-11-23
[SCHEMATIC-183] Update tests - Use magic mock and add parentId 

thomasyu888:
1. I noticed the same thing Gianna observed: the SynapseStorage class automatically fires off a query which isn't great, BUT that's not something we fix right now and is outside the scope of this ticket. This is most likely due to the `perform_query=True` parameter when the class is instantiated
2. You don't have to nest the `patch.object`, but there's that weird syntax (choose your poison situation).
3. Tests are sort of all over the place, but let's try to run the test suites locally for the parts you add code for whenever you modify code (and maybe parts of the integration tests which the code you add impact) - because the tests on GitHub take 6 hours to run, it increases the development time if you just rely on that. (It's completely reasonable to miss some, that's what the CI is here for)
4. `query_fileview` is the function called to get the fileview, so you want to mock that
5. Nit: Use snakecase for everything
7. Adding the HACK adds more lines in the output, which is why Gianna also increased the row count on text output because she also added a fileview query. the positioning of the row output is actually a bit unstable - so I updated some of the tests to not always check the positioning of the output. It's more important that the output is there.

notes about the hack, this is what happens before the HACK
1. The SynapseStorage class is instantiated, when this happens, it queries the latest state of the fileview
2. New files or folders are added within the project which is contained within the scope of the fileview
3. It will throw an error because the fileview isn't re-queried for Dataset not found.


schematic/store/synapse.py
```diff
@@ -705,7 +705,10 @@ def getFilesInStorageDataset(
             ValueError: Dataset ID not found.
         """
         file_list = []
-
+        # HACK: must requery the fileview to get new files, since SynapseStorage will query the last state
+        # of the fileview which may not contain any new folders in the fileview.
+        # This is a workaround to fileviews not always containing the latest information
+        self.query_fileview(force_requery=True)
```
thomasyu888, line 711: Post in the team channel to discuss this hack and there's a message in #synapse channel for us to chime in on

tests/integration/test_store_synapse.py
```diff
@@ -400,38 +400,31 @@ def test_mock_get_files_in_storage_dataset(
         with patch(
             "schematic.store.synapse.CONFIG", return_value=TEST_CONFIG
         ) as mock_config:
-            with patch.object(synapse_store, "syn") as mock_synapse_client:
```
thomasyu888, line 403: This is me misleading Gianna, but there are times when we want a connection with Synapse and there are times we don't.

This is probably a case where it is ok for there to a direct connection with Synapse to simplify the mocking - since you've included in your other `test/test_store.py` the mocked version.

schematic/store/synapse.py
```diff
@@ -19,12 +19,10 @@
 import numpy as np
 import pandas as pd
 import synapseclient
-import synapseutils
```
thomasyu888, line 22: If you're using vscode, you may see unused imports "grayed" out.  It's ok to remove them in PRs, the intent is that each PR that we contribute should put the codebase in a better place.

schematic/store/synapse.py
```diff
@@ -705,7 +705,10 @@ def getFilesInStorageDataset(
             ValueError: Dataset ID not found.
         """
         file_list = []
```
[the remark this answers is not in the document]
BryanFauble: https://github.com/Sage-Bionetworks/schematic/pull/1552

Is necessary to filter out difference between branches in gh runs. It's possible to get the data now, just a bit more difficult to filter it out. As of now the average duration of this function plotted over time:

image

schematic/store/synapse.py
BryanFauble: I wonder if there is something that we can do here with this API:

https://rest-docs.synapse.org/rest/POST/entity/id/table/query/async/start.html

Specifically:
"The last updated on date of the table (lastUpdatedOn) = 0x80"

If this get's updated as data is indexed into the table we could fetch the lastUpdatedOn field before we query the table to know if we need to re-query the table or not.

BryanFauble (approved):
I agree with the changes you've made on this so far, although, as you mention - includes some hacks.

schematic/store/synapse.py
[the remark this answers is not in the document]
BryanFauble: Feel free to pull develop into this feature branch and we can compare the develop branch to this feature branch performance.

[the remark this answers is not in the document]
thomasyu888: Thanks @BryanFauble. Done

schematic/store/synapse.py
thomasyu888: Another thing to note: if we look at the failing test that was the motivation behind this hack, it's because it falls outside of the overall schematic flow. 

So for example: when people upload a bunch of files, they usually first run manifest generate. Now manifest generate will have a "this dataset doesn't exist" situation but then afterwards, the fileview should always consist of the data - theoretically. There have been incidents when it doesn't but that's very little in the grand scheme of things.

So we _could_ remove the hack and trigger the fileview indexing within the test but that's probably best as a team decision.

schematic/store/synapse.py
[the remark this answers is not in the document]
BryanFauble: @thomasyu888 These are the results over the past 5 days. We can see your branch has a much better average execution time for this function. Although, similar to `bwmac/SCHEMATIC-163/error-message-update` for some reason.

image

We can selected on a few fields to filter for what we want, perform an average of the duration for the function, then group by a few fields to get these results.

schematic/store/synapse.py
andrewelamb: Pylint would also catch this, but this module hasn't been gone through and fixed so that is passes, so it isn't included in the pylint check.

andrewelamb (approved):

schematic/store/synapse.py
thomasyu888: @GiaJordan After a team discussion, we decided that it would be better to modify the test to re-query the fileview since this is such an edgecase.  It does throw an error already, and upon multiple runs of this code outside of the test, it would work.  This only doesn't work so smoothly in the test because resources are created and destroyed.

```
FAILED tests/integration/test_metadata_model.py::TestMetadataModel::test_submit_filebased_manifest_file_and_entities_valid_manifest_submitted - LookupError: Dataset syn64313762 could not be found in fileview syn23643253.
FAILED tests/integration/test_metadata_model.py::TestMetadataModel::test_submit_filebased_manifest_file_and_entities_mock_filename - LookupError: Dataset syn64313765 could not be found in fileview syn23643253.
```

linglp (approved):
Looks good to me. Thanks Tom!

thomasyu888:
@GiaJordan feel free to merge this after you review it since I may be on a flight. Thanks everyone for their reviews!

GiaJordan (approved):
Thanks for the changes @thomasyu888 looks good to me!

PR #1555, opened 2024-12-09
[SCHEMATIC-210] Add attribute to nones data model

GiaJordan:
From #1553 
>I spent some time and added the suggestion, but then no matter what I add into the Cancer Type of the MockCompontent manifest nothing happens. Would I be expecting that validate_manifest_rules would be triggering any errors or warning if a value isn't one in the list that it could be?
I also tried modifying one of the CSV files in this test and I couldn't get anything to fail for any random data I add into the CSV: https://github.com/Sage-Bionetworks/schematic/blob/schematic-210-pandas-remove-none/tests/test_validation.py#L73-L81
Would you be willing to add a commit with the changes you're suggesting so I can learn what needs to be done?

For an attribute to be part of a component, it needs to be listed in the component's `Depends On` column. Without the addition and using the model with the addition the column won't be validated with the jsonschema validation(which would validate the `Valid Values` according to the data model). The validation rules should still be performed. 

In the case of the model in #1553 while the manifest can have `None` strings included, the validation that these are actually `None` strings is not happening as part of the validation tests because the attribute has not been added to the component. This doesn't change the behavior of the valid manifest but when there's a manifest with an invalid value, if the attribute has not been added to the component then no error would be raised.


thomasyu888 (commented):
🔥 LGTM - thanks for the quick work! I'll defer to @andrewelamb to final review and wait for tests to pass

GiaJordan:
Just a note that `Invalid_Test_manifest_with_nones.csv` isn't used in any of the tests currently, it was included as an example. Whoever is finishing #1553 could add a test to ensure that the appropriate error gets raised. @thomasyu888 @andrewelamb

andrewelamb (approved):

thomasyu888:
> Just a note that `Invalid_Test_manifest_with_nones.csv` isn't used in any of the tests currently, it was included as an example. Whoever is finishing #1553 could add a test to ensure that the appropriate error gets raised. @thomasyu888 @andrewelamb 

@andrewelamb / @Gibbsdavidl since Bryan is OOO, could I count on the two of you to self organize and decide who adds this test? Thanks for raising that!

GiaJordan:
>🔥 LGTM - thanks for the quick work! I'll defer to @andrewelamb to final review and wait for tests to pass

@thomasyu888 The two tests that use this data model, `test_convert_nan_entries_to_empty_strings` and `test_valid_manifest` are passing for me locally. Would you rather me merge this PR now so @andrewelamb can continue work or wait for the integration test workflow to finish?

PR #1556, opened 2024-12-11
[SCHEMATIC-214] change data model and component

GiaJordan:
I think it'd be better to use the `example.model.csv` model and `Biospecimen` component for the tests of manifest `"None"` values given how `MockComponent` is used and the history of the `example_test_nones.model.csv` model that @thomasyu888 pointed out. 

Included in this PR is updates to the data model (CSV and JSONLD forms) and both valid and invalid manifests that can be used for the integration tests.


thomasyu888 (approved):
🔥 LGTM! Thanks for the quick work!  Will we need to add any additional tests?  Will let Andrew be final reviewer.

GiaJordan:
@thomasyu888 thanks and no prob! 
As Andrew is responding to the other comments covering the tests he can switch to using these manifests. The valid can be added to `test_valid_manifest` and the invalid manifest can be used in place of the current one in his new test. We'll need to modify the other changed data models in the other branch, I didn't want to do so here because the tests still use them

andrewelamb (approved):

PR #1560, opened 2024-12-16
[SCHEMATIC-210] Updates `publish.yml` for Test PyPi deployments (#1559)

BWMac:
**Description:**

This PR updates `publish.yml` to automatically publish releases to test pypi (on pre-release) and to pypi (on normal release). I also added "trusted publisher" relationships (OIDC) between this repo and both pypi repos so permanent access tokens are no longer required for the CI pipeline to run. While I was working on this, I was having issues with the `update_version.py` file and @thomasyu888 suggested deprecating it for simplicity so I did that. We will need to update the version in `pyproject.toml` when publishing a release from now on (which is what the release [instructions](https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/3679911937/Schematic+Release+Branch+workflow#Update-version-number) already said to do). I tested the workflow on both pre-releases and full releases and it behaved as expected. I subsequently deleted the release I made to pypi prod, but here is the [release](https://test.pypi.org/manage/project/schematicpy/release/99.99.9/) I made on test pypi.


BryanFauble:
Is there any documentation update needed to note the version update changes on release? I think we had a confluence page for the release process, contributing.md or the main Readme may also need updates

.github/workflows/publish.yml
```diff
@@ -1,107 +1,92 @@
...
-      - 'v[1-9][0-9].[0-9]+.[0-9]+'
-    branches: [main]
+  release:
+    types: [published]
 
 jobs:
-  pypi_release:
-    runs-on: ubuntu-22.04
+  build:
+    runs-on: ubuntu-latest
```
thomasyu888, line 8: Did we want to lock it to the actual latest version number?

.github/workflows/publish.yml
```diff
@@ -1,107 +1,92 @@
...
-      - name: Check out repository
-        uses: actions/checkout@v4
-
-      - name: Set up Python ${{ env.PYTHON_VERSION }}
-        uses: actions/setup-python@v5
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: ${{ env.PYTHON_VERSION }}
+      - id: build-package
```
BryanFauble, line 19: Is this build-package step equivalent of the poetry build? I'd be concerned if they end up being slightly different, or, This doesn't take into account the poetry lock file anymore.

thomasyu888 (commented):
🔥 LGTM!  Cool trick with the repository-url.  I'll leave it to @linglp / @GiaJordan / @BryanFauble for final review.

.github/workflows/publish.yml
BWMac: It does replace that step. It doesn't use the poetry lock file but it gets the required packages from `pyproject.toml` which is where poetry takes its instructions from to begin with.

.github/workflows/publish.yml
```diff
@@ -1,107 +1,92 @@
...
-      #    post a message to slack
-      #----------------------------------------------  
+      - name: deploy-to-test-pypi
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository-url: ${{ github.event.release.prerelease == true && 'https://test.pypi.org/legacy/' || '' }}
 
-      - name: Post to a Slack channel
-        if: steps.publish-to-pypi.outcome == 'success'
+  slack-notification:
```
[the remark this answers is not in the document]
BWMac: That is a really good point and I don't think it should!

.github/workflows/publish.yml
```diff
@@ -1,107 +1,92 @@
...
+      - name: download-bdist
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ needs.build.outputs.bdist-package-name }}
+          path: dist
 
-      #----------------------------------------------
-      #    post a message to slack
-      #----------------------------------------------  
+      - name: deploy-to-test-pypi
```
linglp, line 74: I noticed this step only includes deployment to test pypi. Are we no longer deploying to PyPI, or is there a step for PyPI deployment that I'm missing?

BWMac: It publishes to pypi when pre-release is not selected. to test pypi when it is selected. Changing the name now

.github/workflows/publish.yml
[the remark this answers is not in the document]
BWMac: changed in latest commit

.github/workflows/publish.yml
BWMac: changed in latest commit

.github/workflows/publish.yml
```diff
@@ -1,107 +1,93 @@
...
+          name: ${{ needs.build.outputs.bdist-package-name }}
+          path: dist
 
-      #----------------------------------------------
-      #    post a message to slack
-      #----------------------------------------------  
+      - name: deploy-to-pypi
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository-url: ${{ github.event.release.prerelease == true && 'https://test.pypi.org/legacy/' || '' }}
```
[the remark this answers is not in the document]
BWMac: when the url is empty it uploads to pypi prod which is the default. So in this case when prerelease is true it goes to test and when it is false (unchecked) it goes to pypi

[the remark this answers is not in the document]
linglp: Okay thank you for confirming!

.github/workflows/publish.yml
linglp: Thanks for the update!

linglp (approved):
Looks good to me. Thanks for all the hard work!

GiaJordan (approved):
LGTM! Thanks @BWMac

PR #1564, opened 2024-12-26
[DPE-1188] Migrate schematic to SigNoz cloud

BryanFauble:
**Problem:**

1. Maintenance of the self-hosted SigNoz solution ended up causing more friction than it was worth.
2. Schematic needed to be updated to use SigNoz cloud

**Solution:**

1. Updating github actions to export telemetry data to SigNoz cloud
2. Updating documentation to reflect this information as well as dropping the need to complete the oauth2 token exchange

**Testing:**

1. Will be testing with this GH action run


thomasyu888 (approved):
🔥 LGTM! Let's make sure to post in team channel to let everyone know of this change from self hosting to cloud. 

Second - does this simplify things in terms of being able to export to other otel services like Jaeger and/or open relic?

BryanFauble:
> Second - does this simplify things in terms of being able to export to other otel services like Jaeger and/or open relic?

It was always the case that we could export data to other otel services or cloud offerings. From all of them that I had looked into though they all follow a similar model where you specify an HTTP header with an API key. In that regard, this means it's easier to swap to whichever cloud provider we want to use. The only caveat is that from python we can only send it to a single location, but, when sending data to an OpenTelemetry collector - That collector can forward the data to 1..* cloud providers. Though - For the most part I assume we only want to send data to a single place.

PR #1565, opened 2025-01-03
[DPE-1151] Update ReadMe instructions and environment variable when exporting to signoz cloud

linglp:
## Changes
* Updated instructions when exporting traces to Signoz Cloud
* Ensured environment variables can be loaded in `__init__.py`

## Tests
* Ensured that after setting environment variables in `.env`, when running tests locally, results can be found on Signoz Cloud
* Ensured that after setting environment variables in `.env`, when running schematic api locally, results can be found on Signoz Cloud
* Ensured that results of test run on Github can be found in Signoz Cloud
* Ensured that after setting environment variables in `.env`, when running schematic CLI locally, results can be found on Signoz Cloud

### DIFF
--- a/schematic/__init__.py
+++ b/schematic/__init__.py
@@ -27,9 +27,13 @@ from schematic.configuration.configuration import CONFIG
 from schematic.loader import LOADER
 from schematic.version import __version__
 from schematic_api.api.security_controller import info_from_bearer_auth
+from dotenv import load_dotenv
 
 Synapse.allow_client_caching(False)
 logger = logging.getLogger(__name__)
+
+# Ensure environment variables are loaded
+load_dotenv()
 
 
 def create_telemetry_session() -> requests.Session:
--- a/README.md
+++ b/README.md
@@ -514,8 +514,9 @@ environment variable should be set:
 - `OTEL_EXPORTER_OTLP_HEADERS=signoz-ingestion-key=<key>`
 - `OTEL_EXPORTER_OTLP_ENDPOINT=https://ingest.us.signoz.cloud`
 
-If a key is required please reach out to DPE and a unique ingestion key will be provided.
+If you do not have access to SigNoz Cloud and require a key, please contact DPE to request a unique ingestion key.
 
+For internal developers with access to SigNoz Cloud, you can obtain an ingestion key by following the step [here](https://signoz.io/docs/ingestion/signoz-cloud/keys/).
 
 # Contributors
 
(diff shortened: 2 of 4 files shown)


BryanFauble (approved):
Great fix. Thanks!

schematic/__init__.py
```diff
@@ -27,10 +27,14 @@
 from schematic.loader import LOADER
 from schematic.version import __version__
 from schematic_api.api.security_controller import info_from_bearer_auth
+from dotenv import load_dotenv
```
thomasyu888, line 30: @linglp does this need to be added to the poetry.lock file?

linglp: we already have `from dotenv import load_dotenv` in conftest, so I don't think we need to add it to the lock file.

thomasyu888: `conftest` is tests. This is now part of `__init__.py` which is part of the API and CLI loading. if `dotenv` is not installed in the general `pip install schematicpy` then this will cause problems

BryanFauble: https://github.com/Sage-Bionetworks/schematic/blob/aa851e3c04689e9574913a1340f923b153c20c8c/pyproject.toml#L91

That's right! It would need to be moved to the regular project dependencies in order to install with the package.

thomasyu888: @linglp what happens when the .dotenv file doesn't exist?  My concern is failing AWS infra deployment

linglp: @thomasyu888 thanks for the question, Tom! I read the source code of `dotenv` here: https://github.com/maxkoryukov/python-dotenv/blob/311f56cf524a9e2b38a65f210170e4e631778b35/dotenv/main.py#L19 and it seems like if .env does not exist, `load_dotenv` will just return None.

thomasyu888: @linglp , this is helpful.

Can you add to the readme that there are two ways to provide environmental variables when using schematic CLI? 

1. `export ...` and/or add to `~/.bashprofile`
2. Create .env file and add the the env variables required.

Given that the schematic API deployment won't ever use `.env` method, I think we don't have to include those instructions in the API deployment docs.

thomasyu888 (approved):
🔥 LGTM! Just minor comments - thanks for the review Bryan!

schematic/__init__.py
linglp: Thank you, Tom! I have added the instructions.

PR #1568, opened 2025-01-15
[schematic-230] propagate user ids as attribute to spans 

linglp:
## Problem
When looking into a trace on Signoz, currently we have trace down user id in __init__ span. This PR is to make sure that user id can be propagated to the top level span and also most of other span. 

example screen shots that demonstrate the problem: 

Here you could see "user id" contained a random string
Screenshot 2025-01-15 at 11 22 19 AM

And spans such as "query file view" do not contain a user id
Screenshot 2025-01-15 at 11 23 59 AM

Note: if a span has already ended we cannot add user id back to it. As an example, `ManifestGenerator::create_manifests` has two children, one is `DataModelParser::parse_model`, and another one is `ManifestGenerator::create_single_manifest`. We could propagate user id back to `ManifestGenerator::create_manifests`, but not `DataModelParser::parse_model` since this step happened before and ended by the time that the user id is available.

# Solution 
Borrowed attribute propagation function [here](https://github.com/BryanFauble/kubefox-python-sdk/blob/5eb7eed24a3c2b2ff09372ca6199658b6077155e/kit/telemetry/trace.py#L16) to propagate attributes from parent to child, and when the span ends, propagate attribute from back to parent. 

Now the user id attribute can be shown in most of the spans. 

Screenshot 2025-01-16 at 3 39 34 PM


schematic/__init__.py
```diff
@@ -140,13 +183,20 @@ def request_hook(span: Span, environ: Dict) -> None:
...
+        auth_header = environ.get("HTTP_AUTHORIZATION", None)
+        access_token = os.getenv("SYNAPSE_ACCESS_TOKEN", None)
+
+        if auth_header and len(auth_header.split(" ")) > 1:
+            token = auth_header.split(" ")[1]
+        else:
+            token = access_token
+
+        if token:
+            user_info = info_from_bearer_auth(token)
```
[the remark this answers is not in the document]
linglp: oh interesting.. when I tested locally, I could see user id be propagated on signoz. Let me try again. Thank you Bryan!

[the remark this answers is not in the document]
BryanFauble: That's correct. This works locally, but the way tokens are generated via oauth do not. That is how users in dca generate a token and pass it to schematic

schematic/store/synapse.py
```diff
@@ -499,23 +506,24 @@ def login(
...
             access_token = os.getenv("SYNAPSE_ACCESS_TOKEN")
 
         # login using a token
         if access_token:
             try:
+                current_span = trace.get_current_span()
+                if current_span.is_recording() and access_token:
+                    user_info = info_from_bearer_auth(access_token)
+                    if user_info:
+                        current_span.set_attribute("user.id", user_info.get("sub"))
```
BryanFauble, line 519: We can't use the bearer token as the source for the user.id. this will not work when deployed to aws behind dca. Setting of the attribute must come from `syn.credentials.owner_id`

linglp: Thanks, Bryan! I think I finally understand. Initially, I thought DCA would obtain a JWT token, decode it, and then pass the decoded token to Schematic. However, upon further thinking, it seems there’s no such "decoding" process. Instead, Schematic simply forwards the JWT token directly to Synapse.

BryanFauble: Schematic does decode the token and verifies the token against the public jwks of Synapse. This is how Synapse verifies that the token it was given was signed by Synapse and we can authenticate the authenticity of the request:
https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic_api/api/security_controller.py#L17-L44

See this PR where I added the logic:
https://github.com/Sage-Bionetworks/schematic/pull/1493

The `sub` claim of the token however, does not contain the profile/owner_id of the actual user. Which is why it must get gotten after the login call when we get the user profile from Synapse.

schematic/__init__.py
```diff
@@ -98,6 +139,9 @@ def set_up_tracing(session: requests.Session) -> None:
     if tracing_export == "otlp":
         exporter = OTLPSpanExporter(session=session)
         trace.get_tracer_provider().add_span_processor(BatchSpanProcessor(exporter))
+        # Add the custom AttributePropagatingSpanProcessor to propagate attributes
+        attribute_propagator = AttributePropagatingSpanProcessor(["user.id"])
+        trace.get_tracer_provider().add_span_processor(attribute_propagator)
```
BryanFauble, line 144: I think the order of span_processors being added needs to be flipped. It may cause things to happen out of order in certain scenarios otherwise.

schematic/__init__.py
```diff
@@ -139,25 +183,6 @@ def request_hook(span: Span, environ: Dict) -> None:
...
-        logger.exception("Failed to set user info in span")
-
-    try:
-        if (request := environ.get("werkzeug.request", None)) and isinstance(
-            request, Request
-        ):
-            for arg in request.args:
-                span.set_attribute(key=f"schematic.{arg}", value=request.args[arg])
-    except Exception:
-        logger.exception("Failed to set request info in span")
```
BryanFauble, line 160: I assume we still want this logic to exist? Was it intentionally removed?

linglp: Thanks for catching this. It was not intentional. Let me add back these lines.

schematic/__init__.py
linglp: Got it. Thank you for pointing that out.

BryanFauble (approved):
LGTM!

schematic/__init__.py
```diff
@@ -139,17 +183,6 @@ def request_hook(span: Span, environ: Dict) -> None:
...
         return
-    try:
-        if auth_header := environ.get("HTTP_AUTHORIZATION", None):
-            split_headers = auth_header.split(" ")
-            if len(split_headers) > 1:
-                token = auth_header.split(" ")[1]
-                user_info = info_from_bearer_auth(token)
-                if user_info:
-                    span.set_attribute("user.id", user_info.get("sub"))
-    except Exception:
```
thomasyu888, line 150: I know you're still working on this, but just a reminder, should this be added back in?

linglp: hmm I don't think we need this part any more because we are not using the `info_from_bearer_auth` function to debug the token. What's the purpose of keeping it?

BryanFauble: This portion does not need to be kept

PR #1573, opened 2025-02-06
[SCHEMATIC-240] stripped google sheet information from open telemetry span status and message 

linglp:
## Problem
When an error involving Google Sheets is logged, the message may include the URL of the Google Sheet being processed and the google sheet link can be found in signoz

## Solution 
Remove the sensitive information in the code. 
Evidence that this is working: 
As you can see, the google sheet links get turned to "REACTED": 
Screenshot 2025-02-06 at 3 21 25 PM

The following can be achieved: 
* Google Sheet links have been removed from all status messages and event exceptions.
* Original error messages are preserved and not completely overridden.

Note: I tried creating a custom processor and experimenting if I could filter out the sensitive information when the spans are still recording using the processor. But something like: `parent_span.set_status(status=Status(status_code=StatusCode.ERROR, description=sanitized_message))` couldn't override the span exception messages in signoz because it only changes the span descriptions, and [this](https://github.com/SigNoz/signoz-otel-collector/blob/main/exporter/clickhousetracesexporter/clickhouse_exporter.go#L386) seems to suggest that signoz uses exception messages to populate the UI.

### DIFF
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -1,10 +1,10 @@
 import json
 import logging
 import os
+import re
 from collections import OrderedDict
 from pathlib import Path
 from typing import Dict, List, Literal, Optional, Tuple, Union
-
 import networkx as nx
 import pandas as pd
 import pygsheets as ps
@@ -33,6 +33,9 @@ from schematic.utils.schema_utils import (
     extract_component_validation_rules,
 )
 from schematic.utils.validate_utils import rule_in_rule_list
+from googleapiclient.errors import HttpError
+from opentelemetry.trace import Status, StatusCode
+from opentelemetry.trace import get_current_span
 
 logger = logging.getLogger(__name__)
 tracer = trace.get_tracer("Schematic")
@@ -1367,7 +1370,8 @@ class ManifestGenerator(object):
         Returns:
             ps.Spreadsheet: A Google Sheet object.
         """
-
+        logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
+        # print('set data frame by url is being called', flush=True)
         # authorize pygsheets to read from the given URL
         gc = ps.authorize(custom_credentials=self.creds)
 
@@ -1375,8 +1379,17 @@ class ManifestGenerator(object):
         # This sheet already contains headers.
         sh = gc.open_by_url(manifest_url)
         wb = sh[0]
-
-        wb.set_dataframe(manifest_df, (1, 1), fit=True)
+        try:
+            wb.set_dataframe(manifest_df, (1, 1), fit=True)
+        except HttpError as ex:
+            span = get_current_span()
+            pattern = r"https://sheets\.googleapis\.com/v4/spreadsheets/[\w-]+"
+            sanitized_message = re.sub(pattern, "REDACTED", str(ex))
+            sanitized_message_b = sanitized_message.encode(encoding="utf-8")
+            span.set_status(
+                Status(status_code=StatusCode.ERROR, description=sanitized_message)
+            )
+            raise HttpError(ex.resp, sanitized_message_b)
 
         # update validation rules (i.e. no validation rules) for out of schema columns, if any
         # TODO: similarly clear formatting for out of schema columns, if any
--- a/schematic/manifest/generator.py
+++ b/schematic/manifest/generator.py
@@ -1370,7 +1370,6 @@ class ManifestGenerator(object):
         Returns:
             ps.Spreadsheet: A Google Sheet object.
         """
-        logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
         # print('set data frame by url is being called', flush=True)
         # authorize pygsheets to read from the given URL
         gc = ps.authorize(custom_credentials=self.creds)
@@ -1382,13 +1381,9 @@ class ManifestGenerator(object):
         try:
             wb.set_dataframe(manifest_df, (1, 1), fit=True)
         except HttpError as ex:
-            span = get_current_span()
             pattern = r"https://sheets\.googleapis\.com/v4/spreadsheets/[\w-]+"
             sanitized_message = re.sub(pattern, "REDACTED", str(ex))
             sanitized_message_b = sanitized_message.encode(encoding="utf-8")
-            span.set_status(
-                Status(status_code=StatusCode.ERROR, description=sanitized_message)
-            )
             raise HttpError(ex.resp, sanitized_message_b)
 
         # update validation rules (i.e. no validation rules) for out of schema columns, if any


schematic/manifest/generator.py
```diff
@@ -33,6 +33,9 @@
     extract_component_validation_rules,
 )
 from schematic.utils.validate_utils import rule_in_rule_list
+from googleapiclient.errors import HttpError
+from opentelemetry.trace import Status, StatusCode
+from opentelemetry.trace import get_current_span
```
andrewelamb, line 38: Are the open telemetry imports used somewhere?

schematic/manifest/generator.py
```diff
@@ -1367,16 +1370,20 @@ def set_dataframe_by_url(
...
         # authorize pygsheets to read from the given URL
         gc = ps.authorize(custom_credentials=self.creds)
 
         # open google sheets and extract first sheet
         # This sheet already contains headers.
         sh = gc.open_by_url(manifest_url)
         wb = sh[0]
-
-        wb.set_dataframe(manifest_df, (1, 1), fit=True)
+        try:
```
andrewelamb, line 1380: I think this is a perfect spot for a comment. I think Schematic tends to be over-commented, but this is somewhere I think warrants it. A comment explaining why this is needed would be great here.

andrewelamb:
Perhaps we should split the Jira issue out, with another to further investiagte doing this via OTEL/Signoz instead of the code?

schematic/manifest/generator.py
```diff
@@ -1367,16 +1370,20 @@ def set_dataframe_by_url(
...
+            wb.set_dataframe(manifest_df, (1, 1), fit=True)
+        except HttpError as ex:
+            pattern = r"https://sheets\.googleapis\.com/v4/spreadsheets/[\w-]+"
```
andrewelamb, line 1383: Will this catch all google sheet urls?

schematic/manifest/generator.py
linglp: I could remove them. I used them for experiment.

linglp:
@andrewelamb I'm not entirely sure if I'm on the right track, so I'd like to wait until @BryanFauble returns to confirm. This implementation removes Google Sheet links in spans, but I'm unsure if the ticket also requires a more general approach to stripping all sensitive information. I'll make further modifications as I discuss more with Bryan.

thomasyu888:
>  not entirely sure if I'm on the right track... This implementation removes Google Sheet links in spans, but I'm unsure if the ticket also requires a more general approach to stripping all sensitive information.

@linglp is this the only place googlesheets are logged? Off the top of your head, do you know of any other sensitive information that is logged?

@andrewelamb what are your thoughts? 

I'll add my personal views after this discussion.

schematic/manifest/generator.py
```diff
@@ -1367,16 +1368,21 @@ def set_dataframe_by_url(
...
 
         # open google sheets and extract first sheet
         # This sheet already contains headers.
         sh = gc.open_by_url(manifest_url)
         wb = sh[0]
-
-        wb.set_dataframe(manifest_df, (1, 1), fit=True)
+        try:
+            wb.set_dataframe(manifest_df, (1, 1), fit=True)
+        except HttpError as ex:
```
thomasyu888, line 1380: Pending further discussion of whether we think this is the right approach, this new exception that's being caught should have a unit test.

This is a good example of tiny design before doing the work would be helpful, that said, sometimes you have to do a little bit before knowing your options.

andrewelamb: Agreed.

andrewelamb:
@thomasyu888 I don't have anythign to add at the moment on data getting into Signoz that shouldn't(Except to agree that it's bad :)  ). This is somethign I'll keep in mind as I start working more cloesly with Signoz however.

schematic/manifest/generator.py
SageGJ: Also my question. Could a change to the google api, that would otherwise be non-breaking, change the format of the URL used so that it's no longer appropriately found in the message string?

schematic/manifest/generator.py
```diff
@@ -1367,16 +1368,21 @@ def set_dataframe_by_url(
...
         gc = ps.authorize(custom_credentials=self.creds)
 
         # open google sheets and extract first sheet
         # This sheet already contains headers.
         sh = gc.open_by_url(manifest_url)
         wb = sh[0]
-
-        wb.set_dataframe(manifest_df, (1, 1), fit=True)
+        try:
+            wb.set_dataframe(manifest_df, (1, 1), fit=True)
```
SageGJ, line 1379: Is this the only call that can trigger an error with the URL in it? I'm thinking of things like `sh = gc.open_by_url(manifest_url)` in `get_dataframe_by_url` or similar.

SageGJ:
What are y'all's thoughts about wrapping/modifying [sys.excepthook](https://docs.python.org/3/library/sys.html#sys.excepthook), catching `HttpError`s within, and sanitizing the messages there?

Something like*
```
import sys
def custom_except_hook(type, value, traceback):
  if type == HttpError:
    # message sanitizing
    # message raising
  else:
    sys.__excepthook__(type, value, traceback)
sys.excepthook = custom_except_hook
```

- would apply to multiple locations where an error is raised that includes a google sheets url
- could be extended to modify different error types that contain other sensitive information
- avoids having to wrap multiple blocks of code in `try: catch:` statements

@linglp @andrewelamb @thomasyu888 

*modified from [here](https://stackoverflow.com/questions/6598053/python-global-exception-handling/6598286#6598286)

schematic/manifest/generator.py
linglp: thanks, I think I added a comment.

linglp:
Thanks for all the discussion here. In retrospect, a design document could help clarify things further. To summarize, the Google Sheet link was originally found in SigNoz traces, and this solution specifically removes it from traces and spans within a single function call—but not logs. My plan was to confirm with Bryan whether removing sensitive information from traces is necessary, as the ticket only mentioned "logs," and whether handling it directly in the function (rather than using a custom trace processor) is an acceptable approach. I can also document what I’ve tried and why those approaches didn’t work separately.

Since the error originates from Google APIs, any part of the system that interacts with them could potentially trigger it. If Bryan confirms this approach is acceptable, I can proceed with adding unit tests and considering how to wrap the exception.

We will likely need to remove Google Sheet links from logs as well. I will keep exploring that part.

thomasyu888:
> What are y'all's thoughts about wrapping/modifying [sys.excepthook](https://docs.python.org/3/library/sys.html#sys.excepthook), catching `HttpError`s within, and sanitizing the messages there?

@GiaJordan here are my thoughts.  Using `sys.excepthook` to filter out sensitive Google API URLs from exception messages can work and is creative, but some considerations

1. Setting `sys.excepthook` changes how all unhandled exceptions are processed globally in one module. I wonder if it would have issues when it's run in a multi-threaded or multi-module environment.
2. This only catches uncaught exceptions. If the `HttpError` is handled somewhere else (e.g., inside a try-except block), this function won't intercept it.
3. There can be potential suppression of useful debug Info. This is a general issue, but we actually want these logs when this is run in the CLI/library, but we just don't want to send the information to SigNoz. @linglp .  For example, users of schematic CLI _should have_ these google sheet links returned to them.

schematic/__init__.py
```diff
@@ -95,6 +101,37 @@ def force_flush(self, timeout_millis: int = 30000) -> None:
...
 
 
+class CustomFilter(LogRecordProcessor):
+    """A custom log record processor that redacts sensitive data from log messages before they are exported."""
+
+    def __init__(self, exporter):
+        self._exporter = exporter
+        self._shutdown = False
+
+    def redact_google_sheet(self, message: str) -> str:
```
thomasyu888, line 111: It's important we keep the logs when people are running the CLI.  If I'm not mistaken, I think some CLI commands rely on the gsheets link being returned.

My guess is that this would fail some of the CLI tests

BryanFauble: I don't think this is a concern for CLI usage. The reason is that the log messages that are sent via OTEL is a (deep?) copy of the data sent to stdout/console. 

If I understand this solution it will only affect the messages making their way into SigNoz, but leave them as is in something like cloudwatch. Which, is probably fine since if someone had access to cloud watch, we're already in serious trouble. But, access to SigNoz is going to be given more freely, so we want to curate it carefully.

However, it should be tested to verify these assumption with the cli usage

schematic/manifest/generator.py
BryanFauble: Instead of an exception being caught in this way I would rather not have us modify any application code to implement a solution here.

I think the idea of using a log or span processor in the OTEL Python SDK is the appropriate solution.

The reason why I say this is: 
Using the OTEL SDK gives us one spot where all redaction logic lives, we do not need to hunt around the code base to find all the spots we need to modify. It also gives us the template to follow and apply to other projects where we want to implement similar functionality.

Additional thoughts @andrewelamb @SageGJ ?

schematic/manifest/generator.py
BryanFauble: A change to the format of the URL would break this regex pattern. The regex could be updated to look for any `/v.` instead of just `/v4`. This is the only thing I think we could reasonable anticipate that may change, and we can resolve that potential issue now ahead of time.

BryanFauble:
Thanks for the discussion everyone (@linglp @andrewelamb @thomasyu888 @linglp )

https://github.com/Sage-Bionetworks/schematic/pull/1573#discussion_r1953137636

Captures my thoughts. I would rather we utilize the OTEL Python SDK to handle the sanitization of the data, rather than the traditional native python ways. By implementing the logic via processors in the Python SDK it will work similar to the concept of how it would be implemented within the OpenTelemtry Collector shown here: https://opentelemetry.io/docs/collector/architecture/#pipelines

Specifically it has the concept of using one or more processors to handle data transformation before that data is exported. Since we are developing code locally without a collector that may always be present, It's important to be able to filter this data out before it leaves the machine where the data is produced. Technically when this code is running in AWS we could use the collector approach, but i wanted us to figure out if we could do it in code before the collector (Lingling is proving that we can in this pull request!)

schematic/__init__.py
```diff
@@ -46,6 +52,107 @@
...
+        )
+        return redacted_span
+
+    def on_end(self, span: ReadableSpan) -> None:
+        """Remove sensitive information in spans by creating new ones with redacted data."""
+        # maybe the goal should be to filter all spans?
+        if span.status.status_code == trace.StatusCode.ERROR:
+            if span.events:
+                redacted_span = self._create_redacted_span(span)
+                self.export.export([redacted_span])
```
BryanFauble, line 145:
```suggestion
        if span.status.status_code == trace.StatusCode.ERROR:
            if span.events:
                redacted_span = self._create_redacted_span(span)
                self.export.export([redacted_span])
                return
        self.export.export([span])
```

BryanFauble: Is this needed? I'm not sure tbh

BryanFauble: I don't think this is going to work with how we have this set up because of:
https://github.com/open-telemetry/opentelemetry-python/blob/a7fe4f8bac7fa36291c6acf86982bbb356e3ae6d/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L173-L175

**Specifically:**

1. We have a `BatchSpanProcessor` that was already responsible for queuing up and exporting the spans to SigNoz
2. Based on the code at the link above the span is sent to each processor, in this case the data is likely being sent twice. Once in the `BatchSpanProcessor` that is still being called, and once here.

BryanFauble: Here is a hack how we can get around this issue:

1. Instead of using a processor to handle this logic, as we can see, won't work we could monkey patch the `_readable_span` function call: https://github.com/open-telemetry/opentelemetry-python/blob/a7fe4f8bac7fa36291c6acf86982bbb356e3ae6d/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L906-L921
2. In the monkey patch we: 1) Start by calling our sensitive data redaction process to strip data out of the span, 2) Return a call to the original function.

By monkey patching this, it would allow us to essentially "slip" in logic before a read-only span has been created.

Let me know what questions you have.

schematic/__init__.py
```diff
@@ -46,6 +58,59 @@
 USER_AGENT |= USER_AGENT_LIBRARY
 
 
+class FilterSensitiveData:
```
BryanFauble, line 61: I'm not sure that this makes sense to be a class since there isn't any state that needs to be tracked. These should probably be functions so you don't need to create an instance of the class like you had done here: `FilterSensitiveData().redacted_sensitive_data_in_exception(attributes)`

linglp: Okay that makes sense. Thanks for the suggestion. I was initially thinking of making it a class so that I could group the functions together. But you are right since I don't need to track any states, a class might not be necessary. I think I can create a separate file called remove_sensitive_data_utils.py and save those functions there.

schematic/__init__.py
```diff
@@ -159,11 +207,48 @@ def set_up_tracing(session: requests.Session) -> None:
...
+    # remove sensitive information in attributes
+    # this avoid exception trace and messages contain sensitive information
+    for event in self._events:
+        attributes = event.attributes
+        redacted_event_attributes = redacted_sensitive_data_in_exception(attributes)
+        redacted_event = Event(
+            name=redact_string(event.name),
+            attributes=redacted_event_attributes,
+            timestamp=event.timestamp,
+        )
```
BryanFauble, line 245: Instead of creating and assigning a new `Event` is it possible to instead set the `name` attribute on `event` instead?

```suggestion
        event._name = redact_string(event.name)
```

The reason for this suggestion is that if there happens to be other fields that get added to the `Event`, this code is now incorrect since data would be dropped at this point in time. By modifying only the fields we want, then we have a lesser burden of maintenance and risk of breaking our code as the library is updated.

schematic/__init__.py
```diff
@@ -159,11 +207,48 @@ def set_up_tracing(session: requests.Session) -> None:
...
+
+    Returns:
+        ReadableSpan: a new readable span that redacts sensitive data
+    """
+    redacted_events = []
+    # remove sensitive information in span status description
+    # this avoid statusMessage on signoz contains sensitive information
+    if self._status.status_code == trace.StatusCode.ERROR:
+        status_description_redacted = redact_string(str(self._status.description))
+        self._status = Status(self.status.status_code, status_description_redacted)
```
BryanFauble, line 234: Same comment from `Event` applied here. Can we set `_description` on `self._status` instead of creating a new `Status` class instance.

schematic/__init__.py
linglp: Got it. Thanks for the suggestion. Let me give that a try.

schematic/__init__.py
linglp: Yeah it worked! Thanks.

schematic/manifest/generator.py
linglp: Thanks Bryan! I updated the regex pattern and the test so that different versions of google sheet can be caught.

BryanFauble (approved):
LGTM! Thanks for the changes

schematic/__init__.py
linglp: I can't find in the documentation about "deep copy", but my understanding is that open telemetry captures the logs and traces as the application runs independently. Since all the tests passed on Github and locally, I will mark it as "resolved".

PR #1580, opened 2025-03-13
[SCHEMATIC-263] Escaping of certain characters in like statement of SynpaseStore.getDatasetFiles method

andrewelamb:
[SCHEMATIC-263]
# **Problem:**

When dataset folders contained a single quote, the `SynpaseStore.getDatasetFiles` method failed because the single quote broke the SQL statement-like clause. 

# **Solution:**

Single quotes and other special characters are now escaped in the SQL statement

# **Testing:**

Unit tests for the new functions that create the SQL statement and do the escaping
Integration tests in Synapse with folders that contain the characters.

[SCHEMATIC-263]: https://sagebionetworks.jira.com/browse/SCHEMATIC-263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Depends on #1582


thomasyu888 (commented):
Thanks for this, I'll defer to the rest of the team to review first, but I'd like to review after they do

schematic/utils/general.py
```diff
@@ -313,3 +315,68 @@ def normalize_path(path: str, parent_folder: str) -> str:
...
+    Args:
+        synapse_path (str): The synapse_path that needs to be escaped
+
+    Returns:
+        str: The like pattern with problematic characters escaped
+    """
+    pattern = re.sub("'", "''", synapse_path)
+    for char in LIKE_PATTERN_SPECIAL_CHARS:
+        pattern = re.sub(char, f"|{char}", pattern)
+    return pattern
```
BryanFauble, line 382: Is there a need to use `re.sub`/regular expressions here? Aren't these simple string replacements like:
```
import re

LIKE_PATTERN_SPECIAL_CHARS = ["%", "_"]

synapse_path = "Chi''ldren's H_ospit____al_"

# Original code:
pattern = re.sub("'", "''", synapse_path)
for char in LIKE_PATTERN_SPECIAL_CHARS:
    pattern = re.sub(char, f"|{char}", pattern)

print(pattern)

# Modification
modified_path = synapse_path.replace("'", "''")
for char in LIKE_PATTERN_SPECIAL_CHARS:
    modified_path = modified_path.replace(char, f"|{char}")

print(modified_path)

```

thomasyu888: Can we try `urrllib.quote` or parse?  This may negate some of this stuff

andrewelamb: @BryanFauble No particular reason. I've made that change.

schematic/utils/general.py
```diff
@@ -313,3 +314,68 @@ def normalize_path(path: str, parent_folder: str) -> str:
...
+      documentation shows.
+
+
+    Args:
+        synapse_path (str): The synapse_path that needs to be escaped
+
+    Returns:
+        str: The like pattern with problematic characters escaped
+    """
+    pattern = synapse_path.replace("'", "''")
```
thomasyu888, line 378: According to API docs: Names may only contain: letters, numbers, spaces, underscores, hyphens, periods, plus signs, apostrophes, and parentheses

@andrewelamb can you check which of those would be problematic as well?  Thanks so much!

andrewelamb: @thomasyu888 At least according to this [documentation](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/web/controller/TableExamples.html) and also my research only `%` and `_` characters are used as special characters. So that's what those are included.  Apostrphes are also escaped because they mess up the quoting.

schematic/utils/general.py
andrewelamb: @thomasyu888 I'll take a look at those and see if they work.

schematic/utils/general.py
andrewelamb: @thomasyu888 Just to be sure I added a test for this folder:

thomasyu888 (approved):
🔥 Thanks for taking the time to check other characters!

BryanFauble (approved):

schematic/utils/general.py
```diff
@@ -313,3 +314,68 @@ def normalize_path(path: str, parent_folder: str) -> str:
...
+    Args:
+        synapse_path (str): The input synapse path to be made into a like statement
+
+    Raises:
+        ValueError: If the input path contains a '|' character
+
+    Returns:
+        str: A SQL like statement
+    """
+    if "|" in synapse_path:
```
linglp, line 345: I could roughly follow the logic here, but can we add some comments in this section to make the logic clearer?

PR #1582, opened 2025-03-14
[SCHEMATIC-266] Resolve errors new in mypy `v1.15.0`

SageGJ:
# Problem
Mypy `v1.15.0` was released which now checks additional packages for type hinting due to new type packages being installed, particularly related to the schemas module. These types were not enforced because they were unable to be checked until recently. With the new developments additional errors are being raised which prevents successful completion of our CI/CD test workflows.

# Solution
Type hinting was corrected where appropriate. These corrections resolved 7 of the 16 errors.
Some errors were modulated due to the type hinting of external libraries (discussed below).
our mypy dependency was also updated to require `1.15.0` as the minimum acceptable version.

## `pkg_resources`
The error from this package was suppressed due to the limited use of the library in the codebase and the inability for mypy to locate stubs for the package despite them being installed. This suppressed 1 of the 16 errors.
## `networkx`
`networkx` is not type hinted, now that a package exists for checking the types of these objects we're getting errors. mypy's behavior has been overridden to follow the untyped imports and attempt to analyze our use of these objects as best it can as described in the [mypy documentation.](https://mypy.readthedocs.io/en/stable/config_file.html#confval-follow_untyped_imports). This resolved 8 of the 16 errors.

# Testing
Resolution was tested by running the mypy command from the workflow file locally and by running the workflow in GitHub actions.


schematic/schemas/data_model_graph.py
```diff
@@ -819,6 +822,7 @@ def get_subgraph_by_edge_type(self, relationship: str) -> nx.DiGraph:
             if key == relationship:
                 rel_edges.append((node_1, node_2))
 
+        relationship_subgraph: nx.DiGraph
         relationship_subgraph = nx.DiGraph()
         relationship_subgraph.add_edges_from(rel_edges)
```
andrewelamb, line 827: Just a nitpick, but you should be able to do this as a one-liner:

`relationship_subgraph: nx.DiGraph = nx.DiGraph()`

thomasyu888 (approved):
🔥 Awesome work here. Feel free to dismiss the security alert as that's your other ticket.

Will pre-approve pending minor mixes!

BryanFauble (approved):

schematic/schemas/data_model_graph.py
SageGJ: @andrewelamb yeah when I looked at the documentation it listed both methods as valid ways to type hint. I opted for doing them on separate lines to avoid having too much on one line within the method itself. I've updated them to be combined like you suggested.

PR #1583, opened 2025-03-17
[schematic-252] Added documentation for manifest generation 

linglp:
# Problem 
Not enough documentation for manifest generation. The interface wasn't user friendly enough 

# Solution 
* updated documentation include instructions on generating a manifest using the CLI and the APIs
* covered most of the use cases

# Test
View the latest doc here: https://schematicpy.readthedocs.io/en/schematic-252_a/manifest_generation.html#

PR #1584, opened 2025-03-17
[schematic-254] added documentation for manifest submission 

linglp:
# Problem 
Not enough documentation for manifest submission. The interface wasn't user friendly enough 

# Solution 
* updated documentation include instructions on submitting a manifest using the CLI and the APIs
* covered most of the use cases

# Test
View the latest doc here: https://schematicpy.readthedocs.io/en/schematic-254/manifest_submission.html#

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -3,6 +3,8 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
+.. _index:
+
 Welcome to Schematic's documentation!
 =====================================
 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -148,3 +148,4 @@ For the entire Python API reference documentation, you can visit the docs here: 
    tutorials
    troubleshooting
    cli_reference
+   manifest_submission
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -148,6 +148,6 @@ For the entire Python API reference documentation, you can visit the docs here: 
    asset_store
    configuration
    tutorials
+   manifest_submission
    troubleshooting
    cli_reference
-   manifest_submission
--- a/docs/source/manifest_submission.rst
+++ b/docs/source/manifest_submission.rst
@@ -168,9 +168,30 @@ Option 2: Use the API
 
 
 
-Use project_scope parameter to expedite submission process
+Expedite submission process (Optional)
 -----------------------------------------------------------
 
 If your asset view contains multiple projects, it might take some time for the submission to finish.
 
 You could expedite the submission process by specifying the project_scope parameter. This parameter allows you to specify the project(s) that you want to submit the manifest to.
+
+To utilize this parameter, make sure that the projects listed there are part of the asset view.
+
+
+Option 1: Use the CLI
+~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: bash
+
+    schematic model -c /path/to/config.yml submit -mp <your csv manifest path> -d <your synapse top level folder id> -vc <your data type> -mrt file_only -no-fa -tcn "class_label" -ps "project_id1, project_id2"
+
+- **-ps**: Specifies the project scope as a comma separated list of project IDs.
+
+
+Option 2: Use the API
+~~~~~~~~~~~~~~~~~~~~~~
+
+1. Locate the **model/submit** endpoint in the **Swagger UI**.
+2. Click **"Try it out"** to enable input fields.
+3. Follow instructions to Submit a Manifest file
+3. Locate "project_scope" parameter and click on **Add string items** to add project IDs.
--- a/docs/source/manifest_submission.rst
+++ b/docs/source/manifest_submission.rst
@@ -12,10 +12,10 @@ Ensure you have a Synapse account and have obtained your Synapse credential by f
 
 - **Install and Configure Schematic**:
   Ensure you have installed `schematic` and set up its dependencies.
-  Refer to the **"Installation Guide for Users"** for detailed instructions.
+  See the :ref:`installation` section for more details.
 
 - **Understand Important Concepts**:
-  Familiarize yourself with key concepts outlined on the :doc:`index.rst` page.
+  Understand Important Concepts: Familiarize yourself with key concepts outlined on the :ref:`index` of the documentation.
 
 - **Configuration File**:
   Learn more about each attribute in the configuration file by referring to the relevant documentation.
--- a/docs/source/manifest_submission.rst
+++ b/docs/source/manifest_submission.rst
@@ -231,7 +231,7 @@ Pre-requisite
 
 1. Ensure that all your manifests, including both the initial manifests and those containing rows to be upserted, include a primary key: <YourComponentName_id>. For example, if your component name is "Patient", the primary key should be "Patient_id".
 2. If you plan to use upsert in the future, select the upsert option during the initial table uploads.
-3. Currently it is required to use -dl/--use_display_label with table upserts.
+3. Currently it is required to use -tcn "display_label" with table upserts.
 
 
 Option 1: Use the CLI
--- a/docs/source/manifest_submission.rst
+++ b/docs/source/manifest_submission.rst
@@ -105,8 +105,7 @@ Option 2: Use the API
 2. Click **"Try it out"** to enable input fields.
 3. Enter the required parameters and execute the request:
 
-   - **schema_url**: The raw URL of your data model.
-     - If your data model is hosted on **GitHub**, use the following formats:
+   - **schema_url**: The raw URL of your data model. If your data model is hosted on **GitHub**, use the following formats:
        - JSON-LD: `https://raw.githubusercontent.com/<your-repo-path>/data-model.jsonld`
        - CSV: `https://raw.githubusercontent.com/<your-repo-path>/data-model.csv`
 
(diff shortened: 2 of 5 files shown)


docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,274 @@
...
+    `<https://python-docs.synapse.org/tutorials/authentication/#prerequisites>`_
+
+**Before Using the Schematic CLI**
+
+- **Install and Configure Schematic**:
+  Ensure you have installed `schematic` and set up its dependencies.
+  Refer to the **"Installation Guide for Users"** for detailed instructions.
+
+- **Understand Important Concepts**:
+  Familiarize yourself with key concepts outlined on the **home page** of the documentation.
```
thomasyu888, line 18: These could actually reference those pages

thomasyu888:
@linglp you may want to switch these to not "draft" PRs if they are ready for review

docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,274 @@
...
+
+**Obtain Synapse Credentials**:
+Ensure you have a Synapse account and have obtained your Synapse credential by following the instructions here:
+    `<https://python-docs.synapse.org/tutorials/authentication/#prerequisites>`_
+
+**Before Using the Schematic CLI**
+
+- **Install and Configure Schematic**:
+  Ensure you have installed `schematic` and set up its dependencies.
+  Refer to the **"Installation Guide for Users"** for detailed instructions.
```
thomasyu888, line 15: You can link to these

docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,274 @@
+Submit a manifest to Synapse
+============================
+
+Prerequisites
+-------------
+
+**Obtain Synapse Credentials**:
```
thomasyu888, line 7: You can probably refer to the installation guide instead of having all of this

docs/source/index.rst
```diff
@@ -148,3 +148,4 @@ For the entire Python API reference documentation, you can visit the docs here:
    tutorials
    troubleshooting
    cli_reference
+   manifest_submission
```
thomasyu888, line 151: nit: Shift this to be above troubleshooting

thomasyu888 (approved):
🔥 LGTM, I'll defer to the rest of the team for final review

docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,274 @@
...
+
+- **Understand Important Concepts**:
+  Familiarize yourself with key concepts outlined on the :doc:`index.rst` page.
+
+- **Configuration File**:
+  Learn more about each attribute in the configuration file by referring to the relevant documentation.
+
+- **Obtain a manifest**:
+  Please obtain a manifest by following the documentation of generating a manifest.
+
```
andrewelamb, line 25: Should we link to the generating a manifest doc here?

docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,274 @@
...
+- **Install and Configure Schematic**:
+  Ensure you have installed `schematic` and set up its dependencies.
+  Refer to the **"Installation Guide for Users"** for detailed instructions.
+
+- **Understand Important Concepts**:
+  Familiarize yourself with key concepts outlined on the :doc:`index.rst` page.
+
+- **Configuration File**:
+  Learn more about each attribute in the configuration file by referring to the relevant documentation.
+
```
andrewelamb, line 22: Is there somewhere we can link to for this?

docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,274 @@
...
+1. Locate the **model/submit** endpoint in the **Swagger UI**.
+2. Click **"Try it out"** to enable input fields.
+3. Enter the required parameters and execute the request:
+
+   - **schema_url**: The URL of your data model.
+     - If your data model is hosted on **GitHub**, the URL should follow this format:
+       - JSON-LD: `https://raw.githubusercontent.com/<your-repo-path>/data-model.jsonld`
+       - CSV: `https://raw.githubusercontent.com/<your-repo-path>/data-model.csv`
+
```
andrewelamb, line 151: I wonder if we should have a section somewhere that describes the data model options, and then link to it from here.

docs/source/manifest_submission.rst
linglp: there is.. but it is hard to do it in this PR because the page of manifest generation does not exist. I will commit to Tom's branch once both PR get merged

docs/source/index.rst
linglp: makes sense. Thanks for pointing that out.

docs/source/manifest_submission.rst
linglp: Yeah, I linked the installation guide.

docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,297 @@
...
+1. Locate the **model/submit** endpoint in the **Swagger UI**.
+2. Click **"Try it out"** to enable input fields.
+3. Enter the required parameters and execute the request:
+
+   - **schema_url**: The URL of your data model.
+     - If your data model is hosted on **GitHub**, use the following formats:
+       - JSON-LD: `https://raw.githubusercontent.com/<your-repo-path>/data-model.jsonld`
+       - CSV: `https://raw.githubusercontent.com/<your-repo-path>/data-model.csv`
+
+   - **data_type**: Specify the data type or schema model for your manifest (e.g., `"Patient"`, `"Biospecimen"`). To skip validation, remove the default inputs.
```
SageGJ, line 108: `Data_type` is only used for validation yes?
I would change this description to
```
Specify the data type to validate your manifest against (e.g., `"Patient"`, `"Biospecimen"`). To skip validation, exclude this parameter.
```

docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,297 @@
...
+
+Option 2: Use the API
+~~~~~~~~~~~~~~~~~~~~~~
+
+.. note::
+
+    During submission, validation is optional. If you have finished validation in previous step, you could skip validation by removing the default inputs.
+
+
+1. Locate the **model/submit** endpoint in the **Swagger UI**.
```
SageGJ: Can we link to this endpoint?

docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,297 @@
...
+2. Click **"Try it out"** to enable input fields.
+3. Enter the required parameters and execute the request:
+
+   - **schema_url**: The URL of your data model.
```
SageGJ, line 103: Why do data models need to follow this format?
In this case I think it's important to call out that that schematic needs the link to the raw version of the file that can be used/read/downloaded directly. Then if a user had a model not on github they'd still have enough information from the documentation to apply it to their use case.

SageGJ:
I think this is a good improvement on our documentation! It's good that the cases of using the cli and the api are both covered. 

For the api however I feel like there's an assumption being made that all users of the api will be going through swaggerUI. I'm thinking of the case where users may want to interact with our api programmatically. For this purpose I think it would be useful to include a description of all the endpoint parameters instead of just the required ones so that the documentation here is complete and they won't also have to go through the UI. Another way this might change is assuming the presence of default values and instead saying what each value is and which ones may be omitted in certain cases. I think this would apply to #1583 as well.

I realize there's a balance with how much we include here and also that we won't want to duplicate everything multiple times but I'm curious what your thoughts on this are? @linglp @thomasyu888

docs/source/manifest_submission.rst
linglp: I have added the link

docs/source/manifest_submission.rst
linglp: link was added.

docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,297 @@
...
+- **-no-fa**: Skips the file annotations upload.
+- **-tcn**: Table Column Names: This is optional, and the available options are "class_label", "display_label", and "display_name". The default is "class_label", but you can change it based on your requirements.
+
+
+Option 2: Use the API
+~~~~~~~~~~~~~~~~~~~~~~
+
+.. note::
+
+    During submission, validation is optional. If you have finished validation in previous step, you could skip validation by removing the default inputs.
```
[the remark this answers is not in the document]
linglp: Okay done.

docs/source/manifest_submission.rst
```diff
@@ -0,0 +1,297 @@
...
+Submit a Manifest File to Synapse
+---------------------------------
+
+.. note::
+
+  You can configure the format of the manifest being submitted by using the `-mrt flag` in the CLI or the `manifest_record_type` in the API.
+
+  For table column names, here's a brief explanation of all the options:
+   - display_name: use raw display name defined in the data model as the column name, no modifications to the name will be made.
+   - display_label: use the display name formatting as the column name. Will strip blacklisted characters (including spaces) when present.
```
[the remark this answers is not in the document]
linglp: Yes. I have added the blacklisted characters in the doc.

docs/source/manifest_submission.rst
linglp: Yes, links were added.

docs/source/manifest_submission.rst
thomasyu888: I think we may have upcoming work to add in the documentation for the different data models.

PR #1585, opened 2025-03-17
update tool for sonarcloud scan

SageGJ:
Resolves [DPE-1270](https://sagebionetworks.jira.com/browse/DPE-1270)

# Problem
Test workflow uses a deprecated tool for sonarcloud scanning.

# Solution
The deprecated tool is used by a tool we use directly, that tool has been updated to the latest version

# Testing
The workflow now completes successfully

[DPE-1270]: https://sagebionetworks.jira.com/browse/DPE-1270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -199,7 +199,7 @@ jobs:
         with:
           files: "coverage.xml"
       - name: SonarCloud Scan
-        uses: SonarSource/sonarqube-scan-action@v4.1.0
+        uses: SonarSource/sonarqube-scan-action@v5.0.0
         if: ${{ always() }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/test.yml
```diff
@@ -199,7 +199,7 @@ jobs:
         with:
```
thomasyu888, line 199: This action will fail by April 1 because it's usage of ubuntu-20.04, we could create another ticket to bump this, or wait and see if this fix will work, then do a bump

SageGJ: @thomasyu888 I think we should create another ticket and bump separately

thomasyu888: Can you merge this in and change that ticket to update ubuntu?

thomasyu888 (approved):
🔥 LGTM.

PR #1592, opened 2025-04-09
[Schematic 272] Standardize Id and entityId before manifest submission

linglp:
# **Problem:**
## Short description of the problem 
When submitting the manifest, if the "id" or "entityId" column is present with a non-standard casing (e.g., "ID", "iD", "entityid"), Schematic accepts the column as-is without normalizing the name. This later led to an issue when updating the table in future manifest submission. 

## More context
The error was: 
```
  File "/app/app/./schematic/store/synapse.py", line 3380, in updateTable
    self.tableToLoad = update_df(existing_table, self.tableToLoad, update_col)
  File "/app/app/./schematic/utils/df_utils.py", line 259, in update_df
    assert index_col in input_df, f"`input_df` lacks `{index_col}` column."
AssertionError: `input_df` lacks `Id` column.
```

The error occurred when the user tried to submit a manifest (with "id" column in lower case) to synapse using the NF data model: https://raw.githubusercontent.com/nf-osi/nf-metadata-dictionary/main/NF.jsonld

Upon further checking, for some reason, the submitted manifest retained the lowercase column name “id” instead of being updated to the capitalized “Id”.  (see version 7 [here](https://www.synapse.org/Synapse:syn65597609.7)) As a result, the table created from the submission also used “id”. In a later step, when updating the table, Schematic assumed the column name to be “Id” (see code reference [here](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/store/synapse.py#L3362)), which prevented it from recognizing and updating the existing “id” column in the table.

# **Solution:**

## Attempts to reproduce the issue
* I attempted to replicate the issue using my own test project.
* Upon generating a manifest through Schematic, I confirmed that all expected columns. Column “id” (any case variation) did not appear. 
* I then submitted this generated manifest back to Synapse using option `table_and_file`.
* As expected, annotations were applied to files, and the "Id" column was correctly added to the annotations and to the submitted manifest file. 
* When I generated a manifest again after submission, as expected, I could see “eTag” and “Id” column. But if the manifest submitted before have column “id” (lower case), the manifest that gets generated again would have column “id”. 

## Uncertainty 
Given that everything worked as expected in my test environment, it's unclear why "id" (lowercase) appeared in the originally submitted manifest that triggered the error in the first place.  I am not sure if any of the following occurred: 

* Manual alteration?
   * Did the user manually change "Id" to "id" in the manifest before submission?
* Pre-existing manual annotations?
    * Were any annotations manually added to the dataset files before the first manifest was generated?
    * If "Id" was manually added as an annotation key before the manifest was created, and use_annotations was enabled during manifest generation, Schematic would attempt to map the existing annotation keys to the data model’s display names—resulting in the column "id" appearing in the generated manifest due to "id" being included in the NF data model. 

## Solution 
Since the source of the "id" column is unclear, I decided to normalize all variations of "id" to "Id" when adding "id" column during submission. Given that the "entityId" column could present similar casing issues, I applied the same normalization to it as well.

# **Testing:**
* added unit test to ensure column normalization is correct.

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -2117,7 +2117,7 @@ class SynapseStorage(BaseStorage):
         for col in manifest.columns:
             if col.lower() == "id":
                 manifest = manifest.rename(columns={col: "Id"})
-            elif col.lower() == "entityid":
+            if col.lower() == "entityid":
                 manifest = manifest.rename(columns={col: "entityId"})
 
         # If 'Id' still doesn't exist, see if uuid column exists
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -2106,12 +2106,24 @@ class SynapseStorage(BaseStorage):
 
     def _add_id_columns_to_manifest(
         self, manifest: pd.DataFrame, dmge: DataModelGraphExplorer
-    ):
-        """Helper function to add id and entityId columns to the manifest if they do not already exist, Fill id values per row.
+    ) -> pd.DataFrame:
+        """
+        Ensures that the manifest DataFrame has standardized 'Id' and 'entityId' columns.
+
+        If any case variation of the 'id' column is present (e.g., 'id', 'ID', 'iD'), it is renamed to 'Id'.
+        If any case variation of the 'entityid' column is present, it is renamed to 'entityId'.
+        If 'Id' is still missing, it will be created as an empty column or derived from a 'Uuid' column,
+        depending on whether 'Uuid' is defined in the schema.
+        Missing values in the 'Id' column are filled with generated UUIDs.
+        If 'entityId' is still missing, it will be created and filled with empty strings.
+        If it is already present, any missing values will be replaced with empty strings.
+
         Args:
-        Returns (pd.DataFrame):
-            Manifest df with new Id and EntityId columns (and UUID values) if they were not already present.
-            If any case variation of "id" column is present, rename it to "Id".
+            manifest (pd.DataFrame): The metadata manifest to be updated.
+            dmge (DataModelGraphExplorer): data moodel graph explorer object
+
+        Returns:
+            pd.DataFrame: The updated manifest with a standardized 'Id' column and an 'entityId' column.
         """
         # Normalize any variation of 'id' to 'Id', "entityid" to "entityId"
         for col in manifest.columns:
@@ -2147,7 +2159,7 @@ class SynapseStorage(BaseStorage):
 
         # Add entityId as a column if not already there or
         # Fill any blanks with an empty string.
-        if not col_in_dataframe("entityId", manifest):
+        if "entityId" not in manifest:
             manifest["entityId"] = ""
         else:
             manifest["entityId"] = manifest["entityId"].fillna("")
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -1568,7 +1568,6 @@ class SynapseStorage(BaseStorage):
         existing_table_id = self.syn.findEntityId(
             name=table_name, parent=table_parent_id
         )
-
         tableOps = TableOperations(
             synStore=self,
             tableToLoad=table_manifest,
@@ -2000,7 +1999,6 @@ class SynapseStorage(BaseStorage):
         # Add metadata to the annotations
         for annos_k, annos_v in metadata.items():
             annos[annos_k] = annos_v
-
         return annos
 
     '''
@@ -2111,13 +2109,20 @@ class SynapseStorage(BaseStorage):
     ):
         """Helper function to add id and entityId columns to the manifest if they do not already exist, Fill id values per row.
         Args:
-            Manifest loaded as a pd.Dataframe
         Returns (pd.DataFrame):
             Manifest df with new Id and EntityId columns (and UUID values) if they were not already present.
+            If any case variation of "id" column is present, rename it to "Id".
         """
+        # Normalize any variation of 'id' to 'Id', "entityid" to "entityId"
+        for col in manifest.columns:
+            if col.lower() == "id":
+                manifest = manifest.rename(columns={col: "Id"})
+            elif col.lower() == "entityid":
+                manifest = manifest.rename(columns={col: "entityId"})
 
-        # Add Id for table updates and fill.
-        if not col_in_dataframe("Id", manifest):
+        # If 'Id' still doesn't exist, see if uuid column exists
+        # Rename uuid column to "Id" column
+        if "Id" not in manifest.columns:
             # See if schema has `Uuid` column specified
             try:
                 uuid_col_in_schema = dmge.is_class_in_schema(
@@ -2128,27 +2133,24 @@ class SynapseStorage(BaseStorage):
 
             # Rename `Uuid` column if it wasn't specified in the schema
             if col_in_dataframe("Uuid", manifest) and not uuid_col_in_schema:
-                manifest.rename(columns={"Uuid": "Id"}, inplace=True)
+                manifest = manifest.rename(columns={"Uuid": "Id"})
             # If no `Uuid` column exists or it is specified in the schema, create a new `Id` column
             else:
                 manifest["Id"] = ""
 
-        # Retrieve the ID column name (id, Id and ID) are treated the same.
-        id_col_name = [col for col in manifest.columns if col.lower() == "id"][0]
-
-        # Check if values have been added to the Id coulumn, if not add a UUID so value in the row is not blank.
+        # Fill in UUIDs in the "Id" column if missing
         for idx, row in manifest.iterrows():
-            if not row[id_col_name]:
+            if not row["Id"]:
                 gen_uuid = str(uuid.uuid4())
-                row[id_col_name] = gen_uuid
-                manifest.loc[idx, id_col_name] = gen_uuid
+                row["Id"] = gen_uuid
+                manifest.loc[idx, "Id"] = gen_uuid
 
-        # add entityId as a column if not already there or
-        # fill any blanks with an empty string.
+        # Add entityId as a column if not already there or
+        # Fill any blanks with an empty string.
         if not col_in_dataframe("entityId", manifest):
             manifest["entityId"] = ""
         else:
-            manifest["entityId"].fillna("", inplace=True)
+            manifest["entityId"] = manifest["entityId"].fillna("")
 
         return manifest
 
@@ -3124,6 +3126,7 @@ class TableOperations:
         )
 
         current_columns = self.synStore.syn.getTableColumns(current_table)
+
         for col in current_columns:
             current_table.removeColumn(col)
 
@@ -3170,6 +3173,7 @@ class TableOperations:
             # adds new columns to schema
             for col in cols:
                 current_table.addColumn(col)
+
             table_result = self.synStore.syn.store(
                 current_table, isRestricted=self.restrict
             )
--- a/tests/unit/test_store_synapse.py
+++ b/tests/unit/test_store_synapse.py
@@ -94,6 +94,11 @@ class TestStoreSynapse:
                 pd.DataFrame({"Id": ["test_value"], "EntityId": ["test_value"]}),
                 pd.DataFrame({"Id": ["test_value"], "entityId": ["test_value"]}),
             ),
+            # Case 5: 'EntityId' and "ID" in wrong mixed case — should be renamed to "Id" and 'entityId'
+            (
+                pd.DataFrame({"iD": ["test_value"], "EntityID": ["test_value"]}),
+                pd.DataFrame({"Id": ["test_value"], "entityId": ["test_value"]}),
+            ),
         ],
     )
     @pytest.mark.parametrize(
--- a/tests/unit/test_store_synapse.py
+++ b/tests/unit/test_store_synapse.py
@@ -94,9 +94,9 @@ class TestStoreSynapse:
                 pd.DataFrame({"Id": ["test_value"], "EntityId": ["test_value"]}),
                 pd.DataFrame({"Id": ["test_value"], "entityId": ["test_value"]}),
             ),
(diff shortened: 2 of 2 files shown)


schematic/store/synapse.py
```diff
@@ -2108,16 +2106,35 @@ def _read_manifest(self, metadataManifestPath: str) -> pd.DataFrame:
...
+
+        If any case variation of the 'id' column is present (e.g., 'id', 'ID', 'iD'), it is renamed to 'Id'.
+        If any case variation of the 'entityid' column is present, it is renamed to 'entityId'.
+        If 'Id' is still missing, it will be created as an empty column or derived from a 'Uuid' column,
+        depending on whether 'Uuid' is defined in the schema.
+        Missing values in the 'Id' column are filled with generated UUIDs.
+        If 'entityId' is still missing, it will be created and filled with empty strings.
+        If it is already present, any missing values will be replaced with empty strings.
 
-        # Add Id for table updates and fill.
```
linglp, line 2119: I changed this to` if "Id" not in manifest.columns:` because at this point, the column should have been normalized.

schematic/store/synapse.py
```diff
@@ -2128,27 +2145,24 @@ def _add_id_columns_to_manifest(
...
+            if not row["Id"]:
                 gen_uuid = str(uuid.uuid4())
-                row[id_col_name] = gen_uuid
-                manifest.loc[idx, id_col_name] = gen_uuid
+                row["Id"] = gen_uuid
+                manifest.loc[idx, "Id"] = gen_uuid
 
-        # add entityId as a column if not already there or
-        # fill any blanks with an empty string.
-        if not col_in_dataframe("entityId", manifest):
```
linglp, line 2148: similarly it here, I changed this to `if "entityId" not in manifest:` because at this point, the column should have been normalized.

schematic/store/synapse.py
```diff
@@ -2128,27 +2145,24 @@ def _add_id_columns_to_manifest(
...
+        # Add entityId as a column if not already there or
+        # Fill any blanks with an empty string.
+        if "entityId" not in manifest:
             manifest["entityId"] = ""
         else:
-            manifest["entityId"].fillna("", inplace=True)
```
linglp, line 2151: address the deprecation warning: 
```
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[JSONLD-manifest_dataframe0-expected0]
  /Users/lpeng/Documents/schematic-git2/schematic/schematic/store/synapse.py:2151: FutureWarning: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.
  The behavior will change in pandas 3.0. This inplace method will never work because the intermediate object on which we are setting values always behaves as a copy.
  
  For example, when doing 'df[col].method(value, inplace=True)', try using 'df.method({col: value}, inplace=True)' or df[col] = df[col].method(value) instead, to perform the operation inplace on the original object.
  
  
    manifest["entityId"].fillna("", inplace=True)

```

SageGJ:
@linglp I agree with your evaluations in the `Uncertainty` section. Both seem plausible and are similar to issues we've faced before. I also agree it's more reliable to fix this during submission; it might also be useful to display a message letting users know that these columns have been modified if we find that's necessary to do so.

andrewelamb:
@linglp Do you know why users are submitting manifests with the mis-cased(is that a word?) column headers? I was under the impression most users start with a manifest generated by Schematic. Do you know if Schematic is doing this?

schematic/store/synapse.py
```diff
@@ -2108,16 +2106,35 @@ def _read_manifest(self, metadataManifestPath: str) -> pd.DataFrame:
...
+                manifest = manifest.rename(columns={col: "Id"})
+            if col.lower() == "entityid":
+                manifest = manifest.rename(columns={col: "entityId"})
+
+        # If 'Id' still doesn't exist, see if uuid column exists
+        # Rename uuid column to "Id" column
+        if "Id" not in manifest.columns:
             # See if schema has `Uuid` column specified
             try:
                 uuid_col_in_schema = dmge.is_class_in_schema(
```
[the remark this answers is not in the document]
linglp: You brought up a good point, @SageGJ. However, I think this might be outside the scope of this PR, which is focused specifically on addressing the "id" issue. Since the uuid column was deprecated a while ago and this line hasn’t caused any problems, I’d suggest we leave it as-is for now.

schematic/store/synapse.py
```diff
@@ -2108,16 +2106,35 @@ def _read_manifest(self, metadataManifestPath: str) -> pd.DataFrame:
...
+        Returns:
+            pd.DataFrame: The updated manifest with a standardized 'Id' column and an 'entityId' column.
+        """
+        # Normalize any variation of 'id' to 'Id', "entityid" to "entityId"
+        for col in manifest.columns:
+            if col.lower() == "id":
+                manifest = manifest.rename(columns={col: "Id"})
+            if col.lower() == "entityid":
+                manifest = manifest.rename(columns={col: "entityId"})
+
```
andrewelamb, line 2134: Since we might use the `uuid` column down below, should we also rename it here like you do with `id` and `entityid`

tests/unit/test_store_synapse.py
```diff
@@ -57,3 +60,55 @@ def test_get_file_entityIds_all_files(self) -> None:
...
+            ),
+            # Case 2: 'Uuid' present — should be renamed to 'Id'
+            (
+                pd.DataFrame({"Uuid": ["test_value"]}),
+                pd.DataFrame({"Id": ["test_value"], "entityId": [""]}),
+            ),
+            # Case 3: 'entityID' in wrong mixed case — should be renamed to 'entityId'
+            (
+                pd.DataFrame({"Id": ["test_value"], "entityId": ["test_value"]}),
+                pd.DataFrame({"Id": ["test_value"], "entityId": ["test_value"]}),
```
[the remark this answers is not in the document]
linglp: Thanks for catching that. The test case should be "pd.DataFrame({"Id": ["test_value"], "entityID": ["test_value"]})", and the expected should be "pd.DataFrame({"Id": ["test_value"], "entityId": ["test_value"]})". Let me correct that.

tests/unit/test_store_synapse.py
```diff
@@ -57,3 +60,55 @@ def test_get_file_entityIds_all_files(self) -> None:
             "Filename": ["test_file_1.txt", "test_file_2.txt"],
             "entityId": ["syn2", "syn3"],
         }
+
+    @pytest.mark.parametrize(
```
andrewelamb, line 64: Could you add a case here where the input df has both "Id" and "Uuid", and the output is unchanged?

andrewelamb: Could you also add a case where the input df has neither `Id` or `entityId` columns and the output has both?

tests/unit/test_store_synapse.py
```diff
@@ -57,3 +60,55 @@ def test_get_file_entityIds_all_files(self) -> None:
...
+                pd.DataFrame({"Id": ["test_value"], "EntityId": ["test_value"]}),
+                pd.DataFrame({"Id": ["test_value"], "entityId": ["test_value"]}),
+            ),
+            # Case 5: 'EntityId' and "ID" are both in wrong mixed case — should be renamed to "Id" and 'entityId'
+            (
+                pd.DataFrame({"iD": ["test_value"], "entityID": ["test_value"]}),
+                pd.DataFrame({"Id": ["test_value"], "entityId": ["test_value"]}),
+            ),
+        ],
+    )
```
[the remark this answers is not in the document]
linglp: Okay. I will change it to the following: 
```
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[CSV-normalize_lowercase_id_to_Id] PASSED    [  7%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[CSV-normalize_uppercase_ID_to_Id] PASSED    [ 14%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[CSV-normalize_mixedcase_iD_to_Id] PASSED    [ 21%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[CSV-rename_Uuid_to_Id] PASSED               [ 28%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[CSV-rename_entityID_to_entityId] PASSED     [ 35%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[CSV-rename_EntityId_to_entityId] PASSED     [ 42%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[CSV-rename_both_iD_and_entityID] PASSED     [ 50%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[JSONLD-normalize_lowercase_id_to_Id] PASSED [ 57%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[JSONLD-normalize_uppercase_ID_to_Id] PASSED [ 64%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[JSONLD-normalize_mixedcase_iD_to_Id] PASSED [ 71%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[JSONLD-rename_Uuid_to_Id] PASSED            [ 78%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[JSONLD-rename_entityID_to_entityId] PASSED  [ 85%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[JSONLD-rename_EntityId_to_entityId] PASSED  [ 92%]
tests/unit/test_store_synapse.py::TestStoreSynapse::test_add_id_columns_to_manifest[JSONLD-rename_both_iD_and_entityID] PASSED  [100%]

```

schematic/store/synapse.py
```diff
@@ -2128,27 +2145,24 @@ def _add_id_columns_to_manifest(
...
                 manifest["Id"] = ""
 
-        # Retrieve the ID column name (id, Id and ID) are treated the same.
-        id_col_name = [col for col in manifest.columns if col.lower() == "id"][0]
-
-        # Check if values have been added to the Id coulumn, if not add a UUID so value in the row is not blank.
+        # Fill in UUIDs in the "Id" column if missing
         for idx, row in manifest.iterrows():
-            if not row[id_col_name]:
+            if not row["Id"]:
```
BryanFauble, line 2155: nit: "Id" could be a made a constant at the top of this python script to reduce the chance that a dev might make a mistake later on and use a different case like "id" or "ID"

schematic/store/synapse.py
```diff
@@ -2128,27 +2145,24 @@ def _add_id_columns_to_manifest(
...
-                row[id_col_name] = gen_uuid
-                manifest.loc[idx, id_col_name] = gen_uuid
+                row["Id"] = gen_uuid
+                manifest.loc[idx, "Id"] = gen_uuid
 
-        # add entityId as a column if not already there or
-        # fill any blanks with an empty string.
-        if not col_in_dataframe("entityId", manifest):
+        # Add entityId as a column if not already there or
+        # Fill any blanks with an empty string.
```
BryanFauble, line 2161: This comment doesn't really add anything tbh

BryanFauble (approved):
I agree with the other comments in the review, but overall the changes make sense. I'm going to go ahead and approve, but feel free to re-request review if there are additional major changes beyond what's been commented.

schematic/store/synapse.py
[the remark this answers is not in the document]
linglp: I saw Andrew's comment [here](https://github.com/Sage-Bionetworks/schematic/pull/1592#discussion_r2039835111) about this too. While I still think this is likely outside the scope of this PR, I went ahead and addressed it to make the function more consistent.

schematic/store/synapse.py
linglp: addressed it per @SageGJ 's comment [here](https://github.com/Sage-Bionetworks/schematic/pull/1592#discussion_r2038267850)

schematic/store/synapse.py
linglp: Okay. Thanks for pointing that out. I could delete it.

tests/unit/test_store_synapse.py
linglp: These tests were added. But I want to note that the conflict resolution between Id and uuid was not previously present. Since it is not difficult to add, I also added it in the PR (but this is also probably outside of the initial scope)

schematic/store/synapse.py
```diff
@@ -2108,16 +2106,35 @@ def _read_manifest(self, metadataManifestPath: str) -> pd.DataFrame:
...
         """
+        Ensures that the manifest DataFrame has standardized 'Id' and 'entityId' columns.
+
+        If any case variation of the 'id' column is present (e.g., 'id', 'ID', 'iD'), it is renamed to 'Id'.
+        If any case variation of the 'entityid' column is present, it is renamed to 'entityId'.
+        If 'Id' is still missing, it will be created as an empty column or derived from a 'Uuid' column,
+        depending on whether 'Uuid' is defined in the schema.
+        Missing values in the 'Id' column are filled with generated UUIDs.
+        If 'entityId' is still missing, it will be created and filled with empty strings.
+        If it is already present, any missing values will be replaced with empty strings.
```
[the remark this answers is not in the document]
linglp: This is a good point. Thank you Bryan!

schematic/store/synapse.py
linglp: Thanks Bryan! This is a good suggestion! I added the global constants at the top of the script.

andrewelamb (approved):

thomasyu888 (approved):
🔥 LGTM!  Fantastic work and first review @SageGJ + additional review @andrewelamb

PR #1593, opened 2025-04-09
[Schematic-272] updated documentation to warn users the reserved word in schematic; added more instructions in trouble shooting guide 

linglp:
# **Problem:**
* trouble shooting guide does not mention about reserved word in schematic
* included fix of schematic 272 in trouble shooting guide 

# **Testing:**
* https://schematicpy.readthedocs.io/en/schematic-272-doc/troubleshooting.html


docs/source/troubleshooting.rst
```diff
@@ -14,7 +14,23 @@ Whether you are using DCA or schematic API or schematic library/CLI, the followi
     1. If you are validating or submitting the manifest, how was the manifest initiatially generated?  If manually and NOT using schematic, there may be errors.
     2. If the manifest was generated by schematic, when was it generated?  Did you download the previously submitted manifest from Synapse and modify it?  Did you download it and resubmit it?  Please run the manifest generate command again to have a fresh manifest.
 
-4. Create a Github issue or reach out to your respective DCC service desks.  What is the schematic or DCA configuration used? Specifically, it's most important to capture the following:
+4. Does your data model has any "reserved word" as attribute names?
```
thomasyu888, line 17: Thanks for this, did you know where to look in the Synapse API docs for the list of reserved keywords?

linglp: I mainly got this list from our documentation here: https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/2473623559/The+Data+Model+Schema#A.-Reserved-use-special-attributes. But I also added "eTag" because it also shouldn't be used in the data model. I will update the documentation here and the confluence doc so that all documentation aligned.

docs/source/troubleshooting.rst
```diff
@@ -13,8 +13,37 @@ Whether you are using DCA or schematic API or schematic library/CLI, the followi
 
     1. If you are validating or submitting the manifest, how was the manifest initiatially generated?  If manually and NOT using schematic, there may be errors.
     2. If the manifest was generated by schematic, when was it generated?  Did you download the previously submitted manifest from Synapse and modify it?  Did you download it and resubmit it?  Please run the manifest generate command again to have a fresh manifest.
+4. Does your data model have any "reserved words" as attribute names?
 
-4. Create a Github issue or reach out to your respective DCC service desks.  What is the schematic or DCA configuration used? Specifically, it's most important to capture the following:
+   The following are reserved words that should not be used as attribute names in your data model:
```
thomasyu888, line 18: There are other reserved keywords that are synapse specific. Take a look at entity view reserved reserved column headers

docs/source/troubleshooting.rst
thomasyu888: There are other keywords that are reserved specifically by Synapse EntityViews - do you know where to look for those?

linglp: Are you thinking of this document? https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/EntityView.html These appear to be reserved words specific to file views, but when users submit a manifest, it's uploaded as a table_and_file, which should not impact the file view?

BryanFauble (approved):
Changes make sense. Thanks for the updates. I left a comment on where we can find additional `reserved words` for tables - Please add them when you get a chance, but overall the changes here are good!

docs/source/troubleshooting.rst
linglp: Thank you for sending me the code. I understand better now! I read [this line](https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/develop/services/workers/src/main/java/org/sagebionetworks/table/worker/UploadPreviewBuilder.java#L84) here that suggested that synapse would check if a column is a reserved column name when creating a table preview. [This line](https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/37c17de7b4e92d0679589f07c9b5ee523ece7062/lib/models/src/main/java/org/sagebionetworks/repo/model/table/TableConstants.java#L125) also suggested that any variations of the reserved name won't work (i.e. RowVersion, row_version, Row_version etc..)  I then experimented in the UI myself by adding reserved column names and confirmed the behavior!

Screenshot 2025-04-10 at 1 33 44 PM


Thank you for the link again! It was great to read the code and learn more about how synapse works!!!

docs/source/troubleshooting.rst
```diff
@@ -13,8 +13,75 @@ Whether you are using DCA or schematic API or schematic library/CLI, the followi
...
+     - When generating a manifest, schematic automatically adds the `entityId` column to the manifest to ensure that the metadata is attached to the correct file in Synapse during the submission step.
+     - When submitting or updating metadata, schematic uses `entityId` to know where the annotations should go.
+
+   - **eTag** (any case variation):
+       - The `eTag` is a version identifier for a file in Synapse. It helps ensure that metadata is being applied to the correct version of an entity.
+       - When submitting or updating metadata, schematic automatically adds the `eTag` column to the manifest.
+
+   Please also note that the following are reserved words for Synapse table columns. Any variations of the following would cause a conflict with Synapse table columns:
+
+   - **ROW_ID** (any case variation):
```
thomasyu888, line 36: @linglp so, you may need to list all of this out, but I think these are also reserved key words: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/EntityView.html.

I'd be great if you could double check, like...
image

andrewelamb:
I think this is great, but I have a question/possible suggestion. Should we differentiate between reserved keywords for Schematic (id, entityid, component), and Synapse Tables/Annotations? I agree that Schematic keywords should never be in the data model. But is that the case for the Synapse Table/Annotation keywords? Those would only be a problem if the DCC wanted to add the data to Synapse tables, and from what I understand, not all DCCs do that. Some DCCs only store data as CSV manifests.

docs/source/troubleshooting.rst
linglp: Thanks for the comment, Tom. This is an interesting case — I tested a few variations like CreatedOn, and it turns out they can actually be added as annotation keys. It might be a good idea to avoid using these altogether, but I can add a note in the documentation to clarify that.

thomasyu888: ^ Ya it's unfortunate that this is "case sensitive"

thomasyu888 (approved):
🔥 LGTM!  Great suggestion @andrewelamb to have the different sections added in

PR #1596, opened 2025-04-16
update lock file

andrewelamb:
# **Problem:**
The Poetry lock file is not up to date with changes made to the TOML file at some point since the last release.

# **Solution:**
Regenerate the lock file


thomasyu888 (commented):
@andrewelamb Thanks for being proactive and fixing this.  I'll defer to the rest of the team for final approval.

BryanFauble (approved):

linglp:
@andrewelamb @GiaJordan I noticed that the lock file may have been modified in this PR: https://github.com/Sage-Bionetworks/schematic/pull/1582/files#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89.

Just wondering—do we have a recommended way to avoid unintended version changes of poetry in the lock file in the future?

linglp (approved):

andrewelamb:
@linglp We don't that I know of.

poetry.lock
```diff
@@ -3995,6 +3996,7 @@ files = [
     {file = "ruamel.yaml.clib-0.2.12-cp311-cp311-manylinux_2_5_i686.manylinux1_i686.manylinux_2_17_i686.manylinux2014_i686.whl", hash = "sha256:811ea1594b8a0fb466172c384267a4e5e367298af6b228931f273b111f17ef52"},
     {file = "ruamel.yaml.clib-0.2.12-cp311-cp311-musllinux_1_1_i686.whl", hash = "sha256:cf12567a7b565cbf65d438dec6cfbe2917d3c1bdddfce84a9930b7d35ea59642"},
     {file = "ruamel.yaml.clib-0.2.12-cp311-cp311-musllinux_1_1_x86_64.whl", hash = "sha256:7dd5adc8b930b12c8fc5b99e2d535a09889941aa0d0bd06f4749e9a9397c71d2"},
+    {file = "ruamel.yaml.clib-0.2.12-cp311-cp311-musllinux_1_2_aarch64.whl", hash = "sha256:1492a6051dab8d912fc2adeef0e8c72216b24d57bd896ea607cb90bb0c4981d3"},
```
thomasyu888, line 3999: Something to note here: it's entirely possible due to your operating system that there are specific python wheels that are installed. 

I'm guessing that you're on a linux/ubuntu computer?

andrewelamb:
@thomasyu888 Yep, as should happen whenever the toml file is updated. My guess is that the toml file got updated again besides that, and the lock file wasn't also updated.

linglp:
@thomasyu888 My understanding is that updating the pyproject.toml (and the package versions there) was necessary to address that ticket. However, if the lock file had been generated using the same version of Poetry, it seems like we might have avoided the issue altogether? I wasn’t quite clear on why the Poetry version needed to be changed in the context of that ticket—was there a specific reason for that?

poetry.lock
thomasyu888: see this comment @andrewelamb . I'm unsure it had to do with anyone forgetting to update their lock file

SageGJ:
Yes it was necessary to update the lock file since the toml was updated. I had a different environments as I was testing out different version installations, it seems like a couple had the newer version of poetry. I wouldn't expect a minor version change to be the cause of the issue though, especially given its already not the version specified in the test workflow

linglp:
@andrewelamb I think it might be clearer for the record if you add the following to the "problem" statement: 

I also double checked that I could reproduce the issue in `develop` before I approved your PR: 
```
(schematicpy-py3.10) lpeng@MacBookPro schematic % poetry install --all-extras
Configuration file exists at /Users/lpeng/Library/Preferences/pypoetry, reusing this directory.

Consider moving TOML configuration files to /Users/lpeng/Library/Application Support/pypoetry, as support for the legacy directory will be removed in an upcoming release.
Installing dependencies from lock file

pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock [--no-update]` to fix the lock file.
```
The error indicates that the pyproject.toml and poetry.lock files are out of sync—based on my experience, this likely happened because the lock file was generated with a different version of Poetry than the one that was used locally. I opened this issue for us to look for options to pin the `poetry` version: https://sagebionetworks.jira.com/browse/SCHEMATIC-280

PR #1597, opened 2025-04-17
Schematic 276 jsonschema cli

SageGJ:
# **Problem:**

- Describe the specific technical problem that this PR solves.
- Include  details such as:
    - How you reproduced the problem
    - What code is affected
    - Any relevant background information

# **Solution:**

- Describe the specific technical solution that this PR provides.
- Include details such as:
    - How you debugged the problem
    - Why you chose this solution
    - Does the solution make any impact beyond the immediate problem
    - Any necessary technical debt incurred
    - Any important design decisions/links to design documentation (if applicable)

# **Testing:**

- Describe any testing that was performed to validate the solution.
- Include details such as:
    - How you tested the solution
    - Any relevant test results (Do not include any test results that may contain sensitive information)
    - Any automated tests that were added to the codebase


schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+    ) -> None:
+        """
+        Gather all components from the data model and store as list in the components attribute.
+        """
+
+        self.components = self._extract_components()
+
+        return
+
+    def _extract_components(
```
SageGJ, line 68: Heretofore schematic didn't have functionality exposed to gather and return all components of the data model without also included non-component attributes: see conversation [here](https://sagebionetworks.slack.com/archives/C028KRF7ZA5/p1744210973172219). This was introduced to avoid the necessity of using a `try: catch:` block.

The method was validated by using the example data model and comparing that to the small list of all components, and by using the [remote HTAN model](https://raw.githubusercontent.com/ncihtan/data-models/refs/heads/main/HTAN.model.jsonld) with many components and comparing the output to a separate [list of all manifest templates ](https://github.com/ncihtan/data-models/blob/main/dca-template-config.json)(components).

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        )
+
+        generator.get_component_json_schema()
+        generator.add_description_to_json_schema()
+        generator.write_json_schema_to_file()
+
+        return generator.component_json_schema
+
+
+class JsonSchemaComponentGenerator:
```
SageGJ, line 136: Jsonschema generation for a single component happens in this class, so this can be a good place to start

schematic/schemas/data_model_json_schema.py
```diff
@@ -405,6 +405,7 @@ def get_json_validation_schema(
             json_schema_dirname = os.path.dirname(json_schema_log_file_path)
             if json_schema_dirname != "":
                 os.makedirs(json_schema_dirname, exist_ok=True)
-            with open(json_schema_log_file_path, "w", encoding="UTF-8") as js_f:
-                json.dump(json_schema, js_f, indent=2)
+            export_json(
```
SageGJ, line 408: Used existing util function to keep sorting within the file more consistent.

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+
+from schematic.models.metadata import MetadataModel
+from schematic.schemas.data_model_json_schema import DataModelJSONSchema
+from schematic.schemas.data_model_parser import DataModelParser
+from schematic.schemas.data_model_graph import DataModelGraph, DataModelGraphExplorer
+from schematic.utils.io_utils import export_json
+
+
+class GeneratorDirector:
+    """
```
andrewelamb, line 20: I'm not crazy about the name of this class. Could we make it something more descriptive?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+    A class that directs the generation of a JSON schema component.
+    """
```
andrewelamb, line 22: Could you fill out this docstring? I'd like to see more description of what this class does.

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def __init__(
+        self,
+        data_model: str,
+        components: Optional[list[str]] = None,
+        output_directory: Optional[str] = None,
+    ):
```
andrewelamb, line 29: Could you fill out this docstring?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+    """
+
+    def __init__(
+        self,
+        data_model: str,
+        component: str,
+        output_directory: Path,
+        parsed_model: Dict[str, Any],
+    ):
+        self.data_model = data_model
```
andrewelamb, line 148: Could you add a docstring here?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        self.parsed_model = parsed_model
+        self._get_data_model_graph_explorer()
+
+        if component is None:
+            raise ValueError("A component must be provided")
+
```
andrewelamb, line 154: This isn't necessary here, as the typing for component isn't `Optional[str]`. You might want to place this in the command, though.

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        # the component can be provided as either a class label or display name
+        # internally all the work is done with the class label
+        component_class_label = self.dmge.get_node_label(component)
+        self.component = component_class_label if component_class_label else component
+
+        self.output_path = self._build_output_path(output_directory)
+
+        return
+
```
andrewelamb, line 163: I don't think this return statement does anything...

andrewelamb (changes requested):

andrewelamb:
@SageGJ This will be useful for the rest of the JSON Schema work that is upcoming. I mentioned it in the code space meeting, and there is interest in it by some of the DCCs. I made an extremely quick pass through some of the code. I'll do a more thorough pass on Monday.

schematic/schemas/commands.py
```diff
@@ -150,3 +152,60 @@ def convert(
...
+def generate_jsonschema(
+    data_model: str, output_directory: str, data_type: Optional[list[str]]
+) -> None:
+    """
+    Command to generate jsonschema files for validation for component(s) of the data model.
+    """
+    # pylint: disable=too-many-locals
+    # pylint: disable=redefined-outer-name
+    # pylint: disable=too-many-branches
+
```
andrewelamb, line 196: Are these pylint disables needed?

tests/test_cli.py
```diff
@@ -186,3 +186,50 @@ def test_submit_file_based_manifest(
...
+                "generate-jsonschema",
+                "--data_model",
+                data_model,
+                "--output_directory",
+                output_directory,
+                "--data_type",
+                data_type,
+            ],
+        )
+        assert result.exit_code == 0
```
andrewelamb, line 235: We're trying to move all our tests to the integration or unit test folder.

andrewelamb (changes requested):

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+    A class that directs the generation of a JSON schema component.
+    """
+
+    def __init__(
+        self,
+        data_model: str,
+        components: Optional[list[str]] = None,
+        output_directory: Optional[str] = None,
+    ):
+        self.data_model = data_model
```
andrewelamb, line 30: Is this a link or a path?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
+"JSON schema file generator"
+
+# pylint: disable=useless-return
+# pylint: disable=line-too-long
+# pylint: disable=attribute-defined-outside-init
+
```
andrewelamb, line 6: There's generally a reason why these are frowned upon, why are they disabled for the whole file?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+    def gather_components(
+        self,
+    ) -> None:
+        """
+        Gather all components from the data model and store as list in the components attribute.
+        """
+
+        self.components = self._extract_components()
+
+        return
```
andrewelamb, line 66: Is this method needed?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        self,
+        data_model: str,
+        components: Optional[list[str]] = None,
+        output_directory: Optional[str] = None,
+    ):
+        self.data_model = data_model
+        self.components = components if components else []
+
+        self.parsed_model: dict = {}
+        if output_directory is None:
```
andrewelamb, line 34: Any reason not to parse the data model here as part of the init?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        generator.write_json_schema_to_file()
+
+        return generator.component_json_schema
+
+
+class JsonSchemaComponentGenerator:
+    """
+    Used to generate a jsonschema for a given component and write it to a file.
+    """
+
```
andrewelamb, line 140: This exists above:
```
 generator.get_component_json_schema()
 generator.add_description_to_json_schema()
 generator.write_json_schema_to_file()
```

Should this have one method that does all three?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+        # Instantiate DataModelGraph
+        data_model_grapher = DataModelGraph(self.parsed_model, "class_label")
+
+        # Generate graph
+        graph_data_model = data_model_grapher.graph
+
+        self.dmge = DataModelGraphExplorer(graph_data_model)
+
+        return
```
andrewelamb, line 190: Thsi return isn't needed

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        Add the description of the component to the JSON schema.
+        """
+        component_description = self.dmge.get_node_comment(node_label=self.component)
+
+        description_dict = {"description": component_description}
+
+        description_dict |= self.incomplete_component_json_schema
+        self.component_json_schema: dict[str, Any] = description_dict
+
+        return
```
andrewelamb, line 232: This return isn't needed

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        if not os.path.exists(output_directory):
+            os.makedirs(output_directory)
+
+        export_json(
+            json_doc=self.component_json_schema,
+            file_path=str(self.output_path),
+            indent=2,
+        )
+
+        return
```
andrewelamb, line 250: this return isn't needed

schematic/schemas/json_schema_component_generator.py
linglp: I am thinking there might be an easier way to extract all the components using the data model graph: 
```
data_model_parser = DataModelParser(path_to_data_model="/Users/lpeng/Documents/schematic-git2/schematic/tests/data/example.model.csv")
parsed_data_model = data_model_parser.parse_model()
data_model_grapher = DataModelGraph(parsed_data_model, data_model_labels="class_label")
graph_data_model = data_model_grapher.graph
dmge = DataModelGraphExplorer(graph_data_model)
edges = graph_data_model.edges(data=True, keys=True)
components = {source for source, target, relation, _ in edges
              if relation == 'requiresDependency' and target == 'Component'}

```
This works for me for the example data component. 
And I am thinking this probably live in the `DataModelGraphExplorer` class.

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        return Path(output_directory, f"{stripped_component}_validation_schema.json")
+
+    def _get_data_model_graph_explorer(
+        self,
+    ) -> None:
+        """'
+        Create a DataModelGraphExplorer object to explore the data model graph.
+        """
+        # Instantiate DataModelGraph
+        data_model_grapher = DataModelGraph(self.parsed_model, "class_label")
```
linglp, line 183: hmm I am curious why here you hard coded this to "class label"?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+        Generate the JSON schema for the component using the DataModelJSONSchema instance.
+        The jsonschema is incomplete and will lack the description of the component.
+        """
+        schema_name = self.component + "_validation"
+
+        metadata_model = MetadataModel(
+            inputMModelLocation=self.data_model,
+            inputMModelLocationType="local",
+            data_model_labels="class_label",
```
linglp, line 204: should the `data_model_labels` here be configurable as well?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        return
+
+    def get_component_json_schema(
+        self,
+    ) -> None:
+        """
+        Generate the JSON schema for the component using the DataModelJSONSchema instance.
+        The jsonschema is incomplete and will lack the description of the component.
+        """
+        schema_name = self.component + "_validation"
```
linglp, line 199: currently the file paths are organized by component names which makes sense if user uses one schema. But what if the user uses two schemas? and those two schemas have the same components?

andrewelamb:
@SageGJ I left a bunch of comments in the code, but I have two main observations:

1. Formatting

At the top of the file that contains the two new classes you have:

```
# pylint: disable=useless-return
# pylint: disable=line-too-long
# pylint: disable=attribute-defined-outside-init
```
These should be removed, and the code should be modified to not trigger those pylint issues.

`useless-return`: There are several methods that just have a blank `return` statement. I don't think there is any reason to have these, and they should be removed.

`attribute-defined-outside-init`: For these there are several cases where somethign like this happens:

```
def init(self):
   self._set_attribute()

def _set_attribute(self):
   attribute = do_something()
   self.attribute = attribute

```

Instead, I would do something like:

```
def init(self):
   self.attribute = self._create_attribute()

def _create_attribute(self):
   attribute = do_something()
   return attribute

```
I think this is better, mainly for organizational reasons. A developer can look in the init section and see all of the attributes instead of having to hunt around different methods to find where they are set

2. Code organization

There are several methods in `GeneratorDirector` that aren't needed and make the class more complicated than it needs to be. I made some suggestions in the comments. 

I think you should ask the question: Should these be classes? I think these would have been perfectly fine as functions, and potentially much simpler. 

Thanks for getting this out there! Like I mentioned before, this will be very useful to use, and the DCC data managers.

tests/data/expected_jsonschemas/expected.BulkRNA-seqAssay.schema.json
```diff
@@ -0,0 +1,167 @@
...
+      "then": {
+        "properties": {
+          "Genome Build": {
+            "enum": [
+              "GRCm38",
+              "GRCh37",
+              "GRCm39",
+              "GRCh38"
+            ]
+          }
```
thomasyu888, line 27: From a feature perspective, this is very interesting.  These references come out of the box?  I hadn't intended for this ticket to include these conditionals, BUT what this looks to to saying is:

1. If File Format is "CSV/TSV", then:
"Genome Build" must be present and must be one of:

"GRCm38", "GRCh37", "GRCm39", "GRCh38"

2. If File Format is "CRAM", then:
"Genome Build" must be present and must be one of:

"GRCm38", "GRCh37", "GRCm39", "GRCh38"

3. If File Format is "BAM", then:
"Genome Build" must be present and must be one of:

"GRCm38", "GRCh37", "GRCm39", "GRCh38"

4. If File Format is "CRAM" (again), then:
"Genome FASTA" must be present and not null or empty.

I wonder if this particular feature should be designed more thoroughly before being added

thomasyu888: So i think what this is saying is that CSV/TSV/CRAM/BAM don't have Genome FASTA?  We'd have to test this out on all the schemas and bind this to synapse to see what this looks like

SageGJ: > These references come out of the box?

Yes, because of how the data model is built. The expected jsonschema files were generated by calling `DataModelJSONSchema.get_json_validation_schema` and adding the description manually to encompass what you included in your example code and what was specified in the ticket. So this particular section would have been included in jsonschemas generated for this data type even before this PR.

> I wonder if this particular feature should be designed more thoroughly before being added

Can you explain more what you mean @thomasyu888?

thomasyu888: Ah - my apologies. I don't think I ran into that in any of the data models I created with that code chunk - but it'd be good to know what fields in the jsonLd led to the creation of this conditional.

Exploration for after the schematic release gets started :)

andrewelamb: We also shouldn't assume that method that creates the JSON Schema is 100% perfect at the moment. When we were discussing the json schems I created in our dpetest org, Christina noticed that there were some issues with the HTAN ones(mialy missing enums). We may have to do some iterations on it once we hand off the JSON Schems we created to the various DCCs and get their feedback on them

schematic/schemas/json_schema_component_generator.py
SageGJ: @andrewelamb sure! I named it as such according to the pattern I followed and the responsibilities of the two classes. What do you feel is confusing about the current name that can be clarified in the new one?

schematic/schemas/json_schema_component_generator.py
SageGJ: It is a string, either to a local file path or a url to a remote data model

schematic/schemas/json_schema_component_generator.py
andrewelamb: If I saw it in the wild, my question would be "What is it generating?" Somehting similar to `JSONSchemaGenerator` would be my preference.

schematic/schemas/json_schema_component_generator.py
andrewelamb: I see, that does make it harder to be more specific. Maybe this should be called `data_model_location` to distinguish it from being an actual data model.

schematic/schemas/json_schema_component_generator.py
SageGJ: I thought it was appropriate given the nature of the errors I was receiving and the convention of disabling them for an entire file in 
* [cli_utils.py](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/utils/cli_utils.py#L3)
* [df_utils.py](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/utils/df_utils.py#L3)
* [general.py](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/utils/general.py#L3)
* [google_api_utils.py](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/utils/google_api_utils.py#L3)
* [schema_utils.py](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/utils/schema_utils.py#L3)
* [schematic/visualization/commands.py](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/visualization/commands.py#L2)
* [schematic/visualization/tangled_tree.py](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/visualization/tangled_tree.py#L3)

and in general the 66 times pylint disables are used throughout the schematic codebase as a whole.

I have removed `# pylint: disable=useless-return` and `# pylint: disable=attribute-defined-outside-init` per your other suggestions for improvement but it was my impression that line length would be covered by the code formatter `black` during the pre-commit hooks we have configured.

schematic/schemas/json_schema_component_generator.py
SageGJ: nope! I meant to remove this method earlier, thanks for catching!

schematic/schemas/json_schema_component_generator.py
SageGJ: That's a good point, I'll make that change, thanks!

schematic/schemas/json_schema_component_generator.py
SageGJ: okay, that makes sense. Thanks!

schematic/schemas/commands.py
SageGJ: They were used in the other schemas command so I included them, but I can try without

schematic/schemas/json_schema_component_generator.py
thomasyu888: This is great context, and it's important to think about the practices that exist in schematic. However, let's take a step back at. I'm just going to be a bit nitpicky on this nitpick empty return statement just to drive across a point.

1. Existing Python standards (PEP8, etc)
2. Python best practices/standards that evolve over time (this is nuanced and trickier - maybe a retro topic)

For the purpose of this PR, let's focus on existing standards. https://peps.python.org/pep-0008/. There is a section on return statements if you Ctrl-F "return statement".  (no empty return statements)

Now, we may ask, should we do `return None` then?  I personally don't think that is necessary if the function is a procedure and generally doesn't return anything. What I read from PEP8 here is that if you have a return statement already in the function, be consistent and `return None` if you have other return statements.

Don't feel responsible to learn all PEP8 standards all at once but- this is the value code reviews bring. Now that we see that, an easy "standards lift" would be to just remove places where we do empty return statements when we touch those parts of the code. So two things 

1. I believe in iterative improvements to the code. The text above is just around an empty return statement, the sentiment is true for all other pieces of code contributed.
2. Not all practices in existing codebases could be following Python and/or coding best practices/standards.

Thanks for enduring my TED text.

andrewelamb: Here is my perspective:

@SageGJ is absolutely correct that there are `pylints disable` sprinkled all around Schematic, so if that was our standard, it would be totally appropriate. I probably put most of them there!

The reason most of them exist, though, was that we(FAIR data at the time) agreed that the code quality needed to be improved. This included lowering tech debt, improving the typing through mypy standards, and linting through pylint and black. So what I did was for about half the code, I went through and fixed 90% of the linting errors. These are the ones that were quick fixes that didn't need to be refactored to fix. For all the ones that would take a more comprehensive fix ([this](https://github.com/Sage-Bionetworks/schematic/blob/develop/schematic/schemas/data_model_json_schema.py#L100) method is the poster child for needing a refactor) I added the `pylint: disable`s so that they wouldn't trigger failures during testing. This helped to ensure we weren't adding to the tech debt. The idea was that at some point in the future, we would be able to get back to those problem methods and refactor them. 

In conclusion :) , we should be looking to remove `pylint: disable`s when we can, and not add them (unless there's a really good reason).

schematic/schemas/commands.py
```diff
@@ -150,3 +152,60 @@ def convert(
...
+    short_help=query_dict(
+        schema_commands, ("schema", "generate-jsonschema", "short_help")
+    ),
+)
+@click_log.simple_verbosity_option(logger)
+@click.option(
+    "--data_model",
+    "-dm",
+    help=query_dict(
+        schema_commands, ("schema", "generate-jsonschema", "data_model_labels")
```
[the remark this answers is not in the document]
SageGJ: The incorrect help text was an error. This param has now been renamed to what was suggested by @andrewelamb

schematic/schemas/json_schema_component_generator.py
SageGJ: At this stage, no. That will be in a subsequent pr

tests/utils.py
```diff
@@ -32,3 +33,30 @@ class CleanupItem:
     name: Optional[str] = None
     parent_id: Optional[str] = None
     action: CleanupAction = CleanupAction.DELETE
+
+
+def dict_sort(item):
```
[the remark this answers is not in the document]
SageGJ: I'm not clear on what you're asking here. I'm not seeing a `export_json` function in `test_utils.py`. There's a test for testing that the function `io_utils.export_json` works as expected?

schematic/schemas/json_schema_component_generator.py
SageGJ: The initial ask was to get a schema with text that conformed to synapse standards, so the class label was used here. Whichever  form of the component name was given, the class label would be derived and used. After development was largely underway @thomasyu888 and I decided we'd need to add support for display names too, which will happen in a subsequent pr

tests/unit/test_json_schema_component_generator.py
```diff
@@ -0,0 +1,381 @@
...
+    GeneratorDirector,
+    JsonSchemaComponentGenerator,
+)
+from schematic.schemas.data_model_parser import DataModelParser
+from tests.utils import json_files_equal
+from unittest.mock import patch
+
+
+@pytest.fixture
+def parsed_example_model(helpers):
```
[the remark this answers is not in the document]
SageGJ: I've moved `parsed_data_model` but had to make updates to `data_model` to support testing the new path functionality. with those changes I think it makes more sense to keep it here since it's specific to this class's behavior

tests/unit/test_json_schema_component_generator.py
```diff
@@ -0,0 +1,381 @@
...
+
+@pytest.fixture
+def data_model(helpers):
+    data_model = helpers.get_data_path("example.model.jsonld")
+
+    yield data_model
+
+
+@pytest.fixture
+def output_directory(helpers):
```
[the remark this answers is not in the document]
SageGJ: see above comment wrt `data_model_location`. This fixture now depends on the other and is more specific to the beavior of this class

schematic/schemas/commands.py
```diff
@@ -150,3 +152,60 @@ def convert(
...
+
+    generator = GeneratorDirector(
+        data_model=data_model, output_directory=output_directory, components=data_type
+    )
+
+    generator.generate_jsonschema()
+
+    # get the end time
+    end_time = time.time()
+
```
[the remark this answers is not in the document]
SageGJ: Great point, thanks for raising. The  log messages were actually from another class. I've added messages to these new ones to be more informative and removed the timing information since it's not very necessary here

schematic/schemas/json_schema_component_generator.py
SageGJ: Thanks for the suggestion @linglp !
I can see how iterating over the edges of the graph and checking the relationship and target is simpler to write. My concern with extracting components this way is the extra time needed to iterate over every edge in a graph (which can be more than just the number of attributes described in a data model) and to do multiple string comparisons at each iteration.

The parsed data model is also a consistent intermediary of loading data models, and by unpacking this dictionary we'd only have to iterate over the top level keys, and the mid level keys of the nested dictionaries, instead of the top level keys, mid level keys, and values of the nested dictionaries. Then, pandas can be used to do comparisons of an entire column at once instead of individually on each attribute, possibly multiple times.

Timing the execution of the two functions illustrates this as well

<details>

<summary>Code Comparison Example</summary>

```
import pandas as pd
import numpy as np
import time

from schematic.schemas.data_model_parser import DataModelParser
from schematic.schemas.data_model_graph import DataModelGraph, DataModelGraphExplorer


def extract_components_df(
    data_model_parser
) -> list[str]:
    # Store the parsed model in the parsed_model attribute
    # The parsed model is a nested dictionary
    parsed_model = data_model_parser.parse_model()

    # To represent each attribute of the nested model dictionary as a column in a dataframe,
    # it must be unpacked and the index reset
    unpacked_model_dict = {}

    for top_key, nested_dict in parsed_model.items():
        for nested_key, value in nested_dict.items():
            unpacked_model_dict[top_key, nested_key] = value

    attrs = pd.DataFrame.from_dict(
        unpacked_model_dict,
        orient="index",
    ).reset_index(drop=True)

    # Get a series of boolean values that can be used to identify with attributes (rows) have 'Component' in the DependsOn column
    string_depends_on = attrs.DependsOn
    string_depends_on = string_depends_on.astype(str)
    depends_on_component = string_depends_on.str.contains("Component")

    # select this subset and return as list of all components in the data model
    components = attrs.loc[depends_on_component, "Attribute"]

    return list(components)

def extract_components_graph(data_model_parser):
    parsed_data_model = data_model_parser.parse_model()
    data_model_grapher = DataModelGraph(parsed_data_model, data_model_labels="class_label")
    graph_data_model = data_model_grapher.graph
    dmge = DataModelGraphExplorer(graph_data_model)
    edges = graph_data_model.edges(data=True, keys=True)
    components = {source for source, target, relation, _ in edges
                if relation == 'requiresDependency' and target == 'Component'}
    
    return components



data_model_parser = DataModelParser(path_to_data_model="https://raw.githubusercontent.com/ncihtan/data-models/refs/heads/main/HTAN.model.jsonld")
t_elapsed_df=np.empty(10,dtype='float')
t_elapsed_graph=np.empty(10,dtype='float')


for i in range(10):
    t_start = time.time()
    extract_components_df(data_model_parser)
    t_elapsed_df[i] = time.time() - t_start

    t_start = time.time()
    extract_components_graph(data_model_parser)
    t_elapsed_graph[i]=time.time() - t_start



print(f"Times elapsed for extracting components from dataframe: {t_elapsed_df} seconds")
print(f"Times elapsed for extracting components from graph: {t_elapsed_graph} seconds")
print(f"Average time elapsed df {np.mean(t_elapsed_df):.2f} seconds")
print(f"Average time elapsed graph {np.mean(t_elapsed_graph):.2f} seconds")
```
</details>

```
Times elapsed for extracting components from dataframe: [26.30762696 31.11916637 30.83139372 31.27199125 31.9875803 30.7575767 27.76651859 25.06767321 31.34337258 29.56928658] seconds
Times elapsed for extracting components from graph: [ 99.60343623 103.96529031 107.35175419 106.22667384 103.00029325 101.33841276 107.22191119 109.73435354 104.1642127  104.0957067 ] seconds
Average time elapsed df 29.60 seconds
Average time elapsed graph 104.67 seconds
```


What are your thoughts? Are there other benefits to using the graph approach?

schematic/schemas/json_schema_component_generator.py
SageGJ: Got it! I removed this and the associated test.
If the class is initialized now without a component then later this exception will be raised
`AttributeError: 'NoneType' object has no attribute 'translate'`

schematic/schemas/json_schema_component_generator.py
SageGJ: resolving per call discussion

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+    """
+
+    def __init__(
+        self,
+        data_model: str,
+        components: Optional[list[str]] = None,
+        output_directory: Optional[str] = None,
+    ):
+        self.data_model = data_model
+        self.components = components if components else []
```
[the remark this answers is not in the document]
SageGJ: resolving per call discussion

tests/utils.py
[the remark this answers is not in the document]
linglp: Sorry, I meant the export_json function in io_utils.

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,311 @@
+"JSON schema file generator"
+
+# pylint: disable=line-too-long
+
+import os
+from pathlib import Path
+from typing import Any, Dict, Optional
```
andrewelamb, line 7: importing `Dict` isn't needed any longer, you cna just use `dict`

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,311 @@
...
+        self.components = components if components else self.gather_components()
+
+        if output_directory is None:
+            self.output_directory = Path(os.getcwd(), "component_jsonschemas")
+        else:
+            self.output_directory = Path(output_directory)
+
+    def generate_jsonschema(
+        self,
+    ) -> list[Dict[str, Any]]:
```
andrewelamb, line 60: just use `dict` (lower case "d") here

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,311 @@
...
+        incomplete_component_json_schema (dict): JSON schema before adding the component description.
+        component_json_schema (dict): Final JSON schema with the description added.
+    """
+
+    def __init__(
+        self,
+        data_model_location: str,
+        component: str,
+        output_directory: Path,
+        parsed_model: Dict[str, Any],
```
andrewelamb, line 172: same as above

tests/utils.py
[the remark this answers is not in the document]
SageGJ: @linglp oh I see what you mean!
So `export_json` does sort the keys because of the parameter usage, but it doesn't sort the values in nested dicts or lists. In this case it means that the list enum values are not necessarily in the same order when comparing to files that are otherwise the same. So it's necessary to sort the keys and values for comparison of the json files to say whether or not they're truly the same

tests/unit/test_json_schema_component_generator.py
```diff
@@ -0,0 +1,381 @@
...
+        self,
+        helpers,
+        specified_component,
+        expected_components,
+        output_directory,
+        example_data_model_path,
+        mocker,
+    ):
+        comopnent_gather_spy = mocker.spy(
+            JsonSchemaGeneratorDirector, "gather_components"
```
andrewelamb, line 90: typo here `comopnent`

tests/unit/test_json_schema_component_generator.py
```diff
@@ -0,0 +1,381 @@
...
+            output_directory=output_directory.given,
+        )
+
+        # WHEN a schema is generated
+        json_schema = generator.generate_jsonschema()
+
+        # AND if no component is specified already
+        # THEN all components should be gathered from the data model
+        if specified_component is None:
+            comopnent_gather_spy.assert_called_once()
```
andrewelamb, line 106: same as above

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        Returns:
+            json_schemas: A list of JSON schemas for each component.
+        """
+        json_schemas = []
+
+        if not self.components:
+            self.gather_components()
+
+        for component in self.components:
+            json_schemas.append(self._generate_jsonschema(component))
```
linglp, line 70: As I think about it more, I am getting a bit confused why we would need to gather all the json schema dictionaries and save them all in a list? Is this `json_schemas` list be used any where? If not, can we do: 
```
for component in self.components:
            self._generate_jsonschema(component)
```
When I think of a "JSON Schema," I typically expect it to be a single dictionary. The fact that this returns a list of dictionaries is a bit confusing to me.

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+            parsed_model: A dictionary representation of the data model.
+        """
+
+        data_model_parser = DataModelParser(self.data_model)
+
+        self.parsed_model = data_model_parser.parse_model()
+
+        return
+
+    def _generate_jsonschema(self, component: str) -> dict[str, Any]:
```
linglp, line 124: Also, since this method is internal, we don't expect users to call it directly. However, as I think about it more, could there be a use case where a user might want to generate the JSON schema for just a single component based on their data model? And thus it makes sense for this method to be public?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -0,0 +1,250 @@
...
+        Write the JSON schema to a file.
+        """
+        output_directory = os.path.dirname(self.output_path)
+        if not os.path.exists(output_directory):
+            os.makedirs(output_directory)
+
+        export_json(
+            json_doc=self.component_json_schema,
+            file_path=str(self.output_path),
+            indent=2,
```
linglp, line 307: Just curious — is there a reason you set the indentation to 2 rather than sticking with the default 4?

tests/unit/test_json_schema_component_generator.py
```diff
@@ -0,0 +1,381 @@
...
+    def test_gather_components(
+        self, helpers, data_model_location: str, expected_components: list[str]
+    ):
+        # GIVEN a data model from a local file or a url
+        if data_model_location.startswith("example"):
+            data_model_location = helpers.get_data_path(data_model_location)
+
+        # WHEN an instance of the JsonSchemaGeneratorDirector class is created
+        generator = JsonSchemaGeneratorDirector(data_model_location=data_model_location)
+        # AND the _extract_components method is called
```
linglp, line 253: this comment can be removed? the name of the function is outdated after your recent commits.

schematic/schemas/json_schema_component_generator.py
linglp: Thanks for doing the comparison. Performance is a valid point. I think in the future, we might even be able to build a function that could take inputs from either a graph or a parsed data model.  But for the first iteration, I think this looks good. Thank you for the hard work.

schematic/schemas/json_schema_component_generator.py
andrewelamb: I don't think this is needed. The user can set the `components` attribute to just the component they want and then use the `generate_jsonschema` method to do this.

schematic/schemas/json_schema_component_generator.py
andrewelamb: @linglp This was part of the spec, "Create all the jsonschemas for all the components".

SageGJ: @linglp like @andrewelamb said it was part of the spec to enable the creation of all components in a data model. 
The jsonschemas are returned in a list so that if users use the class programmatically then they still have access to all the jsonschemas in memory for further processing instead of having to read from a number of files individually.

schematic/schemas/json_schema_component_generator.py
SageGJ: @linglp, @andrewelamb is correct, the way to generate a single component's jsonschema would be to set the `component` attribute to that component and then call `generate_jsonschema`

schematic/schemas/json_schema_component_generator.py
SageGJ: The jsonschema files we have been generating already have an indentation of 2. I opted to keep them consistent here
@linglp

thomasyu888 (commented):
🔥 LGTM!  Thank for the hard work/thoroughness and reviews on this.  I'll defer to the team for final review, but the overall logic is sound to me!

schematic/schemas/json_schema_component_generator.py
linglp: Thanks @SageGJ . This makes sense to me.

tests/integration/test_json_schema_generation.py
```diff
@@ -0,0 +1,71 @@
...
+    yield data_model_jsonld
+
+
+class TestSchemaCli:
+    # TODO: change url target to develop branch
+    @pytest.mark.parametrize(
+        "data_model_location",
+        [
+            "example.model.jsonld",
+            "https://raw.githubusercontent.com/Sage-Bionetworks/schematic/refs/heads/schematic-276-jsonschema-cli/tests/data/example.model.jsonld",
```
linglp, line 31: Here, the json model url contains schematic-276-jsonschema-cli which is the branch name. Can we switch it to the main branch?

tests/test_schemas.py
```diff
@@ -278,16 +278,26 @@ def test_parse_csv_model(
         assert "Attribute" in attr_rel_dictionary[attribute_key]["Relationships"]
 
 
-@pytest.mark.parametrize("data_model", ["example.model.jsonld"], ids=["jsonld"])
+@pytest.mark.parametrize(
+    "data_model",
+    [
+        "example.model.jsonld",
+        "https://raw.githubusercontent.com/Sage-Bionetworks/schematic/refs/heads/develop/tests/data/example.model.jsonld",
```
linglp, line 285: hmm should we point it to main instead? since eventually we are merging to main?

tests/integration/test_json_schema_generation.py
thomasyu888: I would use a permanent link instead of pointing to main or feature branch.

tests/unit/test_json_schema_component_generator.py
```diff
@@ -0,0 +1,381 @@
...
+                    "Patient",
+                    "Biospecimen",
+                    "Bulk RNA-seq Assay",
+                    "MockComponent",
+                    "MockRDB",
+                    "MockFilename",
+                ],
+            ),
+            (
+                "https://raw.githubusercontent.com/ncihtan/data-models/238173f2f193d0b068313d096fcab6ee19c34c3b/HTAN.model.jsonld",
```
linglp, line 145: should we always use the data model in the main branch: https://raw.githubusercontent.com/ncihtan/data-models/refs/heads/main/HTAN.model.jsonld rather than what you have here?

thomasyu888: I think this one is ok because we don't control the data model here, so we may not want our tests to break if the data model gets updated in HTAN

tests/utils.py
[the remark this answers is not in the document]
linglp: Got it. Can we add a doc string to this function and also change the name to something more descriptive? like "normalize_json_schema"?

tests/unit/test_json_schema_component_generator.py
linglp: I see — if their data model changes in the future, our test could fail if we rely on the main branch. That makes sense. I'm still wondering though: if we anticipate their model will change, what’s the benefit of using their current model for testing now? If we leave the test as it is, and their model evolves, wouldn’t we risk no longer validating whether our functions meet their intended use case?

thomasyu888: There's two things here and I'll defer to Gianna if she has further comments.

1. The tests here test that our functions work for the current data model build, so if their data model evolves, that shouldn't impact this function as written
2. If dcc's want to evolve their data model, it is their responsibility to ensure that the updates they make are compliant with the software (ofcourse we can help with tooling like ci/cd)
3. If we change the way these functions behave and it fails with older data model builds, then we would have to update the permanent link according AND communicate out non-backwards incompatibilty. For additive features, we have other data models that are used to test.

Overall - we cannot anticipate every change people may make, and we may have to update this test if there are significant differences within the way we evolve our functions.

SageGJ: This specific link is used per @andrewelamb 's comment [here](https://github.com/Sage-Bionetworks/schematic/pull/1597#discussion_r2064041040)
> If you are going to use a DCCs data model like this, you should either save it in our repo or link to a specific commit. The way it is right now, if they change their datamodel, this could break this test.

We're using this link for two tests:
1. to test that we can initialize an object with a url to a data model and have it set the location and output path appropriately
2. to ensure that when we use a data model other than our own, it can parse out all the components in the data model appropriately 

<br>


1. doesnt depend on the content of the model as much as the fact that it's a url
2. depends on the content of the model and the fact that it's a url. The purpose of the test is to ensure that we can parse current versions of models to retrieve components. like @thomasyu888 said any updates they make it's their responsibility to ensure they're still compatible.

linglp (approved):
Looks good to me. Thanks for all the hard work and discussion.

PR #1599, opened 2025-04-30
Add property descriptions to generated jsonschemas

SageGJ:
# **Problem:**
Property/attribute descriptions were missing from generated jsonschema files

# **Solution:**
Functionality has been added to include those descriptions as well
the link to a data model for a test was also fixed to a permanent link

# **Testing:**
existing tests cover this case

### DIFF
--- a/schematic/schemas/json_schema_component_generator.py
+++ b/schematic/schemas/json_schema_component_generator.py
@@ -284,6 +284,14 @@ class JsonSchemaComponentGenerator:
 
         description_dict |= self.incomplete_component_json_schema
         self.component_json_schema = description_dict
+
+        for attribute, value in self.component_json_schema["properties"].items():
+            if isinstance(value, dict) and "description" not in value:
+                # TODO: Update to give either class label or display name
+                value["description"] = self.dmge.get_node_comment(
+                    node_display_name=attribute
+                )
+
         click.echo(f"Validation JSONschema generated for {self.component}.")
 
     def write_json_schema_to_file(
--- a/tests/data/expected_jsonschemas/expected.Biospecimen.schema.json
+++ b/tests/data/expected_jsonschemas/expected.Biospecimen.schema.json
@@ -4,24 +4,28 @@
   "description": "TBD",
   "properties": {
     "Component": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
       }
     },
     "Patient ID": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
       }
     },
     "Sample ID": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
       }
     },
     "Tissue Status": {
+      "description": "TBD",
       "enum": [
         "Malignant",
         "None",
--- a/tests/data/expected_jsonschemas/expected.BulkRNA-seqAssay.schema.json
+++ b/tests/data/expected_jsonschemas/expected.BulkRNA-seqAssay.schema.json
@@ -120,12 +120,14 @@
   "description": "TBD",
   "properties": {
     "Component": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
       }
     },
     "File Format": {
+      "description": "TBD",
       "enum": [
         "CSV/TSV",
         "FASTQ",
@@ -134,12 +136,14 @@
       ]
     },
     "Filename": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
       }
     },
     "Genome Build": {
+      "description": "TBD",
       "enum": [
         "GRCm38",
         "GRCh37",
@@ -148,8 +152,11 @@
         ""
       ]
     },
-    "Genome FASTA": {},
+    "Genome FASTA": {
+      "description": "TBD"
+    },
     "Sample ID": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
--- a/tests/data/expected_jsonschemas/expected.MockComponent.schema.json
+++ b/tests/data/expected_jsonschemas/expected.MockComponent.schema.json
@@ -4,36 +4,42 @@
   "description": "Component to hold mock attributes for testing all validation rules",
   "properties": {
     "Check Ages": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
       }
     },
     "Check Date": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
       }
     },
     "Check Float": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
       }
     },
     "Check Int": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
       }
     },
     "Check List": {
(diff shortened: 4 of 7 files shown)


schematic/schemas/json_schema_component_generator.py
```diff
@@ -282,8 +282,17 @@ def add_description_to_json_schema(
...
+        self.component_json_schema.update(description_dict)
+
+        for attribute, value in self.component_json_schema["properties"].items():
+            if isinstance(value, dict) and "description" not in value:
+                # https://sagebionetworks.jira.com/browse/SCHEMATIC-284
+                # https://sagebionetworks.jira.com/browse/SCHEMATIC-283
+                value["description"] = self.dmge.get_node_comment(
+                    node_display_name=attribute
+                )
+
```
andrewelamb, line 295: Shouldn't this functionality be in `DataModelJSONSchema.get_json_validation_schema`?

SageGJ: Presently, I don't think so since that class is more for generating schemas for manifest validation during schematic's validation and this class is for creating schema to be used on synapse for more in-depth validation. I think we can consider moving some functionality from these classes to `DataModelJSONSchema` during the refactor, however.

andrewelamb: Sounds good!

schematic/schemas/json_schema_component_generator.py
```diff
@@ -282,8 +282,17 @@ def add_description_to_json_schema(
 
         description_dict = {"description": component_description}
 
-        description_dict |= self.incomplete_component_json_schema
-        self.component_json_schema = description_dict
+        self.component_json_schema.update(self.incomplete_component_json_schema)
+        self.component_json_schema.update(description_dict)
+
+        for attribute, value in self.component_json_schema["properties"].items():
```
andrewelamb, line 288: Do you think there is anyway for this class to get a malformed schema where the "properties" key doesn't exist, or it's not a dict?

SageGJ: I don't think it's likely. It would have to be a component defined without any attributes attached to it. Still I can switch this to
```suggestion
        for attribute, value in self.component_json_schema.get("properties").items():
```
to be safer?

andrewelamb: I'm not sure that's safer (about the same?) here, since `dict.get()` will return `None` if `properties` isn't in the dict. If I wanted to be really safe I would do something like:

```
if "properties" not in self.component_json_schema:
    raise ValuerError(f"component: {self.component} is malformed, missing properties")

if not is_instance(self.component_json_schema["properties"], dict):
    raise ValuerError(f"component: {self.component} is malformed, properties should be an object")

for attribute, value in self.component_json_schema["properties"].items():
```

You would still get an error, but it would be more informative than `KeyError`

tests/data/expected_jsonschemas/expected.Biospecimen.schema.json
```diff
@@ -4,24 +4,28 @@
...
   "properties": {
     "Component": {
+      "description": "TBD",
       "minLength": 1,
       "not": {
         "type": "null"
       }
     },
     "Patient ID": {
+      "description": "TBD",
```
linglp, line 14: Not related to this PR, but why do we have: 
```
      "minLength": 1,
      "not": {
        "type": "null"
      }
```
This seems pretty redundant to me. If minLength is already 1, then the type is not null?

andrewelamb: I think an empty string '' doesn't count as null, so I think this means that there must be a value, AND it must must be lengeth at least 1

schematic/schemas/json_schema_component_generator.py
SageGJ: got it thanks!

andrewelamb: Another alternative would have been to do nothing if properties were malformed:

```
if "properties" in self.component_json_schema and is_instance(self.component_json_schema["properties"], dict):
    for attribute, value in self.component_json_schema["properties"].items():
```

tests/data/expected_jsonschemas/expected.Biospecimen.schema.json
linglp: Thanks, that makes sense.

schematic/schemas/json_schema_component_generator.py
```diff
@@ -282,8 +282,27 @@ def add_description_to_json_schema(
...
+        if not isinstance(self.component_json_schema["properties"], dict):
+            raise ValueError(
+                f"component: {self.component} is malformed, properties should be an object"
+            )
+
+        for attribute, value in self.component_json_schema["properties"].items():
+            if isinstance(value, dict) and "description" not in value:
+                # https://sagebionetworks.jira.com/browse/SCHEMATIC-284
+                # https://sagebionetworks.jira.com/browse/SCHEMATIC-283
+                value["description"] = self.dmge.get_node_comment(
```
linglp, line 302: I saw that `get_node_component` doesn't have a test, so I am a bit worried that the function would fail. Do we need to do error handling here?

andrewelamb: @linglp What do you think should happen if `dmge.get_node_comment` fails?

linglp: I think the description can just be an empty string?

andrewelamb: @thomasyu888 For downstream uses: When there is no description to add, what would be better, no description keyword, or a description keyword with an empty string?

SageGJ: Currently, for components that do not have a specified description, the behavior (that's existed before this pr) is to mark the description as `TBD`

thomasyu888: 1. @andrewelamb can you look into if the JSON schema will work with description as an empty string on Synapse when you register the schema?
2. We can keep the behavior and I can do more user research.  I think we do want to keep the description keyword either way, but I can ask the data modeling group what they would like. I'm assuming that changing the default value from "TBD" to "" won't be too much work?

SageGJ: @thomasyu888 no, it would be easy to switch. However it would affect the processing of every data model in schematic because of where the default value is defined. I'm not sure where else it's used internally at this point but every data model json, jsonschema, and jsonld schematic processes would be affected by such a change, as would any downstream data model processing our users have set up that looks for that string

thomasyu888: Thanks @SageGJ for that context!  I'll be sure to include that in my user research, for now, I think it's safest to NOT change it so we don't break anything unexpectedly.

andrewelamb: @thomasyu888 Will do

BryanFauble (approved):
LGTM

schematic/schemas/json_schema_component_generator.py
andrewelamb: @thomasyu888 Empty strings work in [Synapse](https://repo-prod.prod.sagebase.org//repo/v1/schema/type/registered/dpetest-mockRDB-0.0.3)

PR #1600, opened 2025-05-01
Schematic `v25.4.1`

SageGJ:
Schematic `v25.4.1` release


schematic/schemas/data_model_graph.py
```diff
@@ -359,7 +359,7 @@ def get_component_requirements(
     def get_component_requirements_graph(
         self,
         source_component: str,
-    ) -> nx.DiGraph:
+    ) -> nx.Graph:
```
linglp, line 362: One thing that I noticed is that we now use a mix of nx.MultiDiGraph, nx.DiGraph, and nx.Graph.. I think there probably should be a ticket to clean up and standardize them?

linglp (approved):
looks good to me. Just a minor comment.

schematic/schemas/data_model_graph.py
SageGJ: @linglp that's a good point, could you open a ticket?

PR #1604, opened 2025-05-08
[Schematic-285] DataModelJsonSchema refactor

andrewelamb:
# **Problem:**
The `DataModelJSONSchema` class had a large amount of tech debt and was constructing JSON Schemas in an incorrect way. In addition, the JSON Schemas generated did not have description, type, minimum, or maximum keywords. The `JSONSchemaComponentGenerator` class was adding descriptions.

# **Solution:**
 This PR provides a new function, `create_json_schema`, that also generates a JSON Schema, but does it correctly, leaving the old class for DCA-related purposes. This new function adds a description keyword based on the comment from the data model, and adds type, minimum, or maximum keywords based on the data model validation rules. 

The function has two new parameters. These will make it easier in the future for the CLI to have a display name parameter:

- use_property_display_names: When true, the property names in the JSON Schema are node display names rather than labels
- use_valid_value_display_names: When true the enums(valid values) in the JSON Schema are node display names rather than labels

In addition, the JSON Schema generation code is completely refactored into helper functions to make it clearer how the process works. This should make it much easier to make needed additions in the future. See [this doc](https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/4115365889/Validating+against+Generated+JSON+Schema) for more information.

# **Testing:**
- Unit tests: JSON Schema that are produced locally are validated with instance data
- Integration tests: JSON Schema are uploaded to the dpetest org, bound to folders, annotations added, and then validated


schematic/schemas/json_schema_component_generator.py
thomasyu888, line 1: Can this be plugged directly into the CLI command?   If you feel that that is another ticket, we can certainly add that, but please provide instructions on how to leverage this class to create all the jsonschemas for a data model (Lets start with the AD data model)

schematic/schemas/json_schema_generator.py
thomasyu888, line 1: As a test, can we run some code that does this:

```
schematic schema generate-jsonschema -dml https://raw.githubusercontent.com/adknowledgeportal/data-models/refs/heads/add-scrna-validation-rules/AD.model.jsonld -od . -dt AssayScrnaSeqMetadataTemplate
```

Then in Python

```
import json

import synapseclient
syn = synapseclient.login()
js = syn.service("json_schema")

my_org_name = "tyu.test.schemas" # maybe dpetest here
my_org = js.JsonSchemaOrganization(my_org_name)
with open("AssayScrnaSeqMetadataTemplate_validation_schema.json","r") as f:
    temp = json.load(f)
my_org.create_json_schema(temp, "ad.scrnaseqmetadatatemplate", "0.0.1")
```

This will fail currently

```
SynapseHTTPError: 400 Client Error: 
No enum constant org.sagebionetworks.repo.model.schema.Type.["number","null"]
```

We should add a TODO and/or platform ticket to support this, but we'll have to remove that feature for now.

andrewelamb: @thomasyu888 I can confirm
```
"type": [
        "number",
        "null"
      ]
```
Is the problem. Changing it to:

```
    "RIN": {
      "description": "RNA Integrity Number",
      "oneOf": [
        {"type": "array"},
        {"type" : "null"}
      ]
    }
```
fixes the probem.

schematic/schemas/json_schema_component_generator.py
linglp: I think it is because we still use the old json schema class for the integration test?

schematic/schemas/json_schema_generator.py
andrewelamb: @thomasyu888 I'll start a ticket with Platform, but for now, I made a change to use `oneOf` instead of an array for multiple types.

thomasyu888: Thanks @andrewelamb - that worked! May want to bind the jsonschema too to a folder to check the output.

image

It seems like "type" has to be defined?

thomasyu888: Another find.
image

```json
    "libraryBatch": {
      "description": "Batch library was prepared in",
      "oneOf": [
        {
          "title": "string",
          "type": "string"
        },
        {
          "type": "null"
        }
      ]
    },
```

If you had a title to the "oneOf" types, it shows up nicer in synapse.  Maybe that is a DPE ticket (or done here), but I feel strongly that platform should just support `type: ["string", "null"]`

andrewelamb:
[Platform ticket](https://sagebionetworks.jira.com/browse/PLFM-8982)

schematic/schemas/json_schema_generator.py
andrewelamb: I've added the `title` keyword to all types in a `oneOf` array.

andrewelamb: @thomasyu888 I've bound several of the test schemas to folders [here](https://www.synapse.org/Synapse:syn66511767/files/).

I'm seeing several issues where the JSON Schema seems fine (I validated it using [this](https://jsonschema.dev/). But it doesn't work when bound to a Synapse entity via the ui. This includes the example you showed above where there wasn't a type.

andrewelamb: For example:
Screenshot from 2025-05-12 10-35-07
file:///home/alamb/Pictures/Screenshots/Screenshot%20from%202025-05-12%2010-34-42.png

andrewelamb: Screenshot from 2025-05-12 10-34-42

tests/data/expected_jsonschemas/expected.MockComponent.schema.json
```diff
@@ -1,264 +1,258 @@
...
         "type": "null"
       }
     },
-    "Check Float": {
-      "description": "TBD",
-      "minLength": 1,
-      "not": {
-        "type": "null"
-      }
+    "Check List Strict": {
```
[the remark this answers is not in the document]
linglp: This is also not showing if the field is optional or not. If `"Check List Strict"` is optional, I thought this should show something like: 
```
"oneOf": [
  {
    "type": "array",
    "items": {
      "enum": ["cd", "ab", "ef", "gh"]
    }
  },
  {
    "type": "null"
  }
]
```

tests/data/expected_jsonschemas/expected.MockComponent.schema.json
```diff
@@ -1,264 +1,258 @@
...
-    "Check NA": {
+    "Check Regex Single": {
       "description": "TBD",
-      "minLength": 1,
       "not": {
         "type": "null"
       }
     },
-    "Check Num": {
+    "Check Range": {
```
linglp, line 95: if seems like initially we have `Check Range` as required but now it gets changed to optional?

linglp (commented):
Thanks for the hard work, @andrewelamb. I haven’t finished the full review yet, but I wanted to leave some initial comments before reading further.

My main concern is that previously, validation rules were separate from the JSON Schema, so our focus was only on ensuring the validation logic worked as expected.

However, after the refactor, the JSON Schema has become an integral part of the validation process. This means that if the schema is constructed incorrectly, it could lead to incorrect validation downstream.

I think we need to consider how to write better tests to ensure that the generated JSON Schema enforces the intended validation behavior accurately. 

Let's use "Check Range" as an example. Previous we have: 
```
            GenerateError.generate_content_error(
                val_rule="inRange 50 100 error",
                attribute_name="Check Range",
                dmge=dmge,
                row_num=["3"],
                invalid_entry=["30"],
            )[0]
            in errors
```
this in the test ensures that when the input is 30, it raises an error because it is outside of 50-100 range. But I don't think we are testing if the new schema that we construct below can perform the same validation: 
```
Check Range: {
      "description": "TBD",
      "type": "number",
      "minimum": 50.0,
      "maximum": 100.0
    }
```
In general, we have tests to ensure that the attributes are being transformed as intended, but I’m not sure that this transformation aligns with our intentions for json schema validation.

tests/data/expected_jsonschemas/expected.Biospecimen.schema.json
```diff
@@ -1,44 +1,46 @@
...
       }
     },
+    "Tissue Status": {
+      "description": "TBD",
+      "oneOf": [
+        {
+          "enum": [
+            "Healthy",
+            "Malignant",
+            "None"
```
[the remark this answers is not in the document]
andrewelamb: This gets added to types:
```
{ "type": "null", "title": "null" }
```
when the attribute is not required, but `Tissue Status` is required.

tests/data/expected_jsonschemas/expected.MockComponent.schema.json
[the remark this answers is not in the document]
andrewelamb: `Check List Strict` doesn't have any valid values so it's not an enum in JSON Schema, also it's required.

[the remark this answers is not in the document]
andrewelamb: > I don't think this enforces there needs to be >=2 items in the list. An empty list would pass validation here which is not expected.

I'm interested in this. Should lists have a minimum? Should it be 2? @thomasyu888 @GiaJordan what do you think?

tests/data/expected_jsonschemas/expected.MockComponent.schema.json
andrewelamb: It doesn't look optional to me? It has type: "number", it doesn't have

```
"oneOf" : [
  {"type": "number},
  {"type": "null"}
]
```
And it's in the "required" list at the bottom.

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,682 @@
...
+        jsonld_path: str,
+        graph: nx.MultiDiGraph,
+    ) -> None:
+        self.jsonld_path = jsonld_path
+        self.graph = graph
+        self.dmge = DataModelGraphExplorer(self.graph)
+        self.dmr = DataModelRelationships()
+        self.rel_dict = self.dmr.relationships_dictionary
+
+    def get_json_validation_schema(
```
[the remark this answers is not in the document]
andrewelamb: Thsi is tested [here](https://github.com/Sage-Bionetworks/schematic/blob/SCHEMATIC-285/tests/unit/test_json_schema_generator.py#L118.)

schematic/schemas/json_schema_generator.py
thomasyu888: @andrewelamb You can either tackle it in this PR, or create a ticket to tackle that since it is more of a styling thing

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,682 @@
...
+    """Keeps track of needed information for processing a node in a graph"""
+
+    def __init__(self, initial_nodes: list[str]) -> None:
+        """
+        Arguments:
+            initial_nodes: A list of nodes to start processing
+        """
+        self.root_dependencies: list[str] = initial_nodes
+        self.nodes_to_process = self.root_dependencies.copy()
+        self.current_node = self.nodes_to_process.pop(0)
```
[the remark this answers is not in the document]
andrewelamb: I'll go ahead and add it, but it isn't needed since it will always be a `str` since `nodes_to_process` is a `list[str]`

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,682 @@
...
+
+        node_processor = NodeProcessor(root_dependencies)
+
+        json_schema = JSONSchema(
+            schema_id="http://example.com/" + schema_name,
+            title=schema_name,
+            description=self.dmge.get_node_comment(node_label=datatype),
+        )
+
+        while True:
```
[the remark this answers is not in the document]
andrewelamb: That is much better

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,682 @@
...
+      In the example data model the Patient component has the Diagnosis attribute.
+      The Diagnosis attribute has valid values of ["Healthy", "Cancer"].
+      The Cancer valid value is also an attribute that dependsOn on the
+        attributes Cancer Type and Family History
+      Cancer Type and Family History are attributes with valid values.
+      Therefore: When Diagnosis == "Cancer", Cancer Type and Family History should become required
+
+    Example conditional schema:
+        "if":{
+            "properties":{
```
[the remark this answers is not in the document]
andrewelamb: > "required":[
               "Diagnosis"
            ]

"Diagnosis" is already required in the "required" list see [here.](https://github.com/Sage-Bionetworks/schematic/blob/develop/tests/data/expected_jsonschemas/expected.Patient.schema.json#L133)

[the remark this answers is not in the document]
andrewelamb: > If we're planning to test binding JSON Schema to Synapse, I'd be interested to see whether Synapse correctly interprets conditional JSON Schema constructs like this one. It might be helpful to compile a list of concise conditional schema examples that we can bind to Synapse and use to systematically verify their behavior.

It does! See [this entity](https://www.synapse.org/Synapse:syn66511769).

tests/data/expected_jsonschemas/expected.MockComponent.schema.json
linglp: Sorry my bad. I saw it got listed in "required" here: https://github.com/Sage-Bionetworks/schematic/pull/1604/files/d4b52f71d5a868756276068bf841a6cd900c217e#diff-32abeaa43c00f1f8d0ed3b9b30145d4d2c66bc6550f9bf836006c73a2487e2b7R225

tests/data/expected_jsonschemas/expected.MockComponent.schema.json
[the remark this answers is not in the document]
SageGJ: > Should lists have a minimum? Should it be 2?
So the length comes into play because the list rule does two things

1. validates that the string entry is formatted like a list
2. parses the string entry into an actual `list` type in python

using `list like` vs `list strict` skips check 1 and only performs step 2. for `list like` the minimum length would be 1, whereas for `list strict` the minimum length would be 2. What the minimum length needs to be going forward depends on how jsonschema performs the list check during validation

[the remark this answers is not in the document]
andrewelamb: Just to be clear: In JSON Schema if a property has the `array` type it needs to be an actual array(list) not a string that can be broken into an array. `["a", "b", "c"]` vs `"a,b,c"`. The check that the `array` type does is just that the item is an array, nothing more. These all would pass: `[]`, `[null]`, `["a", "b", "c"]`, `["a,b,c"]`, `[1]`, `[True]`.

schematic/schemas/json_schema_component_generator.py
andrewelamb: @thomasyu888 I've changed it so that the CLI uses the new class. The [CLI integration test](https://github.com/Sage-Bionetworks/schematic/blob/develop/tests/integration/test_json_schema_generation.py) tests that the command works.

andrewelamb: @thomasyu888  I've added a [new integration test](https://github.com/Sage-Bionetworks/schematic/blob/SCHEMATIC-285/tests/integration/test_json_schema_generator.py#L95) that creates the schemas, uploads them in Synapse, bnds them to a folder, annotates them, and validates them.

schematic/schemas/json_schema_component_generator.py
```diff
@@ -258,51 +254,15 @@ def get_component_json_schema(
...
-            )
-
-        for attribute, value in self.component_json_schema["properties"].items():
-            if isinstance(value, dict) and "description" not in value:
-                # https://sagebionetworks.jira.com/browse/SCHEMATIC-284
-                # https://sagebionetworks.jira.com/browse/SCHEMATIC-283
-                value["description"] = self.dmge.get_node_comment(
-                    node_display_name=attribute
-                )
-
```
andrewelamb, line 305: This functionality was moved to JSONSchemaGenerator

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+        if self.current_node is None:
+            raise ValueError("Current node is None")
+        self.processed_nodes.append(self.current_node)
+
+
+class JSONSchemaGenerator:  # pylint: disable=too-few-public-methods
+    "Data Model Json Schema"
+
+    def __init__(
+        self,
```
andrewelamb, line 150: The init method could along could use a more thourough refactor

tests/utils.py
```diff
@@ -59,4 +59,4 @@ def json_files_equal(file1: str, file2: str) -> bool:
         bool: True if the JSON files are equal, False otherwise.
     """
     with open(file1) as f1, open(file2) as f2:
-        return dict_equal(json.load(f1), json.load(f2))
```
andrewelamb, line 62: This function was NOT detecting differences between JSON files.

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+        logger.info("JSON schema successfully generated from schema.org schema!")
+
+        json_schema_dict = json_schema.as_json_schema_dict()
+
+        if write_schema:
+            _write_data_model(json_schema_dict, schema_path, datatype, self.jsonld_path)
+
+        return json_schema_dict
+
+    def _process_node(
```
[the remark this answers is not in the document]
BryanFauble: For these comments is there a tech debt or follow up ticket getting created?

schematic/schemas/json_schema_component_generator.py
```diff
@@ -243,10 +239,10 @@ def get_component_json_schema(
         self,
     ) -> None:
```
BryanFauble, line 240: I usually have an aversion to have a method with the name `get_...` that doesn't return anything. It may not make sense at this point to go and modify this, but I would say that we should keep it in mind for future iterations.

It's find that a class method updates values on the class, but I would also expect that it returns a value so you could easily do:

```
the_value_want = my_class_instance.get_method_to_do_thing()
```

schematic/schemas/json_schema_component_generator.py
```diff
@@ -258,51 +254,15 @@ def get_component_json_schema(
...
             graph=metadata_model.graph_data_model,
         )
 
-        self.incomplete_component_json_schema = (
-            data_model_json_schema.get_json_validation_schema(
-                source_node=self.component, schema_name=schema_name
-            )
+        self.component_json_schema = json_schema_generator.create_json_schema(
+            datatype=self.component, schema_name=schema_name
         )
```
BryanFauble, line 264: The use of creating class instances is a bit puzzling to me. Generally, I'd only recommend using classes when the focus of the class is tracking the state of data, and interacting with that data in some way. The way that this is used, `create_json_schema` doesn't need to be a part of a `JSONSchemaGenerator` class, it just needs to be some function that you can invoke and get a response from.

To reinforce my thought on this, `JSONSchemaGenerator` is created 2 lines above, used here, and then left for garbage collection.

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
+"Data Model Json Schema"
```
BryanFauble, line 1: Is this a module docstring? Doesn't seem to add much IMO

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+    "num": "number",
+    "float": "number",
+    "int": "integer",
+    "bool": "boolean",
+}
+
+
+@dataclass
+class PropertyData:
+    """A Dataclass representing data about a JSON Schema property from its validation rules"""
```
BryanFauble, line 31: Can you expand on the description of what this PropertyData class is for?

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+    """A dataclass representing a JSON Schema"""
+
+    schema_id: str = ""
+    title: str = ""
+    schema: str = "http://json-schema.org/draft-07/schema#"
+    type: str = "object"
+    description: str = "TBD"
+    properties: dict[str, Any] = field(default_factory=dict)
+    required: list[str] = field(default_factory=list)
+    all_of: list[dict[str, Any]] = field(default_factory=list)
```
BryanFauble, line 50: Can you add a docstring to the class (or the fields themselves) that describes the purpose of each attribute?

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+
+    def as_json_schema_dict(self) -> dict[str, Any]:
+        """Returns class as a JSON Schema dictionary, with proper keywords"""
```
BryanFauble, line 53: nit: Proper formatted docstring:

```
        """
        Description of what is going on

        Returns:
            What is being returned
        """
```

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+            "all_of": "allOf",
+        }
+        for old_word, new_word in keywords_to_change.items():
+            json_schema_dict[new_word] = json_schema_dict.pop(old_word)
+        if len(self.all_of) == 0:
+            json_schema_dict.pop("allOf")
+        return json_schema_dict
+
+
+class NodeProcessor:
```
BryanFauble, line 67: nit: Unless this cannot be a dataclass I'd prefer to keep consistent behavior for classes introduced in this code.

BryanFauble, line 67: It would be great to have a diagram created that describes the traversal algorithm and processing algorithm being completed here.

Some diagram options:

1. https://mermaid.js.org/syntax/sequenceDiagram.html
2. https://mermaid.js.org/syntax/stateDiagram.html
3. https://mermaid.js.org/syntax/flowchart.html

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+    def create_json_schema(
+        self,
+        datatype: str,
+        schema_name: str,
+        schema_path: Union[str, None] = None,
+        write_schema: bool = True,
+    ) -> dict[str, Any]:
+        """
+        Consolidated method that aims to gather dependencies and value constraints across terms
+        / nodes in a schema.org schema and store them in a jsonschema /JSON Schema schema.
```
BryanFauble, line 169: Consider writing the docstrings without the previous state that the changes were made from in mind. I assume that "Consolidated method...." is meaning that this is taking a handful of functionality and squashing it together, but if I were to read this a year from now - I don't care that it was constructed from several functions. Rather, I care about what the function does today and right now.

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+        Returns:
+            JSON Schema as a dictionary.
+        """
+
+        # Gets the dependency nodes of the source node. These will be first nodes processed.
+        root_dependencies = self.dmge.get_adjacent_nodes_by_relationship(
+            node_label=datatype,
+            relationship=self.rel_dict["requiresDependency"]["edge_key"],
+        )
+        # if root_dependencies is empty it means that a class with name 'source_node' exists
```
BryanFauble, line 196: Is "class" the right term to use here?

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+        # in the schema, but it is not a valid component
+        if not root_dependencies:
+            raise ValueError(f"'{datatype}' is not a valid component in the schema.")
```
BryanFauble, line 199: Should this exception be expanded to include the notes from your comment above?

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+          In the above example {'Healthy': ['Diagnosis'], 'Cancer': ['Diagnosis']}
+    """
+    # The conditional_value will be the specific value that triggers the conditional dependency.
+    # When the watched_property == conditional_value, the conditional_property will become required
+    for conditional_value in sorted(reverse_dependencies[property_display_name]):
+        if conditional_value in range_domain_map:
+            properties = sorted(range_domain_map[conditional_value])
+            # watched_property is the property such that when the
+            # watched_property == conditional_value
+            # the conditional_property becomes required
```
BryanFauble, line 409: This seems to be repeating the same comment from a few lines above

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+                name=name,
+                property_data=property_data,
+                is_required=is_required,
+                description=description,
+            )
+
+    json_schema.properties.update(schema_property)
+
+    if is_required:
+        json_schema.required += [name]
```
BryanFauble, line 480: This is a good candidate to be a method on the JSONSchema class. Usually it's bad practice to update the values of an argument that you pass into a function

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+
+        logger.info(
+            "The JSON schema file can be inspected by setting the following "
+            "nested key in the configuration: (model > location)."
+        )
+    export_json(json_doc=json_schema_dict, file_path=json_schema_path, indent=2)
+
+
+def _set_conditional_dependencies(
+    json_schema: JSONSchema,
```
BryanFauble, line 350: Same note: https://github.com/Sage-Bionetworks/schematic/pull/1604/files#r2089909192

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+    return schema
+
+
+def _create_simple_property(
+    name: str, property_data: PropertyData, is_required: bool, description: str
+) -> dict[str, Any]:
+    """
+    Creates a JSON Schema property
+    If a property_type is given the type is added to the schema
+    If a property_type is not given and is_required is  "not: {type:null} is added
```
BryanFauble, line 618: This line seems incomplete

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+        if type_rule:
+            property_type = TYPE_RULES.get(type_rule)
+
+        range_rule = _get_in_range_rule_from_rule_list(validation_rules)
+        if range_rule:
+            if property_type not in ["number", "integer"]:
+                property_type = "number"
+            range_min, range_max = _get_ranges_from_range_rule(range_rule)
+
+    return PropertyData(property_type, is_array, range_min, range_max)
```
BryanFauble, line 680: It may make more sense to add `validation_rules` as an attribute on `PropertyData` and accomplish this logic within a `__post__init__` function

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+        The min and max form the rule
+    """
+    range_min: Union[float, None] = None
+    range_max: Union[float, None] = None
+    parameters = rule.split(" ")
+    if len(parameters) > 1 and parameters[1].isnumeric():
+        range_min = float(parameters[1])
+    if len(parameters) > 2 and parameters[2].isnumeric():
+        range_max = float(parameters[2])
+    return (range_min, range_max)
```
BryanFauble, line 702: Is this function being used?

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+
+    Raises:
+        ValueError: When more than one inRange rule is found
+
+    Returns:
+        The inRange rule if one is found, or None
+    """
+    in_range_rules = [rule for rule in rule_list if rule.startswith("inRange")]
+    if len(in_range_rules) > 1:
+        raise ValueError("Found more than one inRange rule in validation rules")
```
BryanFauble, line 721: We should probably include the `in_range_rules` that were found so it's more obvious when debugging what to look for.

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,747 @@
...
+    Raises:
+        ValueError: When more than one type rule is found
+
+    Returns:
+        The type rule if one is found, or None
+    """
+    rule_list = [rule.split(" ")[0] for rule in rule_list]
+    type_rules = [rule for rule in rule_list if rule in TYPE_RULES]
+    if len(type_rules) > 1:
+        raise ValueError("Found more than one type rule in validation rules")
```
BryanFauble, line 744: Same: https://github.com/Sage-Bionetworks/schematic/pull/1604/files#r2089913988

tests/integration/test_json_schema_generator.py
```diff
@@ -0,0 +1,331 @@
...
+):
+    """This yields a Synapse JSON Schema uri"""
+    schema = js_generator.create_json_schema(
+        "Biospecimen", schema_name="Biospecimen_validation", write_schema=False
+    )
+    schema_name = upload_schema_to_synapse2(schema, synapse, schema_org, schema_version)
+    js = synapse.service("json_schema")
+    uri = f"{schema_org}-{schema_name}-{schema_version}"
+    yield uri
+    js.delete_json_schema(f"{schema_org}-{schema_name}")
```
BryanFauble, line 88: I can't recall exactly what would occur, but during the `yield` if an exception or failure occurs in the test would this ever return and hit the `delete_json_schema` call? If not, this should be wrapping in a `try: finally:` block

tests/unit/test_json_schema_generator.py
```diff
@@ -0,0 +1,875 @@
...
+    """Tests for _set_conditional_dependencies"""
+    json_schema = JSONSchema()
+    _set_conditional_dependencies(
+        json_schema=json_schema,
+        conditional_property="property_name",
+        property_display_name="property_name",
+        reverse_dependencies=reverse_dependencies,
+        range_domain_map=range_domain_map,
+    )
+    assert json_schema == expected_schema
```
BryanFauble, line 328: This seems like an incorrect assertion. Types wouldn't match since a JSONSchema class is not  a `dict[str, Any]`

tests/unit/test_json_schema_generator.py
```diff
@@ -0,0 +1,875 @@
+"""Tests for JSON Schema generation"""
```
BryanFauble, line 1: Many of the tests here were a bit hard to follow. There is room for improving clarity and expectations of what each test is setting up (Initial conditions), doing (Your actions), and accomplishing (Your results).

This kind of clarity is why I am a huge fan of following a GIVEN-WHEN-THEN structure, but we don't have to explicitly follow that.

BryanFauble (changes requested):
Great progress!

tests/integration/test_json_schema_generator.py
andrewelamb: @BryanFauble Could you elaborate? I don't think I understand the problem.

BryanFauble: Sure, let's look at a simple example where we have this fixture:
```
@pytest.fixture(name="my_fixture", scope="module")
def my_fixture():
    """doc"""

    # Setting up the yield
    uri = "aaaaaaa"

    # Yield control to the tests
    yield uri

    # Control returns back to this function
    js.delete_json_schema(f"{schema_org}-{schema_name}")
```

Then suppose we have a test that uses the fixture:

```
def test_whatever(my_fixture):
    if True:
        raise ValueError("")
```

What I am concerned about is that `# Control returns back to this function` may not be executed because pytest is going to start failing out and may not actually return control so that you can cleanup the resource with `js.delete_json_schema(f"{schema_org}-{schema_name}")`.


A try, finally block ensures that the finally block is always going to be evaluated even if exceptions still need to bubble up to pytest or the interpreter.

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,822 @@
...
+        Updates the property dict
+
+        Args:
+            property_dict: The property dict to add to the properties dict
+        """
+        self.properties.update(property_dict)
+
+
+@dataclass
+class NodeProcessor:
```
andrewelamb, line 219: This class needs a better name. I'm open to suggestions.

tests/unit/test_json_schema_generator.py
andrewelamb: @BryanFauble I've added `GIVEN-WHEN-THEN` statements to the tests where it seem to make sense to add.

schematic/schemas/json_schema_generator.py
andrewelamb: It is! It's now called from `PropertyClass.__post_init__` (previously `_create_property_data_from_validation_rules)

tests/integration/test_json_schema_generator.py
andrewelamb: Oh I see.

How would the fixture interact with the try/finally block?

```
@pytest.fixture(name="my_fixture", scope="module")
def my_fixture():
    """doc"""
    try:
        # Setting up the yield
        uri = "aaaaaaa"

        # Yield control to the tests
        yield uri

    finally:
            # Control returns back to this function
            js.delete_json_schema(f"{schema_org}-{schema_name}")
      
```

Will the `try` block be aware of exceptions that happen in the test the uri is yielded to?

schematic/schemas/json_schema_generator.py
andrewelamb: @BryanFauble apologies, that link doesn't seem to lead anywhere. Is it a resolved comment?

schematic/schemas/json_schema_generator.py
andrewelamb: @BryanFauble TBH this comment is leftover from the previous class. I had trouble understanding what it means. @linglp @SageGJ Do you know why a class would have no root dependencies? Is that a malformed data-model?

tests/data/expected_jsonschemas/expected.MockComponent.schema.json
```diff
@@ -1,264 +1,258 @@
...
+    "Check Int": {
       "description": "TBD",
-      "minLength": 1,
-      "not": {
-        "type": "null"
-      }
+      "type": "integer"
     },
-    "Check Recommended": {
+    "Check List Enum Strict": {
```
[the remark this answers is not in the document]
andrewelamb: @linglp We're going to leave it like this for now. In the meantime, we will be getting feedback from the various DCCs on how they want this to work and we can update it in the future.

schematic/schemas/json_schema_generator.py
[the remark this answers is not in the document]
andrewelamb: @BryanFauble I think in order to get the other devs unstuck I'll be adding a follow-up ticket.

schematic/schemas/json_schema_component_generator.py
andrewelamb: @BryanFauble If I had more time, I think `JSONSchemaGenerator` would have eventually gotten refactored into a function. I was partially there by moving most of state tracking into `NodeProcessor`. There will be a follow-up ticket to finish doing this.

schematic/schemas/json_schema_component_generator.py
andrewelamb: @BryanFauble @thomasyu888 There's enough interest in turning JSONSchemaComponentGenerator into a function. I'll create a ticket for doing so.

andrewelamb: https://sagebionetworks.jira.com/browse/SCHEMATIC-302

schematic/schemas/json_schema_generator.py
andrewelamb: https://sagebionetworks.jira.com/browse/SCHEMATIC-303

schematic/schemas/json_schema_generator.py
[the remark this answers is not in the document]
andrewelamb: https://sagebionetworks.jira.com/browse/SCHEMATIC-303

linglp (commented):
I saw a lot of progress has been made! I like that there's now test to make sure that conditional dependencies work after we bind a json schema to a synapse entity. But do we have plans to ensure other things that we define in the schema (i.e. enum: 
```
   "Tissue Status": {
      "description": "TBD",
      "enum": [
        "Malignant",
        "None",
        "Healthy"}
```
are also working as expected? For example, for this column, it should raise an error when we add an item that is not in the list defined above.

schematic/schemas/json_schema_generator.py
BryanFauble: > This is a good candidate to be a method on the JSONSchema class. Usually it's bad practice to update the values of an argument that you pass into a function

thomasyu888:
> are also working as expected? For example, for this column, it should raise an error when we add an item that is not in the list defined above.

@linglp This is a good question.  we want to be careful of not testing Synapse jsonschema services.  If they say that the enum works, then we need to trust that it does. If it doesn't work as expected, then it would be a platform ticket.

tests/utils.py
thomasyu888: Do you have an example?

tests/integration/test_json_schema_generator.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    This the version to give all created test schemas
+    """
+    yield "0.0.1"
+
+
+@pytest.fixture(name="js_generator", scope="module")
+def fixture_js_generator() -> Generator[JSONSchemaGenerator, None, None]:
+    """Yields a JSONSchemaGenerator with the example data model"""
+    metadata_model = MetadataModel(
```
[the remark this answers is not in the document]
andrewelamb: The only metadata_model I'm seeing in conftest is in a function and not a fixture. I think it's fine to have one here, since its trivial to to create one.

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,825 @@
...
+            # Determine if current node has conditional dependencies that need to be set
+            if node_display_name in node_processor.reverse_dependencies:
+                _set_conditional_dependencies(
+                    json_schema,
+                    node_processor.current_node,
+                    node_display_name,
+                    node_processor.reverse_dependencies,
+                    node_processor.valid_values_map,
+                )
+                is_node_required = False
```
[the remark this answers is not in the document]
andrewelamb: Yeah, I should add a comment here.  This is because if the current node the function is processing is a reverse dependency, we want it to be a conditional dependency. I included this example in one of the docstrings:

```
      In the example data model the Patient component has the Diagnosis attribute.
      The Diagnosis attribute has valid values of ["Healthy", "Cancer"].
      The Cancer valid value is also an attribute that dependsOn on the
        attributes Cancer Type and Family History
      Cancer Type and Family History are attributes with valid values.
      Therefore: When Diagnosis == "Cancer", Cancer Type and Family History should become required
```
In this case, when we go to `_set_property`, we want to make it's not_required in the schema, but it becomes required ONLY IF Diagnosis == "Cancer", which is what the conditional property sets.

tests/integration/test_json_schema_generator.py
```diff
@@ -0,0 +1,311 @@
...
+        data_model_labels="class_label",
+    )
+    js_generator = JSONSchemaGenerator(
+        jsonld_path=metadata_model.inputMModelLocation,
+        graph=metadata_model.graph_data_model,
+    )
+    yield js_generator
+
+
+@pytest.fixture(name="synapse", scope="module")
```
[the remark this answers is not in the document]
andrewelamb: This is because the one in `conftest.py` is set to `scope="function"`. I need it to be set to `scope="module"` (session would also work). This is because the six schema fixtures are set `scope="module"`.

 This has two important reasons. One, it means that the schemas are only bound once each rather than each time they are called by a function which saves us some time making Synapse calls. Two, it ensures the cleanup of the fixtures happens in the correct order. The Synapse Folder fixture is set to function, which means it gets created and tore down each function call which is wanted to avoid collisions. These folders are bound with the JSON Schemas I mentioned above. These need to be unbound BEFORE the Schema is destroyed.

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,826 @@
...
+        )
+        # if root_dependencies is empty it means that a class with name 'source_node' exists
+        # in the schema, but it is not a valid component
+        if not root_dependencies:
+            raise ValueError(f"'{datatype}' is not a valid component in the schema.")
+
+        node_processor = NodeProcessor(sorted(root_dependencies))
+
+        json_schema = JSONSchema(
+            schema_id="http://example.com/" + schema_name,
```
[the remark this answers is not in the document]
andrewelamb: That's correct, but I believe that belongs in the `$schema` field, not the `$id` field, see [here](https://json-schema.org/learn/getting-started-step-by-step). If you look at the [JSONSchema class](https://github.com/Sage-Bionetworks/schematic/blob/SCHEMATIC-285/schematic/schemas/json_schema_generator.py#L168), you'll see it's the default for that attribute.

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,826 @@
...
+              when it is hosted on the Internet).
+            schema_path: Where to save the JSON Schema file
+            write_schema: whether or not to write the schema as a json file
+
+        Returns:
+            JSON Schema as a dictionary.
+        """
+
+        # Gets the dependency nodes of the source node. These will be first nodes processed.
+        root_dependencies = self.dmge.get_adjacent_nodes_by_relationship(
```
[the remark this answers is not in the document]
andrewelamb: Looks like you are right.

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,826 @@
...
+        logger.info("JSON schema successfully generated from schema.org schema!")
+
+        json_schema_dict = json_schema.as_json_schema_dict()
+
+        if write_schema:
+            _write_data_model(json_schema_dict, schema_path, datatype, self.jsonld_path)
+
+        return json_schema_dict
+
+    def _process_node(
```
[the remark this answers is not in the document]
andrewelamb: [Mermaid chart](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4135059461/SCHEMATIC-285#Mermaid-chart-for-JSONSchemaGenerator._process_node)

tests/integration/test_json_schema_generator.py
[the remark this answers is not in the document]
andrewelamb: I've created a fixture in conftest that uses this function, and now the test above uses it.

tests/data/expected_jsonschemas/expected.Biospecimen.schema.json
```diff
@@ -5,39 +5,41 @@
...
-    "Sample ID": {
+    "SampleID": {
       "description": "TBD",
-      "minLength": 1,
       "not": {
         "type": "null"
       }
     },
-    "Tissue Status": {
+    "TissueStatus": {
```
SageGJ, line 24: You mentioned that now the generator is hardcoded to use ClassLabel over display name. given that we need to support both where do you see the updates being made to add support for display names?

schematic/schemas/json_schema_generator.py
```diff
@@ -0,0 +1,826 @@
...
+        The inRange rule if one is found, or None
+    """
+    in_range_rules = [rule for rule in rule_list if rule.startswith("inRange")]
+    if len(in_range_rules) > 1:
+        raise ValueError(
+            "Found more than one inRange rule in validation rules: ", rule_list
+        )
+    if len(in_range_rules) == 0:
+        return None
+    return in_range_rules[0]
```
SageGJ, line 121: Should we also get `protectAges` rules here (now or later) since its also a range check?

schematic/schemas/json_schema_generator.py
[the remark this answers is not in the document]
SageGJ: the mermaid chart doesn't seem clear to me, is there a way we can more effectively show the relationships between classes?

tests/integration/test_json_schema_generator.py
```diff
@@ -0,0 +1,311 @@
...
+                "#: required key [CancerType] not found",
+                "#: required key [FamilyHistory] not found",
+            ],
+            "patient_json_schema",
+        ),
+    ],
+    ids=[
+        "Patient:valid #1",
+        "Patient:valid #2",
+        "Patient:CancerType and FamilyHistory required when Diagnosis==Cancer",
```
SageGJ, line 267: Can the id here be updated to specify that this case is with invalid data?

tests/unit/test_data_model_graph_explorer.py
```diff
@@ -0,0 +1,46 @@
...
+        ),
+        (
+            "Biospecimen",
+            "requiresDependency",
+            ["Component", "PatientID", "SampleID", "TissueStatus"],
+        ),
+        # requiresDependency will get an attributes dependencies
+        ("Cancer", "requiresDependency", ["CancerType", "FamilyHistory"]),
+    ],
+)
```
SageGJ, line 36: Here and in the other file: can ids be added to the parametrizations where there are none?

schematic/schemas/json_schema_generator.py
andrewelamb: I'll create a separte ticket for `protectAges`

tests/data/expected_jsonschemas/expected.Biospecimen.schema.json
andrewelamb: I have some easy modifications in mind to make whoever's life easier that ends up working on the display name feature. I'll add some comments as well.

schematic/schemas/json_schema_generator.py
thomasyu888: @andrewelamb / @SageGJ Thanks for raising this.  In my opinion, this is really a combination of rules

1. integer range
2. Rule that says, create a governance ticket

number 2 is a Synapse specific ask since we do not implement the actual validation behind the scenes now.  Due to what JSONSchema service supports today, a range of <89 would cause annotations with >=89 become "invalid".  we would need to clarify if that is the behavior or if there needs to be a platform feature.

TLDR: Re-design is required with internal partners

SageGJ: I agree. I remember in the past the working knowledge then was that the desired behavior for "invalid" ages was for them to not pass through or be displayed at all. Which would require a check on our part and coordination with platforms. Also just for clarity, the current implementation of the rule considers "valid" ages to be 18<=x<=89, not <=89

tests/integration/test_json_schema_generator.py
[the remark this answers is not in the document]
linglp: Okay got it. But if that's the case, should we move the one that you create to conftest.py so that we have two fixtures with different scopes in one place?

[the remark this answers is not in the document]
andrewelamb: I think thats a fine solution.

schematic/schemas/json_schema_generator.py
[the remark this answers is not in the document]
linglp: Great. Thanks for the explanation. Could you summarize that to a comment and add it here?

linglp:
@andrewelamb Thanks for the update. Could you summarize the "display name" feature and write it under part of "problem" in the PR description? Currently the PR description only mentioned the tech debt.

schematic/schemas/json_schema_generator.py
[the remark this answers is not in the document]
linglp: Yeah, thanks Andrew. I read more about $id, and I think it just assigns a unique identifier and sets the base URI, so it probably doesn't matter that we don't have "draft 7 schema" in the id. But I am not sure if the link that you provided is for draft 7 schema. I saw that resources for draft-7 schema are here: https://json-schema.org/draft-07#draft-07-documents

andrewelamb:
@linglp  thanks for the reminder!

schematic/schemas/json_schema_generator.py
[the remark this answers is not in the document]
andrewelamb: The `_process_node` function has been totally redone and simplified.

tests/data/example.model.csv
thomasyu888, line 1: What changed about this manifest?

thomasyu888 (commented):
Looking great @andrewelamb , if this is ready for review, please push it into "in review" state and also please do fix the pre-commit error

tests/data/example.model.jsonld
thomasyu888, line 1: Lets make sure all of the manual integration tests that need to be run also passes with these changes

tests/integration/test_create_json_schema.py
```diff
@@ -0,0 +1,296 @@
...
+        datatype="Biospecimen",
+        schema_name="Biospecimen_validation",
+        write_schema=False,
+        use_property_display_names=False,
+    )
+    schema_name = upload_schema_to_synapse(schema, synapse_module_scope, schema_org, schema_version)
+    js = synapse_module_scope.service("json_schema")
+    uri = f"{schema_org}-{schema_name}-{schema_version}"
+    yield uri
+    js.delete_json_schema(f"{schema_org}-{schema_name}")
```
thomasyu888, line 70: If you never call next() again or don’t explicitly close or exhaust the generator, the cleanup line will not run.

tests/data/example.model.csv
andrewelamb: I added a bunch of attributes to the MockComponent datatype for testing

andrewelamb:
@thomasyu888 Everyone is welcome to look/comment now, but I want to take tomorrow to make a final run through and nitpick myself. I'll move it out of draft by the end of the day.

tests/integration/test_create_json_schema.py
andrewelamb: @thomasyu888 Yep, you are correct. I've replaced these with finalizers.

tests/integration/test_json_schema_generator.py
andrewelamb: @BryanFauble I've replaced the post-yield teardowns with finalizers.

schematic/schemas/create_json_schema.py
```diff
@@ -0,0 +1,895 @@
...
+            type_rule = _get_type_rule_from_rule_list(validation_rules)
+            if type_rule:
+                self.type = TYPE_RULES.get(type_rule)
+
+            range_rule = _get_in_range_rule_from_rule_list(validation_rules)
+            if range_rule:
+                if self.type not in ["number", "integer", None]:
+                    raise ValueError(
+                        "Validation rules bust be either 'int' or 'num' when using the inRange rule"
+                    )
```
BryanFauble, line 190: It probably makes sense to include `self.type` in the ValueError exception

schematic/schemas/create_json_schema.py
```diff
@@ -0,0 +1,895 @@
...
+        use_valid_value_display_names: If True, the valid_values in the JSONSchema
+          will be written using node display names
+
+    Returns:
+        JSON Schema as a dictionary.
+    """
+    graph_state = GraphTraversalState(dmge, datatype)
+
+    json_schema = JSONSchema(
+        schema_id="http://example.com/" + schema_name,
```
BryanFauble, line 520: Will this be filled in from somewhere eventually instead of using the static `http://example.com/`?

schematic/schemas/create_json_schema.py
```diff
@@ -0,0 +1,895 @@
...
+        if not graph_state.is_current_node_processed():
+            _process_node(
+                json_schema=json_schema,
+                graph_state=graph_state,
+                use_property_display_names=use_property_display_names,
+                use_valid_value_display_names=use_valid_value_display_names,
+            )
+        graph_state.move_to_next_node()
+
+    logger.info("JSON schema successfully generated from schema.org schema!")
```
BryanFauble, line 535: The usage of the logger is light throughout this process. I would like to see more debug log messages around so we can more effectively watch the process as it is executing & when we need to debug what's going on we could get away with looking at the debug logs instead of needing to run the code locally in a debugger session.

schematic/schemas/create_json_schema.py
```diff
@@ -0,0 +1,895 @@
...
+            data_model_path=jsonld_path, source_node=name
+        )
+        json_schema_dirname = os.path.dirname(json_schema_path)
+        if json_schema_dirname != "":
+            os.makedirs(json_schema_dirname, exist_ok=True)
+
+        logger.info(
+            "The JSON schema file can be inspected by setting the following "
+            "nested key in the configuration: (model > location)."
+        )
```
BryanFauble, line 619: I am a bit confused about this log message. If I am reading this code correctly, the JSON Schema file is created in the same directory as the `json_schema_path`, but the message seems to imply that the file won't be available unless you set the location?

schematic/schemas/create_json_schema.py
```diff
@@ -0,0 +1,895 @@
...
+        }
+
+    Arguments:
+        node: The node to make the property of
+
+    Returns:
+        JSON object
+    """
+
+    array_dict: dict[str, Any] = {"type": "array", "title": "array"}
```
BryanFauble, line 806: If we can avoid `Any` we should try. In this case I think the type is `dict[str, Union[str, dict[str, Union[str, float]]]]` and `items` is `dict[str, Union[str, float]]`.


This same comment applies to a bunch of the cases in the argument and return type hints that use `Any`

tests/conftest.py
```diff
@@ -258,6 +258,27 @@ def syn(syn_token) -> Synapse:
     return syn
 
 
+@pytest.fixture(name="synapse_module_scope", scope="module")
+def fixture_synapse_module_scope() -> Generator[Synapse, None, None]:
```
BryanFauble, line 262: This is not a Generator return type

tests/integration/test_create_json_schema.py
```diff
@@ -0,0 +1,371 @@
...
+        use_property_display_names=False,
+    )
+    schema_name = upload_schema_to_synapse(
+        schema, synapse_module_scope, schema_org, schema_version
+    )
+
+    def delete_schema():
+        js.delete_json_schema(f"{schema_org}-{schema_name}")
+
+    request.addfinalizer(delete_schema)
```
BryanFauble, line 79: This `def delete_schema` is duplicated a bunch. If that could be extracted out and re-used it'll reduce some duplication, but it isn't a huge issue

tests/integration/test_create_json_schema.py
```diff
@@ -0,0 +1,371 @@
...
+    # WHEN validating the annotation against the schema
+    results = js.validate(synapse_folder)
+    # THEN the annotations were valid
+    assert results["isValid"] == is_valid
+    # OR THEN the annotations were invalid
+    if not results["isValid"]:
+        # THEN the results contain "allValidationMessages"
+        assert "allValidationMessages" in results
+        # THEN the "allValidationMessages" match the expected messages
+        assert sorted(results["allValidationMessages"]) == messages
```
BryanFauble, line 371: It's great to see the GIVEN WHEN THEN used here. One small note when I usually have multiple conditions is that I'll use `AND` like:

```suggestion
    # GIVEN a schema in Synapse
    test_schema_uri = request.getfixturevalue(schema_fixture)

    # WHEN binding the schema to a folder
    js.bind_json_schema(test_schema_uri, synapse_folder)

    # AND annotating the folder with a JSON instance of the datatype
    existing_annotations = syn.get_annotations(synapse_folder)
    with open(instance_path, encoding="utf-8") as instance_file:
        instance = json.load(instance_file)
    existing_annotations.update(instance)
    syn.set_annotations(annotations=existing_annotations)
    sleep(4)

    # AND validating the annotation against the schema
    results = js.validate(synapse_folder)
    # THEN the annotations were valid
    assert results["isValid"] == is_valid
    # OR THEN the annotations were invalid
    if not results["isValid"]:
        # AND the results contain "allValidationMessages"
        assert "allValidationMessages" in results
        # AND the "allValidationMessages" match the expected messages
        assert sorted(results["allValidationMessages"]) == messages
```

tests/unit/test_create_json_schema.py
```diff
@@ -0,0 +1,1057 @@
...
+# pylint: disable=too-many-positional-arguments
+
+
+@pytest.fixture(name="test_directory", scope="session")
+def fixture_test_directory(request) -> Generator[str, None, None]:
+    """Yields a directory for creating test jSON Schemas in"""
+    test_folder = "tests/data/json_schema_generator_output"
+
+    def delete_folder():
+        rmtree(test_folder)
```
BryanFauble, line 42: This is likely going to cause some potential race conditions. When using xdist to run tests even `session` scoped fixtures get copies made. In this case the first worker to finish is going to call `rmtree` to delete all data from the test folder. 

This should probably use a UUID4 string in the folder name to prevent collisions from concurrent tests.

tests/unit/test_create_json_schema.py
```diff
@@ -0,0 +1,1057 @@
...
+        # If there are any list type validation rules and one type validation rule
+        #  then is_array is set to True, and the type is set to the
+        #  JSON Schema equivalent of the validation rule
+        ("CheckListString", "string", True, None, None),
+        # If there is an inRange rule the min and max will be set
+        ("CheckRange", "number", False, 50, 100),
+    ],
+    ids=["CheckNone", "CheckString", "CheckList", "CheckListString", "CheckRange"],
+)
+def test_node(
```
BryanFauble, line 167:
```suggestion
def test_node_init(
```

tests/unit/test_create_json_schema.py
andrewelamb: Oh that's interesting. I'll fix that.

BryanFauble (approved):
Awesome changes. I left my last round of review comments, but structurally the code looks good.

Due to the size of changes let's see if we can get 1 final review from @linglp or @SageGJ

schematic/schemas/create_json_schema.py
andrewelamb: I agree that those are more specific and accurate return types. The reason I left them more generic is that other functions could potentiall modify those, and then the type becomes incorrect, and MYPY will throw a type error.

tests/integration/test_create_json_schema.py
andrewelamb: I'll try. I need to experiment with how finalizers work. Can I define them outside the fixture?

BryanFauble: I'm not sure tbh, and it isn't major - So I'd rather that you spend time working on other items and leave this

schematic/schemas/create_json_schema.py
BryanFauble: I'm not seeing cases where it needs to truly be `Any` from what I am seeing. What items did you have in mind that MYPY might throw errors from?

schematic/schemas/create_json_schema.py
andrewelamb: AFAIK this field has no effect on JSON Schema validaiton. So it's really up to what users want. @thomasyu888 What do you think? This becomes the `$id` in the final schema. It's described [here](https://json-schema.org/learn/getting-started-step-by-step#create-a-schema-definition) as :+1: 

```
$id: sets a URI for the schema. You can use this unique URI to refer to elements of the schema from inside the same document or from external JSON documents.
```

Should we allow users to set it via the CLI?

schematic/schemas/create_json_schema.py
andrewelamb: For example:

```
create_schema_dict() -> dict[str, Union[str, int, float]]:
    return {
         "item1": "x",
         "item2":  1,
         "item3":  2.0
    }

add_items_to_dict() -> dict[str, Union[str, int, float, list[str]]]:
    dict_item =  create_schema_dict()
    dict_item["array_item"] = ["y", "z"] <----- mypy type error (cannot assign an array to type of dict[str, Union[str, int, float]])
    return( dict_item)
```

The above is the type of error I was trying to avoid.  You can work around it by doing:

```
create_schema_dict() -> dict[str, Union[str, int, float]]:
    return {
         "item1": "x",
         "item2":  1,
         "item3":  2.0
    }

add_items_to_dict() -> dict[str, Union[str, int, float, list[str]]]:
    dict_item =  create_schema_dict()
    new_dict_item: dict[str, Union[str, int, float, list[str]]]
    new_dict_item["array_item"] = ["y", "z"] <----- mypy type error (cannot assign an array to type of dict[str, Union[str, int, float]])
    return(new_dict_item)
```
but its kind of annoying.

Ideally we didn't have to deal with Python3.9 and we could have this be our type for all these JSON objects:
`JSON: TypeAlias = dict[str, "JSON"] | list["JSON"] | str | int | float | bool | None`

schematic/schemas/create_json_schema.py
thomasyu888: Let's not allow it until enough people ask for it. Thanks!

tests/data/example.model.csv
```diff
@@ -1,55 +1,64 @@
...
+Check Match None,,,,,TRUE,DataProperty,,,matchNone MockComponent.checkMatchNone set error
+Check Match at Least values,,,,,TRUE,DataProperty,,,matchAtLeastOne MockComponent.checkMatchatLeastvalues value
+Check Match Exactly values,,,,,TRUE,DataProperty,,,matchExactlyOne MockComponent.checkMatchExactlyvalues value
+Check Match None values,,,,,TRUE,DataProperty,,,matchNone MockComponent.checkMatchNonevalues value error
+Check Recommended,,,,,FALSE,DataProperty,,,recommended
+Check Ages,,,,,TRUE,DataProperty,,,protectAges
+Check Unique,,,,,TRUE,DataProperty,,,unique error
+Check Range,,,,,TRUE,DataProperty,,,inRange 50 100 error
+Check Date,,,,,TRUE,DataProperty,,,date
+Check NA,,,,,TRUE,DataProperty,,,int::IsNA
```
SageGJ, line 60: The `MockComponent` component is intended primarly for validation rules testing, to ensure that the rules all work as expected and return the expected error messages. Are these new attributes intended for this use as well? If so I would add appropriate metadata to mock manifests used in validation testing and update the tests to look for the new errors. If not, I would move the new attributes to a separate component to avoid confusion in the future.

schematic/schemas/json_schema_component_generator.py
```diff
@@ -243,65 +239,27 @@ def get_component_json_schema(
...
+            schema_name=self.component + "_validation",
             jsonld_path=metadata_model.inputMModelLocation,
-            graph=metadata_model.graph_data_model,
-        )
-
-        self.incomplete_component_json_schema = (
-            data_model_json_schema.get_json_validation_schema(
-                source_node=self.component, schema_name=schema_name
-            )
+            use_property_display_names=False,
```
SageGJ, line 260: Currently with the way `create_json_schema` is called this would utilize property class labels and valid value display names.
Is the intention that:

- valid value display names are used the vast majority of the time and users can change the behavior if they wish
- `use_property_display_names` was added as a parameter because it should be configurable alongside `use_property_display_names` for similar flexibility needs
- `use_valid_value_display_names` is not used here because it should always be used
- or something else

Also, will there be cases where we will want to use different naming conventions for properties and valid values or should they always be the same?

tests/data/example.model.jsonld
SageGJ, line 1: Thanks for including both the csv and jsonld

tests/integration/test_create_json_schema.py
```diff
@@ -0,0 +1,371 @@
...
+from schematic.schemas.data_model_graph import DataModelGraph, DataModelGraphExplorer
+from schematic.schemas.data_model_parser import DataModelParser
+
+
+@pytest.fixture(name="schema_org", scope="module")
+def fixture_schema_org() -> str:
+    """
+    This returns the Synapse org the test Schemas will be created at
+    """
+    return "dpetest"
```
SageGJ, line 25: Will this need to be changed in the future/within tests? Just curious why a fixture over a global

tests/unit/test_create_json_schema.py
```diff
@@ -0,0 +1,1058 @@
...
+    ids=[
+        "Biospecimen",
+        "BulkRNA-seqAssay",
+        "MockComponent",
+        "MockFilename",
+        "MockRDB",
+        "Patient",
+    ],
+)
+def test_create_json_schema(
```
SageGJ, line 472: can this be renamed to `test_create_json_schema_with_class_labels` so that it can be easily differentiated from `test_create_json_schema_with_display_names`

tests/integration/test_create_json_schema.py
```diff
@@ -0,0 +1,371 @@
+"""Tests for JSON Schema generation"""
```
SageGJ, line 1: Could the docstrings for this and `tes_create_json_schema.py` be more specific as to what all each covers?

schematic/schemas/create_json_schema.py
```diff
@@ -0,0 +1,896 @@
...
+        if validation_rules:
+            if rule_in_rule_list("list", validation_rules):
+                self.is_array = True
+
+            type_rule = _get_type_rule_from_rule_list(validation_rules)
+            if type_rule:
+                self.type = TYPE_RULES.get(type_rule)
+
+            range_rule = _get_in_range_rule_from_rule_list(validation_rules)
+            if range_rule:
```
linglp, line 186: nit: The logic here can maybe improved for clarity? How about: 
```
if range_rule:
    if self.type is None:
        self.type = "number"
    elif self.type not in ["number", "integer"]:
        raise ValueError(
            "Validation type must be either 'number' or 'integer' when using the inRange rule, "
            f"but got: {self.type}"
        )
```

schematic/schemas/json_schema_component_generator.py
andrewelamb: > Currently with the way `create_json_schema` is called this would utilize property class labels and valid value display names. Is the intention that:
> * valid value display names are used the vast majority of the time and users can change the behavior if they wish

Yes I'm not sure why someone would want to use class_labels for valid values, but its there just in case.


> * `use_property_display_names` was added as a parameter because it should be configurable alongside `use_property_display_names` for similar flexibility needs

Yes

> * `use_valid_value_display_names` is not used here because it should always be used
> * or something else

Yeah unless there's a need for it, I don't think we need to expose it at the CLI. 


> Also, will there be cases where we will want to use different naming conventions for properties and valid values or should they always be the same?

Nope they may be different. As I mentioned above I think valid_values will 99.9% of the time be display names. If we're doing validaiton in Synapse via annotaitons then properties will have to be class-labels. So in the most common use-case in the near future properties will be class-label and valid-values will be display names.

linglp (approved):
Overall it looks good. Thanks for the hard work. I can approve with just some minor comments.

tests/integration/test_create_json_schema.py
andrewelamb: @SageGJ TBH I don't remember the rationale for making them fixtures, this and the version can both be globals.

schematic/schemas/create_json_schema.py
```diff
@@ -0,0 +1,896 @@
...
+    schema: str = "http://json-schema.org/draft-07/schema#"
+    type: str = "object"
+    description: str = "TBD"
+    properties: dict[str, Any] = field(default_factory=dict)
+    required: list[str] = field(default_factory=list)
+    all_of: list[dict[str, Any]] = field(default_factory=list)
+
+    def as_json_schema_dict(self) -> dict[str, Any]:
+        """
+        Returns class as a JSON Schema dictionary, with proper keywords
```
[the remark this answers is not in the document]
andrewelamb: I wish! Then I wouldn't need the `as_json_schema_dict`. Attributes in dataclasses can't have names that begin with `$`

schematic/schemas/create_json_schema.py
```diff
@@ -0,0 +1,896 @@
...
+    def add_to_all_of_list(self, item: dict[str, Any]) -> None:
+        """
+        Adds a property to the all_of list
+
+        Args:
+            item: The item to add to the all_of list
+        """
+        self.all_of.append(item)
+
+    def update_property(self, property_dict: dict[str, Any]) -> None:
```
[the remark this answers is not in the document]
andrewelamb: Yep that's a good idea. This would only happen if the graph somehow got two nodes with the same name. I'm not sure if that's possible, but better to be safe.

schematic/schemas/create_json_schema.py
andrewelamb: I've resolved most of the Any's but one that would have to have been a really cumbersome type.

PR #1607, opened 2025-05-21
[SCHEMATIC-305] JSONLD parser now parses sms:columnType field

andrewelamb:
# **Problem:**
In order for the columnType to be defined in a JSONLD data model, the `DataModelJSONLDParser` class needs to look for the `sms:columnType` field. Then the value of that field needs to be one of the allowed types.

# **Solution:**

- The `DataModelRelationships` class has a new relationship `columnType` that has a field `allowed_values`.
- The `DataModelRelationships` class has a new method for getting the allowed values.
- The `DataModelJSONLDParser` has a new method for checking the value from the data model against allowed values for a relationship, if the allowed values exist.

# **Testing:**

- A new test data model that is a copy of the example data model, but with `columType` and type validation rules removed. 
- Tests for the new methods.

Depends on: #1610 #1612 - @SageGJ


schematic/schemas/data_model_relationships.py
```diff
@@ -188,6 +191,16 @@ def define_data_model_relationships(self) -> dict:
...
             },
+            "columnType": {
+                "jsonld_key": "sms:columnType",
+                "csv_header": "ColumnType",
+                "node_label": "columnType",
+                "type": str,
+                "required_header": False,
+                "edge_rel": False,
+                "node_attr_dict": {"default": None},
+                "allowed_values": ["string", "integer", "number", "boolean"],
```
thomasyu888, line 202: Susheel was mentioning something like using xkd? to link the values?  I wonder if that's of value for this ticket if we're largely unsure what the json-ld is going to look like in October?

schematic/schemas/data_model_relationships.py
```diff
@@ -188,6 +191,16 @@ def define_data_model_relationships(self) -> dict:
                     "standard": get_label_from_display_name,
                 },
             },
+            "columnType": {
+                "jsonld_key": "sms:columnType",
+                "csv_header": "ColumnType",
```
thomasyu888, line 196: Can you elaborate on what this impacts?

schematic/schemas/data_model_relationships.py
andrewelamb: Yep! I'm happy to do it if we can find the right ontology. I'm good at getting it out there without it for now and adding it in leter if its important enough.

tests/data/example.model_column_type.jsonld
```diff
@@ -0,0 +1,1790 @@
+{
```
linglp, line 1: where is the code that generated this jsonld? `example.model_column_type.jsonld`

schematic/schemas/data_model_relationships.py
linglp: Should this work for a csv version of data model with header `ColumnType` too? Or this feature is only intended for adding "columnType" to json-ld?

tests/data/example.model_column_type.jsonld
andrewelamb: It wasn't generated by code. Its the existing example jsonld modified to include `sms:columnType`.

schematic/schemas/data_model_relationships.py
andrewelamb: It should eventually work for CSV as well. That's an issue @SageGJ  is working on.

andrewelamb: @thomasyu888 Two things: 
1. In the `DataModelJSONLDParser.gather_jsonld_attributes_relationships` method, it loops through all the entries from the JSON-LD, and for each of those loops through each relationship in the `DataModelRelationships`. For each entry/relationship it uses the relationship `jsonld_key` to see if the entry has that field, and if so, pull it out, parses it, and adds it to a dict of entries it found. This dict is eventually what is returned.
2. DataModelRelationships now has a new method `get_allowed_values` that gets the `allowed_values` for the relationship. `DataModelJSONLDParser.gather_jsonld_attributes_relationships` uses this, if it exists, to check the entry from the JSON-LD that it is in the `allowed_values` list.

thomasyu888: Thanks @andrewelamb !

tests/data/example.model_column_type.jsonld
SageGJ: My concern is related to parity between jsonld and csv versions of the same model given the schematic workflow of converting a csv to a jsonld. While I realize users can now choose either as an entrypoint I still think it's important that within our tests, the data we use as examples sets a good example. I think it would be useful to add a csv version of the example model with the columntype column and then convert that to jsonld so that both can serve as examples and we know that our tests of the csv parser and the jsonld parser are indeed testing the same thing in different forms

SageGJ (commented):
One comment other than that LGTM

tests/data/example.model_column_type.jsonld
andrewelamb: @SageGJ I totally agree! But we won't be able to do that until more most of the issues in [this epic](https://sagebionetworks.jira.com/browse/SCHEMATIC-298) are done.

SageGJ: @andrewelamb ahhh right my mistake, I forgot about where this is happening in relation to the other issues. Perhaps we should just do a final check for parity between the two model versions once the rest of the epic is complete

schematic/schemas/data_model_relationships.py
```diff
@@ -235,3 +248,24 @@ def retreive_rel_headers_dict(self, edge: bool) -> dict[str, str]:
...
+            AssertionError: If the allowed values aren't a list
+
+        Returns:
+             A list of allowed values if they exist, otherwise None
+        """
+        if relationship not in self.relationships_dictionary:
+            raise ValueError(f"Relationship: '{relationship}' not in dictionary")
+        allowed_values = self.relationships_dictionary[relationship].get(
+            "allowed_values"
+        )
```
BryanFauble, line 269: Should this be:

```suggestion
        allowed_values = self.relationships_dictionary[relationship].get(
            "allowed_values", None
        )
```
?

BryanFauble (approved):
Flagged 2 areas that may need changes, otherwise this looks good to me

schematic/schemas/data_model_relationships.py
andrewelamb: Isn't `None` the default?

schematic/schemas/data_model_parser.py
```diff
@@ -456,6 +456,12 @@ def gather_jsonld_attributes_relationships(self, model_jsonld: list[dict]) -> di
                             id_jsonld_key=id_jsonld_key,
                             model_jsonld=model_jsonld,
                         )
+                        check_allowed_values(
+                            self.dmr,
+                            entry_id=entry["@id"],
```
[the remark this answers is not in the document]
andrewelamb: Yes, good catch.

thomasyu888 (approved):
🔥 LGTM! Great reviews + work here

andrewelamb:
@SageGJ I think it's fine as is.

schematic/utils/schema_utils.py
```diff
@@ -6,10 +6,14 @@
 import logging
 import os
 import string
-from typing import Literal, Optional, Union
+from typing import Literal, Optional, Union, Any, TYPE_CHECKING
+import pandas as pd
 
 import inflection
 
+if TYPE_CHECKING:
```
linglp, line 14: Is this because DataModelRelationships is very slow to load? Is there benchmark for the time difference of doing: 
```
if TYPE_CHECKING:
      from schematic.schemas.data_model_relationships import DataModelRelationships
```
versus simply importing by doing: 
```
from schematic.schemas.data_model_relationships import DataModelRelationships
```

tests/unit/test_schema_utils.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @pytest.mark.parametrize(
+        "value, relationship",
+        [
+            ("string", "columnType"),
+            ("boolean", "columnType"),
+            ("integer", "columnType"),
+        ],
+    )
+    def test__check_allowed_values(self, value: Any, relationship: str) -> None:
```
linglp, line 42: nit: should be `test_check_allowed_values` instead of `test__check_allowed_values`?

tests/unit/test_schema_utils.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class TestSchemaUtils:
+    """
+    Tests for the schema utils module.
+    """
+
+    def test_model_as_dataframe(self) -> None:
+        """
+        Tests DataModelParser.model_as_dataframe
```
linglp, line 17: nit: I couldn't find `model_as_dataframe`. Perhaps you meant `parsed_model_as_dataframe`?

schematic/schemas/data_model_parser.py
```diff
@@ -232,8 +233,39 @@ def gather_csv_attributes_relationships(
                     attr_rel_dictionary[attribute_name]["Relationships"].update(
                         {relationship: parsed_rel_entry}
                     )
+            if model_includes_column_type:
```
linglp, line 236: This is more of a design question — what you have works fine functionally. However, I was wondering: since columnType is already part of the relationship dictionary, would it make sense to parse it in the same way we handle other relationships in parse_entry?

If the intention is to reserve parse_entry strictly for required headers, perhaps parse_column_type could be refactored into a more general utility for handling optional headers as well?

Also, I noticed there doesn’t seem to be a test for parse_entry — it might be worth adding one for coverage and clarity.

schematic/schemas/data_model_relationships.py
```diff
@@ -188,6 +191,16 @@ def define_data_model_relationships(self) -> dict:
                     "standard": get_label_from_display_name,
                 },
             },
+            "columnType": {
```
linglp, line 194: I think it makes sense to add columnType here, but it can be confusing if because "columnType" is optional while everything in the relationship dictionary is considered as "required". 

As an example, see here: 
```
    def define_required_csv_headers(self) -> list:
        """
        Helper function to retrieve required CSV headers, alert if required header was
          not provided.
        Returns:
            required_headers: lst, Required CSV headers.
        """
        required_headers = []
        for key, value in self.relationships_dictionary.items():
```
As you can see, define_required_csv_headers pulls headers from all entries in the relationships_dictionary. So just by reading the define_required_csv_headers function, it gives the impression that columnType is a required field as well.

linglp (commented):
Sorry—I was still in the middle of typing my comments when you saw the review. I tested the overall functionality and it works well. My feedback is more about the design considerations.

schematic/schemas/data_model_parser.py
andrewelamb: @linglp This was done in @SageGJ's DataModelCSVParser ticket that got merged into this PR I'll let her comment.

tests/unit/test_schema_utils.py
andrewelamb: @linglp This was done in @SageGJ's refactor ticket that got merged into this PR I'll let her comment.

schematic/utils/schema_utils.py
```diff
@@ -499,3 +503,46 @@ def get_json_schema_log_file_path(data_model_path: str, source_node: str) -> str
         prefix = prefix_root
     json_schema_log_file_path = f"{prefix}.{source_node}.schema.json"
     return json_schema_log_file_path
+
+
+def parsed_model_as_dataframe(parsed_model: dict[str, dict[str, Any]]) -> pd.DataFrame:
```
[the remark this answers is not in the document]
andrewelamb: @linglp This was done in @SageGJ's refactor ticket that got merged into this PR I'll let her comment.

PR #1611, opened 2025-05-28
[schematic-292]: add title to property

linglp:
# **Problem:**
"title" in property is missing 

# **Solution:**
* title is added to each property of the jsonschema
* use display name for title


andrewelamb (approved):

linglp:
Thanks @andrewelamb . I will merge now.

PR #1612, opened 2025-05-29
[schematic-304] Allow parsing of `columnType` column for CSV data models

SageGJ:
# **Problem:**

The `columnType` column in a csv data model needs to be  able to be parsed so that it can be used on the backend.

# **Solution:**
Add functionality to parse the new column and add it to the attribute:relationship dictionary like:

```
{'Patient': {"columType": "string", 'Relationships': {'Attribute': 'Patient', 'DependsOn': ['Patient ID', 'Sex', 'Year of Birth', 'Diagnosis', 'Component'], 'Required': False, 'Parent': ['DataType']}}}
```

# **Testing:**

Tests for the csv data model parser have been added according to the feature spec in the assoc ticket

Depends On: #1610


schematic/schemas/data_model_parser.py
```diff
@@ -259,8 +260,38 @@ def gather_csv_attributes_relationships(
...
+        column_type = attr.get("columnType")
+
+        # If no column type specified, we don't want to add any entry to the dictionary
+        if pd.isna(column_type):
+            return {}
+
+        # column types should be case agnostic and valid
+
+        # mypy error ignored due to handling of null values above
+        column_type = column_type.strip().lower()  # type: ignore
```
andrewelamb, line 287: I don't think you are using a mypy error here, because of possible null values. Mypy should know this can't be `Null` because of the check on line 281. It's because it could be anything (besides `Null`), and you are using a string method. 

What you should do here is:

```
column_type = str(column_type).strip().lower()
```
In case pandas reads the values in the column as anything besides strings

andrewelamb (changes requested):

schematic/schemas/data_model_parser.py
SageGJ: I see, thanks!

BryanFauble (approved):
LGTM!

andrewelamb (approved):

PR #1617, opened 2025-06-10
fixes to csv data model parser so data model graph can parse Column…

andrewelamb:
# **Problem:**
The `DataModelGraph` class isn't parsing `columnType` attributes from the `DataModelCSVParser`

# **Solution:**
Changes are made to make `DataModelCSVParser` output look like `DataModelJSONLDParser` output:
- The `columnType` attribute is moved one dict lower into `relationships`
- The `columnType` attribute is renamed `ColumnType`

# **Testing:**

- Unit tests for `DataModelCSVParser` are modified to account for the above changes.
- Unit tests for `DataModelGraphExplorer are added for csv data model input


BryanFauble (approved):
LGTM!


All the more reason for these strings to be in a constants file :)

andrewelamb:
@BryanFauble Ugggh, what is really needed is to standardize the output from the DataModelParser class into one or more dataclasses. They created a complicated nested dict, that has to be a certain way for the `DataModelGraph` class to parse.

SageGJ (approved):

tests/unit/test_data_model_parser.py
```diff
@@ -100,14 +100,16 @@ def test_parse_csv_model_with_column_type(
...
                 # If the expected type is None, we expect the column type to be missing
                 assert (
-                    "columnType" not in result[expected_attribute]
-                ), f"Expected no column type for '{expected_attribute}', but got '{result[expected_attribute].get('columnType')}'"
+                    "ColumnType" not in result[expected_attribute]["Relationships"]
+                ), f"Expected no column type for '{expected_attribute}', but got '{result[expected_attribute]['Relationships'].get('ColumnType')}'"
                 continue
 
             # AND the column type of each attribute should match the expected type if a column type is specified
+            print(result[expected_attribute])
```
linglp, line 108: is this intended?

andrewelamb: Nope, thanks for catching!

PR #1620, opened 2025-06-17
[SCHEMATIC-295] Add ability for JSON Schemas to have format:date and format:uri based on date and url validation rules

andrewelamb:
# **Problem:**
When creating JSON Schemas, the `date` and `url` validation rules were being ignored. 

# **Solution:**
When the  `date` and `url` validation rules are present, the JSON Schema property gets the `format` keyword. The value fo this keyword is either `"format": "date"` or `"format": "uri"`

# **Testing:**
`Date` and `URL` attributes were added to the test data model `JSONSchemaComponent` datatype. The expected generated JSON Schemas have these properties and the expected `format` keyword.

### DIFF
--- a/schematic/schemas/json_schema_validation_rule_functions.py
+++ b/schematic/schemas/json_schema_validation_rule_functions.py
@@ -187,7 +187,7 @@ def check_for_duplicate_inputted_rules(inputted_rules: list[str]) -> None:
     """
     rule_names = get_names_from_inputted_rules(inputted_rules)
     if sorted(rule_names) != sorted(list(set(rule_names))):
-        raise ValueError("Validation Rules contains duplicates: ", inputted_rules)
+        raise ValueError(f"Validation Rules contains duplicates: {inputted_rules}")
 
 
 def check_for_conflicting_inputted_rules(inputted_rules: list[str]) -> None:


schematic/schemas/constants.py
```diff
@@ -29,13 +32,3 @@ class RegexModule(Enum):
...
-
-
-# A dict where the keys are type validation rules, and the values are their JSON Schema equivalent
-TYPE_RULES = {
-    ValidationRule.STR.value: JSONSchemaType.STRING.value,
-    ValidationRule.NUM.value: JSONSchemaType.NUMBER.value,
-    ValidationRule.FLOAT.value: JSONSchemaType.NUMBER.value,
-    ValidationRule.INT.value: JSONSchemaType.INTEGER.value,
-    ValidationRule.BOOL.value: JSONSchemaType.BOOLEAN.value,
-}
```
andrewelamb, line 41: These got moved to the ValidationRule class

schematic/schemas/constants.py
```diff
@@ -3,9 +3,12 @@
 from enum import Enum
 
 
-class ValidationRule(Enum):
+class ValidationRuleName(Enum):
```
andrewelamb, line 6: This got renamed because there is now a ValidationRule class

schematic/schemas/create_json_schema.py
```diff
@@ -15,13 +15,17 @@
...
-    RegexModule,
-    TYPE_RULES,
+    filter_unused_rules,
+    check_for_rule_conflicts,
+    check_for_rule_duplicates,
+    get_in_range_parameters_from_rule,
+    get_regex_parameters_from_rule,
+    get_js_type_from_rule_list,
+    get_rule_from_rule_list,
 )
```
andrewelamb, line 29: The code for getting JSON Schema keywords from validation rules was moved to validation_rule_functions.py

schematic/schemas/create_json_schema.py
```diff
@@ -215,156 +223,59 @@ def _get_validation_rule_based_fields(
...
-    if len(rule_list) > 1:
-        msg = (
-            f"Found more than one '{rule_value}' rule in validation rules: {rule_list}"
-        )
-        raise ValueError(msg)
-    if len(rule_list) == 0:
-        return None
-    return rule_list[0]
-
-
```
andrewelamb, line 367: This functionality was moved to validation_rule_functions.py

schematic/schemas/create_json_schema.py
```diff
@@ -201,168 +207,80 @@ def __post_init__(self) -> None:
...
+        check_for_conflicting_inputted_rules(validation_rules)
 
-    Arguments:
-        rule_list: A list of validation rules
+        if get_rule_from_inputted_rules(ValidationRuleName.LIST, validation_rules):
+            js_is_array = True
 
-    Raises:
-        ValueError: When more than one type rule is found
+        js_type = get_js_type_from_inputted_rules(validation_rules)
```
linglp, line 243: I found the logic here difficult to follow.  Why `get_js_type_from_inputted_rules` is not returning the correct `js_type` in the beginning and we have to override it later? if a rule has both date and regex, it seems like js_type will get overwritten first by `if date_rule` and then by `if regex_rule`

linglp: I also think maybe the mapping between json schema type and validation rules can be more explicit? Just by reading the code here, it is hard to process what the mapping is.

schematic/schemas/validation_rule_functions.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+@dataclass
+class ValidationRule:
+    """
+    This class represents a validation rule to be used for creating JSON Schemas
+
+    Attributes:
+        name: The name of the validation rule
+        js_type: The JSON Schema type this rule indicates
```
linglp, line 15: Nit: When I first read the docstring, I found it a bit unclear what js_type was referring to. Based on reading through the code, I think it's meant to map validation rule to the JSON Schema type this rule is compatible with? 

If that’s the case, it might help to clarify this in the docstring — perhaps with a reference to the list of types defined by JSON Schema [here](https://json-schema.org/understanding-json-schema/reference/type#type-specific-keywords).

linglp: Could you also update the docstring of other functions that use `js_type` parameter?

schematic/schemas/validation_rule_functions.py
```diff
@@ -0,0 +1,386 @@
...
+
+    name: ValidationRuleName
+    js_type: Optional[JSONSchemaType]
+    is_type_rule: bool
+    incompatible_rules: list[ValidationRuleName]
+    parameters: Optional[list[str]] = None
+
+
+# A List of validation rules currently collected to make a JSON Schema
+_VALIDATION_RULES = [
```
linglp, line 29: I think this list is not as efficient as the dictionary that we had before. Every time we will have to loop through the list just to get the name of the validation rule in `_get_rule_by_name`. Can we change this structure back to a dictionary? and the key can just be "ValidationRuleName". Here's an example: 
```
    ValidationRuleName.LIST: ValidationRule(
        name=ValidationRuleName.LIST,
        js_type=None,
        is_type_rule=False,
        incompatible_rules=[],
    )
```

schematic/schemas/validation_rule_functions.py
```diff
@@ -0,0 +1,386 @@
...
+    ValidationRule(
+        name=ValidationRuleName.LIST,
```
linglp, line 31: why here `LIST` does not map to array: https://json-schema.org/understanding-json-schema/reference/type#type-specific-keywords? is it because `array` is not a supported key word in synapse?

schematic/schemas/create_json_schema.py
```diff
@@ -201,168 +207,80 @@ def __post_init__(self) -> None:
...
-    if len(parameters) > 1 and parameters[1].isnumeric():
-        range_min = float(parameters[1])
-    if len(parameters) > 2 and parameters[2].isnumeric():
-        range_max = float(parameters[2])
-    return (range_min, range_max)
-
+        A tuple containing fields for a Node object:
+        - js_is_array: Whether or not the Node should be an array in JSON Schema
+        - js_type: The JSON Schema type
+        - js_format: The JSON Schema format
```
linglp, line 223: Could you clarify how js_format differs from js_type? From what I can tell, it looks like we currently only support format = "uri" and format = "date" — is that correct? If so, would it be possible to expand a bit in the docstring on what format means and how it's expected to be used?

linglp (changes requested):
Hi @andrewelamb ! Thanks for your hard work. My main concerns are: 
1) the look up of validation rule name using a list seems to be inefficient. See `_get_rule_by_name` and `_VALIDATION_RULES`
2) I found some of the variable names a bit unclear on first read. For example, `js_is_array`, `js_type`, and `js_format`. It was a bit confusing that `js_is_array` is separate from `js_type`, since "array" is technically a JSON Schema type. Also, I'm not entirely sure how js_format is used or when it's expected to be set. Would it be possible to clarify the distinction between these variables, either in comments or the docstring? Or is it possible to consolidate some of them so that we have less variables to maintain? 
3) in `_get_validation_rule_based_fields`, we overwrite `js_type` for a couple of times. I am curious if there's a way to make `get_js_type_from_inputted_rules` more "authoritative" so that the logic becomes easier to follow.

schematic/schemas/validation_rule_functions.py
andrewelamb: Great question! You are correct about array. See this design doc:

schematic/schemas/create_json_schema.py
andrewelamb: Could you elaborate? Both `type` and `format` are keywords in JSON Schema, so these like the other attributes being returned will define the properties JSON Schema.

schematic/schemas/validation_rule_functions.py
```diff
@@ -0,0 +1,386 @@
...
+
+@dataclass
+class ValidationRule:
+    """
+    This class represents a validation rule to be used for creating JSON Schemas
+
+    Attributes:
+        name: The name of the validation rule
+        js_type: The JSON Schema type this rule indicates
+        is_type_rule: Whether or not this rule directly corresponds to a JSON Schema type
```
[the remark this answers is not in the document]
andrewelamb: After making some other changes, I no longer needed the `is_type_rule` attribute.

schematic/schemas/create_json_schema.py
```diff
@@ -201,168 +208,77 @@ def __post_init__(self) -> None:
...
+        - js_minimum: If the type is numeric the JSON Schema minimum
+        - js_maximum: If the type is numeric the JSON Schema maximum
+        - js_pattern: If the type is string the JSON Schema pattern
+    """
+    if not validation_rules:
+        return (False, None, None, None, None, None)
+
```
SageGJ, line 236: I think it would be better to initialize all of the return variables to `None` at the beginning, and then return them by name, not by value. Then, you could eliminate all of `else` clauses because they're already set and the code will be simpler and eaiser to understand for other devs.

schematic/schemas/create_json_schema.py
```diff
@@ -201,168 +208,77 @@ def __post_init__(self) -> None:
...
+    validation_rules = filter_unused_inputted_rules(validation_rules)
```
SageGJ, line 237: The use of "inputted_rules" feels cumbersome to read after this point, especially since the rules in use further on are not the raw inputs but a filetered subset of them

schematic/schemas/validation_rule_functions.py
SageGJ, line 1: Can this module have a more descriptive name? At first glance it feels redundant compared to `utils/validate_rules_utils.py`. Including something related to jsonschemas in the name I think would help

tests/data/example.model.csv
```diff
@@ -53,14 +53,16 @@ MockRDB,,,"Component, MockRDB_id, SourceManifest",,FALSE,DataType,,,
 MockRDB_id,,,,,TRUE,DataProperty,,,int
 SourceManifest,,,,,TRUE,DataProperty,,,
 MockFilename,,,"Component, Filename",,FALSE,DataType,,,
-JSONSchemaComponent,Component to hold attributes for testing JSON Schemas,,"Component, No Rules, No Rules Not Required, Enum, Enum Not Required, List String, List String, List InRange, List Not Required, List Enum Not Required, String Not Required, ",,FALSE,DataType,,,
+JSONSchemaComponent,Component to hold attributes for testing JSON Schemas,,"Component, No Rules, No Rules Not Required, String, String Not Required, Enum, Enum Not Required, Date, URL, InRange, Regex, List, List Not Required, List Enum, List Enum Not Required, List String, List InRange",,FALSE,DataType,,,
 No Rules,,,,,TRUE,DataProperty,,,
```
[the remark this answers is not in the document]
SageGJ: could you make these additions to the other example models as well for parity?

tests/unit/test_create_json_schema.py
```diff
@@ -569,7 +459,12 @@ def test_create_json_schema_with_class_label(
         schema_path=test_path,
         use_property_display_names=False,
     )
-    assert json_files_equal(expected_path, test_path)
+    with open(expected_path, encoding="utf-8") as file1, open(
+        test_path, encoding="utf-8"
+    ) as file2:
+        expected_json = json.load(file1)
+        test_json = json.load(file2)
+    assert expected_json == test_json
```
SageGJ, line 468: did something happen here with the old funciton?

andrewelamb: It works fine most of the time, but when you do `pytest -vv` you don't get the detailed difference when looking at things like dicts from JSON files.

thomasyu888 (commented):
🔥 Great work @andrewelamb and reviews @Sage-Bionetworks/schematic-developers ! I'll defer to @SageGJ as final reviewer after all comments addressed. 

Looking forward to the documentation that shows how the validationRules map to these!

schematic/schemas/create_json_schema.py
andrewelamb: Yes this was to differentiate them from the `ValidationRule` class in the `validation_rule_functions.py` module. I'm open to a better name.

schematic/schemas/create_json_schema.py
linglp: Could you let me know what comment is this?

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,367 @@
...
+def check_for_conflicting_inputted_rules(inputted_rules: list[str]) -> None:
+    """Checks that each rule has no conflicts with any other rule
+
+    Arguments:
+        inputted_rules: A list of validation rules
+
+    Raises:
+        ValueError: If a rule is in conflict with any other rule
+    """
+    rule_names = _get_names_from_inputted_rules(inputted_rules)
```
linglp, line 180: Since `inputted_rules` are the same for both `check_for_conflicting_inputted_rules` and `check_for_duplicate_inputted_rules`, I am wondering can this step be called only once?

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,367 @@
...
+            ValidationRuleName.INT,
+            ValidationRuleName.FLOAT,
+            ValidationRuleName.BOOL,
+            ValidationRuleName.NUM,
+        ],
+    ),
+    "regex": ValidationRule(
+        name=ValidationRuleName.REGEX,
+        js_type=JSONSchemaType.STRING,
+        incompatible_rules=[
```
linglp, line 74: Why regex is compatible with InRange?

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,367 @@
...
+            ValidationRuleName.FLOAT,
+            ValidationRuleName.BOOL,
+            ValidationRuleName.NUM,
+        ],
+    ),
+    "url": ValidationRule(
+        name=ValidationRuleName.URL,
+        js_type=JSONSchemaType.STRING,
+        incompatible_rules=[
+            ValidationRuleName.DATE,
```
linglp, line 64: hmm URL is incompatible with `inRange` as well?

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,367 @@
...
+    "list": ValidationRule(
+        name=ValidationRuleName.LIST,
+        js_type=None,
+        incompatible_rules=[],
+    ),
+    "date": ValidationRule(
+        name=ValidationRuleName.DATE,
+        js_type=JSONSchemaType.STRING,
+        incompatible_rules=[
+            ValidationRuleName.URL,
```
linglp, line 53: missing inRange here?

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,367 @@
...
+            ValidationRuleName.DATE,
+            ValidationRuleName.REGEX,
+            ValidationRuleName.STR,
+            ValidationRuleName.BOOL,
+        ],
+    ),
+    "num": ValidationRule(
+        name=ValidationRuleName.NUM,
+        js_type=JSONSchemaType.NUMBER,
+        incompatible_rules=[
```
linglp, line 130: Also missing inRange

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,367 @@
...
+            ValidationRuleName.REGEX,
+            ValidationRuleName.STR,
+            ValidationRuleName.BOOL,
+        ],
+    ),
+    "int": ValidationRule(
+        name=ValidationRuleName.INT,
+        js_type=JSONSchemaType.INTEGER,
+        incompatible_rules=[
+            ValidationRuleName.URL,
```
linglp, line 120: Also missing InRange?

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,367 @@
...
+    name: ValidationRuleName
+    js_type: Optional[JSONSchemaType]
+    incompatible_rules: list[ValidationRuleName]
+    parameters: Optional[list[str]] = None
+
+
+# A dictionary of current Schematic validation rules
+#   where the keys are name fo the rule in Schematic
+#   and the values are ValidationRule objects
+_VALIDATION_RULES = {
```
linglp, line 43: I am trying to think if there's a better way rather than manually maintaining a list of validation rules, especially when: 
maintaining manual, duplicated, and asymmetric incompatible_rules lists is brittle and error-prone. Especially when:
* Rule A lists Rule B as incompatible, but B doesn’t reciprocate.
* Someone adds a new rule but forgets to update other rules’ incompatible_rules.

Can we try to maintain a map of incompatible rule pair? is that better?

tests/unit/test_json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,355 @@
...
+        (["regex", "int"]),
+    ],
+    ids=[
+        "Multiple type rules",
+        "Multiple format rules",
+        "InRange and str rules",
+        "Regex and int rules",
+    ],
+)
+def test_check_for_conflicting_inputted_rules_with_conflicts(
```
linglp, line 107: the test cases here are fine. I am thinking if one rule has multiple incompatible rules, can those incompatible rules all be surfaced?

schematic/schemas/create_json_schema.py
```diff
@@ -201,168 +208,85 @@ def __post_init__(self) -> None:
...
+        is not None
+    )
 
-    Arguments:
-        rule_list: A list of validation rules
+    js_type = get_js_type_from_inputted_rules(validation_rules)
 
-    Raises:
-        ValueError: When more than one type rule is found
+    if get_rule_from_inputted_rules(ValidationRuleName.URL, validation_rules):
```
linglp, line 256: I am a bit worried if this design would be scalable. Currently we have: 
```
        if get_rule_from_inputted_rules(ValidationRuleName.LIST, validation_rules):
            js_is_array = True

        if get_rule_from_inputted_rules(ValidationRuleName.URL, validation_rules):
            js_format = "uri"

        if get_rule_from_inputted_rules(ValidationRuleName.DATE, validation_rules):
            js_format = "date"

        in_range_rule = get_rule_from_inputted_rules(
            ValidationRuleName.IN_RANGE, validation_rules
        )
        if in_range_rule:
            js_minimum, js_maximum = get_in_range_parameters_from_inputted_rule(
                in_range_rule
            )

        regex_rule = get_rule_from_inputted_rules(
            ValidationRuleName.REGEX, validation_rules
        )
        if regex_rule:
            js_pattern = get_regex_parameters_from_inputted_rule(regex_rule)
```
I think this is readable, but I am worried that every time we add a new rule, we will essentially have to add an if statement here.

linglp: Also, this line here: 
```
    inputted_rules = [
        rule for rule in inputted_rules if rule.startswith(rule_name.value)
    ]
```
is being called every time `get_rule_from_inputted_rules` is called. The function is essentially to figure out if a certain ValidationRuleName is in the inputted rules? 

I am thinking what if we group all the rules first? So that we have something like: 
```
{ValidationRuleName.LIST: [list], 
ValidationRuleName.URL:[url, other-url-related-input-rules]
```
And then we could just do: 
```
if ValidationRuleName.LIST in grouped_rules:
    js_is_array = True
if ValidationRuleName.URL in grouped_rules:
    js_format = "uri"
```

linglp (changes requested):
@andrewelamb Thanks for your hard work. I glanced through the code again, and I am mostly concerned about: 
* the manual list of incompatible_rules that we have to maintain for each validation rule. For example, I saw in your code that `inRange` is incompatible with `Regex`, but `Regex` didn't list `InRange` as incompatible. I think it is a bit confusing to maintain incompatible lists in two places. Can we maintain a list of incompatible pairs rather than two lists? 
* I was also thinking — could we reduce the number of times we call get_rule_from_inputted_rules? Right now, each time a new rule is added, it requires its own if block that calls this function. That feels a bit repetitive and hard to maintain. Also, within get_rule_from_inputted_rules, there's a step that extracts the validation rule name from each inputted rule. I wonder if that logic could be split out into its own utility function — that way, we could pre-group the inputted rules by their rule name.

schematic/schemas/create_json_schema.py
```diff
@@ -1011,3 +935,5 @@ def _set_type_specific_keywords(schema: dict[str, Any], node: Node) -> None:
         schema["maximum"] = node.maximum
     if node.pattern is not None:
         schema["pattern"] = node.pattern
+    if node.format is not None:
```
linglp, line 938: Maybe worth refactoring this function `_set_type_specific_keywords`? 
Currently we have: 
```
def _set_type_specific_keywords(schema: dict[str, Any], node: Node) -> None:
    """Sets JSON Schema keywords that are allowed if type has been set

    Arguments:
        schema: The schema to set keywords on
        node (Node): The node the corresponds to the property which is being set in the JSON Schema
    """
    if node.minimum is not None:
        schema["minimum"] = node.minimum
    if node.maximum is not None:
        schema["maximum"] = node.maximum
    if node.pattern is not None:
        schema["pattern"] = node.pattern
    if node.format is not None:
        schema["format"] = node.format
```
What about: 
```
for attr in ["minimum", "maximum", "pattern", "format"]:
    value = getattr(node, attr)
    if value is not None:
        schema[attr] = value
```

schematic/schemas/create_json_schema.py
andrewelamb: > I am a bit worried if this design would be scalable. Currently we have:

I hear you on this. But this is as big as it's going to get. We aren't adding any more validation rules to Schematic, and we probably won't be translating anymore existing validation rules into JSON Schema keywords. In fact, in the near future, we'll start shrinking this list. For example, the work we are doing on `columnType` will eventually lead to us removing the type validation rules from here.

linglp: Okay thanks. I am curious is there a design doc that we could potentially reference to see where things are/how things are connected and how things will be shifted in the near future? i.e using columnType?

schematic/schemas/create_json_schema.py
andrewelamb: I like the idea, but this function has changed to :

```
def _set_type_specific_keywords(schema: dict[str, Any], node: Node) -> None:
    """Sets JSON Schema keywords that are allowed if type has been set

    Arguments:
        schema: The schema to set keywords on
        node (Node): The node the corresponds to the property which is being set in the JSON Schema
    """
    if node.minimum is not None:
        schema["minimum"] = node.minimum
    if node.maximum is not None:
        schema["maximum"] = node.maximum
    if node.pattern is not None:
        schema["pattern"] = node.pattern
    if node.format is not None:
        schema["format"] = node.format.value
```

schematic/schemas/json_schema_validation_rule_functions.py
andrewelamb: I agree, but not sure what the solution is. That said, like I mentioned in a different comment this should be the last time we add any validation rules to be converted into JSON Schema keywords. There just aren't anymore Schematic validation rules that have a direct translation into JSON Schema like `date` or `url`.

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,367 @@
...
+        ],
+        parameters=["module", "pattern"],
+    ),
+    "inRange": ValidationRule(
+        name=ValidationRuleName.IN_RANGE,
+        js_type=JSONSchemaType.NUMBER,
+        incompatible_rules=[
+            ValidationRuleName.URL,
+            ValidationRuleName.DATE,
+            ValidationRuleName.REGEX,
```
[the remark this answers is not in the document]
andrewelamb: Not sure what this means...

schematic/schemas/json_schema_validation_rule_functions.py
andrewelamb: inRange is definitely compatible with integers

schematic/schemas/json_schema_validation_rule_functions.py
andrewelamb: InRnage is compatible with numbers

schematic/schemas/create_json_schema.py
linglp: How about: 
```
for attr in ["minimum", "maximum", "pattern"]:
    value = getattr(node, attr)
    if value is not None:
        schema[attr] = value

if node.format is not None:
    schema["format"] = node.format.value
```

SageGJ:
> 🔥 Great work @andrewelamb and reviews @Sage-Bionetworks/schematic-developers ! I'll defer to @SageGJ as final reviewer after all comments addressed.

@andrewelamb just let me know when you're ready for me to take a last look

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,394 @@
...
+    """
+    return [
+        rule
+        for rule in inputted_rules
+        if _get_name_from_inputted_rule(rule) in [e.value for e in ValidationRuleName]
+    ]
+
+
+def check_for_duplicate_inputted_rules(inputted_rules: list[str]) -> None:
+    """Checks that there are no duplicate rules by name
```
linglp, line 167: nit: Checks that there are no rules with duplicate names?

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,394 @@
...
+    if rule and rule.parameters:
+        return dict(zip(rule.parameters, rule_values))
+    return None
+
+
+def _get_name_from_inputted_rule(inputted_rule: str) -> str:
+    """Gets the name from an inputted rule
+
+    Arguments:
+        inputted_rule: An inputted validation rule form a data model
```
linglp, line 376: nit: a typo here? "form" -> "from"

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,394 @@
...
+
+    Returns:
+        The name of the inputted rule
+    """
+    return inputted_rule.split(" ")[0]
+
+
+def _get_rules_by_name(names: list[str]) -> list[ValidationRule]:
```
linglp, line 384: nit: should this function be called `_get_rules_by_names` since the input and output are both plural?

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,394 @@
...
+    """
+    rule_name = _get_name_from_inputted_rule(inputted_rule)
+    rule_values = inputted_rule.split(" ")[1:]
+    rule = _VALIDATION_RULES.get(rule_name)
+    if rule and rule.parameters:
+        return dict(zip(rule.parameters, rule_values))
+    return None
+
+
+def _get_name_from_inputted_rule(inputted_rule: str) -> str:
```
linglp, line 372: Nit: I think `_get_name_from_inputted_rule` and `_get_rules_by_name` are very similar. What about change this function to `_extract_name_from_inputted_rule`

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,394 @@
...
+    return pattern
+
+
+def get_validation_rule_names_from_inputted_rules(
+    inputted_rules: list[str],
+) -> list[ValidationRuleName]:
+    """Gets a list of ValidationRuleNames from a list of inputted validation rules
+
+    Args:
+        inputted_rules: A list of inputted validation rules form a data model
```
linglp, line 328: nit: typo? `form` -> `from`?

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,394 @@
...
+        for rule_name_string in get_names_from_inputted_rules(inputted_rules)
+    ]
+    return [rule.name for rule in validation_rules if rule is not None]
+
+
+def get_names_from_inputted_rules(inputted_rules: list[str]) -> list[str]:
+    """Gets the names from a list of inputted rules
+
+    Arguments:
+        inputted_rules: A list of inputted validation rules form a data model
```
linglp, line 344: also a typo here? `form` -> `from`?

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,394 @@
...
+
+    Raises:
+        ValueError: If there are multiple type rules in the list
+
+    Returns:
+        The JSON Schema type if a type rule is found, otherwise None
+    """
+    rule_names = get_names_from_inputted_rules(inputted_rules)
+    validation_rules = _get_rules_by_name(rule_names)
+    json_schema_types = {
```
linglp, line 245: Nit: can we add a comment for this? something like: Collects all unique js_type values from the validation_rules list, skipping any rules where js_type is None

linglp (commented):
Hi @andrewelamb ! It looks good overall, just some `nit` comments.

schematic/schemas/create_json_schema.py
andrewelamb: @linglp theres not really a good design doc for this, I'll work on one

thomasyu888: @andrewelamb one thought is to create a ticket to add in the live documentation page the data model impact upcoming. 

For example
* Deprecation of basic type validation rules in favor to columnType (it's fine if people have basic type validationRules, but it just won't be used eventually)
* Deprecation of cross manifest validation rules
* Etc.

This way we are looking ahead.

schematic/schemas/create_json_schema.py
```diff
@@ -15,14 +15,20 @@
...
+    check_for_conflicting_inputted_rules,
+    check_for_duplicate_inputted_rules,
+    get_in_range_parameters_from_inputted_rule,
+    get_regex_parameters_from_inputted_rule,
+    get_js_type_from_inputted_rules,
+    get_rule_from_inputted_rules,
+    get_validation_rule_names_from_inputted_rules,
+    get_names_from_inputted_rules,
 )
+from schematic.schemas.constants import JSONSchemaType, JSONSchemaFormat
```
SageGJ, line 31: It looks like `ValidationRuleName` should be imported from `schematic.schemas.constants` instead of `schematic.schemas.json_schema_validation_rule_functions`.

schematic/schemas/create_json_schema.py
```diff
@@ -201,168 +210,91 @@ def __post_init__(self) -> None:
...
+        in_range_rule = get_rule_from_inputted_rules(
+            ValidationRuleName.IN_RANGE, validation_rules
+        )
+        if in_range_rule:
+            js_minimum, js_maximum = get_in_range_parameters_from_inputted_rule(
+                in_range_rule
+            )
+        else:
+            js_minimum = None
+            js_maximum = None
```
SageGJ, line 280: These lines are unnecessary since both variables have been set to `None` above

schematic/schemas/create_json_schema.py
```diff
@@ -201,168 +210,91 @@ def __post_init__(self) -> None:
...
         )
-        raise ValueError(msg)
-    if len(rule_list) == 0:
-        return None
-    return rule_list[0]
+        if regex_rule:
+            js_pattern = get_regex_parameters_from_inputted_rule(regex_rule)
+        else:
+            js_pattern = None
+
```
SageGJ, line 289: These lines are unnecessary since the variable has been set to `None` above

schematic/schemas/create_json_schema.py
```diff
@@ -201,168 +210,91 @@ def __post_init__(self) -> None:
...
 
-    Arguments:
-        rule: A ValidationRule enum
-        rule_list: A list of validation rules
+        if ValidationRuleName.URL in validation_rule_names:
+            js_format = JSONSchemaFormat.URI
+        elif ValidationRuleName.DATE in validation_rule_names:
+            js_format = JSONSchemaFormat.DATE
+        else:
+            js_format = None
```
SageGJ, line 270: I'm assuming this exists but wanted to confirm; there is a check ensureing that both `url` and `date` are not specified for the same attribute right?

tests/data/example.model.column_type_component.csv
SageGJ, line 1: Thanks for updating this example model! Could you update `example.model.column_type_component.invalid.csv` and `example.model.column_type_component.invalid.jsonld` as well?

SageGJ (changes requested):
Additons looks good, just a left a few comments before finalizing. Thanks again!

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,395 @@
...
+def _get_rules_by_names(names: list[str]) -> list[ValidationRule]:
+    """Gets a list of ValidationRules by name of they exist
+
+    Args:
+        names: A list of names of ValidationRules
+
+    Returns:
+        A list of ValidationRules
+    """
+    rules = [_VALIDATION_RULES.get(name) for name in names]
```
linglp, line 394: hmm I have the same concern here. If a name is not a rule in `_VALIDATION_RULES`, this will also fail silently.

schematic/schemas/json_schema_validation_rule_functions.py
```diff
@@ -0,0 +1,395 @@
...
+    """Gets a list of ValidationRuleNames from a list of inputted validation rules
+
+    Args:
+        inputted_rules: A list of inputted validation rules from a data model
+
+    Returns:
+        A list of ValidationRuleNames
+    """
+    validation_rules = [
+        _VALIDATION_RULES.get(rule_name_string)
```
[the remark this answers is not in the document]
andrewelamb: I can add a warning. I can't raise an error here because there are validation rules like `protectAges` that are valid in Schematic, but aren't used for JSON Schema creation. These rules would be dropped from the list, and we don't want to raise an error for those.

thomasyu888 (approved):
🔥 LGTM!  Helping in with the final final review (on top of Gianna's final review) since lingling has other program initiatives.  

Appreciate the reviews and addition of warnings for validation rules not supported (maybe that's a good example of a Jira ticket but it was a good addition)

linglp (approved):
LGTM! Thank you @andrewelamb

PR #1621, opened 2025-06-17
[Schematic-318] Consolidate column type example data models

SageGJ:
# **Problem:**

There were two separate versions of the data model used to specify column type for attributes. This can cause confusion when users look to these as an example. The goal of this PR was to reduce the number of different data models and have the column type model jsonld and csv versions be the same data model specification.

# **Solution:**
## Previous Models

- `tests/data/example.model_column_type.jsonld`
  - initial model created to test jsonld parser since it was developed first
- `example.model.column_type_component.csv`
  - model developed to test csv parser developed after jsonld parser
- `example.model.column_type_component.jsonld`
  - model created after development on `DataModelJsonLD` that allowed a jsonld to be output with the column type information
- `example.model.column_type_component.invalid.csv`
  - model that would raise an error when parsed due to presence of invalid column type specification 

## Current Models

- `example.model.column_type_component.csv`
  - model developed to test csv parser developed after jsonld parser
- `example.model.column_type_component.jsonld`
  - model created after development on `DataModelJsonLD` that allowed a jsonld to be output with the column type information
- `example.model.column_type_component.invalid.csv`
  - model that would raise an error when parsed due to presence of invalid column type specification 
- `example.model.column_type_component.invalid.jsonld`
  - jsonld model with same invalid column type specification as the csv version

## Subsequent Changes
The jsonld parser tests have been modified to use the jsonld versions of the existing csv versions, so now they are checking for the exact same data.

The data model graph explorer tests covering column type information were consolidated to two functions and use the relevant data model files. As part of the new parametrization getting the node column type from the csv using display name is now a newly covered test case.

# **Testing:**

New tests were added and both old and new complete successfully.


tests/unit/test_data_model_parser.py
```diff
@@ -1,66 +1,110 @@
...
-from schematic.utils import schema_utils
 from schematic.utils.schema_utils import parsed_model_as_dataframe
 
 # pylint: disable=protected-access
 
 
-class TestDataModelJsonLdParser:
-    """Unit tests for DataModelJsonLdParser"""
+class TestDataModelJSONLDParser:
+    """Unit tests for DataModelCSVParser"""
```
andrewelamb, line 17: Was this change intentional?

thomasyu888 (approved):
🔥 LGTM!  Thanks for consolidating.

I'll defer to you and @andrewelamb to sync up on how ensuring the tests pass to merge this!

andrewelamb (approved):

SageGJ:
Merging after okay from @andrewelamb

PR #1625, opened 2025-07-08
[SCHEMATIC-308] Add ability for JSON Schema type to be set by columnType attribute

andrewelamb:
[SCHEMATIC-308]

# **Problem:**
When creating a JSON Schema, the type is inferred from existing validation rules if they exist. It would be better if the type were set explicitly in the data model.

# **Solution:**
The JSON Schema creation function will look for the columnType attribute in the data model graph and use it to set the JSON Schema type if it exists. Otherwise, it will continue to infer the type from the validation rules. Infering from validaiton rules has been deprecated and will be sunsetted in the future.

# **Testing:**
Unit tests for both when the node has the columType attribute and when it doesn't.

[SCHEMATIC-308]: https://sagebionetworks.jira.com/browse/SCHEMATIC-308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schematic/schemas/create_json_schema.py
+++ b/schematic/schemas/create_json_schema.py
@@ -12,6 +12,7 @@ import logging
 import os
 from typing import Union, Any, Optional
 from dataclasses import dataclass, field, asdict
+import warnings
 
 from schematic.schemas.data_model_graph import DataModelGraphExplorer
 from schematic.utils.schema_utils import get_json_schema_log_file_path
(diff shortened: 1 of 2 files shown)


SageGJ (approved):
LGTM
new code logic looks sound and the tests cover the variety of cases appropriately.

PR #1627, opened 2025-07-09
Fix json schema bug

andrewelamb:
# **Problem:**
When validating the PR for adding columnType, I discovered a bug. The CLI tool `generate-jsonschema` was used on the `JSONSchemaComponent` in the test data model, and an exception was raised. This was due to the `DataModelGraph,get_node_column_type` method outputing the wrong type. It was outputing a string instead of the `JSONSchemaType` enum. 

# **Solution:**
`DataModelGraph.get_node_column_type` is now outputing the enum.

# **Testing:**

- The tests for `DataModelGraph,get_node_column_type` were corrected
- Additional tests in `create_json_schema` were added


schematic/schemas/data_model_graph.py
```diff
@@ -956,7 +954,10 @@ def get_node_column_type(
         """
         node_label = self._get_node_label(node_label, node_display_name)
         rel_node_label = self.dmr.get_relationship_value("columnType", "node_label")
-        return self.graph.nodes[node_label][rel_node_label]
+        type_string = self.graph.nodes[node_label][rel_node_label]
+        if type_string is None:
+            return type_string
+        return JSONSchemaType(type_string)
```
andrewelamb, line 961: This is where the bug was. This was returning the type as a string, btu should ahve been returning it as the enum.

tests/unit/test_create_json_schema.py
```diff
@@ -616,6 +616,58 @@ def test_create_json_schema_with_display_names(
...
+        use_property_display_names=False,
+    )
+    with open(expected_path, encoding="utf-8") as file1, open(
+        test_path, encoding="utf-8"
+    ) as file2:
+        expected_json = json.load(file1)
+        test_json = json.load(file2)
+    assert expected_json == test_json
+
+
```
andrewelamb, line 670: This tests using the data model with columnType

thomasyu888 (approved):
🔥 LGTM!  Thanks for fixing these tests and doing thorough testing! I'll defer to @SageGJ for final review!

SageGJ (commented):
LGTM! 
Only ask is to also update the invalid column type models as well before merge. Thanks for catching!

PR #1629, opened 2025-07-10
[DPE-1399] add maximum werkzeug version

SageGJ:
# **Problem:**

A our version of flask was incompatible with the latest werkzeug version

# **Solution:**

The version of werkzeug was pinned

# **Testing:**

pre-release API tests pass after update


SageGJ:
@andrewelamb if you're referring to the `poetry.lock` then yes. That file is generated automatically by poetry based on the `pyproject.toml` so if changes were made there then they're necessary

andrewelamb:
@SageGJ  I was able to add the werkzeug version restriction here:

SageGJ:
@andrewelamb okay cool that works

PR #1633, opened 2025-07-15
Display schema column too small error in query_fileview

afwillia:
# **Problem:**

A DCA user reported a `Forbidden` error trying to access folders within a project. This was caused by a 403 error returned from the `/storage/projects/` endpoint. The user and myself both had admin permission to the file view and folders, so it wasn't a 403 issue. Looking up the file view on Synapse, there was an error banner stating one of the schema columns was too short. The user rolled back the schema changes and resolved the issue.

image

# **Solution:**

This PR adds a `elif` statement to check `query_table` errors for the specific text returned by Synapse when a schema column is too short. Avoiding the default 403 error for this case.
Screenshot 2025-07-15 at 3 34 21 PM

### DIFF
--- a/schematic/store/synapse.py
+++ b/schematic/store/synapse.py
@@ -419,6 +419,10 @@ class SynapseStorage(BaseStorage):
                     raise ValueError(
                         f"The columns {missing_column} specified in the query do not exist in the fileview. Please make sure that the column names are correct and that all expected columns have been added to the fileview."
                     )
+                elif "The column size needs to be at least" in exception_text:
+                    raise ValueError(
+                      f"Error in the fileview schema. " + exception_text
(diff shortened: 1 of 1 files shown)


thomasyu888:
@afwillia , is there a workaround for this? If so, we will no longer be addressing non critical bugs. @andrewelamb can you assess this?

andrewelamb:
@afwillia just to be clear there's nothing wrong with your solution. If we hadn't frozen DCA development this would be a fine solution.

afwillia:
@thomasyu888 that's fine as this schema issue will raise an error either way. I imagine in most cases it shouldn't be an issue because users will check the file view on Synapse and notice the error.

PR #1635, opened 2025-07-25
[SCHEMATIC-277] fix bug where columnType mistaken conflicts with no type validation type

andrewelamb:
# **Problem:**

A bug was found while creating JSON Schema, where the data model was using the columnType column. This bug occurred when there was an explicit type set in `columnType`, and the only validation rules are those that don't imply an implicit type, such as `list`. When this happened, there would be an explicit type, but the implicit type would be None, which would cause the exception to be mistakenly raised. Now this only occurs if there is both an explicit and an implicit type, and they are different.

# **Solution:**
The exception is now only raised if there are both an explicit and implicit type, and they conflict.

# **Testing:**

- Added a test that would have caused the exception before, but doesn't now.

### DIFF
--- a/schematic/schemas/create_json_schema.py
+++ b/schematic/schemas/create_json_schema.py
@@ -294,11 +294,8 @@ def _get_validation_rule_based_fields(
                 f"derived from its validation rules: {validation_rules}"
             )
             raise ValueError(msg)
-        # If there is no explicit type and there is an implicit type...
         if implicit_js_type:
-            # then the implicit type is used...
             js_type = implicit_js_type
-            # and a warning is raised since this behavior is deprecated
             msg = (
                 f"No explicit type set for property: '{name}', "
                 "using validation rules to set the type. "


schematic/schemas/create_json_schema.py
```diff
@@ -281,30 +281,31 @@ def _get_validation_rule_based_fields(
...
-                    "You should set the columnType for this property in your data model."
-                )
-                warnings.warn(msg)
+        # If there is an explicit and implicit type set and the implicit type conflicts with the
+        # explicit type, then an exception is raised
+        if (
+            explicit_js_type
+            and implicit_js_type
+            and explicit_js_type != implicit_js_type
+        ):
```
andrewelamb, line 290: The new check checks that there is both an explicit and an implicit type,  and that they conflict.

BryanFauble: IMO The github comment is overkill - There is already an in-line comment explaining what is occurring, and the code is relatively straight forward.

I think there are a few in-line comments throughout this that don't really need to be there, like `# and a warning is raised since this behavior is deprecated`

BryanFauble (approved):
Only 1 comment that doesn't impact the logic of functionality, otherwise looks good to me!

tests/unit/test_create_json_schema.py
```diff
@@ -283,6 +283,16 @@ def test_get_validation_rule_based_fields_no_explicit_type(
             None,
             None,
         ),
+        (
+            ["list"],
```
linglp, line 287: Shouldn't this be ["list", "str"] if we are testing list string?

linglp (commented):
Hi @andrewelamb! Thanks for your hard work! Your solution makes sense to me. 

I think this should probably be a separate ticket, but when I am testing `["inRange 50 100", "int"]` using function: `_get_validation_rule_based_fields_with_explicit_type`, I got an error: 
```
        }
        if len(json_schema_types) > 1:
>           raise ValueError(
                f"Validation rules contain more than one implied JSON Schema type: {inputted_rules}"
            )
E           ValueError: Validation rules contain more than one implied JSON Schema type: ['inRange 50 100', 'int']

schematic/schemas/json_schema_validation_rule_functions.py:263:
```

I believe the issue arises because inRange implicitly uses the Number type, while Integer uses the Integer type. These are treated as conflicting based on the current logic. However, technically they could be combined, since the following code doesn't explicitly include `inRange` in `incompatible_rules`: 
```
    "int": ValidationRule(
        name=ValidationRuleName.INT,
        js_type=JSONSchemaType.INTEGER,
        incompatible_rules=[
            ValidationRuleName.URL,
            ValidationRuleName.DATE,
            ValidationRuleName.REGEX,
            ValidationRuleName.STR,
            ValidationRuleName.BOOL,
            ValidationRuleName.NUM,
            ValidationRuleName.FLOAT,
        ],
    ),
```

tests/unit/test_create_json_schema.py
andrewelamb: @linglp Close! We are testing the case where the explicit type set by setting that as the `columType` (see line 288 just below) AND the only validation rule is a list rule.

andrewelamb:
> Hi @andrewelamb! Thanks for your hard work! Your solution makes sense to me.
> 
> I think this should probably be a separate ticket, but when I am testing `["inRange 50 100", "int"]` using function: `_get_validation_rule_based_fields_with_explicit_type`, I got an error:
> 
> ```
>         }
>         if len(json_schema_types) > 1:
> >           raise ValueError(
>                 f"Validation rules contain more than one implied JSON Schema type: {inputted_rules}"
>             )
> E           ValueError: Validation rules contain more than one implied JSON Schema type: ['inRange 50 100', 'int']
> 
> schematic/schemas/json_schema_validation_rule_functions.py:263:
> ```
> 
> I believe the issue arises because inRange implicitly uses the Number type, while Integer uses the Integer type. These are treated as conflicting based on the current logic. However, technically they could be combined, since the following code doesn't explicitly include `inRange` in `incompatible_rules`:
> 
> ```
>     "int": ValidationRule(
>         name=ValidationRuleName.INT,
>         js_type=JSONSchemaType.INTEGER,
>         incompatible_rules=[
>             ValidationRuleName.URL,
>             ValidationRuleName.DATE,
>             ValidationRuleName.REGEX,
>             ValidationRuleName.STR,
>             ValidationRuleName.BOOL,
>             ValidationRuleName.NUM,
>             ValidationRuleName.FLOAT,
>         ],
>     ),
> ```
Yep you are correct, we should address this. I'll start a new ticket.

tests/unit/test_create_json_schema.py
linglp: Got it — I see where I got confused. The id is `List String`, but it's not meant to test the validation rule ["list", "str"]. Instead, it's intended to test that the column type is being set as String.

linglp (approved):
LGTM!

PR #1642, opened 2025-09-08
[SCHEMATIC-360] Add script to write jsonschemas to a synapse table

thomasyu888:
# **Problem:**

Currently, we maintain a Confluence page manually with the JSON schemas we generate that can be used for Curator MVP.  This won't scale as we create more JSON schemas. 

# **Solution:**
Add ad-hoc script to write jsonschemas that DPE generates to a synapse table for Curator MVP.  This is partially doing some of schematic-360 work and adds a 'script' directory

# **Testing:**

View Synapse table here: https://www.synapse.org/Synapse:syn69735275/tables/

### DIFF
--- a/scripts/write_jsonschemas_to_table.py
+++ b/scripts/write_jsonschemas_to_table.py
@@ -28,6 +28,7 @@ def main():
         "sage.schemas.v2571",
         "sage.schemas.v2581",
         "MultiConsortiaCoordinatingCenter",
+        "org.synapse.nf",
     ]
     js = syn.service("json_schema")
     to_write_schemas = []
@@ -59,6 +60,9 @@ def main():
                     if organization_name == "MultiConsortiaCoordinatingCenter":
                         # only include the latest version of MCC schemas
                         dcc = "MC2"
+                        datatype = version["schemaName"]
+                    elif organization_name == "org.synapse.nf":
+                        dcc = "NF-OSI"
                         datatype = version["schemaName"]
                     else:
                         dcc = version["schemaName"].split(".")[0]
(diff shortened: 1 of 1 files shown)


scripts/write_jsonschemas_to_table.py
```diff
@@ -0,0 +1,54 @@
...
+                        "version": version.semantic_version,
+                        "link": f"https://repo-prod.prod.sagebase.org/repo/v1/schema/type/registered/{version.uri}",
+                    }
+                )
+
+    df = pd.DataFrame(to_write_schemas)
+    # exclude HTAN1, HTAN2, and NF schemas as they have their own JSONschema organizations
+    df = df[~df["dcc"].isin(["htan", "htan2", "nf"])]
+    table = Table(id="syn69735275").get(include_columns=True)
+    table.upsert_rows(values=df, primary_keys=["uri"])
```
BryanFauble, line 88: Great to see this upsert functionality getting some use!

thomasyu888:
This could probably be added to the extensions package instead of this repo

scripts/write_jsonschemas_to_table.py
```diff
@@ -0,0 +1,71 @@
...
+        - name: the full name of the JSONschema version
+        - dcc: the name of the DCC
+        - datatype: the name of the datatype
+        - uri: the URI of the JSONschema version
+        - version: the semantic version of the JSONschema version
+        - link: a link to the JSONschema version in the Synapse repo
+    """
+    syn = synapseclient.login()
+    # json_schema_organizations = ["sage.schemas.v2571", "sage.schemas.v2581"]
+    json_schema_organizations = ["sage.schemas.v2571", "sage.schemas.v2581"]
```
andrewelamb: Would it make sense to have the orgs as a parameter to the main function so that if/when we add new orgs we wouldn't have to update this?

PR #1643, opened 2025-10-30
Fix json creation bugs

andrewelamb:
# **Problem:**

- Users were getting incorrect warnings about not setting types in the "columnType" column.
- Users were getting incorrect errors about type conflicts with the `inRange` rule

# **Solution:**
Since using validation rules to determine type is deprecated and will be sunsetted in the Curator extension version, I've made some temporary fixes to get users going:

- The logic for the type warning was fixed
- The type error was changed to a warning so users could ignore it


BryanFauble (approved):
LGTM!

